PR #8, opened 2026-04-22
feat: migrate console conversations to postgres

thecopy:
**Tests:** 676 passed, 0 failed

Migrate sessions, conversations, messages, and socket sessions from DynamoDB to PostgreSQL. Remove `aiodynamo` dependency and all DynamoDB configuration.

### Deployment Changes

Here are the env vars you can remove from the console-backend deployment:

| Env var | Service | Action |
|---------|---------|--------|
| `DYNAMODB_SESSIONS_TABLE` | console-backend | Remove |
| `DYNAMODB_CONVERSATIONS_TABLE` | console-backend | Remove |
| `DYNAMODB_MESSAGES_TABLE` | console-backend | Remove |

### Changes

- Add SQL migration `039_create_session_and_conversation_tables.sql` with 4 tables: `sessions`, `socket_sessions`, `conversations`, `messages`
- Rewrite `SessionService`, `ConversationService`, `MessagesService`, `SocketSessionService` to use PostgreSQL via `sqlalchemy.text()`
- Update Pydantic models: replace `ttl: int` with `expires_at: datetime`
- Update in-memory service variants to match new model signatures
- Switch `service_instances.py` to use `USE_IN_MEMORY_STORE` env var instead of `DYNAMODB_SESSIONS_TABLE`
- Remove `DynamoDBConfig`, `aiodynamo[httpx]` dependency, and `DYNAMODB_USERS_TABLE` references from config, env templates, and `start-local.sh`
- Fix test migrations path and update test fixtures for PG-backed services

## Checklist


packages/console-backend/playground_backend/models/socket_session.py
```diff
@@ -1,27 +1,26 @@
...
 
-    socket_id: str  # The Socket.IO session ID (sid) with 'socket:' prefix as DynamoDB key
+    socket_id: str  # The Socket.IO session ID (sid) with 'socket:' prefix
     user_id: str  # User's ID (sub from OIDC)
     http_session_id: str  # HTTP session ID for linking back to user session
     agent_url: str | None = None  # Agent URL for cache lookup
     custom_headers: dict[str, str] = Field(default_factory=dict)  # Custom HTTP headers
     is_initialized: bool = False  # Whether client has been initialized
     created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
-    ttl: int  # Unix timestamp for DynamoDB TTL (auto-cleanup)
```
aartaria, line 24: removing ttl means we will cumulate socket session forever in postges propbably we need to schedule a clean up job

aartaria (changes requested):
generally looks good.

though removing the ttl in the socket session model might lead to cumulating records forever in the table

packages/console-backend/playground_backend/models/socket_session.py
aartaria: @copilot please create a follow up issue about this one

PR #9, opened 2026-04-23
feat: Add IObjectStorageService abstraction for pluggable storage backends

AlexMetsai:
Thought to give this a try in a planning session with copilot. Unfortunately the generated refactoring code ended up being quite large, we might need some more time to inspect and test.

update: re-started from the latest main


packages/console-backend/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "itsdangerous>=2.2.0",
     "cachetools>=5.5.0",
     "ringier-a2a-sdk>=0.1.0",
+    "agent-common>=0.7.0",
```
AlexMetsai: which means that we now need common for the backend (storage tool)

scripts/release-helpers.sh
```diff
@@ -36,14 +36,15 @@ pkg_dir() {
...
+    object-storage)          echo "packages/object-storage" ;;
     *) echo "ERROR: Unknown package '$1'" >&2; return 1 ;;
   esac
 }
 
 pkg_type() {
   case "$1" in
     console-frontend|client-slack-frontend|client-slack|client-email|client-google-chat) echo "node" ;;
-    agent-creator|agent-runner|orchestrator-agent|console-backend|catalog-worker|ringier-a2a-sdk|voice-agent|soffice-worker) echo "python" ;;
+    agent-creator|agent-runner|orchestrator-agent|console-backend|catalog-worker|ringier-a2a-sdk|object-storage|voice-agent|soffice-worker) echo "python" ;;
```
aartaria: shall it be here too?

packages/console-backend/console_backend/routers/file_router.py
```diff
@@ -211,3 +212,63 @@ async def serve_local_file(
         raise HTTPException(status_code=404, detail="File not found")
 
     return FileResponse(abs_path)
+
+
+@router.get("/download/{bucket}/{file_path:path}")
```
aartaria, line 217: 🤔 feels there is no protection about which objects and keys one can access ... being local probably is OK

packages/console-backend/console_backend/services/file_storage_service.py
```diff
@@ -171,30 +190,23 @@ async def upload_file(
             filename=filename,
         )
 
-        # Determine size without consuming the stream permanently
-        upload.file.seek(0, os.SEEK_END)
-        size = upload.file.tell()
-        upload.file.seek(0)
```
aartaria, line 177: this might be inefficient?

packages/console-backend/tests/test_notification_transaction_isolation.py
```diff
@@ -8,7 +8,7 @@
 from unittest.mock import AsyncMock, patch
 
 import pytest
-from aiomoto import mock_aws
+from moto import mock_aws
```
aartaria, line 11: this is suspicious makes me think that we are doing some synchronous operation in the code

packages/console-backend/tests/test_sub_agent_service.py
```diff
@@ -18,7 +18,10 @@
 os.environ.setdefault("ECS_CONTAINER_METADATA_URI", "true")
 
 import pytest
-from aiomoto import mock_aws
+from moto import mock_aws
```
aartaria, line 21: suspicious

aartaria (approved):
LGTM

scripts/release-helpers.sh
aartaria: in the sense I believe it should not

AlexMetsai: True, it should not be, just like agent-common.

packages/console-backend/console_backend/services/file_storage_service.py
AlexMetsai: 05b13f8c556dc5b049c6f13e2ddb8a5dea3ea104

talking with copilot, final resolution was:

**Keep `bytes`, document the limit.** Here's why objectively:

- The performance difference is one extra allocation (`await upload.read()` vs boto reading from the file handle). Both hold the full content in memory simultaneously. For files within the infra-enforced limit, this is noise.
- Adding `BinaryIO` support solves a problem that doesn't exist at the current scale, and the correct solution for when it *does* exist (presigned upload URLs) is architecturally different anyway.
- A simpler interface has ongoing maintenance value for every future backend implementation.

I'd add a short note in the interface docstring and reply to the reviewer. Want me to:

1. Add a size note to the `IObjectStorageService.upload()` docstring
2. Add a comment in file_storage_service.py at the `await upload.read()` line explaining the tradeoff


----

also said that for large files the previous approach would be anyway problematic:
1. **The old code wasn't truly streaming either** — `put_object(Body=upload.file)` with aiobotocore doesn't do multipart chunked upload. It still reads the full body into memory before sending. The "efficiency" difference between passing a file handle vs `bytes` is negligible for sub-100MB files.

packages/console-backend/console_backend/routers/file_router.py
AlexMetsai: True, I think it might be difficult to create user-level object access perms, won't it? Should we rely that users are considered trusted?
(we could also have a hash in the path, plus a low ttl)

packages/console-backend/tests/test_notification_transaction_isolation.py
AlexMetsai: Moto's `@mock_aws` should work with both boto3 and aibotocore. But the real smell is elsewhere:

[backend::catalog/sync.py uses synchronous boto3](https://github.com/ringier-data/nannos/blob/feature/abstraction_level_for_storage/packages/console-backend/console_backend/catalog/sync.py#L464)
[backend::scheduler_router.py uses synchronous boto3](https://github.com/ringier-data/nannos/blob/feature/abstraction_level_for_storage/packages/console-backend/console_backend/routers/scheduler_router.py#L78)

Let's create a ticket and improve incrementaly? (e.g. improve async)

PR #46, opened 2026-06-03
feat: installations and a2a client notifications

andreibazhenau-itechart:
closes #

## Checklist


packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -241,3 +243,55 @@ async def get_owner_client_id(self, db: AsyncSession, channel_id: int) -> str |
...
+        """Idempotently create or update a channel keyed by ``(client_id, installation_id)``.
+
+        Returns ``(channel, created)`` where ``created`` is True when a new row was inserted.
+        Mutable fields (name, description, webhook_url, secret, group_ids) are overwritten
+        on update. ``installation_id`` MUST be set on ``data``; callers should branch to
+        :meth:`create_channel` when it is not.
+        """
+        assert data.installation_id is not None, "installation_id required for upsert"
+
+        existing = await self.get_by_installation(db, client_id, data.installation_id)
```
aartaria, line 246: **Check-then-act race on concurrent registration.** `get_by_installation` -> `create_channel` is not atomic. Two replicas booting at once both see `None` and both `INSERT`; the partial unique index `delivery_channels_client_installation_uidx` correctly rejects the second insert, but the resulting `IntegrityError` is unhandled and surfaces as a 500 to the losing replica.

Fix: use `INSERT ... ON CONFLICT (client_id, installation_id) DO UPDATE`, or catch the unique-violation and fall back to the update path.

packages/client-slack/src/services/awsSsmInstallationSecretService.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (existing) return existing;
+
+    const name = this.parameterName(installationId);
+    const generated = randomBytes(32).toString('hex');
+    await this.client.send(
+      new PutParameterCommand({
+        Name: name,
+        Value: generated,
+        Type: 'SecureString',
+        Overwrite: false,
```
aartaria: **`ParameterAlreadyExists` race is unhandled.** `resolve()` does read-then-`PutParameter` with `Overwrite: false`. Two instances provisioning the same `installationId` concurrently both read `null` and both PUT; the loser's `send()` rejects with `ParameterAlreadyExists`, which propagates out of `resolve()`/`getOrCreate()` and fails that boot's registration for the installation. Self-healing on the next boot, but produces a spurious error.

Fix: catch `ParameterAlreadyExists` in `resolve()`, re-read, and return the now-present value. (Same code is duplicated in client-google-chat — keep in sync.)

packages/client-slack/src/slackApp.ts
```diff
@@ -71,6 +76,38 @@ export async function startSlackApp(config: Config) {
...
+     * success (so the caller can route the notification to the correct team)
+     * or `null` if no installation matched.
+     */
+    const validateNotificationToken = async (token: string) => {
+      const installations = await storage.botInstallation.listAll();
+      for (const bot of installations) {
+        if (!bot.isActive) continue;
+        try {
+          const secret = await installationSecretService.get(bot.botName);
+          if (secret && token === secret) return bot;
```
aartaria, line 96: **Non-constant-time secret comparison.** `token === secret` short-circuits on the first differing byte, leaking a timing side-channel on the notification secret (same pattern at a2aNotificationAuth.ts:30 in client-google-chat). Entropy is high (`randomBytes(32)`) so practical risk is low, but this is a textbook anti-pattern.

Fix: `crypto.timingSafeEqual` on equal-length Buffers (guard the length-mismatch case first).

packages/client-slack/src/slackApp.ts
```diff
@@ -71,6 +76,38 @@ export async function startSlackApp(config: Config) {
...
+     * for any active bot installation. Resolves true on first match.
+     */
+    /**
+     * Verify a notification token by matching it against the SSM-stored secret
+     * for any active bot installation. Returns the matching installation on
+     * success (so the caller can route the notification to the correct team)
+     * or `null` if no installation matched.
+     */
+    const validateNotificationToken = async (token: string) => {
+      const installations = await storage.botInstallation.listAll();
```
aartaria, line 91: **Per-request sequential SSM scan; misses are never cached.** On a cold cache, every inbound callback awaits `installationSecretService.get()` serially for each active installation — up to N sequential SSM `GetParameter` round-trips on the request hot path. Worse, `get()` only caches non-`null` results, so any installation without a provisioned secret (and every invalid-token request) re-reads from SSM on *every* request indefinitely.

Fix: parallelize the lookups (`Promise.all`/`Promise.any`) and negatively-cache misses with a short TTL, or invert the lookup so a token maps to an installation without scanning.

packages/client-slack/src/slackApp.ts
```diff
@@ -71,6 +76,38 @@ export async function startSlackApp(config: Config) {
...
+    // Per-installation notification secrets are stored in AWS SSM Parameter Store;
+    // the registrar generates them on first registration and the callback validator
+    // resolves them here to authenticate inbound A2A notifications.
+    const installationSecretService = new AwsSsmInstallationSecretService(
+      new SSMClient({ region: config.aws.region }),
+      config.installationSecret.ssmPrefix
+    );
+
+    /**
+     * Verify a notification token by matching it against the SSM-stored secret
```
aartaria: **Duplicate JSDoc block.** Two stacked doc comments precede `validateNotificationToken` (lines 88-91 and 92-96). The first is stale ("Resolves true on first match") and contradicts the actual return type (the matching installation or `null`). Remove the stale first block.

packages/client-slack/src/services/installationSecretService.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Promise cache so callers can invoke `getOrCreate` repeatedly without
+ * re-hitting the backend.
+ *
+ * Implementations:
+ *   - AwsSsmInstallationSecretService — see ./awsSsmInstallationSecretService.ts
+ *
+ * The same module is duplicated verbatim in client-google-chat — keep them in sync.
+ */
+
+export abstract class InstallationSecretService {
```
aartaria, line 23: **Verbatim cross-package duplication.** This module — plus `awsSsmInstallationSecretService.ts` and most of `installationRegistrar.ts`/`getServiceToken` — is duplicated byte-for-byte between client-slack and client-google-chat (the header even says "keep them in sync"). That's a standing maintenance hazard: every fix must land in two places and risks divergence. Consider extracting these into a shared package; the registrar's only client-specific part is how it enumerates tenants, which can be a callback.

packages/client-slack/src/storage/providers/postgres/pgUserAuthStorage.ts
```diff
@@ -163,4 +171,78 @@ export class PgUserAuthStorage {
     this.logger.info(`Token found for userId=${userId}, teamId=${teamId}, valid=${isValid}`);
     return isValid;
   }
+
+  /**
+   * Find a user auth record by OIDC subject identifier
+   */
+  async findByOidcSub(oidcSub: string): Promise<UserAuthToken | null> {
```
aartaria: **Dead code.** `findByOidcSub` is added to both the interface (`storage/types.ts`) and this implementation, but has no callers — the handler only uses `findByOidcSubAndTeam`. Drop it (and the interface entry) unless it's intended for imminent use; the two methods also share nearly identical row-mapping boilerplate that could fold into a private helper.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -241,3 +243,55 @@ async def get_owner_client_id(self, db: AsyncSession, channel_id: int) -> str |
...
+        if existing is None:
+            created = await self.create_channel(db=db, actor=actor, client_id=client_id, data=data)
+            return created, True
+
+        update = DeliveryChannelUpdate(
+            name=data.name,
+            description=data.description,
+            webhook_url=data.webhook_url,
+            secret=data.secret,
+            group_ids=data.group_ids,
```
aartaria: haven't we agreed to drop `group_ids`?

aartaria (commented):
Automated `/code-review` (medium effort).

PR #50, opened 2026-06-05
Fix: fail fast with actionable errors when MCP endpoint returns

zyma:
when an MCP endpoint returns HTML instead of JSON/SSE, fail fast with an actionable error instead of hanging

## Checklist

### DIFF
--- a/packages/agent-common/agent_common/agents/dynamic_agent.py
+++ b/packages/agent-common/agent_common/agents/dynamic_agent.py
@@ -46,7 +46,11 @@ from langgraph.errors import GraphInterrupt
 from langgraph.graph.state import CompiledStateGraph
 from langgraph.store.postgres.aio import AsyncPostgresStore
 from ringier_a2a_sdk.oauth import OidcOAuth2Client
-from ringier_a2a_sdk.utils.mcp_errors import format_mcp_error, is_retryable_mcp_error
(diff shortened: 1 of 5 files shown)


packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -628,7 +595,15 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
-                tools = await client.get_tools()
+                # Best-effort: surface the first connection's URL/slug to the
+                # streamable-HTTP guard so unexpected-content-type failures
+                # log with caller context. ``get_tools()`` discovers across all
+                # connections; per-server enrichment happens in the patched
+                # transport via the contextvar.
+                first_slug = next(iter(connections), None)
+                first_url = getattr(connections[first_slug], "url", None) if first_slug else None
+                async with guarded_streamable_http(url=first_url, server_slug=first_slug):
+                    tools = await client.get_tools()
```
aartaria, line 606: feels this PR is trying to fix a bug in the `mcp` which has already been addressed in https://github.com/modelcontextprotocol/python-sdk/pull/2005

Unfortunately 2005 hasn't made it yet to a stable release, though I would still tend to wait, given I don't see this issue too critical

PR #52, opened 2026-06-05
fix: reliable final-response delivery from sub-agents

zyma:
Problem: Several paths could send the client an empty completed response:

The first streaming chunk was always sent with append=True, so the A2A SDK dropped it (nonexistent artifact index) and short replies like "Hello!" never arrived.
The terminal completed status carried no message body - if a client failed to parse the streamed artifact frame, the user saw nothing.

## Checklist

### DIFF
--- a/packages/orchestrator-agent/app/core/executor.py
+++ b/packages/orchestrator-agent/app/core/executor.py
@@ -579,7 +579,9 @@ class OrchestratorDeepAgentExecutor(AgentExecutor):
 
             while True:
                 streaming_artifact_id = str(uuid.uuid4())
-                first_chunk_sent = False  # Track if we've sent the initial artifact
+                first_chunk_sent = False  # Track if we've sent the initial MAIN artifact chunk
+                first_intermediate_chunk_sent = False  # Track if we've sent the initial INTERMEDIATE artifact chunk
+                streamed_chars = 0  # Total chars streamed via the MAIN artifact (code points, not wire bytes; for completion diagnostics)
                 deferred_terminal_item = None
 
                 async for item in self.agent.stream(message_parts, user_config, config=config, resume=resume_value):
@@ -594,6 +596,11 @@ class OrchestratorDeepAgentExecutor(AgentExecutor):
                     metadata = item.metadata or {}
                     if metadata.get("streaming_chunk"):
                         is_final = True  # Doesn't matter for streaming chunks
+                        # Track MAIN-artifact bytes only (intermediate sub-agent thoughts
+                        # go to a separate "-thought" artifact and aren't part of the
+                        # main response stream the client renders).
+                        if not metadata.get("intermediate_output") and item.content:
+                            streamed_chars += len(item.content)
                     else:
                         current_state = graph.get_state(config)  # type: ignore
                         if hasattr(current_state, "interrupts") and current_state.interrupts:
@@ -601,15 +608,17 @@ class OrchestratorDeepAgentExecutor(AgentExecutor):
                         else:
                             is_final = True
 
-                    # Pass first_chunk_sent flag and update it after each chunk
-                    first_chunk_sent = await self._handle_stream_item(
+                    # Pass per-artifact first_chunk_sent flags and update after each chunk
+                    first_chunk_sent, first_intermediate_chunk_sent = await self._handle_stream_item(
                         item,
                         updater,
                         task,
                         is_final=is_final,
                         streaming_artifact_id=streaming_artifact_id,
                         first_chunk_sent=first_chunk_sent,
+                        first_intermediate_chunk_sent=first_intermediate_chunk_sent,
                         active_extensions=requested_extensions,
+                        streamed_chars=streamed_chars,
                     )
 
                 # Check for steering messages that arrived after the last abefore_model
@@ -710,14 +719,16 @@ class OrchestratorDeepAgentExecutor(AgentExecutor):
                             is_final = False
                         else:
                             is_final = True
-                    await self._handle_stream_item(
+                    first_chunk_sent, first_intermediate_chunk_sent = await self._handle_stream_item(
                         deferred_terminal_item,
                         updater,
                         task,
                         is_final=is_final,
                         streaming_artifact_id=streaming_artifact_id,
                         first_chunk_sent=first_chunk_sent,
+                        first_intermediate_chunk_sent=first_intermediate_chunk_sent,
                         active_extensions=requested_extensions,
+                        streamed_chars=streamed_chars,
                     )
                 break  # Done — no re-invocation needed
         except asyncio.CancelledError:
@@ -776,13 +787,18 @@ class OrchestratorDeepAgentExecutor(AgentExecutor):
         is_final: bool,
         streaming_artifact_id: str = "",
         first_chunk_sent: bool = False,
+        first_intermediate_chunk_sent: bool = False,
         active_extensions: set[str] | None = None,
-    ) -> bool:
+        streamed_chars: int = 0,
+    ) -> tuple[bool, bool]:
         """Handle a stream item from the agent and update the task accordingly.
 
         Streaming chunks (metadata.streaming_chunk=True) are emitted as
-        TaskArtifactUpdateEvents with append=True for ALL chunks (including first).
-        The backend accumulates all artifact-update messages into a single response.
+        TaskArtifactUpdateEvents. The FIRST chunk for a given artifact_id must
+        be a create (append=False) so the A2A SDK registers the artifact;
+        subsequent chunks use append=True. Main content and intermediate
+        (sub-agent thought) chunks use separate artifact IDs and are therefore
+        tracked independently.
         Status updates and terminal events use TaskStatusUpdateEvents.
 
         Args:
@@ -791,10 +807,11 @@ class OrchestratorDeepAgentExecutor(AgentExecutor):
             task: Current task
             is_final: Whether this is the final state
             streaming_artifact_id: Stable artifact ID for streaming chunks
-            first_chunk_sent: Whether we've already sent the first chunk
+            first_chunk_sent: Whether the first main-artifact chunk has been sent
+            first_intermediate_chunk_sent: Whether the first intermediate-artifact chunk has been sent
 
         Returns:
-            Updated first_chunk_sent flag
+            Updated (first_chunk_sent, first_intermediate_chunk_sent) tuple
         """
         # item is an AgentStreamResponse object
         state = item.state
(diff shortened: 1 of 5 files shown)


packages/orchestrator-agent/app/handlers/stream_handler.py
```diff
@@ -480,6 +532,19 @@ def parse_agent_response(final_state: Any) -> AgentStreamResponse:
                     a2a_tracking = final_state.get("a2a_tracking", {})
                     recently_called = StreamHandler._extract_recently_called_subagents(final_state)
 
+                    # GUARD: Only the agents invoked in the CURRENT turn matter for the
+                    # "all blocked" override. If no sub-agent was called this turn (e.g.
+                    # a short greeting like "hi" → "Hello!"), trust the LLM's completed
+                    # state — otherwise stale blocked entries from earlier turns in
+                    # a2a_tracking would suppress the final response.
+                    if not recently_called:
```
aartaria: **Cleanup (altitude): outer `if not recently_called` guard duplicates an invariant already inside `_check_all_agents_blocked`**

`_check_all_agents_blocked` already starts with `if not recently_called: return False, None` (line 121), so calling it with an empty set already produces `all_blocked=False` and falls through to the `completed` return anyway. The outer guard adds a special-case bypass at a different altitude. If `_check_all_agents_blocked`'s empty-set handling is ever changed, this outer guard silently swallows that change.

Consider removing the outer guard and letting `_check_all_agents_blocked` own the invariant.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -935,10 +968,25 @@ def _ext_active(uri: str) -> bool:
                     context_id=task.context_id,
                     task_id=task.id,
                 )
+                await updater.update_status(
```
aartaria, line 990: **Bug (high): HITL branch never closes an open streaming artifact**

When `first_chunk_sent=True` and a HITL interrupt fires, the artifact stream is left permanently open — no `add_artifact(..., last_chunk=True)` is called before this `update_status`. The non-HITL `else` branch (lines 999–1010) correctly seals the artifact first; this branch needs the same:
```python
                if first_chunk_sent:
                    await updater.add_artifact(
                        [Part(root=TextPart(text=""))],
                        artifact_id=streaming_artifact_id,
                        append=True,
                        last_chunk=True,
                        metadata={},
                    )
                await updater.update_status(
                    TaskState.input_required,
                    msg,
                )
```

Note: **do not** add `final=True` here — see comments on lines 1019 and 1059.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -947,39 +995,115 @@ def _ext_active(uri: str) -> bool:
...
+                        streaming_artifact_id,
+                        streamed_bytes,
+                        len(final_answer),
+                        state,
+                    )
+                    status_metadata["final_answer_source"] = "fallback"
+                await updater.update_status(
+                    TaskState.input_required,
+                    msg,
+                    final=True,
```
aartaria: **Cleanup: `final=True` is deprecated in A2A v1.x — remove it**

A2A spec PR #1308 removes `final` from `TaskStatusUpdateEvent` entirely as redundant — stream termination should be inferred from the task state, not an explicit flag. The HITL branch (line 971) already omits `final=True` and is the more forward-compatible pattern. Callers should never rely on setting `final=True` to close the stream.

The stream closes cleanly when `execute()` returns and the framework calls `queue.close()`. Remove `final=True` here for consistency with the HITL branch and the v1.x spec direction.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -947,39 +995,115 @@ def _ext_active(uri: str) -> bool:
...
+                status_metadata["final_answer_source"] = "fallback"
             await updater.update_status(
                 TaskState.auth_required,
                 new_agent_text_message(
-                    content,
+                    final_answer,
                     task.context_id,
                     task.id,
                 ),
+                final=True,
```
aartaria: **Cleanup: `final=True` is deprecated in A2A v1.x — remove it**

A2A spec PR #1308 removes `final` from `TaskStatusUpdateEvent` entirely as redundant — stream termination should be inferred from the task state, not an explicit flag. The HITL branch (line 971) already omits `final=True` and is the more forward-compatible pattern. Callers should never rely on setting `final=True` to close the stream.

Same as line 1019: remove `final=True` here.

packages/orchestrator-agent/app/handlers/stream_handler.py
```diff
@@ -365,52 +365,85 @@ def parse_agent_response(final_state: Any) -> AgentStreamResponse:
                 # - FinalResponseSchema (Bedrock structured output)
                 # - Other tools (file operations, etc.)
                 # Only extract from ToolMessages that correspond to "task" tool calls (sub-agents)
+                subagent_content = None
+                # Diagnostic: per-subagent-call info for enriched warning
+                # Entries: {"tool_call_id", "subagent_type", "has_tool_message", "extracted_length"}
+                subagent_call_diag: list[dict] = []
```
aartaria: **Cleanup: remove the diagnostic scaffolding — it's debugging code, not production logic**

`subagent_call_diag`, `diag_by_id`, the forward pass over AIMessages to collect them, and the full O(N) walk after `subagent_content` is found all exist only to produce a richer warning log line. The actual functional change — skip empty ToolMessage extractions instead of stopping at the first match — is straightforward and doesn't need any of it:
```python
for msg in reversed(current_turn_messages):
    if not isinstance(msg, ToolMessage):
        continue
    tool_call = tool_call_map.get(msg.tool_call_id)
    if not (tool_call and tool_call.get("name") == "task"):
        continue
    try:
        if isinstance(msg.content, str):
            parsed = json.loads(msg.content)
            if isinstance(parsed, dict):
                extracted = parsed.get("message", msg.content)
            elif isinstance(parsed, list):
                extracted = extract_text_from_content(parsed)[0]
            else:
                extracted = msg.content
        elif isinstance(msg.content, list):
            extracted = extract_text_from_content(msg.content)[0]
        else:
            extracted = msg.content
    except json.JSONDecodeError:
        extracted = msg.content
    if isinstance(extracted, str) and extracted:
        subagent_content = extracted
        logger.info(...)
        break
```

packages/orchestrator-agent/app/handlers/stream_handler.py
```diff
@@ -365,52 +365,85 @@ def parse_agent_response(final_state: Any) -> AgentStreamResponse:
...
+                                diag_entry["has_tool_message"] = True
+
+                            # Extract content (sub-agent content may be JSON-wrapped).
+                            try:
+                                if isinstance(msg.content, str):
+                                    parsed_content = json.loads(msg.content)
+                                    if isinstance(parsed_content, dict):
+                                        extracted = parsed_content.get("message", msg.content)
+                                    elif isinstance(parsed_content, list):
+                                        extracted = extract_text_from_content(parsed_content)[0]
```
aartaria: **Bug (latent): `IndexError` from `[0]` not caught by the `json.JSONDecodeError` handler**

The `try/except` only catches `json.JSONDecodeError`. If `extract_text_from_content` ever returns an empty sequence the `[0]` raises `IndexError`, propagating uncaught and aborting stream_handler with a 500. Same issue at line 427.

```python
extracted = (extract_text_from_content(parsed_content) or [""])[0]
```

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -947,39 +995,115 @@ def _ext_active(uri: str) -> bool:
                 } or None
                 if interrupt_metadata:
                     msg.metadata = interrupt_metadata
-            await updater.update_status(
-                TaskState.input_required,
-                msg,
-            )
+                status_metadata = dict(metadata) if metadata else {}
+                if first_chunk_sent:
```
aartaria: **Cleanup: `input_required` and `auth_required` terminal branches duplicate ~20 lines verbatim**

Both share the same pattern: copy metadata → optional artifact-close + log → set `final_answer_source: fallback` → `update_status`. When the log format or key name changes, one branch will be missed.

Consider a `_close_streaming_artifact_and_respond(...)` helper.

packages/orchestrator-agent/app/handlers/stream_handler.py
aartaria: **Cleanup (altitude): outer `if not recently_called` guard duplicates an invariant already inside `_check_all_agents_blocked`**

`_check_all_agents_blocked` already returns `(False, None)` for an empty set (line 121), so the outer guard is redundant. If that inner behaviour ever changes, this outer guard silently swallows it. Consider removing it and letting `_check_all_agents_blocked` own the invariant.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -616,22 +618,29 @@ async def execute(
                     metadata = item.metadata or {}
                     if metadata.get("streaming_chunk"):
                         is_final = True  # Doesn't matter for streaming chunks
+                        # Track MAIN-artifact bytes only (intermediate sub-agent thoughts
+                        # go to a separate "-thought" artifact and aren't part of the
+                        # main response stream the client renders).
+                        if not metadata.get("intermediate_output") and item.content:
+                            streamed_bytes += len(item.content)
```
aartaria: **Cleanup: `streamed_bytes` counts Unicode code-points, not UTF-8 bytes**

`len(item.content)` returns character count. For CJK or emoji content each character is 3–4 wire bytes. Rename to `streamed_chars` or use `len(item.content.encode())`.

aartaria (changes requested):

packages/orchestrator-agent/app/core/executor.py
zyma: Removed

packages/orchestrator-agent/app/core/executor.py
zyma: Cleared

packages/orchestrator-agent/app/core/executor.py
zyma: Cleared

packages/orchestrator-agent/app/handlers/stream_handler.py
zyma: Cleared

packages/orchestrator-agent/app/handlers/stream_handler.py
zyma: Guarded all 5 sites  with    ... or [""])[0] to kill the uncaught IndexError.

packages/orchestrator-agent/app/core/executor.py
zyma: 1. Added a new helper method — _close_streaming_artifact_and_respond(...)
2. Replaced three call sites to use that helper instead of repeating the block:

packages/orchestrator-agent/app/handlers/stream_handler.py
zyma: removed the redundant outer if not recently_called: guard

packages/orchestrator-agent/app/core/executor.py
zyma: Renamed on streamed_chars

aartaria (approved):

aartaria (approved):

PR #54, opened 2026-06-08
fix: Improve voice agent first reply latency

jalmari-sor:
Several fixes to the voice agent:
Adds risk score for the tooling and verbal confirmation to prevent unwanted actions.
Adds initial wake up text to have a faster first response from the agent.
Refactor initialization order to properly warmup the gemini instance before calling and simplify session keying.

## Checklist

### DIFF
--- a/packages/voice-agent/voice_agent/a2a_agent.py
+++ b/packages/voice-agent/voice_agent/a2a_agent.py
@@ -394,12 +394,19 @@ class VoiceAgent(BaseAgent):
                 )
 
         voice_name = self._active_sessions[session_key]["agent"].voice_name
+        audio_in = self._active_sessions[session_key]["audio_in"]
         event_out = self._active_sessions[session_key]["event_out"]
         yield AgentStreamResponse(
             state=TaskState.TASK_STATE_WORKING,
             content="Voice session initialized",
             metadata={"session_id": session_key, "voice_name": voice_name},
         )
+
+        # Trigger the opening greeting immediately by injecting a text turn.
+        # Without this, Gemini waits silently for user audio to clear VAD,
+        # causing 3-4 s of dead air after the callee answers.
+        await audio_in.put("[call_connected]")
+        logger.debug("Injected greeting trigger for session %s", session_key)
 
         try:
             while True:
--- a/packages/voice-agent/voice_agent/a2a_agent.py
+++ b/packages/voice-agent/voice_agent/a2a_agent.py
@@ -79,6 +79,7 @@ from ringier_a2a_sdk.utils.a2a_part_conversion import a2a_parts_to_content
 from voice_agent.agent import SYSTEM_PROMPT as DEFAULT_SYSTEM_PROMPT
 from voice_agent.agent import GeminiLiveAgent
 from voice_agent.call_bridge import (
+    _CALL_ANSWERED,
     _CALL_FUTURES,
     _PENDING_CALLS,
     OutboundCallRequest,
@@ -659,7 +660,8 @@ class VoiceAgent(BaseAgent):
           8. Yield ``completed`` with the formatted transcript.
         """
         public_url: str = os.environ.get("PUBLIC_URL") or os.environ.get("VOICE_AGENT_BASE_URL", "")
-        timeout: int = int(os.getenv("CALL_TIMEOUT_SECONDS", "60"))
+        connect_timeout: int = int(os.getenv("CALL_CONNECT_TIMEOUT_SECONDS", "60"))
+        duration_timeout: int = int(os.getenv("CALL_DURATION_TIMEOUT_SECONDS", "3600"))
 
         # Validate PUBLIC_URL — Twilio must be able to reach /twilio/voice via this URL.
         # localhost / 127.0.0.1 will cause Twilio to fail silently when the call is answered.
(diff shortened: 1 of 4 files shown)


packages/voice-agent/voice_agent/a2a_agent.py
```diff
@@ -672,34 +686,62 @@ async def _stream_phone_call(
...
-                voice_name=voice_name,
-                mcp_tools=mcp_tools,
-                access_token=access_token,
-                phone_number=phone_number,
-            )
-        )
+        # Rename prewarm entries from the temporary key to the real call_sid
+        # so the rest of the pipeline (MCP check, twilio_stream, _end_session)
+        # can look up the session by call_sid as normal.
+        if prewarm_key in self._active_sessions:
```
aartaria: **Bug: race condition — prewarm session may not be re-keyed in time**

`_prewarm_audio_session` runs concurrently and writes to `_active_sessions[prewarm_key]` only after its own `await` chain completes. This re-key check executes right after `make_outbound_call()` returns — with no synchronisation between the two. When the task has not yet registered the session, the check is `False` and `_active_sessions[call_sid]` is never populated.

When `twilio_stream` later calls `_start_audio_session(call_sid)`, the session is not found, a second cold-start session is created, and the prewarmed session under `prewarm_key` is never cleaned up (leaking a live Gemini WebSocket connection).

**Fix:** eliminate the rename entirely — pass `call_sid` to `_prewarm_audio_session` only after `make_outbound_call()` returns (i.e. the two operations are still concurrent, but the task is created *after* call_sid is known and keyed directly under it).

packages/voice-agent/voice_agent/a2a_agent.py
```diff
@@ -672,34 +686,62 @@ async def _stream_phone_call(
...
+        # not yet known; we rename both dicts to call_sid once the call is up.
+        # asyncio.create_task inherits the current context so contextvars
+        # (user token etc.) are available inside _prewarm_audio_session.
+
+        # Merge context_messages into system_prompt here so the Gemini session
+        # is created with the full prompt regardless of whether the call is
+        # answered before or after prewarm completes (prewarmed path ignores
+        # init_config in _start_audio_session).
+        effective_prompt = system_prompt or DEFAULT_SYSTEM_PROMPT
+        if context_messages:
```
aartaria: **Cleanup: context_messages → system_prompt merging is duplicated**

The same `'\n'.join(context_messages)` + f-string pattern appears here (lines 701–703) and again in `twilio_transport.py` lines 159–162. Any change to the merge format (adding a header, escaping, trimming) must be applied in both places. Consider extracting a shared helper, e.g.:

```python
def _build_effective_prompt(base: str, context_messages: list[str] | None) -> str:
    if not context_messages:
        return base
    return f"{base}\n\nContext:\n{'\n'.join(context_messages)}"
```

aartaria (changes requested):

packages/voice-agent/voice_agent/agent.py
```diff
@@ -218,8 +257,27 @@ async def _init_mcp_tools(
...
+                        # First call — gate it: ask Gemini to get verbal confirmation.
+                        _awaiting_confirmation.add(_name)
+                        logger.info("Write tool %r intercepted — requesting confirmation", _name)
+                        return (
+                            f"CONFIRMATION_REQUIRED: Before calling {_name}, tell the user "
+                            f"exactly what you are about to do and ask 'Shall I proceed?' "
+                            f"Only call {_name} again once they have clearly said yes."
+                        )
+                    # Second call — user has confirmed; execute and clear the gate.
+                    _awaiting_confirmation.discard(_name)
```
aartaria: what if you deny? I believe there are several edge cases, two that come to my mind

* If Gemini retries the tool call with other arguments it would just go through
* Assuming Gemini would not immediately re-attempt the tool call, `_awaiting_confirmation` would though never be cleared and during the same conversation conversation if it would try again with other arguments it would just go through.

aartaria (changes requested):

packages/voice-agent/voice_agent/agent.py
```diff
@@ -121,6 +120,35 @@ def build_gemini_client() -> genai.Client:
...
+)
+
+# Keywords that indicate a tool has write/mutate side-effects.
+_WRITE_KEYWORDS = frozenset(
+    {
+        "create", "update", "delete", "remove", "post", "write", "send", "push",
+        "modify", "edit", "close", "merge", "publish", "commit", "open", "add",
+        "insert", "patch", "put", "submit", "deploy", "release",
+    }
+)
```
aartaria, line 166: we could in principle extend it a bit to be closer to the orchestrator behaviour: e.g. a lighter/adapted version of  [`score_tool_risk`](https://github.com/ringier-data/nannos/blob/2b4d90456ba8b9218737831ea98b4083c9e94d71/packages/agent-common/agent_common/core/tool_risk_scorer.py#L93). I would still keep the keyword fallback (which is present in `score_tool_risk` too) since we score tools just in time, so many tool calls might not have a cache hit.

packages/voice-agent/voice_agent/agent.py
```diff
@@ -449,6 +480,77 @@ async def _exec(
...
+        # These are registered as FunctionDeclarations alongside the MCP tools
+        # but are handled entirely in-process — no MCP round-trip, no risk scoring.
+
+        async def search_stored_result(args: dict) -> str:
+            result_id = args.get("result_id", "")
+            pattern = args.get("pattern", "")
+            text = _stored_results.get(result_id)
+            if text is None:
+                return f"No stored result with id '{result_id}'."
+            lines = text.splitlines()
```
aartaria, line 493: what if it is a single really long line? e.g. single line json?

aartaria (approved):
just tried the voice agent... works way better than before ... I just rebased and resolved conflicts to make it work...

I'm just curious about two things

* don't we need `behavior=types.Behavior.NON_BLOCKING` ? in fact I didn't felt too bad even if stopped for a couple of seconds before answering...
* I asked like 10 times to create an issue and it always asked confirmation but I just saw once the log
```
{"time": "2026-06-16T15:47:28.743Z", "level": "INFO", "name": "voice_agent.agent", "message": "Tool call: github_issue_write({'repo': 'rcplus-alloy-infrastructure-agents', 'method': 'create', 'title': 'Risky Actions are not surfacing', 'owner': 'alloy-ch'}) (turn=10)"}
{"time": "2026-06-16T15:47:28.746Z", "level": "INFO", "name": "voice_agent.agent", "message": "Risky tool 'github_issue_write' intercepted \u2014 requesting confirmation"}
{"time": "2026-06-16T15:47:28.747Z", "level": "INFO", "name": "voice_agent.agent", "message": "Tool 'github_issue_write' response ready (scheduling=INTERRUPT)"}
{"time": "2026-06-16T15:47:28.756Z", "level": "INFO", "name": "voice_agent.agent", "message": "Gemini: turn 10 complete"}
```

is it really the deterministic gate which block the tool call or is the instructions? If would be gate + instructions I would expect to be asked twice, or am I not seeing it the right way?

PR #68, opened 2026-06-16
fix(console-backend): stop conversation reload from duplicating/splitting the answer

aartaria:
Fixes two distinct persistence-layer bugs that made a turn render wrong **on reload** (live was fine). Both are follow-ups to #67.

`_process_a2a_response` persists a turn via two paths: the accumulated streaming-artifact buffer (`insert_message`) and standalone messages (`save_agent_response`). Two defects in that split:

## 1. Terminal fallback persisted as a duplicate
The orchestrator streams the answer (persisted from the buffer) **and** re-sends it in the terminal status as a fallback for non-streaming clients, tagged `final_answer_source: "fallback"`. The skip-conditions never checked the flag, so both were stored → the answer replayed **twice** on reload.

**Fix:** skip `save_agent_response` for terminal status-updates tagged `final_answer_source: "fallback"` (the streamed copy is already persisted). Non-streamed answers aren't tagged, so they're still persisted.

## 2. First artifact chunk persisted standalone (content split)
The first chunk of a streamed artifact is `append=False` (it creates the artifact). Accumulation was gated on `is_artifact_append` (append=True only), so the first chunk was skipped from the buffer **and** fell through to `save_agent_response` as its own message. The content was stored as two messages — standalone first chunk + accumulated remainder — which reload rendered as **two fragments** (e.g. an orchestrator thought split mid-sentence: "The user wants me to delegate" | "a poem generation task…", or a truncated leading bubble).

**Fix:** gate accumulation and the standalone-save skip on `is_artifact_update` (any append) so the first chunk is assembled with the rest into one message. Live streaming is unchanged (first chunk still emits via the normal path).

`#67` exposed #2 by routing the orchestrator's prose to the thinking channel, turning it into a persisted intermediate-output thought.

## Tests
`tests/test_socket_message_handling.py` — 3 new cases: fallback terminal is **not** persisted; non-fallback terminal **is**; the `append=False` first chunk is accumulated, not saved standalone. Full file passes (10 tests).

## Note
Both are symptoms of the same architectural seam (dual emission + per-path persistence, plus separate live vs. reload reconstruction). Tracked separately: single-source emission and single-source (frontend-owned) reconstruction would eliminate the class.

### DIFF
--- a/packages/console-backend/app.py
+++ b/packages/console-backend/app.py
@@ -706,7 +706,8 @@ async def _process_a2a_response(
                 message_extensions = status_message.get("extensions", []) if isinstance(status_message, dict) else []
                 is_work_plan = "urn:nannos:a2a:work-plan:1.0" in message_extensions
                 is_feedback_request = "urn:nannos:a2a:feedback-request:1.0" in message_extensions
-                is_artifact_append = response_data.get("kind") == "artifact-update" and response_data.get("append")
+                is_artifact_update = response_data.get("kind") == "artifact-update"
+                is_artifact_append = is_artifact_update and response_data.get("append")
                 # Handle both camelCase and snake_case, check explicitly for boolean value
                 # Don't use 'or' because False would fallback to checking second field
                 last_chunk_value = response_data.get("lastChunk")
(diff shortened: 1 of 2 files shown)


aartaria:
Parking in favor of the architectural fix (not in prod, so no need to ship these model-A interim patches). The first-chunk accumulation here is a patch on the backend-assembles model we're replacing, and the terminal-fallback dedup is better handled by single-source emission. Superseded by the single-source streaming design (doc PR incoming).

PR #101, opened 2026-06-29
feat: incoming calls

jalmari-sor:
closes #

## Checklist


packages/console-backend/console_backend/services/voice_session_service.py
```diff
@@ -0,0 +1,189 @@
...
+            result = await db.execute(text("""
+                UPDATE voice_sessions SET gemini_session_handle = :handle, updated_at = :now
+                WHERE id = :id
+                RETURNING id
+            """), {"id": session_id, "handle": gemini_session_handle, "now": datetime.now(timezone.utc)})
+            found = result.mappings().first() is not None
+            await db.commit()
+            if not found:
+                logger.warning("update_handle: no voice session found with id=%s", session_id)
+            return True
```
aartaria: **[correctness] `update_handle` always returns `True`, even when the session doesn't exist.**

Line 153 computes `found`, line 155-156 logs a warning when not found, then line 157 unconditionally `return True`. The router does `if not ok: raise HTTPException(500)`, so a PATCH for a stale/invalid `session_id` returns **200 OK** and the Gemini resumption handle is silently never persisted.

The computed `found` is dead — this looks like it should be `return found`. (`complete_session`/`fail_session` share the always-`True` shape but don't compute rowcount at all, so they're a weaker variant of the same issue.)

🤖 automated review

packages/voice-agent/voice_agent/a2a_agent.py
```diff
@@ -476,6 +490,13 @@ async def _start_audio_session(
                         content="Tool authorization required — SMS sent with instructions.",
                         metadata={"type": "mcp_auth_failed", "authorize_url": authorize_url},
                     )
+                elif event_type == "session_resumption_handle":
+                    handle = event.get("handle", "")
+                    voice_session_id = self._active_sessions.get(session_key, {}).get("voice_session_id")
+                    if voice_session_id and handle:
+                        from voice_agent.console_client import update_session_handle  # noqa: PLC0415
+                        asyncio.create_task(update_session_handle(voice_session_id, handle))
```
aartaria: **[correctness] Fire-and-forget `create_task` result is not referenced — may be garbage-collected mid-flight.**

Per the CPython docs, the event loop keeps only a *weak* reference to a bare `asyncio.create_task(...)`, so an unreferenced task can be collected before it completes — silently dropping the handle update. This compounds the `update_handle` issue: the handle-persistence path can fail with no error anywhere. (Same pattern at `twilio_transport.py:354` for `complete_voice_session`.)

Keep a reference, e.g. store in a `set` with a `task.add_done_callback(set.discard)`.

🤖 automated review

packages/voice-agent/voice_agent/inbound.py
```diff
@@ -0,0 +1,487 @@
...
+
+    # Store interim state
+    state = InboundCallState(
+        call_sid=CallSid,
+        user=user,
+        from_number=from_number,
+        sub_agents=sub_agents,
+        digit_map=digit_map,
+    )
+    _INBOUND_PENDING[CallSid] = state
```
aartaria, line 297: **[correctness / resource leak] `_INBOUND_PENDING` grows unbounded.**

This entry is added on every inbound webhook, but it's only removed by `pop_inbound_state` on the Media Streams `"start"` event. A caller who is offered the resume/menu prompt and hangs up (or never presses a digit, or drops after the `<Connect><Stream>` TwiML but before the WS connects) leaves the `InboundCallState` in the dict forever. There's no TTL/eviction → unbounded memory growth under repeated abandoned/spam calls.

Consider a TTL-based eviction or cleanup on the reject/expiry paths.

🤖 automated review

packages/console-backend/console_backend/services/voice_session_service.py
```diff
@@ -0,0 +1,189 @@
...
+        user_id: str,
+        phone_number: str,
+        sub_agent_id: int | None = None,
+        call_sid: str | None = None,
+        use_session_memory: bool = False,
+    ) -> VoiceSession | None:
+        try:
+            now = datetime.now(timezone.utc)
+            session_id = str(uuid.uuid4())
+            query = text("""
```
aartaria: **[convention — CRITICAL] Direct SQL writes bypass the repository/audit pattern.**

All four mutations in this service (`create_session` here, `update_handle` L148, `complete_session` L166, `fail_session` L180) use raw `db.execute(text("INSERT/UPDATE ..."))`. `packages/console-backend/AGENTS.md:61` states:

> **CRITICAL: All database write operations (INSERT/UPDATE/DELETE) MUST use the repository pattern to ensure automatic audit logging.**

and L492: *"Never bypass the repository pattern for data mutations."* These writes produce no audit log. The `AuditedRepository` base is used by ~10 existing repositories — voice_sessions mutations should go through one too.

🤖 automated review

packages/voice-agent/voice_agent/inbound.py
```diff
@@ -0,0 +1,487 @@
...
+
+    # Offer to resume the most recent session (<1h, any agent) BEFORE the picker.
+    try:
+        resumable = await get_latest_resumable_session(user.id)
```
aartaria: **[efficiency] Independent backend calls in the inbound webhook run sequentially.**

`list_sub_agents_for_menu` (L248) and `get_latest_resumable_session` (here, L275) both depend only on `user.id`, yet are awaited one after the other. `asyncio.gather` would remove a full backend round-trip from the webhook's latency — directly felt as dead air before the caller hears the menu.

🤖 automated review

packages/voice-agent/voice_agent/console_client.py
```diff
@@ -0,0 +1,249 @@
...
+
+_CONSOLE_BACKEND_URL = os.getenv("CONSOLE_BACKEND_URL", "http://localhost:5001")
+_OIDC_ISSUER = os.getenv("OIDC_ISSUER", "")
+_CLIENT_ID = os.getenv("OIDC_CLIENT_ID", "voice-agent")
+_CLIENT_SECRET = os.getenv("OIDC_CLIENT_SECRET", "")
+
+_TOKEN_MARGIN_SECONDS = 60
+
+
+class _TokenCache:
```
aartaria: **[reuse] Re-implements client-credentials token fetch + caching that the SDK already provides.**

`_TokenCache` (monotonic expiry, margin, lock) + `_fetch_client_credentials_token` hand-roll what `ringier_a2a_sdk.oauth.client.OidcOAuth2Client` already does — and that client is already used in this same package (`a2a_agent.py:192`). Reusing it removes the duplicated caching/expiry logic and the drift risk between two service-token paths.

🤖 automated review

aartaria (commented):
🤖 **Automated code review by Claude** (`/code-review`, high effort)

Review of the incoming-calls / voice-agent feature against `main`. 7 findings posted inline below, ranked by severity (3 correctness, 1 convention, 3 efficiency/reuse). These are suggestions from an automated pass — please use judgement; some may be intentional trade-offs.

_Generated with [Claude Code](https://claude.com/claude-code)_

packages/voice-agent/voice_agent/console_client.py
```diff
@@ -0,0 +1,249 @@
...
+    sub_agent_id: int | None = None
+    gemini_session_handle: str | None = None
+    use_session_memory: bool = False
+
+
+async def lookup_user_by_phone(phone_number: str) -> UserInfo | None:
+    """Return user info for a phone number, or None if not registered."""
+    headers = await _service_headers()
+    url = f"{_CONSOLE_BACKEND_URL}/api/v1/voice/users/by-phone/{phone_number}"
+    async with httpx.AsyncClient(timeout=10.0) as client:
```
aartaria: **[efficiency] New `httpx.AsyncClient` per call — no connection pooling.**

Every function here opens `async with httpx.AsyncClient()`, paying connection setup + TLS handshake on each request (an inbound call makes 3+ sequential calls). A module-level shared client — the pattern `soffice_client.py` already uses — reuses connections.

🤖 automated review

PR #104, opened 2026-07-01
fix(agent-common): attribute in-process sub-agent LLM costs to the sub-agent

aartaria:
## Problem

Local (in-process) sub-agents' gateway spend logs are **misattributed to the orchestrator** — `sub_agent_id` is missing in the litellm `LiteLLM_SpendLogs` metadata for any interactive orchestrator→sub-agent turn. Reproduced with task-scheduler and confirmed generic across agents (general-purpose, agent-creator, etc.).

## Root cause

There are two cost-attribution sinks:

1. **App-side `CostLogger`** — reads `sub_agent_id` from the `sub_agent:{id}` **LangGraph tag** added by `LocalA2ARunnable._instrument()`. ✅ Works.
2. **LiteLLM gateway spend logs** — the proxy reads `sub_agent_id` from the `x-litellm-spend-logs-metadata` header, which the shared httpx hook stamps from the `current_sub_agent_id` **ContextVar** (`ringier_a2a_sdk.cost_tracking.attribution`). ❌ Broken.

The executor sets the attribution ContextVars once per orchestrator turn, with `sub_agent_id` left unset (the orchestrator has none). When the orchestrator dispatches into a **local** sub-agent, the sub-agent's LLM calls run in-process on that same context and nobody ever updates `current_sub_agent_id` — so those tokens reach the gateway tagged as the orchestrator. (The scheduled-job path was fine because it sets attribution explicitly before invoking.)

## Fix

`LocalA2ARunnable.astream()` — the single chokepoint all local sub-agents (and `ainvoke`, which delegates to it) flow through — now sets `current_sub_agent_id` to the agent's own `sub_agent_id` for the duration of the run and restores the caller's value in `finally`.

- Restore uses `set()` rather than `token.reset()`: `astream()` is an async generator and a `Token` created in one context can't be reset from another after a `yield`.
- Built-in agents with no integer `sub_agent_id` (e.g. file-analyzer) leave the caller's attribution untouched.
- Leak-safe and nests correctly (a sub-agent invoked while another's id is active restores that outer id, not blanket-clears it).

Confirmed end-to-end that the proxy reads `sub_agent_id` from this header (`packages/litellm-proxy/custom_logger.py`).

## Tests

Adds 3 regression tests in `test_local_a2a_runnable.py`: sets-during-run/restores-after, nested-dispatch restore, and no-int-id no-op. Full agent-common suite passes (653).

### DIFF
--- a/packages/agent-common/agent_common/a2a/base.py
+++ b/packages/agent-common/agent_common/a2a/base.py
@@ -937,6 +937,25 @@ class LocalA2ARunnable(CostTrackingMixin, BaseA2ARunnable):
         Raises:
             NotImplementedError: If subclass doesn't implement _astream_impl()
         """
+        # Gateway cost-attribution for in-process sub-agents.
+        # A local sub-agent's LLM calls reach the Model Gateway on the *caller's*
+        # attribution context (the orchestrator turn, where sub_agent_id is unset),
+        # so their litellm spend logs would be misattributed to the orchestrator.
+        # The app-side CostLogger already gets sub_agent_id from the ``sub_agent:``
+        # LangGraph tag (see _instrument), but the gateway's
+        # ``x-litellm-spend-logs-metadata`` header is stamped from the
+        # ``current_sub_agent_id`` ContextVar, which nothing updates when the
+        # orchestrator dispatches into a local sub-agent. Set it to this agent's own
+        # id for the duration of the run, then restore the caller's value so the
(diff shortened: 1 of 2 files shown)


aartaria:
Superseded by #105, which fixes the same bug at a more robust layer (a `wrap_model_call` middleware that derives attribution from each model call's own tags, so there's no invocation boundary to forget) and includes the follow-up cleanup (collapsing the duplicate `current_sub_agent_id` ContextVar). Closing in favor of #105.

PR #128, opened 2026-08-06
test(orchestrator-agent): enable and standardize the integration test suite

AlexMetsai:
Closes the Framework Evaluation, Standardization, Documentation, Observability and
Backfilling requirements of "Improve and Enable Nannos Integration Tests", and makes
a start on the test-rigor feedback from #202.

## The suite was off

`tests/integration/` was excluded with `--ignore`, which hid the directory entirely.
An a2a-sdk v1.1 migration had broken an import in `test_a2a_streaming.py` and nothing
noticed for months. It is now collected on every run and deselected by marker
(`-m "not integration"` in addopts), so breakage surfaces immediately while ordinary
runs stay fast. The fixtures were also rewritten to be gateway-native: models are
discovered from `/v1/model/info` rather than hardcoded per provider.

## Two tiers, one assertion vocabulary

`tests/support/` holds the shared vocabulary. `extraction.py` reads turn outcomes
through `current_turn_messages`, which this PR promotes from
`StreamHandler._extract_current_turn_messages` to a public function in `app.handlers`
— so tests and production call one implementation and assertions cannot drift from
what the orchestrator actually does.

Scenarios live in `tests/datasets/` and drive both tiers through the same
`assert_scenario`:

- **mock tier** (`test_scenarios_mock.py`) — the model is scripted to behave
  perfectly, so a failure means the *scenario* is wrong. Milliseconds, every PR.
- **real tier** (`test_scenarios_real.py`) — nothing scripted; the model routes on
  the sub-agent descriptions. The only tier that can answer "does routing work".

10 scenarios in two datasets, split by subject so the real tier's per-scenario spend
stays separable: `core_routing.yaml` (5 routing, 2 attachment) and
`failure_propagation.yaml` (3).

`scripted_model.py` exists because every stock LangChain fake inherits a `bind_tools`
that raises `NotImplementedError`. It records bound tools, which turned it into an
asset rather than a workaround.

## Observability and gating

Per-test wall-clock and token counts, aggregate pass ratio, and total cost, printed
as a terminal report. Gating is statistical, not retry-based: each test runs **once**
and only the ratio is enforced (default 0.75, `EVAL_MIN_PASS_RATIO`). Individual
failures stay visible instead of being silently retried away.

**Baseline, measured 2026-09-10 on the merged branch:** **98% (50/51)**, 571s,
1,974,054 tokens, against a 75% threshold. Gate passed. Mock tier alongside it: 1023
passed in 11.9s.

Read it with three caveats:

- **Two providers, not five.** The local gateway currently exposes only
  `bedrock_converse` and `vertex_ai`, so `one_model_per_provider()` picks
  `claude-sonnet-4-6` and `gemini-3-flash-preview`. Azure is gone. That makes this
  **not comparable** to the older 95% (39/41) figure, and it makes the ratio noisier
  than the threshold assumes: one flaky scenario moves it ~2 points instead of ~1.
- **51 judged, not 53 collected.** The two `test_static_tools_bound_to_graph` tests
  make no model call and are excluded from judging, so the denominator is not the
  test count.
- **One test is 40% of the bill, and it is the failing one.**
  `reports_subagent_failure_as_failure` on gemini spent **782,577 tokens over 254s** —
  more than the other 50 tests combined spend on either provider. A per-test token
  budget is the obvious follow-up: the ratio gate says nothing about spend, and a
  looping test is both the most expensive and the least informative kind of failure.

**The failure set turned over completely between the two runs**, which is itself the
most useful thing this baseline says.

Fixed since 2026-09-07:

- `relays_subagent_request_for_input[claude-sonnet-4-6]` — **my scenario bug, fixed.**
  The query said only "post the quarterly summary" without supplying one, so sonnet
  correctly ran `ls`, found nothing, and asked the user rather than delegating. It
  still reached `input_required`, by a route the scenario was not testing; gemini
  delegated and passed. The dataset was accidentally measuring how literally a model
  reads an ambiguous request. The query now carries the content inline, plus an
  `instructions` needle so a pass cannot come from delegating an empty message. Passes
  on both providers.
- `test_multiturn_context_preservation[gemini-3-flash-preview]` — **did not
  reproduce.** Previously gemini misrouted "what is my name?" to `file-analyzer`,
  whose `input_required` response correctly tripped the completion guard. This run it
  passed. That is the second single-observation gemini routing failure in this suite
  not to recur (the "2 + 2" one below is the first), which is a caveat on the whole
  category: with two providers and one sample each, a routing failure seen once is
  weak evidence. Tracing it did surface two real contract defects in
  `_check_all_agents_blocked` — nothing reads `TASK_STATE_WORKING`, and `is_complete`
  defaults to `True` under a comment reading "Defensive: assume incomplete if missing"
  — filed as #227. Neither is reachable through any sub-agent this repo ships, and
  neither caused this failure.
- **`forwards_attachment_to_subagent[gemini]` stopped looping.** It was 607,515 tokens
  over 192.3s, 28% of the previous bill; it is now **50,281 over 27.0s**, a 12x drop.
  Most likely #215, which stopped rate-limited tool calls burning a sub-agent's
  recursion budget. Not investigated further, but it is the clearest evidence here that
  the real tier tracks production behaviour rather than its own fixtures.

New, and a genuine product finding:

- `reports_subagent_failure_as_failure[gemini-3-flash-preview]` — **reproduced.** The
  scenario fails `revenue-analyst` with "query timed out after 30s against the
  warehouse" and expects the turn to report `failed`. gemini instead retried the same
  failing sub-agent, then escalated to `general-purpose`, then produced nothing at
  all. The watchdog aborted at a 60s inter-chunk timeout, auto-resumed once, and gave
  up at 180s.

  The retry is defensible; the silence is not. sonnet passes the same scenario in 9s.

  Confirmed rather than assumed, given the two non-reproducing failures above. Run
  twice back to back: **254.0s / 782,577 tokens, then 254.8s / 410,414 tokens.** The
  tokens halved and the duration did not move — because 254s is not work, it is
  `60 + 180` seconds of watchdog plus ~15s of turn. That also makes a second point
  worth its own look: *any* stalled turn costs a user four minutes before they are
  told anything.

  The second run additionally hit `[ERROR_TYPE: system_error]` on
  `write_todos(status="failed")` — a status the orchestrator's own prompt asks about
  (`config.py:584`) and the A2A vocabulary maps (`TODO_STATE_MAP`), but langchain's
  todo schema rejects. Run 1 stalled without reaching it, so it is a separate finding,
  not the cause. Filed as #230 (todo status) and #231 (the stall).

  This is what the real tier is for. The mock tier passes this scenario, because a
  scripted model does the sensible thing by construction. Left failing rather than
  xfailed.

## Test rigor

The reviewer's point on #202 — that a test can be "too mocked" to prove anything —
had an instance in this repo, in the same subject area. `test_hitl_reject_*` built
its own `StateGraph`, docstring conceding it *"replicates the topology from
langchain.agents.factory"*, including its own router. That router **was** the
protection under test: a rejected tool call is not removed from the AIMessage
(langchain's `_process_decision` returns the tool call, not `None`) but answered with
a synthetic error `ToolMessage`, so non-execution depends entirely on a router
skipping already-answered calls. The test supplied that router and proved its own
copy worked.

Measured rather than argued: disabling the orchestrator's HITL guard, so **every
guarded tool executes with no approval**, left all 796 pre-existing tests green. The
rewrite fails 7 of its 8.

It now runs through `scripted_graph()` — real `GraphFactory`, real middleware stack,
real `ConditionalHumanInTheLoopMiddleware`, real router, only the model scripted. Net
−237 lines. That also showed why the old tests could never have caught it: their tool
names, `dangerous_tool` and `safe_tool`, both score **0.7**, under the 0.8 threshold,
so neither is guarded under risk scoring at all — they worked only via static
`interrupt_on` config. The new tools are named for the scores they get, and one test
pins the scoring so the rest cannot drift onto a different guarded set.

`AGENTS.md` gains the rule this came from, with that measurement as its worked
example: **a test must not replace the system whose behaviour it asserts.** Plus the
pin-plus-recorded-evidence pattern #202 itself demonstrates, for claims only a real
provider can settle.

## Production bugs this found

- **`MAX_RECURSION_LIMIT` was truncating ordinary turns.** The budget counts
  LangGraph super-steps, not model calls, and every middleware hook is its own node —
  measured at `3 + 8 × model_calls`, so 50 allowed six model calls. Live traces showed
  the orchestrator completing a two-delegation request correctly, then having the
  answer discarded and replaced with the "shall I continue?" prompt.

  **Fixed on `main`, not here.** The reviewer asked for it as its own PR, so it went
  out as #210 and is no longer in this diff: the limit is now derived from the compiled
  graph by classifying node names, rather than pinned to a hand-maintained multiplier
  that goes stale the moment a middleware with model hooks is added. #216 tracks the
  same bug in `agent-runner`, `agent-common` and `ringier-a2a-sdk`, where it is worse —
  a sub-agent that hits its limit returns a partial result with no signal that it was
  truncated.
- **`.gitignore` was fail-open.** It enumerated specific `.env` names, so
  `.env.integration` was tracked — in a public repo. Replaced with a pattern that
  ignores all variants and re-includes templates. History verified: nothing was ever
  committed.
- **`start-local.sh` bind-mounted `/tmp` paths**, which on macOS silently creates an
  empty directory inside the Docker VM instead of mounting the file. Killed the proxy,
  and broke Vertex auth with nothing in the logs.
- **Three a2a-sdk v1.1 misses**, invisible while the directory was ignored.
- **`ringier-a2a-sdk` documented an env var nothing reads.** Five docstrings across
  `langgraph.py`, `langgraph_anthropic.py`, `langgraph_bedrock.py` and
  `langgraph_google.py` told operators to configure the recursion limit with
  `LANGGRAPH_RECURSION_LIMIT`; `_get_default_recursion_limit()` reads
  `MAX_RECURSION_LIMIT`. Anyone following the docstrings set a variable with no
  effect and silently kept the default of 50 — the same number, in the same wrong
  unit, that truncated orchestrator turns above. Corrected here, since it is a
  five-line `sed` found while auditing the same subject; `_get_default_recursion_limit`
  also gains a note that the name is shared across three packages with different
  defaults. #229 phase 1 deletes three of those files anyway.

## Review feedback addressed

- **Sub-agent names in scenarios were architecturally wrong.** They delegated to
  `agent-runner` and `slack-client`; the orchestrator can reach neither. agent-runner
  is called by console-backend's scheduler and runs sub-agent jobs itself;
  client-slack (the package name is the other way round) calls the orchestrator and
  posts its reply, serving no agent card. Nothing in the harness validates a
  sub-agent name, so both tiers passed either way — the cost was paid by readers
  taking the dataset as documentation of two delegation paths that do not exist.
  Renamed to real targets, with the constraint now stated in the dataset header and
  in `MockSubAgent`'s docstring, which previously recommended both wrong names.
  Root `AGENTS.md` is in the diff for the same reason: its architecture diagram
  listed agent-runner under Sub-Agents, contradicting the package table nine lines
  below and most likely seeding the mistake.

## Not included

- **No CI.** This repo has no `.github/workflows` at all, so nothing executes either
  tier anywhere. The mock tier is built for it — 1023 tests, ~12s, no gateway and no
  credentials — and wiring it is the single biggest remaining gap. Deliberately left
  out of this PR: it is one small independent file, and it would be better reviewed on
  its own. Note that a repo-wide job would be red immediately — `agent-common` has 9
  failing tests and one broken collection (`foundry_sdk` not installed) on `main`
  today, unrelated to this branch.
- **The previously-noted gemini "2 + 2" failure did not reproduce.** It was listed
  here as a known real-tier failure: gemini-3-flash-preview delegating four times to
  answer "what is 2 + 2", correct answer by the wrong path. In the run above
  `answers_directly_without_delegating[gemini-3-flash-preview]` **passed** in 3.0s on
  11,544 tokens. Either something on this branch or in `main` fixed it, or it was
  flaky from the start and the single observation was over-read. Not chased further;
  noting it so the claim is not left standing unexamined.
- **Two-level interrupt proxying is still hand-built.** `TestTwoLevelInterruptProxying`
  has the same defect as the tests rewritten above, and now carries a KNOWN GAP
  docstring saying so. Left standing rather than deleted, since it is the only
  coverage of that flow; fixing it needs a `MockSubAgent` that raises
  `GraphInterrupt` through the real dispatch path.
- **`message_formatting` obedience**, the one backfill item deliberately dropped: the
  mock tier cannot test it by construction, and a real-tier version would assert on
  model wording, which this PR's own anti-flakiness rules say to avoid.

## Running it

```bash
cd packages/orchestrator-agent
uv run pytest -q                       # 1023 tests, ~12s, no gateway
uv run pytest -m integration -q        # needs a Model Gateway; see the conftest docstring
```

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — **no.** This is tests and docs
      only. The one behavioural change it originally carried, the env-var switch from
      `MAX_RECURSION_LIMIT` to `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`, was extracted
      at the reviewer's request and shipped in #210; merging `main` removed it from
      this diff. Two production files remain, both sides of one refactor:
      `app/handlers/stream_handler.py` and `app/handlers/__init__.py`, promoting
      `_extract_current_turn_messages` to the public `current_turn_messages` so tests
      and the stream handler share one implementation. The rest is `.gitignore`,
      `scripts/start-local.sh`, `pyproject.toml`/`uv.lock`, two `AGENTS.md`, and
      docstrings in `ringier-a2a-sdk`.


packages/orchestrator-agent/app/models/config.py
```diff
@@ -304,7 +304,22 @@ class AgentSettings:
...
+    # exploration between delegations. An ordinary two-delegation request
+    # ("look up X and post it to Slack") therefore exhausted the budget and the
+    # user was asked to continue a task that had already completed correctly.
+    #
+    # 200 allows ~25 model calls, still far below deepagents' 1000 default, so a
+    # runaway loop is still bounded. This is a stopgap because the unit is wrong:
+    # a super-step budget silently tightens for everyone whenever a middleware is
+    # added. The proper fix is to bound model calls (or price middleware into the
+    # limit) — tracked separately.
+    MAX_RECURSION_LIMIT = int(os.getenv("MAX_RECURSION_LIMIT", "200"))
```
AlexMetsai: Off topic for the PR but ended up here; should we go with it?

alloy-cicd (commented):

Reviewed the full diff (28 files, +3506/−207) with 8 parallel finder passes (line-by-line, removed-behavior, cross-file tracing, conventions, reuse, simplification, efficiency, altitude) followed by adversarial verification of single-source findings. The mock tier itself was confirmed green during review, and several suspicious hunks were cleared (a2a `Part`/`new_text_message` API usage, `.gitignore` coverage, `start-local.sh` path handling, empty-parametrization skip behavior). One candidate finding was **refuted** during verification and is not reported: parametrizing every chat model with `ThinkingLevel.low` is safe because both gateway configs set `drop_params: true` and `test_thinking_level.py` pins the always-forward contract.

**Inline comments (11):** 4 correctness issues around the new marker-based gating and env-file layering, 1 doc contradiction the PR itself introduces, 1 design concern on the shared `MAX_RECURSION_LIMIT` env var, and 5 quality/reuse items in the new test support layer.

**Minor notes not worth their own threads:**
- `tests/integration/conftest.py:557` — the `request.node.stash_usage_recorder` line (and its docstring claim) is dead; the terminal summary reads `_EVAL`, which the fixture teardown already feeds.
- `tests/integration/conftest.py:624-658` — `models_sharing_a_provider()` and `one_model_per_provider()` duplicate the same provider-grouping loop; one `_models_by_provider()` helper would serve both.
- `tests/test_extraction.py:28` — `_task_call`/`_final_call` re-implement `task_call`/`final_response` already exported by `tests/support/graph_harness.py`.
- `tests/test_scenarios_mock.py:53` and `tests/test_scenarios_real.py:122` — the try/except-print(`describe_outcome`)-raise wrapper is duplicated verbatim; folding the diagnostics into `assert_scenario` itself would delete both.
- `tests/support/scripted_model.py:119` — `_tool_name()` is a third copy of the tool-name unwrapper (also in `agent_common.core.graph_utils._tool_name_of` and `ringier_a2a_sdk.middleware.tool_schema_cleaning._get_tool_name`); promoting the agent-common one would give a single canonical implementation.

packages/orchestrator-agent/pyproject.toml
```diff
@@ -77,14 +78,24 @@ addopts = [
...
     "--strict-markers",
     "--disable-warnings",
-    "--ignore=tests/integration",
+    # Integration tests are collected (so breakage in them is visible) but
+    # deselected by default, since they need a reachable Model Gateway and make
+    # real, billable LLM calls. Run them with:  uv run pytest -m integration
+    # They previously used --ignore, which hid the whole directory from
+    # collection — an a2a-sdk migration broke an import there and nothing noticed.
+    "-m",
+    "not integration",
```
alloy-cicd, line 94: **[correctness] A CLI `-m` expression now re-enables billable integration tests.** Replacing `--ignore=tests/integration` with `-m "not integration"` means any user-supplied marker expression *replaces* the default filter (pytest `-m` is last-wins). Verified empirically: `uv run pytest -m slow --collect-only -q` selects 21 tests, **all** under `tests/integration/` (every integration module carries `pytest.mark.slow`, plus `@pytest.mark.langsmith` variants). With gateway coordinates in `.env` — which this PR's docs instruct developers to set up — a casual `-m slow` or `-m langsmith` run makes real LLM calls (~4 min, ~$1.40 per the PR's own numbers), which the old `--ignore` made impossible regardless of markers. Consider an explicit opt-in guard: e.g. in `pytest_collection_modifyitems`, skip integration items unless the markexpr actually requests `integration` (or an env flag is set).

packages/orchestrator-agent/tests/integration/conftest.py
```diff
@@ -370,21 +530,130 @@ def _make(model_type: ModelType, thinking_level: ThinkingLevel | None = None) ->
...
+    if not _EVAL.judged:
+        return
+
+    if _EVAL.gate_failure_reason():
+        session.exitstatus = 1
+    elif exitstatus == 1:
+        # Ratio met: the failures present are tolerated sampling noise. Only
+        # override a plain test-failure status — leave collection errors,
+        # interrupts and internal errors alone.
+        session.exitstatus = 0
```
alloy-cicd, line 594: **[correctness] The exit-status rewrite can turn setup errors green.** `pytest_runtest_logreport` (line 566) only records `report.when == "call"` outcomes, so a test that ERRORs in fixture setup/teardown never enters `_EVAL` — it is neither judged nor failed as far as the gate is concerned. If the remaining tests meet the ratio, `gate_failure_reason()` returns `None` and this `elif exitstatus == 1` branch rewrites pytest's failing exit status to 0: a fixture regression that ERRORs a whole module goes green in CI. The same rewrite also masks genuine *unit-test* failures in any mixed run where at least one integration test was judged (e.g. `uv run pytest -m ""`). Suggested fix: also record non-call failures (`if report.failed and report.when != "call"` → count the nodeid as failed), and/or only downgrade `exitstatus` when every failing/erroring nodeid is accounted for in `_EVAL` as a judged, non-strict failure.

packages/orchestrator-agent/tests/integration/conftest.py
```diff
@@ -112,15 +139,36 @@ def _restore_real_credentials() -> None:
...
+_GATEWAY_KEYS = ("LLM_GATEWAY_URL", "LLM_GATEWAY_API_KEY")
+
+
+def _gateway_env_from_files() -> dict[str, str]:
+    """Gateway coordinates from .env.integration, falling back to the orchestrator .env."""
+    for path in (_ENV_INTEGRATION_FILE, _ORCHESTRATOR_ENV_FILE):
+        values = _load_env_file(path)
+        found = {k: v for k, v in values.items() if k in _GATEWAY_KEYS}
+        if found:
+            return found
```
alloy-cicd: **[correctness] `.env.integration` and the orchestrator `.env` don't layer consistently between discovery and runtime — two confirmed failure modes.**

(a) Here: the loop returns as soon as *any* gateway key is found in a file, so a `.env.integration` containing only `LLM_GATEWAY_API_KEY` shadows the `LLM_GATEWAY_URL` sitting in `packages/orchestrator-agent/.env`. `assert_gateway_configured()` then raises inside `_discover_models()`, `ALL_MODELS` is empty, and the whole tier skips as "No Model Gateway" despite a valid split configuration.

(b) The converse in `_restore_real_credentials` (lines 97–102, just above this hunk): it short-circuits on *any* non-empty `.env.integration`, while discovery falls through per-file. A leftover `.env.integration` without gateway keys (e.g. from the old, now-deleted template) lets discovery succeed via the orchestrator `.env` — tests are collected and run — but `_discover_models`' `finally` block rolls the gateway vars back out of `os.environ` and the `integration_environment` fixture never re-injects them, so every test errors mid-call with `RuntimeError: LLM_GATEWAY_URL is not set` instead of skipping, pointing the developer at configuration they verifiably have.

Fix: merge per-key across both files, and use the same merged loader in both `_gateway_env_from_files` and `_restore_real_credentials`.

packages/orchestrator-agent/tests/integration/conftest.py
```diff
@@ -129,124 +177,189 @@ def _restore_real_credentials() -> None:
...
+            return []
+    finally:
+        for key, value in saved.items():
+            if value is None:
+                os.environ.pop(key, None)
+            else:
+                os.environ[key] = value
+
+
+ALL_MODELS: list[ModelType] = _discover_models()
```
alloy-cicd, line 214: **[efficiency] Gateway discovery runs at conftest import — every plain unit run now pays it.** Since integration tests are deselected by marker rather than `--ignore`, this directory is collected on every `uv run pytest`, and `ALL_MODELS = _discover_models()` performs a live HTTP fetch against `LLM_GATEWAY_URL` at collection time (plus up to the ~2s urlopen timeout, or a full TCP connect timeout for a firewalled/VPN-only host, when the gateway is down). The advertised ~8s unit loop silently becomes network-dependent for exactly the developers who follow this PR's `.env` setup instructions. Deferring discovery — e.g. lazily inside `pytest_generate_tests`, or a memoized function invoked only when integration items are actually selected — keeps parametrization working while keeping unit runs network-free.

packages/orchestrator-agent/AGENTS.md
```diff
@@ -160,13 +160,177 @@ How each type reaches the model (all traffic goes through the gateway as a langc
...
+- **`a2a_tracking[...]["state"]` is the protobuf enum name** (`TASK_STATE_COMPLETED`),
+  not the lowercase `task_state` vocabulary.
+- **Sub-agents require a parent config.** `LocalA2ARunnable` refuses to run without
+  one rather than inventing user ids.
+- **`UserConfig.sub_agents` must be assigned post-construction.** It is annotated
+  `list[CompiledSubAgent]` whose `runnable` is a `Runnable`, which no A2A runnable
+  actually is; production only works because `executor.py` assigns after construction,
+  bypassing validation. Passing them to the constructor raises.
+- **A turn is capped at ~6 model calls.** Each costs ~8 LangGraph super-steps because
+  every middleware hook is its own node, against `MAX_RECURSION_LIMIT=50`. Multi-step
```
alloy-cicd: **[docs] This PR's fact sheet contradicts this PR's config change.** This line states a turn is capped at ~6 model calls against `MAX_RECURSION_LIMIT=50`, but the same PR raises the default to 200 (~25 model calls) in `app/models/config.py:322` — so the doc is wrong from the moment of merge, not stale drift. Since this repo's workflow treats AGENTS.md as authoritative for coding agents, a reader will size scenarios (or "fix" `test_step_budget.py`'s headroom assertion) against a cap that no longer exists. Update to 200/~25, or better, reference the config constant instead of repeating the number.

packages/orchestrator-agent/app/models/config.py
alloy-cicd: **[design] `MAX_RECURSION_LIMIT` is a shared env var name with divergent defaults across packages.** ringier-a2a-sdk and agent-runner default to 50, agent-common's dynamic_agent to 75, and this PR moves the orchestrator to 200 — all reading the same env name. Two consequences: a deployment that pins `MAX_RECURSION_LIMIT=50` (plausible, since 50 was the documented default this PR fixes) silently negates the fix in production while `test_step_budget.py` stays green in CI; conversely, an operator exporting 200 to apply the fix also quadruples the runaway-loop bound of every sub-agent. Consider a package-specific name (e.g. `ORCHESTRATOR_MAX_RECURSION_LIMIT`), and deriving the value instead of hardcoding 200: `BASE_STEPS + STEPS_PER_MODEL_CALL * MAX_MODEL_CALLS_PER_TURN` would give the arithmetic currently scattered across this comment, `test_step_budget.py`, and AGENTS.md a single source of truth.

packages/orchestrator-agent/tests/integration/conftest.py
```diff
@@ -129,124 +177,189 @@ def _restore_real_credentials() -> None:
...
+
+
+def model_available(model_type: ModelType) -> bool:
+    """Whether the gateway serves this alias."""
+    return model_type in ALL_MODELS
+
+
+# Legacy name kept so the older integration modules keep importing cleanly. It no
+# longer inspects credentials — see the note above.
+has_credentials = model_available
```
alloy-cicd: **[simplification] `has_credentials = model_available` is a lying alias, and `requires_gateway`/`skip_marker_for` (lines 269–283) have zero call sites.** Every parametrized `model_type` comes from `ALL_MODELS`, and `model_available` is just membership in `ALL_MODELS` — so the 13 `if not has_credentials(...): pytest.skip(...)` guards across `test_model_streaming.py`, `test_a2a_streaming.py`, and `test_edge_cases.py` can never fire, yet keep readers hunting for AWS/Azure/GCP credentials the process no longer uses — the exact confusion this PR's own comments retire elsewhere. Since the PR already edits all three modules, the mechanical rename/deletion is cheap. `requires_gateway`/`skip_marker_for` are never used anywhere (skipping is handled by `pytest_collection_modifyitems` plus empty parametrization); deleting both leaves behavior identical.

packages/orchestrator-agent/tests/support/usage.py
```diff
@@ -0,0 +1,59 @@
...
+
+from __future__ import annotations
+
+from typing import Any
+
+from langchain_core.callbacks import BaseCallbackHandler
+from langchain_core.outputs import LLMResult
+
+
+class UsageRecorder(BaseCallbackHandler):
```
alloy-cicd: **[reuse] `UsageRecorder` re-implements langchain-core's stock `UsageMetadataCallbackHandler`.** Verified importable in the pinned langchain-core 1.4.8: it aggregates `usage_metadata` per model name and is maintained upstream as provider shapes change — and the gateway models are built with `stream_usage=True` (agent_common `model_factory.py:363`) precisely so `usage_metadata` is present. If the `llm_output` fallback here is genuinely needed for some gateway shape, that ~6-line fallback is the only part worth keeping locally.

packages/orchestrator-agent/tests/support/extraction.py
```diff
@@ -0,0 +1,296 @@
...
+    ``StreamHandler`` does, so a harness bug surfaces as an empty result rather
+    than an AttributeError deep inside an assertion.
+    """
+    if not isinstance(values, dict):
+        return []
+    messages = values.get("messages") or []
+    return list(messages) if all_turns else _current_turn_messages(messages)
+
+
+def message_text(message: Any) -> str:
```
alloy-cicd, line 110: **[reuse] `message_text()` duplicates `BaseMessage.text`.** Verified against the installed langchain-core 1.4.8: `AIMessage(content=[{'type':'thinking',...},{'type':'text','text':'hi'}]).text == 'hi'` — identical behavior for exactly the thinking-block case the docstring defends. Calling `msg.text` (with a small shim for the non-message inputs this helper accepts) drops ~25 lines and tracks new provider block shapes upstream for free.

packages/orchestrator-agent/tests/support/extraction.py
```diff
@@ -0,0 +1,296 @@
...
+from dataclasses import dataclass, field
+from typing import Any
+
+from langchain_core.messages import AIMessage, ToolMessage
+
+from app.handlers import StreamHandler
+
+# The production definition of "this turn" — imported rather than reimplemented
+# so eval assertions cannot drift from the behaviour the executor relies on.
+_current_turn_messages = StreamHandler._extract_current_turn_messages
```
alloy-cicd: **[altitude] Test support now aliases a class-private production method.** Sharing `StreamHandler._extract_current_turn_messages` is the right instinct (no drift), but turn-scoping now has two first-class consumers reached through a private attribute: any StreamHandler refactor breaks the whole `tests/support` layer at import, and nothing signals to a production maintainer that the method has external consumers. Promoting it to a module-level public function (used by both StreamHandler and this module) keeps the no-drift guarantee with an honest visibility contract.

packages/orchestrator-agent/tests/support/graph_harness.py
```diff
@@ -0,0 +1,144 @@
...
+
+    In-memory checkpointer and store stand in for DynamoDB and Postgres, the
+    same substitution the integration conftest makes. ``_create_model`` is
+    shadowed on the instance rather than monkeypatched on the class, so the
+    swap is scoped to this factory and needs no fixture.
+
+    The model type only steers provider-shaped decisions (structured-output
+    strategy, prompt caching); no model of that name is ever contacted.
+    """
+    factory = GraphFactory(config=AgentSettings(), cost_logger=None)
```
alloy-cicd: **[reuse/robustness] The private-attribute patching recipe is duplicated, and the graph is rebuilt per test.** (a) These `factory._checkpointer/_store/_store_setup_complete/_static_tools_cache/_create_model` pokes replicate the `patched_graph_factory` fixture in `tests/integration/conftest.py:446` — when GraphFactory grows or renames a private attribute, one copy breaks (or worse, keeps passing against a half-initialized factory, defeating "the graph is production code"). Extract one shared `build_patched_factory(...)`, or better, give GraphFactory explicit injection kwargs (`checkpointer=`, `store=`, `model_factory=`, `static_tools=` with production defaults) so both tiers use one supported seam. (b) Each `scripted_graph()` call re-runs `AgentSettings()` env parsing plus a full `create_deep_agent` compile — ~23 times across the mock tier, several seconds of pure rebuild growing linearly with the scenario set; a module-scoped compiled graph with a per-test script/checkpointer reset would cut most of it.

packages/orchestrator-agent/tests/integration/conftest.py
AlexMetsai: Confirmed both halves, fixed. Went with your second suggestion rather than the first, since recording non-call failures alone wouldn't close the mixed-run case: note_problem(nodeid, when) now records every failing report in any phase and directory, and may_downgrade_exit_status() requires a passing gate and that every failure be a sampled, non-strict, call-phase integration failure. Setup errors, teardown errors, unit failures in a mixed run, and strict failures all block the downgrade now.

The summary grew a third outcome so a ratio-passed-but-still-red run explains itself, and unaccounted_problems is in the JSON artifact. 8 new tests in tests/test_eval_report.py cover each case, including that a genuine sampled flake is still absolved.

Caveat: the full scenario needs a live gateway, so it's verified at the EvalSession level rather than end to end — the hooks are a thin shell with one call and no other route to exitstatus = 0.

packages/orchestrator-agent/tests/integration/conftest.py
AlexMetsai: Confirmed, both modes, and fixed as suggested — one loader, layered per key, used by _gateway_env_from_files and _restore_real_credentials alike. It's in tests/support/gateway_env.py so the layering is testable without a conftest import.

Your guess about the deleted template was better than a guess: .env.integration.template was removed in this branch (7ef9e85), so anyone who checked it out earlier and copied it has exactly the leftover file that triggers (b), with nothing left to explain it.

Kept one asymmetry on purpose and pinned it with tests: the orchestrator .env is filtered to credential keys since it's generated app config (DATABASE_URL, PORT), while .env.integration is not filtered at all. Also note fake-credential stripping is now unconditional — the old early return skipped it whenever .env.integration merely existed, leaving AWS_ACCESS_KEY_ID=testing in place.

19 tests cover both modes plus the invariant that discovery and runtime agree. Verified against a real .env with a stale override present: both now see both coordinates; the old logic saw neither.

packages/orchestrator-agent/tests/integration/conftest.py
AlexMetsai: Confirmed, and worse than the 2s you estimated: I measured +4.8s on a plain pytest --co, because DNS resolution isn't bounded by urlopen(timeout=2) — an unresolvable in-cluster hostname (a .env for the deployed proxy, developer off VPN) took collection from 3.06s to 7.88s.

Went with your second option rather than pytest_generate_tests, which would mean unwinding 12 parametrize decorators across 4 modules. The blocker was getting the markexpr before the conftest import; I probed the ordering and the root tests/conftest.py's pytest_configure does run first, so it stashes the expression and _discover_models() returns [] when the tier wasn't requested — the same path as a gateway being down.

One subtlety worth flagging for review: answering before collection means evaluating hypothetical marker sets, and a set that doesn't match the expression makes the question vacuous (-m langsmith against a set without langsmith read as "requested"). Each candidate set now has to match first. There's a test asserting the pre-collection and per-item answers agree, since divergence would skip discovery and then select the tests.

Verified import breakage still surfaces on a plain run — planted a bad import and got 1 error, exit 2 — since that's the property the marker-based selection exists to preserve. Collection counts unchanged; -m integration and RUN_INTEGRATION_TESTS=1 still probe.

AlexMetsai: eea8bcfac74c6159309e5ebdb7e3b7e7cb5fa18c

packages/orchestrator-agent/tests/integration/conftest.py
AlexMetsai: 420dd3bd6657a8a3bbf2fedb702702050a9650e0

packages/orchestrator-agent/tests/integration/conftest.py
AlexMetsai: 1d62f13d576fb5f3bdcf128f5f583295fdd67960

packages/orchestrator-agent/pyproject.toml
AlexMetsai: 9ede8043675464fffcbd7745f03a50379d0b88f1

packages/orchestrator-agent/AGENTS.md
AlexMetsai: 9e59904dcbb0399fa20e4d472f12bce1a124dfb4

packages/orchestrator-agent/app/models/config.py
AlexMetsai: 51b86caea4e90035b5f5e1c4bf51d0d04df4c1c8

packages/orchestrator-agent/tests/integration/conftest.py
AlexMetsai: 60fca1492a85b0c6a00d263634733983ed1950ee

packages/orchestrator-agent/tests/support/usage.py
AlexMetsai: 7a147c781d59ce1ffa3932bb77fdb5b5f2160f4b

packages/orchestrator-agent/tests/support/extraction.py
AlexMetsai: 54dbe7a59750f5855c92bcd0301a6fc7fdbcb3ff

packages/orchestrator-agent/tests/support/extraction.py
AlexMetsai: 48a31bd8e13d12947db37d4d72308e54240b5322

packages/orchestrator-agent/tests/support/graph_harness.py
AlexMetsai: 09b9bf4a6536fd5b9bf4d76371d6df53d4ecbcc8

(a) agreed and fixed — one patched_factory() in tests/support/graph_harness, used by both tiers; the conftest fixture is down to a single line. Took the shared-helper option rather than adding injection kwargs to GraphFactory: kwargs are the better end state, but that's production API for a test's benefit and deserves its own change. Your "or worse, keeps passing" point is now guarded — PATCHED_ATTRIBUTES is declared and parametrized over in tests/test_graph_harness.py against a real unpatched factory, so a rename fails by name.

(b) measured it, and I'd rather not. AgentSettings() costs nothing — it's a plain class, so env parsing happens once at import; 23 instantiations is 0.0000s. scripted_graph() is 21ms, 0.49s across 23 calls, against a 1.91s tier and an 8.3s suite. And a module-scoped graph needs the scripted model swapped on shared state, which is the cross-test leakage this tier exists to rule out — so it trades a guaranteed isolation property for half a second. Your linear-growth point stands though: at 21ms each it takes ~50 more scenarios to cost a second, so that's the threshold to revisit. Noted in the docstring.

AlexMetsai:
Prod code touched is 3 places essentially (the rest are tests and support):

1. `stream_handler.py`: no significant change
2. `config.py`: changes regarding max model calls
3. `handlers/init.py`: no significant change


Most of these new tests run on a plain pytest, network-free, 796 green locally. The integration tier doesn't run anywhere yet (by design). It's opt-in behind RUN_INTEGRATION_TESTS (explicit -m).

packages/orchestrator-agent/tests/test_scenarios_mock.py
AlexMetsai, line 1: @aartaria 

Why is this more than a glorified unit test? It's because **only the model is faked**. A proper `GraphFactory` is built from `patched_factory`, and then `patched_factory(model=model)._create_graph(model_type, thinking_level)`, so we compile the production graph. Real middleware stack, real DynamicToolDispatchMiddleware, real HITL middleware, real state channels.

Thus this mock test touches wiring that message-level unit tests cannot reach.

AlexMetsai: The mock tier derives the script from the expectations. The `scripted_responses(scenario)` builds a model that does precisely what the scenario says it should, so it cannot tell whether the orchestrator routes correctly, the routing decision is handled to it. But it does tell us that a scenario is well-formed, that the named sub agents register and dispatch, the expected tools are visible by the orchestrator, that instruction substrings survive the hand-off, that the assertions read the state channels correctly.

So a malformed scenario should fail in milliseconds instead of burning through llm calls.

packages/orchestrator-agent/tests/datasets/core_routing.yaml
AlexMetsai, line 1: - mock tier: `scripted_responses(scenario)` ->  `ScriptedChatModel` -> `scripted_graph(model)` -> `ainvoke` -> `assert_scenario(scenario, state, agents)`. On failure it prints `describe_outcome(state)` first, because "expected slack-client" with no idea what happened is the difference between a five-minute and a fifty-minute diagnosis.
- real tier: "scenarios X one-model-per-provider", with the provider read from the gateway's `model_info`.

AlexMetsai: - scenarios that delegate
    - `routes_to_slack_client`
    - `routes_to_agent_runner_for_data `
    - `forwards_attachment_to_subagent `
- scenaris that should **not** delegate
    - `answers_directly_without_delegating `
    - `uses_time_tool_for_current_time `
    - `attachment_does_not_force_delegation `


`chains_runner_then_slack` also delegates, and on top asserts order

AlexMetsai:
Summing up:

Testing stuff:
- brought up integration dir, fixed that conftest predated the gateway
- PR enables routing testing
- one dataset, one assertion vocabulary, two tiers, mock/real. Basically they get the same stuff, but differ in what is substituted (`core_routing.yaml`)
- 10 modules added in `tests/support`
- own tests
- `test_scenarios_real.py` and a rewritten `integration/conftest.py`
- `test_scenarios_mock.py` (not under integration folder since it doesn't bill)

prod/existing-code changes:
- `stream_handler.py` should have no behavioural changes
- `handlers/__init__.py` also should have no meaningful impact
- `config.py` — the only real behavioural change. New `MAX_MODEL_CALLS_PER_TURN` (25); the recursion limit is now derived from it (`2 + 8 × 25 = 202`) instead of being a raw step count. The shared `MAX_RECURSION_LIMIT` env var is no longer read, since it's set to different values in four packages, so the orchestrator now has its own `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN` and warns loudly if the old name is set.

keeping track of spending:
- `addopts = -m "not integration"` deselects them but also collects them
- `pytest_collection_modifyitems` explicit tier
- if no reachable gateway, skips with reason, but doesn't fail
- `remember_markexpr` skips the import-time gateway completely when it's not asked from the tier

Decision/Gate
- pass ratio decides the session, 0.75 for now. every test runs once, no reruns (rerunning hides flakiness instead of measuring it)
- `@pytest.mark.strict` overrides the ratio: a marked test failing fails the run
- failures the ratio never sampled can't be absolved by it. e.g. a fixture erroring in setup produces no call report, so it keeps the run red
- per-test seconds + tokens in the terminal summary, JSON artifact via `EVAL_REPORT_PATH`

Findings
- `chains_runner_then_slack` aborted with GraphRecursionError at 50 on both providers
  - action: it's a real production ceiling, just a two-delegation chain exceeds it. Adapted `config.py`
  - verified: passes on both models after the fix, ~13s each. Note `logs/eval-report.json` predates the fix and still shows it red
- The abort was undiagnosable, `stream()` swallows the error and returns without capturing state
  - action: now `run_turn()` falls back to the checkpointer and dumps one line per message
- Gemini over-delegated
  - action: opened issue
- Claude delegated to the built-in file-analyzer, read /attachments/, found nothing, and correctly reported failure — gemini didn't bother
  - action: fixed scenario, dropped `task_state`
- Gemini answered "Two plus two is four." instead of 4
  - action: dropped `response_contains`, I guess we will bump into a lot of these in AI answers. We could evaluate the correctness in many ways, but I would keep it out of scope for this bulk of work
- 13 `has_credentials` guards could never fire, dropped some dead code

AlexMetsai:
Three things that need a decision, none of which I want to make silently. Only the
first blocks this PR.

## 1. `config.py` — the one production change, and it needs a call

This is the only non-test change in the diff, so it deserves an explicit yes or no
rather than riding along.

**The bug.** The turn budget was `MAX_RECURSION_LIMIT`, which counts LangGraph
**super-steps**, not model calls. Every middleware hook is its own graph node, so one
model call costs several super-steps — measured at `2 + 8 × model_calls`. A limit of
50 therefore allowed **six model calls per turn**. Live traces showed the orchestrator
completing a two-delegation request correctly and then having the finished answer
discarded and replaced with the "shall I continue?" prompt.

**What the PR does.** The budget is now expressed in the unit people actually reason
about — `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`, default 25 — and converted to
super-steps internally. `MAX_RECURSION_LIMIT` is deliberately no longer read: it is a
*shared* env name, defaulted to 50 by ringier-a2a-sdk and agent-runner and to 75 by
agent-common's dynamic_agent. One value cannot serve all four. A deployment pinning
50 would silently reinstate the truncation; an operator raising it enough for the
orchestrator would quadruple every sub-agent's runaway-loop bound. Setting the old
name now logs a warning naming the replacement, rather than being ignored in silence.

**Why it is still a stopgap.** The multiplier is measured, not designed. Add a
middleware and the same nominal budget silently buys fewer model calls.
`tests/test_step_budget.py` counts the super-steps of a real graph run and pins
`BASE_STEPS` / `STEPS_PER_MODEL_CALL`, so that can no longer change invisibly — a
failure there means the per-call cost genuinely rose and the constant needs updating.
That converts a silent regression into a loud one; it does not make the unit right.
The real fix is for the orchestrator to bound model calls directly rather than infer
them from a step count, which is a bigger change than this PR should carry.

**The question:** merge the stopgap now, or hold the PR until the proper budget
lands? My recommendation is merge — the current behaviour is a user-visible
correctness bug, the pin makes the workaround safe, and the follow-up is tracked. But
it is a behavioural change to production and I would rather it be chosen than
inherited.

☝️ I think Claude puts too much emphasis on this, I would just merge.

## 2. `MAX_RECURSION_LIMIT` in deployments — probably a non-issue

Flagging so nobody is surprised, not because I think it bites. (👈 funny choise of words from Claude 😆 )

Nothing in this repo sets it — I checked every yaml, template and shell script,
including `example-k8s-deployment/base/orchestrator.yaml`. If the infra repo sets it
for the orchestrator, the effect after this merge is:

- the orchestrator ignores it and uses its own default (25 model calls), which **is**
  the fix, not a regression;
- a warning appears at startup naming the replacement variable;
- sub-agents keep reading it exactly as before — unchanged, since they are the
  services the name legitimately belongs to.

So the worst case is a new log line. Worth one grep in the infra repo before merge so
the warning does not arrive unexplained, but it is not a blocker and there is nothing
to coordinate.

## 3. Two findings that belong in their own tickets

Both surfaced while writing tests. Neither is caused by this PR and neither should be
fixed in it.

**a) There is no prompt rule for what to report when a human rejects a tool call.**

`app/models/config.py` says nothing about it — grep for `reject`, `denied` or
`declin` returns nothing. The middleware behaviour is well defined (the call does not
execute, and the model receives a synthetic error `ToolMessage` explaining the
rejection), but what the orchestrator then *tells the user* is unspecified. The model
decides, per model.

The three candidates all mislead in different ways: `failed` reports a malfunction
when the user deliberately exercised a control; `completed` risks hiding that the
action never happened; `input_required` implies something more is needed when nothing
is. My inclination is `completed` with the skipped tool named in the response, but
that is a product call.

Concretely, this is why `test_hitl_reject_turn.py` asserts non-execution and the error
result — both mechanical and safe — but deliberately does **not** assert the reported
`task_state`. There is no rule to assert against, and pinning current model behaviour
would freeze an accident into a test.

**b) Sub-agent descriptions have no selector, unlike tools.**

`app/middleware/dynamic_tool_dispatch.py` → `_build_agent_list` iterates the **entire**
`subagent_registry` unconditionally, wrapping each agent in `<agent name="...">` with
its skills inlined. `ToolsetSelectorMiddleware` exists precisely so hundreds of
*tools* are not all bound to the model — there is no sub-agent equivalent. So the
system prompt grows linearly with the number of sub-agents a user has, with no cap
and no filtering.

I do not know whether this bites in practice, and I would rather measure than
speculate: what is the largest registry in production today, and does anything limit
it? If the answer is "a handful", this is a note for later. If some tenants have
dozens, it is eating context on every turn and is worth a look. A cheap
credential-free test registering N sub-agents would answer it, which is written up in
`TEST_RIGOR_PLAN.md` rather than added here.

packages/orchestrator-agent/tests/datasets/core_routing.yaml
```diff
@@ -0,0 +1,185 @@
...
+#   expecting e.g. send_slack_message can only fail.
+# - `expect.instructions` substring-matches what a sub-agent was handed. Keep
+#   needles to stable specifics (a name, an identifier) — never whole phrases,
+#   which the model rewords freely.
+# - `expect.response_contains` matches the user-visible final message.
+#
+# Keep expectations to what the orchestrator genuinely controls. Asserting on
+# wording the model chooses makes a flaky test, not a strict one.
+
+- id: routes_to_slack_notifier
```
aartaria: question: would these run under which model? shall those run under each model tier?

packages/orchestrator-agent/tests/datasets/failure_propagation.yaml
```diff
@@ -0,0 +1,88 @@
...
+#   `fails:`       sub-agent returns `failed` with this message
+#   `needs_input:` sub-agent returns `input_required` with this message
+# Both replace `reply`, and both go through the same production response
+# builders a real sub-agent uses.
+#
+# Keep `response_contains` off the failure text itself. The model paraphrases a
+# sub-agent's error freely, and asserting on its wording is the flaky-test trap
+# `AGENTS.md` warns about; `task_state` is the falsifiable part.
+
+- id: reports_subagent_failure_as_failure
```
aartaria: gut feeling wise these tests don't deserve an integration test. Those bahaviours shall be just mechanical: `sub-agent state -> orchestrator state`. And completely covered by unit tests. Or am I missing something?

.gitignore
```diff
@@ -306,9 +306,12 @@ build/
 coverage/
 
 # Environment variables
-.env
-.env.local
-.env.*.local
+# Ignore every .env variant by default and re-include the committed templates.
+# This is a public repo: enumerating specific names fails open, so a future
+# .env.staging with real keys would be tracked (.env.integration already was).
+.env*
```
aartaria: shall we add a tracked tests/integration/.env.integration.template pre-filled with the start-local defaults or a just recipe that writes the file if absent, or both?

PR #129, opened 2026-08-10
feat(clients): task-cancellation slash commands for Slack and Google Chat

aartaria:
Closes https://github.com/alloy-ch/rcplus-alloy-infrastructure-agents/issues/224

## Checklist

## What

Adds a `cancel` command (alias: `stop`) to both chat clients, giving them parity with the web client's stop button:

- **Slack**: `/nannos cancel` — new subcommand in the existing slash-command dispatcher.
- **Google Chat**: `cancel` app command — the command dispatcher now matches on the first token of `argumentText` so trailing text doesn't break matching.

Task IDs are internal and unknown to users, so the command takes no arguments. It cancels the user's running task(s) where it is issued: the current **thread** in Google Chat, the current **channel** in Slack (slash command payloads carry no `thread_ts`, so channel scope is the finest granularity available). Tasks running elsewhere are left alone and the reply points the user to where they can cancel them.

## How

- Both `A2AClientService`s gain a `cancelTask(taskId, accessToken)` method that calls the standard A2A `tasks/cancel` JSON-RPC method — the same protocol the console backend uses for the web stop button, so the orchestrator's existing `on_cancel_task` propagation (sub-agent cancels + terminal `canceled` status event) is reused unchanged.
- The tasks to cancel are resolved from the in-flight task store, filtered to the current thread (Google Chat) or channel (Slack).
- Auth uses the existing per-user orchestrator token exchange (`getOrchestratorToken`); unauthenticated users get a login prompt.
- If the orchestrator replies with `TaskNotFoundError`/`TaskNotCancelableError` (-32001/-32002), the stale in-flight record is deleted and the user is told the task already finished. On success the in-flight record is left alone — the active stream handler (or the recovery loop) observes the terminal `canceled` state and cleans up, exactly as it does for any other terminal state.

## Tests

- `packages/client-slack/test/listeners/nannosCancel.test.ts` (8 tests)
- `packages/client-google-chat/test/handlers/commandHandlerCancel.test.ts` (8 tests)

Both packages: `tsc --noEmit` clean, full jest suites green (slack 58, google-chat 25). `npm run lint` fails in both packages due to the missing private `@alloy-ch/eslint-config-alloy` package — pre-existing, reproduces on `main`.

### DIFF
--- a/packages/client-google-chat/src/handlers/commandHandler.ts
+++ b/packages/client-google-chat/src/handlers/commandHandler.ts
@@ -182,15 +182,15 @@ async function handleLoginCommand(
 }
 
 /**
- * Handle "cancel" command – aborts a running task via the same A2A cancel
+ * Handle "cancel" command – aborts running tasks via the same A2A cancel
  * protocol the web client's stop button uses.
  *
- * The task is resolved from the current thread when possible; otherwise from
- * the user's in-flight tasks (a task ID argument disambiguates).
+ * Task IDs are internal to the system, so the tasks to cancel are resolved
+ * from where the command is issued: the user's running tasks in the current
+ * thread (there is almost always exactly one).
  */
 export async function handleCancelCommand(
   appCommand: AppCommand,
-  args: string,
   chatService: GoogleChatService,
   inFlightTaskStore: IInFlightTaskStore,
   userAuthService: UserAuthService,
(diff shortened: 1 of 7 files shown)


packages/client-google-chat/src/handlers/commandHandler.ts
```diff
@@ -208,18 +293,25 @@ export async function handleAppCommand(
...
 ): Promise<void> {
-  const { chatService, contextStore, inFlightTaskStore, userAuthService } = deps;
+  const { chatService, contextStore, inFlightTaskStore, userAuthService, a2aClientService } = deps;
 
   const logger = Logger.getLogger('handleAppCommand');
   logger.info(`App command argument=${appCommand.commandArgument} from user ${appCommand.userId} in space ${appCommand.spaceId}`);
 
-  switch (appCommand.commandArgument) {
+  // Dispatch on the first token so trailing text doesn't break command matching
+  const [commandName] = appCommand.commandArgument.split(/\s+/);
```
andreibazhenau-itechart, line 302: what is the purpose of this instruction.
Previously it worked like: `/nannos login` `/nannos debug` and etc

packages/client-google-chat/src/handlers/commandHandler.ts
```diff
@@ -175,6 +181,84 @@ async function handleLoginCommand(
...
+
+  const reply = (text: string) => chatService.sendPrivateTextMessage(projectId, spaceId, userId, text, threadId);
+
+  const tasks = await inFlightTaskStore.getByUser(projectId, userId);
+  const targets: InFlightTask[] = tasks.filter((t) => t.threadId === threadId);
+
+  if (targets.length === 0) {
+    const elsewhere =
+      tasks.length > 0
+        ? ` You have ${tasks.length} running task(s) in other threads — use the *cancel* command there to cancel them.`
```
andreibazhenau-itechart, line 212: Related with comment below. Previously it worked like: `/nannos login` `/nannos debug` and etc. Why right now it's just `cancel`?

andreibazhenau-itechart (approved):
Have you tested it on any env?

PR #131, opened 2026-08-12
fix(orchestrator-agent): don't scan successful tool results for auth errors

### LINKED ISSUE #130: AuthErrorDetectionMiddleware false-positives on successful tool results containing auth-flavored text
## Summary

`AuthErrorDetectionMiddleware.awrap_tool_call` in `packages/orchestrator-agent/app/middleware/auth_error_middleware.py` scans the content of **every** tool result — including successful ones — for a hardcoded list of free-text phrases (`"authentication required"`, `"authorization required"`, `"access denied"`, etc.). If a successful result's payload happens to contain one of those phrases anywhere in its data, the middleware misclassifies the run as needing auth and raises a `GraphInterrupt`, aborting an otherwise-successful task.

## Repro

An `eval` tool call against Salesforce returned valid JSON (Notes/Tasks/`ADMEIRA_Kontaktbericht__c` records) via stdout, with no `<error>` tag in the QuickJS sandbox output — i.e. a clean success. `<result>null</result>` was just because the script printed via `console.log` instead of returning a value.

Despite this, the run was interrupted with:

```
task_state: 8 (TASK_STATE_AUTH_REQUIRED)
tool: 'eval'
error_code: 'auth-required'
auth_url: ''
message: '<stdout>...=== RESULTS ===...</stdout><result>null</result>'
```

One of the returned CRM free-text fields (business notes about contract/representation authorization) almost certainly contained a substring matching one of the hardcoded patterns (likely "authorization required" or "access denied"), triggering the loose free-text fallback in `_detect_auth_error`.

## Root cause

`awrap_tool_call` (`auth_error_middleware.py:356`) calls `self._detect_auth_error(result.content)` unconditionally, with no check on whether the tool call actually failed. The sibling `ErrorClassificationMiddleware` (`error_classification_middleware.py:158`) guards its own classification with `if status != "error" and not _looks_like_error(content): return result` — `AuthErrorDetectionMiddleware` has no equivalent guard.

The loose free-text fallback in `_detect_auth_error` (`auth_error_middleware.py:598-603`) then matches any of 9 hardcoded phrases anywhere in the content, regardless of whether the content represents an error at all.

## Suggested fix

Gate the free-text fallback (and ideally the whole `_detect_auth_error` call) on the tool result actually indicating failure, e.g. `getattr(result, "status", None) == "error"`, mirroring the guard already used in `ErrorClassificationMiddleware`. Reserve free-text phrase matching for genuine error/exception content, not arbitrary successful payload data.

## Files

- `packages/orchestrator-agent/app/middleware/auth_error_middleware.py:356` (unconditional call site)
- `packages/orchestrator-agent/app/middleware/auth_error_middleware.py:598-603` (loose free-text fallback)
- `packages/orchestrator-agent/app/middleware/error_classification_middleware.py:158` (reference guard pattern to mirror)

aartaria:
closes ringier-data/nannos#130

## Summary
`AuthErrorDetectionMiddleware.awrap_tool_call` scanned every tool result's content for auth-flavored free text ("access denied", "authorization required", etc.), even on successful results. A successful payload that happened to contain one of those phrases anywhere in its data — e.g. a Salesforce CRM note about contract/representation authorization — falsely triggered a `GraphInterrupt` with `error_code: "auth-required"`, aborting a run that had actually completed successfully.

## Fix
Added `AuthErrorDetectionMiddleware._looks_like_tool_error`, which gates the content-based fallback (`_detect_auth_error`) on the result actually looking like a failure — `status == "error"`, or the content itself containing an error-ish keyword (`error`, `exception`, `failed`, `failure`, `traceback`, `unauthorized`, `forbidden`). This mirrors the existing `_looks_like_error` guard already used by the sibling `ErrorClassificationMiddleware`. Applied at both call sites (`ToolMessage` and `Command` branches). The structured JSON (`errorCode: "need-credentials"`) and A2A metadata detection paths are unaffected — they're unambiguous signals that don't need gating.

## Test plan
- [x] Added a regression test reproducing the exact false-positive shape (successful `ToolMessage`, default `status="success"`, content containing "access denied" as ordinary business text) — asserts no interrupt fires.
- [x] All 12 existing tests in `test_auth_error_middleware.py` still pass, including ones that rely on detection firing for genuine errors without an explicit `status="error"`.
- [x] Full `orchestrator-agent` suite: 619 passed.
- [x] `ruff check` clean on changed files.

## Checklist
- [x] Does this PR introduce a breaking change? 🚨 No


packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -359,8 +365,12 @@ async def awrap_tool_call(
...
 
-                # If no A2A auth requirement, fall back to content-based detection
-                if not auth_metadata:
+                # If no A2A auth requirement, fall back to content-based detection —
+                # but only for results that plausibly represent a failure, so a
+                # successful tool's payload is never scanned for auth-flavored text.
+                if not auth_metadata and self._looks_like_tool_error(
+                    getattr(result, "status", None), result.content
+                ):
                     auth_metadata = self._detect_auth_error(result.content)
```
aartaria, line 213: **Correctness: the new gate makes 8 of the 9 `_AUTH_TEXT_PATTERNS` effectively unreachable unless `status=="error"` is set explicitly.**

`_looks_like_tool_error` (lines 533-546) only lets `_detect_auth_error` run when `status == "error"` or the content contains one of `("error", "exception", "failed", "failure", "traceback", "unauthorized", "forbidden")` (line 90). Checking each string in `_AUTH_TEXT_PATTERNS` (lines 74-84) for overlap with that keyword list: only `"401 unauthorized"` contains a keyword (`unauthorized`). The other eight — `"authentication required"`, `"authorization required"`, `"secondary authorization"`, `"need credentials"`, `"need-credentials"`, `"please authorize"`, `"login required"`, and — notably — `"access denied"` itself — contain none.

Concrete failure: a real tool returns `ToolMessage(content="Please authorize this action to continue.", ...)` with the default `status="success"` (many tools never set `status="error"` explicitly, they just report failure via content). `_looks_like_tool_error` now returns `False`, so `_detect_auth_error` is never invoked and the auth-required interrupt never fires — silently regressing exactly the "Text patterns" support the class docstring still advertises (lines 143-145), and ironically doing so for the very `"access denied"` pattern this PR sets out to fix (a genuine `"Access Denied"`-only failure without an explicit error status would now also be missed). This same gate is duplicated at the `Command` branch (lines 422-425).

Suggest either widening `_ERROR_LOOKING_KEYWORDS` to cover the auth-specific vocabulary (e.g. include `"denied"`, `"authoriz"`, `"credential"`, `"required"`), or restructuring the gate so it can't starve the very patterns it's supposed to let through.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -83,6 +83,12 @@
     "access denied",
 )
 
+# Keywords that mark a *successful* tool result as plausibly error-shaped anyway
+# (e.g. a tool that reports failure via content without setting status="error").
+# Gates `_detect_auth_error` so it never scans arbitrary successful payload data
+# (a CRM note mentioning "access denied" in passing must not trigger an interrupt).
+_ERROR_LOOKING_KEYWORDS = ("error", "exception", "failed", "failure", "traceback", "unauthorized", "forbidden")
```
aartaria: **Reuse: this duplicates `_looks_like_error` from `error_classification_middleware.py` verbatim.** `ErrorClassificationMiddleware._looks_like_error` (packages/orchestrator-agent/app/middleware/error_classification_middleware.py:96-103) already implements the exact same 7-keyword tuple and lower-cased containment check, and is already used with the same `status != "error" and not looks_like_error(...)` shape at that file's line 158 (which the PR description explicitly says this mirrors). Importing that helper instead of redefining `_ERROR_LOOKING_KEYWORDS`/`_looks_like_tool_error` here would avoid the two lists silently drifting apart if one is tuned later and the other isn't.

packages/orchestrator-agent/tests/test_auth_error_middleware.py
```diff
@@ -168,6 +168,29 @@ async def test_detect_auth_error_secondary_authorization_text(middleware):
...
+    content = (
+        '{"Note": [{"Body": "Kunde hat schriftlich bestätigt: access denied für weitere Angebote."}]}'
+    )
+    result_msg = ToolMessage(content=content, tool_call_id="tc-6")  # status defaults to "success"
+    handler = AsyncMock(return_value=result_msg)
+
+    result = await middleware.awrap_tool_call(request, handler)
+
+    handler.assert_awaited_once_with(request)
+    assert result is result_msg  # Passed through without interrupt
```
aartaria, line 221: **Test-coverage gap: no test exercises the now-narrowed true-positive path end-to-end.** This new test only proves the "access denied" false positive is suppressed. Nothing in the suite drives one of the other free-text patterns (e.g. `"please authorize"`, `"login required"`, `"authentication required"`) through `awrap_tool_call` with `status` left at its default `"success"`, so the regression described in the `_looks_like_tool_error` gate (auth_error_middleware.py:371-374) isn't caught by CI. `test_detect_auth_error_text_patterns` (line 104) calls `_detect_auth_error` directly, bypassing the new gate entirely, and additionally prefixes every pattern with `"Error: "`, which coincidentally supplies a gate keyword — masking that the gate would otherwise block it. Worth adding a case that goes through `awrap_tool_call` with a bare free-text pattern and default status to confirm true positives still fire.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -517,6 +530,21 @@ def _check_a2a_auth_metadata(
...
+
+        A successful ToolMessage (``status="success"``, the default) can carry
+        arbitrary business data — e.g. a CRM note mentioning "access denied" in
+        passing — that must never be run through ``_detect_auth_error``'s loose
+        free-text fallback. Only scan when the result is explicitly flagged as
+        an error, or its content itself reads like one.
+        """
+        if status == "error":
+            return True
+        text = cls._content_to_text(content).lower()
```
aartaria: **Minor efficiency: `_content_to_text(content)` is computed and lower-cased here, then `_detect_auth_error` (line 593 and again at line 627) normalises/lower-cases the same content again right after** on every call where the gate passes. No correctness impact, just duplicate work on the (larger) content payload on every gated tool result.

aartaria (commented):

Reviewed the diff for correctness and cleanup issues at high recall. The fix itself (gating the free-text fallback on `status=="error"` or an error-ish keyword) correctly suppresses the reported false positive, but the specific keyword list chosen doesn't line up well with the existing `_AUTH_TEXT_PATTERNS` list, which narrows true-positive detection more than intended. Details inline. Posting as a comment only — no approval/change request, that's for the human reviewer.

aartaria:
Updated per the automatic review's findings (all four addressed by removing the flawed approach rather than patching around it):

- The `status`/keyword gate I'd added was itself broken: 8 of 9 free-text patterns shared no keyword with the gate, so most real auth-text cases (including "access denied", the exact pattern this PR set out to fix) would silently stop interrupting unless a tool explicitly set `status="error"`.
- Root-caused instead: removed the loose free-text fallback (`_AUTH_TEXT_PATTERNS`) and the gate entirely. Detection is now structured-only — the `errorCode: "need-credentials"` JSON, which is the actual deterministic signal the gatana MCP tool gateway emits and is specific enough to run ungated (as it always did pre-fix). Free-text auth phrasing is ambiguous against arbitrary tool payload data, so it's now left for the LLM to interpret — consistent with how this middleware already treats A2A 401s and LLM-reported errors per its own docstring.
- This also resolves the reuse/duplication and efficiency comments, since the gate helper that duplicated `_looks_like_error` no longer exists.
- Added a test asserting the real gatana gateway JSON shape still interrupts reliably regardless of `ToolMessage.status`, and a test asserting free-text patterns alone are intentionally no longer detected (documenting the behavior change).

620 tests pass, ruff clean.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -359,7 +353,7 @@ async def awrap_tool_call(
                 a2a_metadata = additional_kwargs.get("a2a_metadata")
                 auth_metadata = self._check_a2a_auth_metadata(a2a_metadata, subagent_type or tool_name)
 
-                # If no A2A auth requirement, fall back to content-based detection
+                # If no A2A auth requirement, fall back to content-based detection.
                 if not auth_metadata:
                     auth_metadata = self._detect_auth_error(result.content)
```
aartaria, line 213: **Behavior regression beyond the linked issue's ask**: this removes the free-text auth fallback entirely, not just gates it on `status == "error"` as issue #130 suggested (mirroring `ErrorClassificationMiddleware._looks_like_error`). Consequence: a genuinely failing tool call whose error is plain text rather than the gatana `errorCode:"need-credentials"` envelope — e.g. a `requests`/`httpx` 401 (`"401 Client Error: Unauthorized for url ..."`) or a `ToolException("Access denied")` — will no longer trigger `interrupt()`/the resumable auth flow at all, even though it is an unambiguous, real auth failure (not the false-positive-on-success case #130 reported). This is confirmed by this PR's own new/renamed tests (`test_regular_tool_with_free_text_401_passes_through`, `test_detect_auth_error_free_text_alone_not_detected`), which assert this content now falls through with no auth metadata regardless of the tool's actual success/failure state. Worth confirming this broader scope-reduction (vs. the minimal `status=="error"` gate originally suggested) is intentional, since it silently weakens a feature advertised in the module's own "Key Features" docstring for any tool/gateway that doesn't emit the structured JSON.

packages/orchestrator-agent/tests/test_auth_error_middleware.py
```diff
@@ -162,12 +170,69 @@ async def test_detect_auth_error_list_content_blocks(middleware):
...
+    content = (
+        '{"errorCode":"need-credentials",'
+        '"authorizeUrl":"https://gatana.ai/api/v1/mcp-servers/oauth/gt_ADsagJ9hdU/begin",'
+        '"message":"This tool requires secondary authorization. You must tell the end-user '
+        'to please go to the authorizeUrl. After this is done, you can retry the tool call and it will work."}'
+    )
+    result_msg = ToolMessage(content=content, tool_call_id="tc-7")  # status defaults to "success"
+    handler = AsyncMock(return_value=result_msg)
+
+    with pytest.raises(Exception):
```
aartaria, line 244: Minor test-quality nit: `with pytest.raises(Exception):` here (new test `test_gatana_gateway_need_credentials_triggers_interrupt_regardless_of_status`) is broad enough to pass even if `interrupt()` never fires and some unrelated exception is raised instead (e.g. a typo/AttributeError before reaching the interrupt call). Asserting on the specific `GraphInterrupt`/`GraphBubbleUp` type (or at minimum inspecting the exception) would make this regression-proof; right now it mirrors a pre-existing weak pattern elsewhere in the file rather than tightening it.

aartaria (commented):
Automated review findings for PR #131.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -595,11 +597,4 @@ def _detect_auth_error(self, content: Any) -> Dict[str, Any] | None:
             logger.info(f"[AUTH MIDDLEWARE] Detected embedded auth error. Auth URL: {authorize_url}")
             return {"auth_url": authorize_url, "auth_message": error_message, "error_code": "need-credentials"}
```
aartaria, line 456: **Residual false-positive path not fully closed by this PR**

The embedded-detection branch just above (`if "need-credentials" in content:`) is a bare substring match, not a JSON-structure check. This PR's stated goal is to eliminate false-positive interrupts on arbitrary business-data content (issue #130), and the new docstring claims the structured payload "is specific enough to never appear in legitimate tool output by accident" — but that only holds for step 1 (whole-content JSON parse). Step 2 will still fire on any successful ToolMessage whose content merely contains the literal token `need-credentials` outside of a JSON error envelope, e.g.:

```json
{"ticket_notes": "customer support case mentions a need-credentials error from a partner API, unrelated to this session"}
```

This has no `errorCode` key (step 1 fails), but the substring check in step 2 matches, so `_detect_auth_error` returns here with a synthetic `auth_url=""` / generic message, and `awrap_tool_call` calls `interrupt()`, aborting an otherwise-successful run — the same class of false positive #130 was filed for, just triggered by a different token than "access denied".

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -359,7 +353,7 @@ async def awrap_tool_call(
                 a2a_metadata = additional_kwargs.get("a2a_metadata")
                 auth_metadata = self._check_a2a_auth_metadata(a2a_metadata, subagent_type or tool_name)
 
-                # If no A2A auth requirement, fall back to content-based detection
+                # If no A2A auth requirement, fall back to content-based detection.
                 if not auth_metadata:
```
aartaria, line 212: **Dead default value after this change**

A few lines below (and at the equivalent `Command` branch, and in the `ToolException` handler further down), `auth_metadata.get("error_code", "auth-required")` can no longer produce `"auth-required"`: `_detect_auth_error` only ever returns `error_code="need-credentials"` now that the free-text branch (which used to be the sole source of `"auth-required"`) is gone, and `_check_a2a_auth_metadata` only returns `"a2a-auth-required"`. `auth_metadata` is never returned without an explicit `error_code` key, so the `.get(..., "auth-required")` default is now unreachable. Not a crash, but it's misleading — a reader (or a downstream consumer keying off `error_code`) could reasonably believe `"auth-required"` is still an emittable value from this middleware; it isn't anymore.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -70,18 +73,6 @@
...
-    "authentication required",
-    "authorization required",
-    "secondary authorization",
-    "need credentials",
-    "need-credentials",
-    "please authorize",
-    "login required",
-    "401 unauthorized",
-    "access denied",
-)
```
aartaria, line 80: **Minor: three consecutive blank lines**

Removing `_AUTH_TEXT_PATTERNS` left three blank lines before the `AuthErrorState` class definition, instead of the two-blank-line convention used elsewhere in this module between top-level definitions. Cosmetic only (ruff's default rule set doesn't flag it here), but worth a quick cleanup pass.

aartaria:
Addressed the three review comments from the latest pass:

1. **[Residual false-positive path](https://github.com/ringier-data/nannos/pull/131#discussion_r3766274781)** — step 2's embedded-error detection was a bare substring check for `"need-credentials"` anywhere in the content, so business data merely mentioning that word in prose (not as part of a real error envelope) would still false-positive an interrupt — the same class of bug #130 was filed for, just a different token. Tightened to match the actual `"errorCode":"need-credentials"` field via regex, and added a regression test with the exact scenario from the comment (a support-ticket note mentioning the word in passing).
2. **[Dead default value](https://github.com/ringier-data/nannos/pull/131#discussion_r3766277875)** — `auth_metadata.get("error_code", "auth-required")` was unreachable dead code once the free-text branch (the only source of `"auth-required"`) was removed. Changed the default to `"need-credentials"`, the only value the structured-JSON path can actually produce, at all three call sites.
3. **[Blank lines](https://github.com/ringier-data/nannos/pull/131#discussion_r3766278986)** — cleaned up.

621 tests pass, ruff clean.

PR #135, opened 2026-08-12
fix(agent-common): degrade gracefully on Gatana MCP gateway errors

### LINKED ISSUE #134: Gatana MCP gateway errors crash embedded execute-only sub-agents
## Problem

When the Gatana MCP gateway (`https://alloych.gatana.ai`) returns a non-2xx (observed: `400 Bad Request` due to a disabled Gatana account), embedded execute-only sub-agents crash the entire turn with `a2a.utils.errors.InternalError`, instead of degrading gracefully.

Two related gaps:

1. `ToolDiscoveryService.fetch_available_servers` (`packages/orchestrator-agent/app/core/discovery.py`) already catches and swallows this — but `DynamicLocalAgentRunnable._discover_mcp_tools` (`packages/agent-common/agent_common/agents/dynamic_agent.py`) re-raised non-retryable errors, and nothing upstream caught it. Pre-embed this was less visible because the orchestrator was always the top-level graph; embedded execute-only sub-agents run as the top-level graph (see #116), so a dead gateway now takes the whole turn down.
2. Neither failure path logged the response body — `httpx`'s `raise_for_status()` discards it, so `400 Bad Request` alone doesn't say *why* (e.g. `{"message":"Your account has been disabled..."}`).

## Fix (implemented, tests green)

- `_discover_mcp_tools` now returns `[]` instead of raising on non-retryable/exhausted-retry errors (existing retry-with-backoff for 502/503/504 untouched).
- New `get_mcp_http_status_error()` / `get_mcp_error_body()` helpers in `ringier_a2a_sdk/utils/mcp_errors.py` extract and safely log the response body (handles `httpx.ResponseNotRead` for the MCP streamable-http client, which closes the stream before the body can be read).
- When discovery degrades, the sub-agent's system prompt gets a `<tool_availability_warning>` addendum so the LLM can tell the user its tools are temporarily unavailable, instead of just silently doing less.

## Test plan

- [x] `ringier-a2a-sdk`: 331 passed, 2 skipped
- [x] `agent-common` (`test_dynamic_agent.py`): 23 passed
- [x] `orchestrator-agent` (discovery tests): 29 passed

aartaria:
closes #134

## Summary
- `_discover_mcp_tools` degrades to `[]` for genuine MCP gateway/transport failures — now covering both `httpx.HTTPError` **and** `mcp.shared.exceptions.McpError` (the MCP SDK's own error for JSON-RPC-level failures, e.g. a synthesized 404 or `initialize`/`tools/list` errors when the gateway is up but a downstream MCP server is down — the canonical way a *gateway* specifically degrades). Arbitrary bugs in our own code still crash loudly, as they should.
- A failure exchanging the **user's own token** is retried like a transient gateway error but never degraded — routed through a typed `_TokenExchangeError` (via a single `_exchange_token_for()` helper) rather than a boolean flag, so the "never degrade" invariant is structural.
- Fixed a stale-pin bug: `_mcp_discovery_error` is now only cleared on an actual successful resolution, not unconditionally at the top of every attempt — a degrade followed by a later raising call no longer gets silently treated as "resolved."
- `flatten_exceptions` is public and shared with `voice-agent`. `log_mcp_gateway_error()` consolidates gateway-response logging and redacts 401/403 bodies (deliberately not 400 — this PR's own motivating case is a safe, actionable 400).
- The sub-agent's system prompt gets a fixed, parameter-free `<tool_availability_warning>` when degraded — never the raw gateway error text.
- Fixed the same `rstrip("/mcp")` character-set bug in two places, and found (by actually running the old code) that my own regression test didn't cover the bug it claimed to — fixed the test too.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 No

## Test plan
- [x] `ringier-a2a-sdk`: `uv run pytest -q` — 359 passed, 2 skipped
- [x] `agent-common`: `uv run pytest -q` (excl. 2 pre-existing environment-credential gaps, unrelated) — 703 passed
- [x] `orchestrator-agent`: `uv run pytest -q` — 625 passed
- [x] `voice-agent`: `uv run pytest -q` — 8 passed
- [x] `console-backend`: `uv run pytest -q tests/test_mcp_router.py` — 14 passed

## Follow-ups filed (not blocking this PR)
- nannos#136 — pre-existing `combined_middlewares`/`None` crash in the tool-catalog + client-action path (`feature/nannos-embedded` only, doesn't exist on `main`)
- nannos#137 — design: surface discovery-time auth failures as a distinct signal instead of a generic `InternalError`
- nannos#138 — design: scope degraded-instance retries to the gateway step only (currently re-runs the whole lazy init every call while degraded); consolidate the MCP gateway URL derivation duplicated across 3 files

## Discarded review findings (with reasoning, not silently dropped)
- Widening 401/403 body redaction to include 400: **rejected** — this PR's own motivating case (disabled Gatana account) arrives as a 400 with a safe, actionable body.
- `console_gateway_server_access_service.py`'s `api_base_url`: flagged as a third buggy copy of the URL-derivation logic — checked, it's actually already correct (different but valid implementation). No fix needed there; the duplication itself is tracked in nannos#138.
- Redundant OIDC round-trips on retry; "no cooldown during a sustained outage" (now tracked concretely in nannos#138 with a design sketch).

### DIFF
--- a/packages/agent-common/agent_common/agents/dynamic_agent.py
+++ b/packages/agent-common/agent_common/agents/dynamic_agent.py
@@ -129,6 +129,22 @@ def is_console_backend_tool(name: str) -> bool:
     return name.startswith(CONSOLE_BACKEND_TOOL_PREFIXES)
 
 
+class _TokenExchangeError(Exception):
+    """Marks a failure from exchange_token() as a distinct type, not a string flag.
+
+    _discover_mcp_tools must never degrade a failure exchanging the user's own
+    token into the gateway's "temporarily unavailable" path (their credentials,
+    not the gateway being down). Routing every exchange_token() call through
+    _exchange_token_for (the only place that raises this) makes that invariant
+    structural: a future third exchange call site that skips the helper is a
+    visible bug (unclassified as auth), not a silently-forgotten flag.
+
+    Always raised as ``raise _TokenExchangeError(...) from original_exc`` —
+    callers recover the original via ``__cause__`` for logging/classification,
+    then re-raise it bare (not ``from None``) so its own chain stays intact.
+    """
+
+
 def _validate_tool_schema(tool: BaseTool) -> BaseTool:
     """Validate and fix MCP tool schema for OpenAI API compatibility.
 
@@ -601,6 +617,21 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
             "</self_improvement>"
         )
 
+    async def _exchange_token_for(self, target_client_id: str) -> str:
+        """Exchange self.user_token for a token scoped to target_client_id.
+
+        The only place that calls self.oauth2_client.exchange_token() during
+        MCP discovery — see _TokenExchangeError for why that matters.
+        """
+        try:
+            return await self.oauth2_client.exchange_token(
+                subject_token=self.user_token,
+                target_client_id=target_client_id,
+                requested_scopes=["openid", "profile", "offline_access"],
+            )
+        except Exception as e:
+            raise _TokenExchangeError(f"Exchanging token for {target_client_id} failed") from e
+
     async def _discover_mcp_tools(self) -> List[BaseTool]:
         """Discover tools from MCP servers with authentication.
 
@@ -638,7 +669,6 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
         """
         import asyncio
 
-        self._mcp_discovery_error = None
         mcp_gateway_url = self.mcp_gateway_url
         mcp_gateway_client_id = self.mcp_gateway_client_id
         mcp_tool_names = set(self.config.mcp_tools or [])
@@ -657,14 +687,6 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
         delay = 1.0
 
         for attempt in range(max_retries):
-            # Set when a failure below came from exchange_token() rather than
-            # the MCP gateway/transport itself — a different failure class
-            # (the user's own credentials, not the gateway being down) that
-            # must not be degraded to []. Reused as-is by is_retryable_mcp_error
-            # for the retry decision, so a transient OIDC hiccup still gets
-            # retried the same as a transient gateway one; only a non-retryable
-            # or exhausted-retries auth failure skips the degrade and raises.
-            is_auth_exchange_error = False
             try:
                 # Build connections dict for MultiServerMCPClient
                 connections: dict[str, StreamableHttpConnection] = {}
@@ -675,15 +697,7 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                     headers: dict[str, str] = {}
                     if self.oauth2_client and self.user_token:
                         logger.debug(f"Exchanging token for MCP gateway access for {self.name}")
-                        try:
-                            mcp_gateway_token = await self.oauth2_client.exchange_token(
-                                subject_token=self.user_token,
-                                target_client_id=mcp_gateway_client_id,
-                                requested_scopes=["openid", "profile", "offline_access"],
-                            )
-                        except Exception:
-                            is_auth_exchange_error = True
-                            raise
+                        mcp_gateway_token = await self._exchange_token_for(mcp_gateway_client_id)
                         headers["Authorization"] = f"Bearer {mcp_gateway_token}"
                         logger.info(f"Successfully exchanged token for MCP gateway ({self.name})")
                     else:
@@ -703,15 +717,7 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                     console_headers: dict[str, str] = {}
                     if self.oauth2_client and self.user_token:
                         logger.debug(f"Exchanging token for console backend access for {self.name}")
-                        try:
-                            console_token = await self.oauth2_client.exchange_token(
-                                subject_token=self.user_token,
-                                target_client_id=self.console_backend_client_id,
-                                requested_scopes=["openid", "profile", "offline_access"],
-                            )
-                        except Exception:
-                            is_auth_exchange_error = True
-                            raise
+                        console_token = await self._exchange_token_for(self.console_backend_client_id)
                         console_headers["Authorization"] = f"Bearer {console_token}"
                         logger.info(f"Successfully exchanged token for console backend ({self.name})")
                     elif self.user_token:
@@ -746,24 +752,41 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
 
                 if attempt > 0:
                     logger.info(f"Successfully discovered MCP tools for {self.name} on attempt {attempt + 1}")
+                # Only cleared on an actual successful resolution — not
+                # unconditionally at the top of the method — so a call that
+                # raises (auth failure, non-transport error) before reaching
+                # here leaves a prior degrade's error in place. Otherwise the
+                # _ensure_agent guard below would see a cleared error next to
+                # still-cached degraded tools and wrongly treat that as
+                # "resolved," permanently skipping the retry it's meant to allow.
+                self._mcp_discovery_error = None
                 return validated_tools
 
             except Exception as e:
+                # A _TokenExchangeError wraps the real failure (from exchange_token
+                # via _exchange_token_for) purely to route it through this one
+                # except clause structurally — classify/log/retry on the
+                # original cause, never the wrapper's own generic message.
+                is_auth_exchange_error = isinstance(e, _TokenExchangeError)
+                classify_target = e.__cause__ if is_auth_exchange_error else e
+
                 # Check if this is a retryable error
-                is_retryable = is_retryable_mcp_error(e)
+                is_retryable = is_retryable_mcp_error(classify_target)
 
                 if not is_retryable or attempt >= max_retries - 1:
                     if is_auth_exchange_error:
-                        # Don't degrade — propagate so the turn fails loudly,
-                        # same as before this degrade path existed for
-                        # gateway/transport errors (see docstring).
+                        # Don't degrade — propagate the original cause (not the
+                        # wrapper) so the turn fails loudly, same as before this
+                        # degrade path existed for gateway/transport errors (see
+                        # docstring). Bare raise, not "from None": preserves
+                        # classify_target's own __cause__/__context__ chain intact.
                         logger.error(
                             f"Token exchange failed while discovering MCP tools for {self.name} "
-                            f"(attempt {attempt + 1}): {e}"
+                            f"(attempt {attempt + 1}): {classify_target}"
                         )
-                        raise
+                        raise classify_target
 
-                    if not is_mcp_transport_error(e):
+                    if not is_mcp_transport_error(classify_target):
                         # Not a gateway/network failure at all — a bug in our own
                         # code (e.g. a tool-schema validation error) raised from
                         # inside this try block. Must not be silently reported to
@@ -775,17 +798,15 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                         raise
 
                     # Non-retryable error or exhausted retries — degrade gracefully
-                    # instead of crashing the whole turn (see docstring). This is
-                    # always the loop's last possible outcome: every path either
-                    # returns here or above, so there is no unreachable tail after it.
-                    error_msg = format_mcp_error(e)
+                    # instead of crashing the whole turn (see docstring).
+                    error_msg = format_mcp_error(classify_target)
                     if is_retryable:
                         logger.error(
                             f"Failed to discover MCP tools for {self.name} after {attempt + 1} attempts: {error_msg}"
                         )
                     else:
-                        logger.error(f"Non-retryable error discovering MCP tools for {self.name}: {e}")
-                    log_mcp_gateway_error(logger, e, context=f"for {self.name} ")
+                        logger.error(f"Non-retryable error discovering MCP tools for {self.name}: {classify_target}")
+                    log_mcp_gateway_error(logger, classify_target, context=f"for {self.name} ")
                     # Recorded so _ensure_agent can warn the LLM (and, through it,
                     # the user) that some tools are missing — otherwise the agent
                     # just looks silently less capable with no way to explain why.
@@ -794,11 +815,18 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
 
                 # Retryable error - wait and retry
                 logger.warning(
-                    f"Transient error discovering MCP tools for {self.name} (attempt {attempt + 1}/{max_retries}): {e}. "
-                    f"Retrying in {delay:.1f}s..."
+                    f"Transient error discovering MCP tools for {self.name} (attempt {attempt + 1}/{max_retries}): "
+                    f"{classify_target}. Retrying in {delay:.1f}s..."
                 )
                 await asyncio.sleep(delay)
                 delay *= 2  # Exponential backoff
+
+        # Unreachable today (every branch above returns or raises on the last
+        # attempt) — kept as an executable invariant rather than only a
+        # comment, so a future change to the loop structure (e.g. a stray
+        # `continue`) fails loudly instead of silently returning None into
+        # _discovered_tools with _mcp_discovery_error left unset.
+        raise AssertionError("unreachable: MCP discovery retry loop exited without returning or raising")
 
     # Tools that sub-agents always get from console-backend MCP for self-improvement
     _CONSOLE_SELF_IMPROVEMENT_TOOLS = frozenset(
--- a/packages/agent-common/agent_common/agents/dynamic_agent.py
+++ b/packages/agent-common/agent_common/agents/dynamic_agent.py
@@ -128,17 +128,6 @@ def is_console_backend_tool(name: str) -> bool:
     return name.startswith(CONSOLE_BACKEND_TOOL_PREFIXES)
 
 
-class _McpTokenExchangeError(Exception):
-    """Marks a failure from exchange_token(), not from the MCP server itself.
-
-    _discover_mcp_tools degrades gracefully on MCP transport/gateway errors —
-    but a failure exchanging the user's own token is a different failure class
-    (their credentials, not the gateway being down) and must not be caught by
-    that same degrade path silently. Raised with the original exception as
-    __cause__ and unwrapped back to it before propagating past this module.
-    """
-
-
 def _validate_tool_schema(tool: BaseTool) -> BaseTool:
     """Validate and fix MCP tool schema for OpenAI API compatibility.
 
@@ -625,15 +614,18 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
         If config.mcp_tools is set, only those tools are returned (whitelist filtering).
         The discovered tools override orchestrator tools entirely when whitelist is specified.
 
-        Implements retry logic with exponential backoff for transient errors (502, 503, 504).
-        Non-retryable MCP transport/gateway errors (4xx, exhausted retries) degrade
-        gracefully to an empty tool list rather than failing the turn — a dead/
-        misconfigured gateway (e.g. Gatana) shouldn't crash embedded execute-only
-        sub-agents that depend on this as their only tool source, mirroring
-        ToolDiscoveryService.fetch_available_servers. Failures exchanging the
-        user's own token are a different failure class (their credentials, not
-        the gateway being down) and are NOT degraded — they propagate so the turn
-        fails loudly, same as before this degrade path existed.
+        Implements retry logic with exponential backoff for transient errors (502, 503, 504),
+        applied uniformly to MCP transport/gateway errors and to exchange_token()
+        failures alike (a transient OIDC hiccup gets retried the same as a transient
+        gateway one). Once non-retryable or retries are exhausted, the two failure
+        classes diverge: a dead/misconfigured MCP transport/gateway (4xx, e.g. a
+        disabled Gatana account) degrades gracefully to an empty tool list rather
+        than failing the turn — it shouldn't crash embedded execute-only sub-agents
+        that depend on this as their only tool source, mirroring
+        ToolDiscoveryService.fetch_available_servers. A failure exchanging the
+        user's own token is a different failure class (their credentials, not the
+        gateway being down) and is NOT degraded — it propagates so the turn fails
+        loudly, same as before this degrade path existed.
 
         Returns:
             List of discovered BaseTool instances (filtered by whitelist if specified),
@@ -664,6 +656,14 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
         delay = 1.0
 
         for attempt in range(max_retries):
+            # Set when a failure below came from exchange_token() rather than
+            # the MCP gateway/transport itself — a different failure class
+            # (the user's own credentials, not the gateway being down) that
+            # must not be degraded to []. Reused as-is by is_retryable_mcp_error
+            # for the retry decision, so a transient OIDC hiccup still gets
+            # retried the same as a transient gateway one; only a non-retryable
+            # or exhausted-retries auth failure skips the degrade and raises.
+            is_auth_exchange_error = False
             try:
                 # Build connections dict for MultiServerMCPClient
                 connections: dict[str, StreamableHttpConnection] = {}
@@ -680,8 +680,9 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                                 target_client_id=mcp_gateway_client_id,
                                 requested_scopes=["openid", "profile", "offline_access"],
                             )
-                        except Exception as auth_exc:
-                            raise _McpTokenExchangeError(str(auth_exc)) from auth_exc
+                        except Exception:
+                            is_auth_exchange_error = True
+                            raise
                         headers["Authorization"] = f"Bearer {mcp_gateway_token}"
                         logger.info(f"Successfully exchanged token for MCP gateway ({self.name})")
                     else:
@@ -707,8 +708,9 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                                 target_client_id=self.console_backend_client_id,
                                 requested_scopes=["openid", "profile", "offline_access"],
                             )
-                        except Exception as auth_exc:
-                            raise _McpTokenExchangeError(str(auth_exc)) from auth_exc
+                        except Exception:
+                            is_auth_exchange_error = True
+                            raise
                         console_headers["Authorization"] = f"Bearer {console_token}"
                         logger.info(f"Successfully exchanged token for console backend ({self.name})")
                     elif self.user_token:
@@ -745,18 +747,21 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                     logger.info(f"Successfully discovered MCP tools for {self.name} on attempt {attempt + 1}")
                 return validated_tools
 
-            except _McpTokenExchangeError as auth_exc:
-                # A different failure class from the gateway/transport errors below
-                # (the user's own credentials, not the gateway being down) — don't
-                # degrade, propagate the original exception as before this path existed.
-                logger.error(f"Token exchange failed while discovering MCP tools for {self.name}: {auth_exc}")
-                raise auth_exc.__cause__ from None
-
             except Exception as e:
                 # Check if this is a retryable error
                 is_retryable = is_retryable_mcp_error(e)
 
                 if not is_retryable or attempt >= max_retries - 1:
+                    if is_auth_exchange_error:
+                        # Don't degrade — propagate so the turn fails loudly,
+                        # same as before this degrade path existed for
+                        # gateway/transport errors (see docstring).
+                        logger.error(
+                            f"Token exchange failed while discovering MCP tools for {self.name} "
+                            f"(attempt {attempt + 1}): {e}"
+                        )
+                        raise
+
                     # Non-retryable error or exhausted retries — degrade gracefully
                     # instead of crashing the whole turn (see docstring). This is
                     # always the loop's last possible outcome: every path either
@@ -1002,10 +1007,14 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
         6. Build self._agent ONLY if sandbox is not active (sandbox agents build
            a fresh graph per invocation in _astream_impl)
 
-        After first call, this is a no-op (guarded by _cached_tools sentinel).
+        After a call that resolved cleanly, this is a no-op (guarded by
+        _cached_tools). A call that degraded (_mcp_discovery_error set) is
+        NOT treated as resolved — the guard lets the next call retry
+        discovery instead of permanently pinning this instance at zero MCP
+        tools because of one transient gateway blip.
         """
-        # Already resolved — skip
-        if self._cached_tools is not None:
+        # Already resolved — skip, unless the last attempt degraded.
+        if self._cached_tools is not None and self._mcp_discovery_error is None:
             return
 
         # Discover MCP tools if whitelist is configured AND no injected tools
@@ -1013,6 +1022,14 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
         if self.inject_all_tools is None:
             if self.config.mcp_tools and len(self.config.mcp_tools) > 0 and self._discovered_tools is None:
                 self._discovered_tools = await self._discover_mcp_tools()
+                if self._mcp_discovery_error:
+                    # Degraded, not resolved — leave _discovered_tools unset (not
+                    # cached as "[]", the same as a real empty whitelist match)
+                    # so the next call's guard above re-triggers discovery instead
+                    # of treating this instance as permanently tool-less.
+                    # _get_effective_tools() below still sees `[]` for *this*
+                    # call via `self._discovered_tools or []`.
+                    self._discovered_tools = None
 
         tools = self._get_effective_tools()
 
--- a/packages/agent-common/agent_common/agents/dynamic_agent.py
+++ b/packages/agent-common/agent_common/agents/dynamic_agent.py
@@ -49,6 +49,7 @@ from langgraph.store.postgres.aio import AsyncPostgresStore
 from ringier_a2a_sdk.oauth import OidcOAuth2Client
 from ringier_a2a_sdk.utils.mcp_errors import (
     format_mcp_error,
+    is_mcp_transport_error,
     is_retryable_mcp_error,
     log_mcp_gateway_error,
 )
@@ -759,6 +760,17 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
                         logger.error(
                             f"Token exchange failed while discovering MCP tools for {self.name} "
                             f"(attempt {attempt + 1}): {e}"
+                        )
+                        raise
+
+                    if not is_mcp_transport_error(e):
+                        # Not a gateway/network failure at all — a bug in our own
+                        # code (e.g. a tool-schema validation error) raised from
+                        # inside this try block. Must not be silently reported to
+                        # the user as "temporarily unavailable" forever; crash
+                        # loudly like any other programming error.
+                        logger.error(
+                            f"Unexpected (non-MCP-transport) error discovering MCP tools for {self.name}: {e}"
                         )
                         raise
 
--- a/packages/agent-common/agent_common/agents/dynamic_agent.py
+++ b/packages/agent-common/agent_common/agents/dynamic_agent.py
@@ -49,9 +49,8 @@ from langgraph.store.postgres.aio import AsyncPostgresStore
 from ringier_a2a_sdk.oauth import OidcOAuth2Client
 from ringier_a2a_sdk.utils.mcp_errors import (
     format_mcp_error,
-    get_mcp_error_body,
-    get_mcp_http_status_error,
     is_retryable_mcp_error,
+    log_mcp_gateway_error,
 )
 from ringier_a2a_sdk.utils.mcp_progress import on_mcp_progress
 from ringier_a2a_sdk.utils.streaming import (
@@ -127,6 +126,17 @@ CONSOLE_BACKEND_TOOL_PREFIXES = ("console_", "scheduler_")
 def is_console_backend_tool(name: str) -> bool:
     """Return True if ``name`` is served by the console-backend MCP, not the gateway."""
     return name.startswith(CONSOLE_BACKEND_TOOL_PREFIXES)
+
+
+class _McpTokenExchangeError(Exception):
+    """Marks a failure from exchange_token(), not from the MCP server itself.
+
+    _discover_mcp_tools degrades gracefully on MCP transport/gateway errors —
+    but a failure exchanging the user's own token is a different failure class
+    (their credentials, not the gateway being down) and must not be caught by
+    that same degrade path silently. Raised with the original exception as
+    __cause__ and unwrapped back to it before propagating past this module.
+    """
 
 
 def _validate_tool_schema(tool: BaseTool) -> BaseTool:
@@ -494,6 +504,36 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
 
         return "\n\n" + "\n".join(parts)
 
+    def _build_tool_availability_addendum(self) -> str:
+        """Build the tool-availability warning for the system prompt.
+
+        Set when _discover_mcp_tools degraded to zero tools rather than failing
+        the turn (see its docstring) — without this, the agent just looks
+        silently less capable, with no way to tell the user a gateway is down
+        instead of the capability never having existed.
+
+        Deliberately a fixed, parameter-free sentence rather than interpolating
+        self._mcp_discovery_error: that text is gateway/exception-controlled
+        (a prompt-injection surface at system-prompt privilege), can embed
+        internal URLs the model would otherwise be told to relay to the user,
+        and varies turn to turn during a flapping gateway — which would
+        invalidate the prompt-cache prefix on every degraded turn. The detailed
+        message is already logged server-side, in _discover_mcp_tools.
+
+        Returns:
+            Formatted string to append to the system prompt, or empty string
+        """
+        if not self._mcp_discovery_error:
+            return ""
+        return (
+            "\n\n<tool_availability_warning>\n"
+            "Some of your integration tools failed to load. "
+            "If the user asks for something that would need those tools, tell them the "
+            "integration is temporarily unavailable due to a backend error (don't guess, "
+            "improvise, or silently refuse as if the capability never existed).\n"
+            "</tool_availability_warning>"
+        )
+
     def _build_self_improvement_addendum(self) -> str:
         """Build the self-improvement decision tree for the system prompt.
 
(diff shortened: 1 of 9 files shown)


packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/mcp_errors.py
```diff
@@ -47,6 +47,60 @@ def is_retryable_mcp_error(error: Exception) -> bool:
...
+    can log the server's actual response body instead of just the status code —
+    raise_for_status() alone discards it.
+
+    Args:
+        error: Exception raised during MCP connection
+
+    Returns:
+        The httpx.HTTPStatusError if one is found, else None
+    """
+    if hasattr(error, "__class__") and error.__class__.__name__ == "ExceptionGroup":
```
alloy-cicd: **The unwrap misses the common failure shape this PR was written to diagnose.** This matches only a group whose class name is exactly `"ExceptionGroup"` and walks one level deep. anyio task groups in `mcp.client.streamable_http` frequently nest (`ExceptionGroup(ExceptionGroup(HTTPStatusError))`), and mixed-cancellation groups surface as `BaseExceptionGroup` — in both shapes this returns `None` and the new `"MCP gateway response ... body="` log line never fires, silently.

The existing `is_retryable_mcp_error`/`format_mcp_error` share the same idiom (this is now the third copy in this file), but for them the miss degrades to a generic message; here it removes the diagnostic entirely. The repo already has a correct recursive helper — `_flatten_exceptions` in `packages/voice-agent/voice_agent/agent.py:407` (`isinstance(exc, BaseExceptionGroup)` + recursion). Suggest promoting it into this module and implementing this function as:

```python
return next((e for e in _flatten_exceptions(error) if isinstance(e, httpx.HTTPStatusError)), None)
```

(The `hasattr(error, "__class__")` guard is also always true, and every package here requires Python ≥3.12, so `isinstance(error, BaseExceptionGroup)` is available natively.) Bonus: `is_retryable_mcp_error` and `format_mcp_error` can then reuse the same flattening and lose their duplicated one-level ladders.

alloy-cicd (commented):

The graceful-degradation goal is sound, but two findings undercut the PR's own purpose: the new `get_mcp_http_status_error` misses nested/`BaseExceptionGroup` shapes (so the new body logging silently doesn't fire in the common streamable-HTTP failure shape), and the degraded `[]` result is latched by `_ensure_agent`'s caching sentinels with `_mcp_discovery_error` never cleared. Details inline.

Notes that don't anchor to diff lines:

- **Pre-existing bug in the touched function** — `fetch_available_servers` builds its URL with `self.config.MCP_GATEWAY_URL.rstrip("/mcp")` (`discovery.py:231`). `str.rstrip` strips a *character set*, not a suffix: any gateway host ending in `m`/`c`/`p` (e.g. `https://example.com/mcp` → `https://example.co`) produces a wrong servers URL whose failure this PR's `return []` now swallows silently. `removesuffix("/mcp")` is the fix; worth doing while instrumenting this exact error path.
- **Tests** — the checklist claims thorough tests were added, but the PR touches only the three source files; no test covers `get_mcp_http_status_error`/`get_mcp_error_body`, the new `return []` degrade path, or the `<tool_availability_warning>` addendum (the test plan runs pre-existing suites). The unreachable-nested-group case in particular would have been caught by a unit test constructing `ExceptionGroup("x", [ExceptionGroup("y", [HTTPStatusError(...)])])`.
- **Auth-shaped failures are now invisible** — the `except Exception` in `_discover_mcp_tools` also covers the two `oauth2_client.exchange_token(...)` calls and 401/403 gateway responses, all non-retryable. An expired grant or misconfigured `MCP_GATEWAY_CLIENT_ID` no longer fails the turn; every affected user just silently gets a tool-less agent told to blame a "temporary backend error", so a permanent auth/config problem is reported as transient and nothing triggers re-auth or alerting. Consider re-raising (or at least distinguishing in the warning/logs) 401/403 and token-exchange failures.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -691,15 +699,26 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
+                    if http_err is not None:
+                        logger.error(
+                            f"MCP gateway response for {self.name} ({http_err.request.url}): "
+                            f"{http_err.response.status_code} body={get_mcp_error_body(http_err)!r}"
+                        )
+                    # Recorded so _ensure_agent can warn the LLM (and, through it,
+                    # the user) that some tools are missing — otherwise the agent
+                    # just looks silently less capable with no way to explain why.
+                    self._mcp_discovery_error = error_msg
+                    return []
```
alloy-cicd, line 814: **The degraded state is latched for the runnable's lifetime with no path back.** `_ensure_agent` stores this `[]` in `self._discovered_tools` (line 964 guards on `is None`, and `[]` is not `None`) and then sets `_cached_tools`/`_cached_system_prompt`, after which the whole method is a no-op (line 958). `_mcp_discovery_error` is never cleared anywhere. Before this PR the raise propagated before those sentinels were set, so the next construction retried.

Today the orchestrator rebuilds runnables per turn (`build_runtime_context` from `app/core/agent.py:342`), so the blast radius is one turn — but the per-invocation sandbox graph rebuild in `_astream_impl` reuses `_cached_tools`/`_cached_system_prompt`, and any longer-lived embedding of this class keeps telling the user the integration is down after a 3-second gateway blip has recovered.

Two cheap hardenings: reset `self._mcp_discovery_error = None` at the top of `_discover_mcp_tools`, and consider not conflating "discovery failed" with "discovered zero tools" (e.g. keep `_discovered_tools = None` on failure with a short negative-cache timestamp) so a reused instance can recover.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -710,7 +729,9 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
                 delay *= 2  # Exponential backoff
 
         # Should never reach here, but just in case
-        raise last_error or Exception(f"Failed to discover MCP tools for {self.name}")
+        logger.error(f"Failed to discover MCP tools for {self.name}: {last_error}")
+        self._mcp_discovery_error = format_mcp_error(last_error) if last_error else "unknown error"
```
alloy-cicd: **Dead code:** this tail is unreachable — on the last iteration (`attempt >= max_retries - 1`) the except block always returns at line 721, and success returns at line 693. `last_error` is now written only to feed this dead path, and its degrade copy diverges from the live one (raw `{last_error}`, no body logging, an "unknown error" string that can never appear). Suggest deleting the tail (or reducing it to a bare `return []`) and dropping the `last_error` bookkeeping, so the retry loop has exactly one failure exit.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -1039,6 +1060,21 @@ def _to_skill_def(s) -> AgentSkillDef:
...
             logger.debug(f"Added tool-catalog addendum to {self.name} system prompt")
 
+        # MCP gateway discovery degraded to zero tools rather than failing the turn
+        # (see _discover_mcp_tools) — without this, the agent just looks silently
+        # less capable, and the user/operator has no way to tell a broken gateway
+        # apart from the agent simply refusing to help.
+        if self._mcp_discovery_error:
+            system_prompt += (
+                "\n\n<tool_availability_warning>\n"
+                f"Some of your integration tools failed to load: {self._mcp_discovery_error}\n"
```
alloy-cicd: **Interpolating `format_mcp_error` verbatim into the system prompt has two costs:**

1. The formatted string embeds the internal gateway URL and status (e.g. `"MCP server returned HTTP 403 for https://…gatana…/mcp"`), and the fallback branch is the raw `str(error)` — gateway-controlled text. The instruction to relay unavailability to the user makes internal hostnames/status codes user-visible chat output, and an error message containing markup/instructions lands inside the prompt unescaped.
2. The warning text varies with gateway state (502 vs 503 vs timeout vs absent), so during a flapping gateway the otherwise-static system-prompt prefix changes turn to turn, invalidating the LLM prompt cache exactly when the system is already unhealthy.

A constant, URL-free sentence ("Some of your integration tools are temporarily unavailable due to a backend error…") keeps the user-facing behavior while logging the detailed `error_msg` server-side only — the log line at `_discover_mcp_tools` already carries it. For symmetry with `_build_preferences_addendum`/`_build_playbook_addendum` (and testability), a `_build_tool_availability_addendum()` seam would also fit this file's conventions.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -242,6 +247,12 @@ async def fetch_available_servers(self, token: str) -> List[Dict[str, Any]]:
 
         except Exception as e:
             logger.error(f"Failed to fetch MCP servers: {e}", exc_info=True)
+            http_err = get_mcp_http_status_error(e)
```
alloy-cicd: **Duplication + log-content caution.** This extract-and-log block is byte-for-byte the one in `dynamic_agent.py:711-716` (modulo the `self.name` prefix), and both call sites had to import the two helpers. A single `log_mcp_gateway_error(error, context)` next to the helpers in `mcp_errors.py` would keep the log format greppable/alertable from one place and stop the third copy-paste at the next call site — note this module already routes error reporting through `_log_discovery_error` (line 174), which has an `httpx.HTTPStatusError` branch this bypasses.

Also: unlike the streaming path, this is a plain buffered `httpx.AsyncClient.get`, so the body *is* readable — and a 401/403 error body from the gateway can echo auth context (token diagnostics, session identifiers) which now lands verbatim at ERROR level. Worth considering redaction or DEBUG level for the body on 4xx auth responses.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -223,7 +227,7 @@ async def fetch_available_servers(self, token: str) -> List[Dict[str, Any]]:
         try:
             # Call the MCP gateway API to get server list
             # Extract base URL from MCP_GATEWAY_URL (remove /mcp path)
-            base_url = self.config.MCP_GATEWAY_URL.rstrip("/mcp").rstrip("/")
+            base_url = self.config.MCP_GATEWAY_URL.removesuffix("/mcp").rstrip("/")
```
alloy-cicd: **Bug: operations are in the wrong order — trailing-slash URLs regress.**

`removesuffix("/mcp")` runs before `rstrip("/")`, so `MCP_GATEWAY_URL="https://gw.example/mcp/"` is a no-op for `removesuffix` (the string ends with `/`, not `/mcp`) and yields `https://gw.example/mcp` → `servers_url` becomes `https://gw.example/mcp/api/v1/mcp-servers`, the gateway 404s, and the `except Exception` below silently returns `[]` — the whole server catalog disappears for a config shape the old (admittedly buggy) char-set `rstrip("/mcp")` did handle.

```python
base_url = self.config.MCP_GATEWAY_URL.rstrip("/").removesuffix("/mcp")
```

strips the trailing slash first and preserves the intent of the fix for hosts ending in `m`/`c`/`p`.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -684,22 +745,35 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
                         )
                     else:
                         logger.error(f"Non-retryable error discovering MCP tools for {self.name}: {e}")
-                    raise
+                    log_mcp_gateway_error(logger, e, context=f"for {self.name} ")
+                    # Recorded so _ensure_agent can warn the LLM (and, through it,
+                    # the user) that some tools are missing — otherwise the agent
+                    # just looks silently less capable with no way to explain why.
+                    self._mcp_discovery_error = error_msg
+                    return []
```
alloy-cicd, line 814: **Bug: the degraded empty result is cached permanently — the runnable never recovers.**

Before this PR, a discovery failure raised, so `self._discovered_tools` stayed `None` and `_cached_tools` was never set — the next invocation re-ran discovery. Now this returns `[]`, and in `_ensure_agent`:

- `self._discovered_tools = []` is not `None`, so the `_discovered_tools is None` guard never re-triggers discovery;
- the `_cached_tools is not None` sentinel makes `_ensure_agent` a no-op thereafter;
- the tool-availability warning is baked into `_cached_system_prompt`.

So a 30-second Gatana blip during a runnable's first invocation pins it at zero MCP tools with a permanent "temporarily unavailable" warning for its entire lifetime, even after the gateway recovers. It also makes the `self._mcp_discovery_error = None` reset at the top of `_discover_mcp_tools` unreachable in production — only `test_stale_discovery_error_cleared_on_next_call` exercises it, testing a recovery path the real call graph can't produce.

Consider not caching the degraded result (e.g. leave `_discovered_tools`/`_cached_tools` unset when `_mcp_discovery_error` is set, or return `(tools, error)` and let `_ensure_agent` decide), so the next invocation retries discovery.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -684,22 +745,35 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
                     logger.info(f"Successfully discovered MCP tools for {self.name} on attempt {attempt + 1}")
                 return validated_tools
 
-            except Exception as e:
-                last_error = e
+            except _McpTokenExchangeError as auth_exc:
```
alloy-cicd: **Behavior change: token-exchange failures lost their retry.**

Before this PR, an `exchange_token` error fell into the generic `except Exception`, went through `is_retryable_mcp_error`, and a transient 503/timeout from the OIDC token endpoint was retried up to 3× with backoff. Now every exchange failure is wrapped in `_McpTokenExchangeError`, caught here first, and re-raised immediately on attempt 1 — the retryable check is never consulted for this failure class.

The comment says "propagate the original exception as before this path existed", but the propagation is *not* the same as before: the OAuth server can flap just like the gateway, and those turns previously survived via retry. If skipping retry for auth failures is intentional (plausible — bad credentials won't get better), the docstring/comment should say so explicitly rather than claiming parity; if not, the retryable check should still apply to exchange errors before propagating.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -684,22 +745,35 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
+                # A different failure class from the gateway/transport errors below
+                # (the user's own credentials, not the gateway being down) — don't
+                # degrade, propagate the original exception as before this path existed.
+                logger.error(f"Token exchange failed while discovering MCP tools for {self.name}: {auth_exc}")
+                raise auth_exc.__cause__ from None
```
alloy-cicd: **`raise auth_exc.__cause__ from None` mutates the original exception's causal chain.**

`raise X from None` sets `X.__cause__ = None` and `X.__suppress_context__ = True` on the re-raised object. If `exchange_token` raised its own wrapped error (e.g. `raise TokenExchangeError(...) from httpx.HTTPStatusError(...)`), the propagated exception has its underlying HTTP error stripped, so the failed-state traceback logged upstream loses the status/URL that explains *why* the exchange failed.

The `from None` is presumably there to suppress the `_McpTokenExchangeError` context noise, but the cure also deletes the original `__cause__`. A plain `raise auth_exc.__cause__` keeps the chain (at the cost of a "during handling…" line), or restructuring per the comment on the wrap blocks removes the need for the unwrap entirely. Also note: if `_McpTokenExchangeError` is ever raised without `from` (or `__cause__` is otherwise `None`), this becomes `raise None` → `TypeError: exceptions must derive from BaseException`, masking the real auth error.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -619,11 +674,14 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
-                            requested_scopes=["openid", "profile", "offline_access"],
-                        )
+                        try:
+                            mcp_gateway_token = await self.oauth2_client.exchange_token(
+                                subject_token=self.user_token,
+                                target_client_id=mcp_gateway_client_id,
+                                requested_scopes=["openid", "profile", "offline_access"],
+                            )
+                        except Exception as auth_exc:
+                            raise _McpTokenExchangeError(str(auth_exc)) from auth_exc
```
alloy-cicd: **The wrap/unwrap machinery could disappear, and token exchanges shouldn't re-run per retry.**

Two things converge here:

1. This try/except block is duplicated byte-for-byte at the console-backend exchange below (only `target_client_id` differs).
2. Because the exchanges sit inside the `for attempt` loop, every retry of a transient 502/503 from `client.get_tools()` re-executes up to two OAuth token-exchange round trips against the OIDC provider, even though the tokens from attempt 1 are still valid.

Since `_McpTokenExchangeError` exists solely to tunnel the exception out of a try block the exchange never needed to be inside, hoisting connection-building (both exchanges) above the retry loop — or into a small `async def _exchange_token_for(target_client_id)` helper called before the `try` — removes the sentinel class, both wrap sites, and the `__cause__` unwrap at line 753 in one move, and stops the redundant OIDC round trips. (This interacts with the retry-loss question raised at line 748 — whichever retry semantics you choose for exchange failures, this structure makes them explicit instead of encoded in a marker exception.)

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/mcp_errors.py
```diff
@@ -1,8 +1,26 @@
 """MCP error handling utilities for retry logic and user-friendly error messages."""
 
+import logging
+
 import httpx
 
 
+def _flatten_exceptions(error: BaseException) -> list[BaseException]:
```
alloy-cicd: **Knowing duplicate — consider making this the single copy.**

The docstring says this "mirrors the existing helper in packages/voice-agent/voice_agent/agent.py" (byte-for-byte equivalent at `agent.py:407`). voice-agent already depends on `ringier-a2a-sdk` (editable path dep in its pyproject.toml, and it imports `ringier_a2a_sdk.oauth.client`), so the import is feasible today: export this as a public `flatten_exceptions` and have voice-agent use it, deleting its private copy. Otherwise the two flatteners will drift independently — a future fix (e.g. `__cause__` chain handling or cycle protection) lands in one and not the other, and voice-agent classifies the identical exception tree differently from every other package.

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/mcp_errors.py
```diff
@@ -23,30 +41,87 @@ def is_retryable_mcp_error(error: Exception) -> bool:
...
-
-    # Handle direct httpx exceptions
-    if isinstance(error, httpx.HTTPStatusError):
-        status_code = error.response.status_code
-        return status_code in (502, 503, 504)
-    elif isinstance(error, httpx.TimeoutException):
-        return True
+    for leaf in _flatten_exceptions(error):
+        if isinstance(leaf, httpx.HTTPStatusError):
+            return leaf.response.status_code in (502, 503, 504)
```
alloy-cicd: **Minor (pre-existing semantics, but the rewrite is the moment to fix it): first leaf masks later retryable leaves.**

The loop returns on the *first* `HTTPStatusError` leaf, so `ExceptionGroup([HTTPStatusError(400), ConnectTimeout()])` — plausible when an anyio task group fails over two MCP connections (gateway + console) — returns `False` and a genuinely transient timeout is never retried. The old code had the same first-match behavior, so this isn't a regression, but since the function was rewritten anyway:

```python
return any(
    isinstance(leaf, httpx.TimeoutException)
    or (isinstance(leaf, httpx.HTTPStatusError) and leaf.response.status_code in (502, 503, 504))
    for leaf in _flatten_exceptions(error)
)
```

treats a group as retryable if *any* leaf is.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -242,6 +246,7 @@ async def fetch_available_servers(self, token: str) -> List[Dict[str, Any]]:
 
         except Exception as e:
             logger.error(f"Failed to fetch MCP servers: {e}", exc_info=True)
+            log_mcp_gateway_error(logger, e)
```
alloy-cicd: **Minor: double ERROR logging for one failure.**

`logger.error(..., exc_info=True)` on the line above already emits the full traceback (including the HTTPStatusError's status and URL); this call emits a second ERROR record repeating status + URL. The only new information is the response body. During a gateway outage every `fetch_available_servers` call writes two ERROR records, doubling log volume and duplicating any alerting keyed on ERROR level. Consider one combined log call (e.g. append `get_mcp_error_body(...)` to the existing message when `get_mcp_http_status_error(e)` returns non-None).

alloy-cicd (commented):

The degrade-gracefully direction is sound and the new `mcp_errors` helpers (recursive flattening, body capture with the streaming caveat, 401/403 redaction) are well thought out. The review found **2 correctness bugs**, **2 behavior changes worth an explicit decision**, and a few maintainability items, posted as inline comments:

**Correctness**
1. `discovery.py:230` — the `removesuffix`/`rstrip` order regresses trailing-slash `MCP_GATEWAY_URL` configs (silently returns zero servers).
2. `dynamic_agent.py:776` — a degraded empty tool list is cached permanently by `_ensure_agent`, so a transient gateway blip pins the runnable at zero tools (with the "temporarily unavailable" warning baked into the cached system prompt) for its whole lifetime.

**Behavior changes / diagnosability**
3. `dynamic_agent.py:748` — token-exchange failures lost their retry: transient 503s/timeouts from the OIDC endpoint now fail the turn on attempt 1, contradicting the "same as before" comment.
4. `dynamic_agent.py:753` — `raise auth_exc.__cause__ from None` wipes the original exception's own `__cause__`, discarding the underlying HTTP status/URL from auth-failure tracebacks.

**Maintainability** — duplicated `exchange_token` wrap blocks re-run on every retry attempt; `_flatten_exceptions` duplicates the voice-agent helper; first-leaf masking in `is_retryable_mcp_error`; double ERROR logging in `fetch_available_servers`.

Not inlined (minor): the broad `except Exception` degrade path also swallows programming errors (e.g. a bug in `_validate_tool_schema` would masquerade as a gateway outage); the retry loop is now a diverged near-copy of `ToolDiscoveryService._get_tools_with_retry` (which did not get `log_mcp_gateway_error`); the `_http_error` test helper now exists in three test files (a shared conftest helper would do); and the comment at `dynamic_agent.py:761-763` asserting the loop's control-flow shape will silently go stale on the next edit.

*Submitted as COMMENT — the human reviewer decides.*

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -685,21 +748,37 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
+                        # Don't degrade — propagate so the turn fails loudly,
+                        # same as before this degrade path existed for
+                        # gateway/transport errors (see docstring).
+                        logger.error(
+                            f"Token exchange failed while discovering MCP tools for {self.name} "
+                            f"(attempt {attempt + 1}): {e}"
+                        )
+                        raise
+
+                    # Non-retryable error or exhausted retries — degrade gracefully
```
alloy-cicd, line 800: **Correctness:** the degrade path triggers for *any* exception that isn't a token-exchange failure — not just gateway/transport errors. A `ValueError` from `_validate_tool_schema`, a `TypeError` out of `MultiServerMCPClient` after a dependency bump, an `AttributeError` in the whitelist filtering — all are non-retryable and non-auth, so they log, set `_mcp_discovery_error`, and return `[]`. A permanent programming bug is then reported to the user forever as “temporarily unavailable due to a backend error” and never fails a turn again (pre-PR it propagated loudly), while `log_mcp_gateway_error` is a no-op for it (no `HTTPStatusError` to find).

The docstring scopes degradation to “a dead/misconfigured MCP transport/gateway (4xx…)” — consider making the code match, e.g. only degrade when `get_mcp_http_status_error(e) is not None` or the leaves are httpx transport errors (`ConnectError`/`TimeoutException`), and re-raise everything else.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -931,17 +1007,29 @@ async def _ensure_agent(self) -> None:
...
+        After a call that resolved cleanly, this is a no-op (guarded by
+        _cached_tools). A call that degraded (_mcp_discovery_error set) is
+        NOT treated as resolved — the guard lets the next call retry
+        discovery instead of permanently pinning this instance at zero MCP
+        tools because of one transient gateway blip.
         """
-        # Already resolved — skip
-        if self._cached_tools is not None:
+        # Already resolved — skip, unless the last attempt degraded.
+        if self._cached_tools is not None and self._mcp_discovery_error is None:
```
alloy-cicd, line 1057: **Design/efficiency:** while the gateway stays down, this guard makes every invocation re-run the *entire* `_ensure_agent` body — not just MCP discovery: `_discover_console_self_improvement_tools()` (another token exchange + MCP round trip), `resolve_skills_for_agent()` (store reads), playbook addendum, prompt rebuild, and graph compile, plus the 3-attempt discovery with 1s+2s backoff sleeps. During a sustained outage that's roughly 3–5s of added latency and extra gateway/store load on every turn of every affected sub-agent, with no cooldown — and orchestrator delegations can invoke the same instance several times per turn. Retrying is the right call; consider bounding it (e.g. re-attempt discovery at most once per N seconds via a monotonic timestamp) and/or scoping the retry to the failed discovery step so the memoized skills/graph work isn't redone.

Also: `packages/agent-common/AGENTS.md` (key design decision 1) still documents “After first call, it's a no-op” — worth updating in this PR since that's exactly the invariant being changed.

alloy-cicd (commented):

Reviewed `b825ca1` against `main`. The core approach looks sound and is well-tested: gateway/transport failures degrade to an empty tool list instead of crashing embedded sub-agents, token-exchange failures still propagate, and the tool-availability addendum deliberately avoids interpolating gateway-controlled text (good call, including the prompt-cache rationale). The `removesuffix("/mcp")` fix and the recursive `flatten_exceptions` consolidation are strict improvements with regression tests.

5 inline comments below (1 correctness, 1 design/efficiency, 3 minor). Three repo-wide observations that can't be anchored to diff lines:

- **The `rstrip("/mcp")` bug fixed here is still live in console-backend**: `packages/console-backend/console_backend/routers/mcp_router.py:470` has the identical base-URL corruption (`config.mcp_gateway.url.rstrip("/mcp").rstrip("/")`). Any gateway host whose last label ends in `m`/`c`/`p` gets truncated, and the request carries the user's exchanged bearer token to the wrong host. Same one-line fix applies — worth including here or as an immediate follow-up.
- **Retry-loop triplication**: this loop is now the third near-identical copy (also `orchestrator-agent/app/core/discovery.py::_get_tools_with_retry` and `ringier_a2a_sdk/agent/langgraph.py`), and the PR's improvements (gateway body logging, nested-group handling on exhaustion) land only in this copy — the other two still produce the bodyless log this PR exists to fix. Consider extracting a shared `retry_mcp_call(...)` into `ringier_a2a_sdk.utils.mcp_errors`, which both other call sites already import from.
- **Scope note**: the availability warning only covers the whitelist (`mcp_tools`) discovery path. The GP/orchestrator path (`inject_all_tools`) sources tools from `fetch_available_servers`, which still degrades to `[]` silently — and the executor caches that empty result for the discovery-cache TTL — so the silent-capability-loss behavior this PR fixes for sub-agents remains for the agent users talk to most. Fine as a follow-up; flagging so it's tracked.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -619,11 +674,15 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
-                            target_client_id=mcp_gateway_client_id,
-                            requested_scopes=["openid", "profile", "offline_access"],
-                        )
+                        try:
+                            mcp_gateway_token = await self.oauth2_client.exchange_token(
+                                subject_token=self.user_token,
+                                target_client_id=mcp_gateway_client_id,
+                                requested_scopes=["openid", "profile", "offline_access"],
+                            )
+                        except Exception:
```
alloy-cicd: **Simplification:** this 8-line wrapper is duplicated verbatim at the console-backend exchange below, and the “auth failures must never degrade” invariant only holds if every `exchange_token` call inside the `try` remembers to set the flag — a future third call site would silently lose the protection, degrading exactly the failure class the docstring says must never degrade, with nothing structural to catch it. A small helper (e.g. `async def _exchange_for(self, target_client_id: str) -> str` that raises a typed `TokenExchangeError`, or typing the exception in the OAuth client itself) would encode the classification in the exception type: both call sites become one `await`, and the handler becomes `except TokenExchangeError: raise` with no mutable flag.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -242,6 +247,7 @@ async def fetch_available_servers(self, token: str) -> List[Dict[str, Any]]:
...
```
alloy-cicd: **Minor (log noise):** the line above already logs this exception at ERROR with `exc_info=True`, so every gateway outage now emits two ERROR records per call. In this non-streaming path the response body was already reachable (the response is buffered), so the helper's main added value here is the 401/403 redaction. Consider collapsing to one record — e.g. drop the first line to `logger.debug(..., exc_info=True)` and let `log_mcp_gateway_error` be the single ERROR, or skip the helper in this caller.

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/mcp_errors.py
```diff
@@ -21,32 +40,93 @@ def is_retryable_mcp_error(error: Exception) -> bool:
...
+        logger: The caller's module logger
+        error: Exception raised during MCP connection
+        context: Optional short prefix identifying the caller (e.g. "for my-agent "),
+            included verbatim before the URL in the log line
+    """
+    http_err = get_mcp_http_status_error(error)
+    if http_err is None:
+        return
+    status = http_err.response.status_code
+    body = "<redacted: auth error body withheld from logs>" if status in (401, 403) else get_mcp_error_body(http_err)
```
alloy-cicd, line 165: **Minor (security):** status-based redaction may be the wrong discriminator. The PR's own motivating case — a disabled Gatana account — can arrive as a 400 whose body echoes request/session context, and that body lands verbatim (up to 2000 chars) in log storage via the non-redacted branch; same for a 500 from an auth proxy. Consider redacting all 4xx, or scrubbing token-shaped substrings, rather than trusting 401/403 to be the only statuses whose bodies carry credentials.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -685,21 +749,48 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
+                        # Don't degrade — propagate so the turn fails loudly,
+                        # same as before this degrade path existed for
+                        # gateway/transport errors (see docstring).
+                        logger.error(
+                            f"Token exchange failed while discovering MCP tools for {self.name} "
+                            f"(attempt {attempt + 1}): {e}"
+                        )
+                        raise
+
+                    if not is_mcp_transport_error(e):
```
alloy-cicd: **Correctness (verified against the pinned `mcp` 1.28.1 SDK):** `is_mcp_transport_error` only matches `httpx.HTTPError` leaves, but realistic gateway failures reach this branch as `mcp.shared.exceptions.McpError` with no httpx leaf at all:

- `streamable_http.py` intercepts an HTTP **404** before `raise_for_status()` and synthesizes `JSONRPCError("Session terminated")`, raised as `McpError` — so a route-miss/stale-session during a Gatana redeploy never becomes an `httpx.HTTPStatusError`.
- Any JSON-RPC error the gateway returns during `initialize`/`tools/list` (gateway up, downstream MCP server down — the canonical degradation mode for an MCP *gateway*) also surfaces as `McpError`.

Both crash the turn via this `raise`, and the comment labels them "a bug in our own code" when they are unambiguously remote-side failures. This preserves pre-PR behavior (gap, not regression), but it defeats the degrade path for a realistic class of gateway outages. Consider treating `McpError` leaves as transport errors in `is_mcp_transport_error` (or adding them to the degrade condition here).

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -579,16 +615,30 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
+            or an empty list if MCP transport/gateway discovery fails after retries.
 
         Raises:
-            Exception: If MCP discovery fails (will result in failed state)
+            Exception: If exchanging the user's token fails (the original
+                exception, not wrapped) — this is not a gateway/transport error.
         """
         import asyncio
 
+        self._mcp_discovery_error = None
```
alloy-cicd: **Correctness:** clearing `self._mcp_discovery_error` unconditionally at the top of every discovery attempt creates two failure modes with the new `_ensure_agent` guard (`_cached_tools is not None and _mcp_discovery_error is None`):

1. **Degrade → raise → stale pin (within one turn):** invocation 1 degrades (error set, degraded `_cached_tools`/prompt cached); invocation 2 re-enters, this line clears the flag, then the auth-exchange or non-transport branch raises. Invocation 3+ sees `_cached_tools` set and error `None`, so the guard short-circuits and the intended retry never happens for the rest of the turn — contradicting the design comment at the guard. (Scope is bounded: `build_runtime_context` rebuilds the runnable each `stream()` call, so this resets on the next user turn, and the cached prompt does keep the warning addendum.)
2. **Parallel task-call race:** the orchestrator can dispatch parallel `task` calls to the same sub-agent (see `dynamic_tool_dispatch.py`'s parallel-call handling), and `_ensure_agent` has no lock. Two overlapping calls can interleave so that a degraded call's empty tool list overwrites a successful call's tools while this reset clears the error flag — ending with error `None`, cached tools missing all MCP tools, and no warning, permanently for the instance.

A fix for both: don't reset the flag here — set it only at the two terminal outcomes of this method (success → `None`, degrade → message), and consider an `asyncio.Lock` around `_ensure_agent`'s init section.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -931,17 +1019,29 @@ async def _ensure_agent(self) -> None:
...
+        After a call that resolved cleanly, this is a no-op (guarded by
+        _cached_tools). A call that degraded (_mcp_discovery_error set) is
+        NOT treated as resolved — the guard lets the next call retry
+        discovery instead of permanently pinning this instance at zero MCP
+        tools because of one transient gateway blip.
         """
-        # Already resolved — skip
-        if self._cached_tools is not None:
+        # Already resolved — skip, unless the last attempt degraded.
+        if self._cached_tools is not None and self._mcp_discovery_error is None:
```
alloy-cicd, line 1057: **Efficiency:** while `_mcp_discovery_error` is set, this guard never short-circuits, so every subsequent invocation of the same instance re-runs the *entire* lazy init — gateway discovery (up to 3 attempts with 1s+2s sleeps on retryable errors), `_discover_console_self_improvement_tools` (its own token exchange + MCP round trip, unrelated to the gateway failure), skills resolution from the Postgres store, playbook reads, and a full graph recompile — with no cooldown.

Since instances are per-turn, the blast radius is repeated `task` calls to the same sub-agent within one orchestrator turn (e.g. parallel or sequential fan-out): each degraded call pays ~3s of backoff plus a console round trip, a DB query, and a graph compile that produced identical results the call before. Consider scoping the retry to the gateway-discovery step only (keep console tools, skills, and playbook cached in their own fields) so a degraded retry is a single probe rather than a full re-init.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -604,11 +654,17 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
 
         for attempt in range(max_retries):
+            # Set when a failure below came from exchange_token() rather than
+            # the MCP gateway/transport itself — a different failure class
+            # (the user's own credentials, not the gateway being down) that
+            # must not be degraded to []. Reused as-is by is_retryable_mcp_error
+            # for the retry decision, so a transient OIDC hiccup still gets
+            # retried the same as a transient gateway one; only a non-retryable
+            # or exhausted-retries auth failure skips the degrade and raises.
+            is_auth_exchange_error = False
```
alloy-cicd: **Minor (robustness):** `is_auth_exchange_error` is a loop-scoped flag set by two identical `except Exception: is_auth_exchange_error = True; raise` wrappers and consumed ~90 lines later. The classification is a property of the exception, so a typed wrapper makes it impossible to forget: a small `_exchange_token_for(target_client_id, label)` helper that raises `TokenExchangeError(...) from e`, with `except TokenExchangeError` in the handler (retry classification on `e.__cause__`). If a third exchange site is ever added without the flag, an expired-credential failure would silently take the degrade path and be reported to the user as "integration temporarily unavailable" — exactly the confusion the auth carve-out exists to prevent — and nothing would catch it.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -685,21 +749,48 @@ async def _discover_mcp_tools(self) -> List[BaseTool]:
...
+                        # loudly like any other programming error.
+                        logger.error(
+                            f"Unexpected (non-MCP-transport) error discovering MCP tools for {self.name}: {e}"
+                        )
+                        raise
+
+                    # Non-retryable error or exhausted retries — degrade gracefully
+                    # instead of crashing the whole turn (see docstring). This is
+                    # always the loop's last possible outcome: every path either
+                    # returns here or above, so there is no unreachable tail after it.
```
alloy-cicd: **Minor (defensive):** the removed `raise last_error or ...` tail was the executable guarantee that this `-> List[BaseTool]` method can't fall off the end of the retry loop; its replacement is this comment. That holds today (`max_retries = 3`, and the last attempt always returns or raises), but agent-common has no mypy gate, so if the loop structure is ever changed (configurable retries, an added `continue`), an implicit `None` return would land in `_discovered_tools` with `_mcp_discovery_error` unset — a silently tool-less agent with no warning and no log. A one-line `raise AssertionError("unreachable: retry loop exited without returning")` after the loop keeps the invariant executable.

packages/orchestrator-agent/tests/test_discovery.py
```diff
@@ -132,6 +132,38 @@ def test_initialization(self):
         assert service.config == config
         assert service.oauth2_client == oauth2_client
 
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        ("gateway_url", "expected_servers_url"),
+        [
+            # A host ending in one of the stripped characters (m/c/p) is exactly
+            # the case rstrip("/mcp") used to corrupt — regression coverage.
+            ("https://alloych.gatana.ai/mcp", "https://alloych.gatana.ai/api/v1/mcp-servers"),
```
alloy-cicd, line 147: **Test coverage (verified by execution):** this comment is inaccurate and the suite doesn't actually cover the regression. `alloych.gatana.ai` ends in `i`, not one of `m`/`c`/`p`, so the old `rstrip("/mcp")` produces the identical (correct) result for **all three** parametrized cases — including the trailing-slash one. Reintroducing the bug would pass this test.

The shape `rstrip("/mcp")` actually corrupts is a host ending in one of those characters, most commonly any `.com` domain: `"https://gateway.example.com/mcp".rstrip("/mcp")` → `"https://gateway.example.co"`. Suggest adding `("https://gateway.example.com/mcp", "https://gateway.example.com/api/v1/mcp-servers")` and fixing the comment. The identical fix in `console_backend/routers/mcp_router.py` has no test at all.

packages/console-backend/console_backend/routers/mcp_router.py
```diff
@@ -466,8 +466,11 @@ async def list_mcp_servers(
 
     try:
         # Call the Gatana gateway's /api/v1/mcp-servers endpoint to get the server list
-        # Extract base URL from mcp_gateway.url (remove /mcp suffix if present)
-        base_url = config.mcp_gateway.url.rstrip("/mcp").rstrip("/")
+        # Extract base URL from mcp_gateway.url (remove /mcp suffix if present). Strip
+        # the trailing slash first: rstrip("/mcp") strips a character *set* (/,m,c,p),
+        # not the literal suffix, so a host ending in one of those letters gets
+        # corrupted, and removesuffix("/mcp") alone is a no-op on ".../mcp/".
+        base_url = config.mcp_gateway.url.rstrip("/").removesuffix("/mcp")
```
alloy-cicd, line 473: **Reuse:** this is now the third live copy of the "/mcp suffix → API base" derivation — the same expression is at `orchestrator-agent/app/core/discovery.py:232`, and `console_backend/services/mcp_gateway_server_access_service.py:58-64` (`api_base_url`, feeding `GatanaClient` for SCIM/team resolution) implements the same semantics a third way. This PR fixed two copies and left the third untouched; the next URL-shape change will likely miss one and surface as unexplained 404s from a single service. The SDK already holds shared gateway URL logic (`ringier_a2a_sdk/utils/gateway.py`, currently LLM-gateway only) — a `mcp_gateway_api_base(url)` helper next to the new `mcp_errors` utilities would collapse all three.

Related asymmetry: the new tests bless a trailing-slash `MCP_GATEWAY_URL` for the server-list path, but the raw value still flows unmodified into `discovery.py:394`'s query-string URL and `dynamic_agent.py:697`'s `StreamableHttpConnection` — so a trailing-slash config value is now half-supported. A shared normalization point would fix that too.

alloy-cicd (commented):

Reviewed the degrade-gracefully change across `dynamic_agent.py`, `mcp_errors.py`, `discovery.py`, `mcp_router.py`, and the voice agent. The overall direction is sound and the auth-vs-transport distinction is well reasoned. Seven inline findings below: two correctness issues in the new degrade path (the `McpError` gap and the discovery-state reset), one test-coverage gap that was verified by actually running the old code against the new test cases, one efficiency concern, one duplication concern, and two minor robustness notes.

Not anchorable inline (outside the diff), for completeness:
- `console_backend/routers/mcp_router.py:386` and `:505` still log raw `e.response.text` with no truncation and no 401/403 redaction — the exact policy this PR establishes in `log_mcp_gateway_error`. Consider switching both to the new helper.
- `ringier_a2a_sdk/agent/langgraph.py:469` (the SDK's own MCP discovery failure handler) still logs only `format_mcp_error(e)`, so the motivating disabled-account 400 body stays invisible on that path; a `log_mcp_gateway_error` call there would close the loop.
- `console_backend/services/mcp_gateway_server_access_service.py:57-64` is a third copy of the `/mcp` → API-base derivation (see the inline comment on `mcp_router.py`).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (bc6152cc0e4dc6c25c0f27fd44b023a8b65ca4f0..8ccc65b0d32142bd96a98cc67653a00790e6843d)

Every previous finding is resolved or explicitly deferred with reasoning, and the delta introduces nothing new. **No inline comments this round.**

**Ledger of round-4 findings:**

1. **`McpError` bypasses the degrade path** — ✅ verified-fixed. `is_mcp_transport_error` now matches `(httpx.HTTPError, McpError)` with a defensive import; I verified the `McpError = ()` ImportError fallback behaves correctly (nested empty tuple in `isinstance` never matches), and re-ran the 4 new tests locally (green).
2. **Discovery-error reset → stale pin / race** — ✅ verified-fixed. The unconditional clear at method entry is gone; the flag is now cleared only on actual successful resolution, and the new `test_discovery_error_survives_a_later_raising_call` regression test covers the degrade-then-raise sequence (re-ran locally, green). *Residual, non-blocking:* `_ensure_agent` still has no lock, so concurrent task calls to the same instance with divergent outcomes can last-writer-pin a degraded (warning-bearing) cache for the remainder of one turn — far weaker than the original finding (the warning survives and it self-heals next turn); fits naturally into the nannos#138 design work.
3. **Full lazy-init re-run while degraded** — ⏸ rejected-by-author / deferred to nannos#138 with a design sketch. Reasonable disposition; not re-raised.
4. **Regression test didn't cover the `rstrip("/mcp")` bug** — ✅ verified-fixed. The new `.com`-host case is exactly the shape that fails under the old code (verified by execution last round), and the misleading comment is corrected. Re-ran the parametrized suite locally (green).
5. **URL-derivation drift (3 copies)** — ⏸ deferred to nannos#138; the author checked `api_base_url` and it is indeed already correct (consistent with my own verification — the finding was about duplication, not a bug there). Not re-raised.
6. **`is_auth_exchange_error` flag → typed exception** — ✅ verified-fixed. `_TokenExchangeError` raised only from the single `_exchange_token_for()` helper makes the never-degrade invariant structural; classification and retry now run on `__cause__`, so transient OIDC errors keep their retry, and the bare-object re-raise preserves the original cause chain (the `from None` mistake from round 3 is not repeated).
7. **Retry-loop fall-through invariant prose-only** — ✅ verified-fixed with an executable `AssertionError` tail.

**Delta review (new issues):** none found. The `classify_target` routing is correct in all three handler branches (auth raise, non-transport raise, degrade), `format_mcp_error`/`log_mcp_gateway_error` handle the `McpError` shape gracefully, and the test additions are sound. Local re-runs: `ringier-a2a-sdk/tests/test_mcp_errors.py` 28 passed; `agent-common` discovery/ensure_agent subset 3 passed; `orchestrator-agent` URL-derivation cases 4 passed.

Issue #136, opened 2026-08-12
combined_middlewares is None when tool-catalog middleware appends on the client-action path


aartaria:
## Problem

On `feature/nannos-embedded` (PR #116), `DynamicLocalAgentRunnable._build_graph` fails with:

```
AttributeError: 'NoneType' object has no attribute 'append'
```

at `packages/agent-common/agent_common/agents/dynamic_agent.py:1220`:

```python
combined_middlewares = combined_middlewares or None
...
if exposed_catalog is not None and not code_interpreter_ptc_enabled():
    combined_middlewares.append(ToolCatalogMiddleware(exposed_catalog))
```

`combined_middlewares or None` collapses an empty list to `None`, but the tool-catalog branch below unconditionally calls `.append()` on it — so any agent with an empty middleware list up to that point (the client-action path, per repro below) crashes as soon as it also needs the tool-catalog middleware.

Repro: `uv run pytest -q tests/test_tool_catalog.py::TestDynamicAgentCatalogMode::test_build_graph_forwards_context_registry_flag tests/test_tool_catalog.py::TestDynamicAgentCatalogMode::test_build_graph_adds_catalog_middleware_without_ptc` in `packages/agent-common`, on `feature/nannos-embedded`.

This code path does not exist on `main` (it's specific to the embed-sdk client-action work), so it doesn't block anything there — confirmed by running the same tests against a fresh `origin/main` worktree (16/16 pass).

## Fix

Move the `combined_middlewares or None` collapse to after the tool-catalog (and any other) append, or guard the append with `combined_middlewares = combined_middlewares or []` instead of `or None`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

aartaria:
🤖 Claude Code session — [resume in VSCode](https://vscode.dev/redirect?url=vscode%3A%2F%2Fanthropic.claude-code%2Fopen%3Fsession%3Db0c7a20d-4c03-4d3f-9ba2-9b9c48d265cc) _(resumes on aartaria's machine, with the nannos workspace open)_

**Session state** _(updated 2026-08-13)_
- [x] reproduce failing tests on `feature/nannos-embedded`
- [x] fix `combined_middlewares` None-collapse in `dynamic_agent.py`
- [x] tests green locally (`packages/agent-common`, 741 passed)
- [x] PR opened: ringier-data/nannos#139 (base: `feature/nannos-embedded`)
- [x] auto-review: clean, zero findings
[...]

PR #139, opened 2026-08-13
fix(agent-common): don't collapse combined_middlewares to None before catalog append

aartaria:
closes ringier-data/nannos#136

## Checklist

## Summary

`_build_graph` collapsed `combined_middlewares` to `None` (`or None`) *before* the tool-catalog branch, which unconditionally calls `.append()` on it — so any agent with an empty middleware list up to that point (the client-action path) crashed with `AttributeError: 'NoneType' object has no attribute 'append'` when it also needed the tool-catalog middleware.

Fix: keep `combined_middlewares` a list all the way to the `build_sub_agent_graph` call. No `or None` needed at the call site either — `build_sub_agent_graph` guards with `if extra_middlewares:`, so an empty list and `None` behave identically.

Covered by the existing repro tests (`tests/test_tool_catalog.py::TestDynamicAgentCatalogMode`), which also assert `extra_middlewares` is iterable. Full `agent-common` suite: 741 passed.

### DIFF
--- a/packages/agent-common/agent_common/agents/dynamic_agent.py
+++ b/packages/agent-common/agent_common/agents/dynamic_agent.py
@@ -1193,7 +1193,6 @@ class DynamicLocalAgentRunnable(StructuredResponseMixin, LocalA2ARunnable):
             from agent_common.middleware.client_objects_middleware import ClientObjectsMiddleware
 
             combined_middlewares.append(ClientObjectsMiddleware())
-        combined_middlewares = combined_middlewares or None
 
         # Catalog mode without PTC: contribute the native discovery surface
         # (search_tools/describe_tool) and the call_tool dispatch. Inserted with the


alloy-cicd (commented):

**No findings.** The single-line deletion of `combined_middlewares = combined_middlewares or None` checks out:

- With the line removed, `combined_middlewares` is always a list, so the subsequent `combined_middlewares.append(ToolCatalogMiddleware(...))` (`dynamic_agent.py:1211`) can no longer hit `None` — matching the PR description.
- The only behavior the deleted line provided was normalizing `[]` → `None` before `build_sub_agent_graph(extra_middlewares=...)`. The sole consumer (`packages/agent-common/agent_common/core/graph_utils.py:1761`) guards with `if extra_middlewares:`, where `[]` and `None` are equivalent, so nothing is lost.
- `_build_graph`'s signature and return are unchanged; callers are unaffected. Tests in `packages/agent-common/tests/test_tool_catalog.py` now pin the list-typed behavior.

Clean review — no inline comments.

PR #141, opened 2026-08-13
feat(orchestrator): identity-scoped tools with first-use consent gate (Gate 3)

### LINKED ISSUE #140: Orchestrator middleware: force-populate + hide reserved nannos__user_identity tool argument, gated by first-use consent
## Context

Some MCP hosts can only offer Nannos a single shared/service-account credential (no per-user OAuth)

The one reliable channel for getting verified per-user identity into such a tool is the tool-call **argument** payload (not headers, not the credentials object — both were ruled out). This issue is the generic, reusable orchestrator-side mechanism that makes that channel safe to use.

## What to build

**1. `nannos__user_identity` convention + middleware.**
Any MCP tool whose input schema declares a field named `nannos__user_identity` becomes an *Identity-scoped tool*. Orchestrator middleware (likely alongside `DynamicToolDispatchMiddleware`/`wrap_tool_for_ptc` in `packages/orchestrator-agent`) must, for such tools:
- Hide the `nannos__user_identity` field from the tool schema the model sees (it must never be a model-controllable argument).
- Force-populate it from the verified `GraphRuntimeContext.email` at dispatch time, overwriting whatever value (if any) the model attempted to supply. Same "no LLM round-trip" principle already used for `pending_file_blocks" (see `GraphRuntimeContext` in `packages/orchestrator-agent/app/models/config.py`).
- Payload is the email only — no other `GraphRuntimeContext` fields, to keep PII exposure minimal.

**2. First-use consent gate (Gate 3).**
This is the first real implementation of a consent gate scoped specifically to the identity-disclosure capability class:
- Before the *first* dispatch of any Identity-scoped tool for a given `(user, tool)` pair, prompt for explicit consent ("this tool will use your email address to scope Salesforce access to your own records" or similar, tool-specific wording) and remember the answer.
- On rejection: block only the identity-requiring tool(s) — not the whole integration/server — surfaced to the orchestrator as `auth_required` (reuse the existing sub-agent `auth_required" state handling already in the orchestrator system prompt; no new state-machine branch needed).
- Fail **closed** or timeout/no-answer (never proceed without an explicit grant).
- Do **not** gate this via the per-user `tool_bypass_rules` HITL config — that table encodes a user's own risk tolerance for action-approval, a different axis from "is this integration trustworthy enough to receive PII," and conflating them gives wrong answers in both directions.

aartaria:
## Summary

Implements the generic orchestrator-side mechanism for **Identity-scoped tools** (ADR 0006, shared-credential tier): any MCP tool whose input schema declares the reserved field `nannos__user_identity` gets the verified user identity injected safely, gated by first-use consent.

closes ringier-data/nannos#140

### 1. `nannos__user_identity` convention (`app/middleware/identity_scoped.py`)

- **Wrapped at discovery time** — in `agent_common.core.identity_scoped`, applied where MCP tools materialize (orchestrator `ToolDiscoveryService` **and** the dynamic sub-agent's own MCP rediscovery), with an idempotent backstop in `build_runtime_context` — so:
  - the reserved field is hidden from **every** model-facing schema render — regular tool dicts, toolset selector, and PTC signature renders all derive from the wrapper's cleaned `args_schema`;
  - **every execution path** goes through the wrapper — regular ToolNode dispatch, PTC `eval` (`wrap_tool_for_ptc` wraps the wrapper), and sub-agents.
- At execution the wrapper force-populates the field from the verified `GraphRuntimeContext.email`, **overwriting** any model-supplied value (no LLM round-trip, same principle as `pending_file_blocks`). Payload is the email only.
- **Fails closed**: without a remembered consent grant (or without a verified email on the context) it returns an `auth_required` error payload instead of executing — this is the enforcement backstop on paths where no interrupt is possible (PTC `eval`, sub-agents).

### 2. First-use consent gate — Gate 3

- `IdentityConsentMiddleware.aafter_model` raises a **HITL-shaped `interrupt()`** before the first dispatch per `(user, tool)` — reuses the existing approval-widget payload shape (`action_requests`/`review_configs`/`_call_id`) and the executor's existing resume map, so no frontend or executor changes are needed for the prompt itself.
- Answers are remembered per `(user, tool)` (keyed by tool name alone — server-slug resolution is path-dependent):
  - in-session on `GraphRuntimeContext.identity_consent_grants`;
  - durably via the same pattern as bypass rules: `_pending_identity_consents` → `persist_identity_consents` → new `PUT /api/v1/auth/me/settings/identity-consent` → new `user_settings.identity_consent_grants` jsonb column (migration 078).
- **Rejection** blocks only the identity-requiring tool (artificial `ToolMessage` with `error_code: auth_required`, which the orchestrator's existing auth-required handling understands); other tools of the integration stay available. A remembered denial blocks without re-prompting (`remove` on the endpoint resets it).
- **Timeout/no-answer fails closed** by construction (graph stays interrupted; a missing decision in the resume payload defaults to reject without remembering a denial).
- Deliberately **not** gated via `tool_bypass_rules` — different axis (ADR 0006); consent grants live in their own column/endpoint.
- Ordering: placed after the risk-HITL middleware in the stack list so its `aafter_model` runs **first** — a consent-rejected call is dropped before it can be risk-prompted.

## Out of scope / follow-ups

- Settings UI for viewing/resetting remembered consent answers (the endpoint already supports `remove`; grants surface in `GET /me/settings`).
- Interactive consent prompt *inside* PTC `eval` / sub-agent dispatch — those paths fail closed with `auth_required` until consent is granted via a direct orchestrator call.

## Testing

- `packages/orchestrator-agent`: 19 new tests (detection, schema hiding, force-populate/overwrite, fail-closed variants, consent middleware approve/reject/remembered-denial/mismatch) — full suite **644 passed**.
- `packages/console-backend`: 6 new endpoint tests (grant/deny/remove/persistence/axis separation) — plus `test_tool_bypass_router` (16) and user-settings suites (32) still green.

## Auto-review round 1

All 14 findings triaged in 42e17f57 (11 fixed, 3 declined with reasons — see inline replies). Highlights: blocked calls now stay paired with their artificial ToolMessage (no orphaned tool_results); wrapper moved to agent-common and applied at both discovery points (closes the sub-agent bypass); grants + email threaded into sub-agent contexts; one consent question per tool per message; consent metadata rides the existing `_risk_metadata` envelope.


packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+                            ),
+                            tool_call_id=tool_call["id"],
+                            name=tool_name,
+                            status="error",
+                        )
+                    )
+                continue
+            revised_tool_calls.append(tool_call)
+
+        last_ai_msg.tool_calls = revised_tool_calls
```
alloy-cicd: 🔴 **Orphaned `tool_result` → provider 400.** Denied (lines 371–382) and rejected/defaulted (393–407) calls are *removed* from `revised_tool_calls`, yet an artificial `ToolMessage` answering that `tool_call_id` is appended. The persisted history then contains a `role:"tool"` message whose id has no matching tool call in the AIMessage — via the LiteLLM gateway this fails the next model invocation (Anthropic: 400 `unexpected tool_use_id found in tool_result blocks`), and the corrupted history is checkpointed, poisoning every subsequent turn of the thread.

Note this is *not* the upstream pattern: langchain's `HumanInTheLoopMiddleware._process_decision` returns `(tool_call, tool_message)` on reject — the call is deliberately **kept** so the tool_use/tool_result pairing survives, and `ConditionalHumanInTheLoopMiddleware` inherits that. This middleware copied the loop shape but inverted the reject semantics.

Fix: keep the denied/rejected tool_call in `revised_tool_calls` and pair it with the artificial `auth_required` ToolMessage.

packages/orchestrator-agent/app/utils.py
```diff
@@ -148,6 +148,14 @@ def build_runtime_context(
         elif isinstance(tool, dict):
             tool_registry[tool.get("name", str(tool))] = tool
 
+    # Wrap identity-scoped tools (reserved nannos__user_identity field) BEFORE
+    # anything captures the registry (sub-agents, GP agent, PTC), so no
+    # model-facing schema ever shows the reserved field and every execution
+    # path goes through the consent-gated, email-injecting wrapper (ADR 0006).
+    from .middleware.identity_scoped import wrap_identity_scoped_tools
+
+    wrap_identity_scoped_tools(tool_registry)
```
alloy-cicd: 🔴 **Sub-agent MCP rediscovery bypasses the wrapper entirely.** This is the only production call site of `wrap_identity_scoped_tools`, and it only covers the orchestrator's `tool_registry`. A `LocalLangGraphSubAgentConfig` with non-empty `mcp_tools` triggers `_discover_mcp_tools` (agent-common `dynamic_agent.py:635/744`), which fetches **fresh raw BaseTool instances** from the gateway and — per its own docstring — overrides orchestrator tools entirely. Those instances are never wrapped: `nannos__user_identity` stays model-visible and model-writable, with no consent gate and no forced verified-email overwrite. Since whitelisting an integration tool on a sub-agent is the intended way to use it, this defeats the ADR 0006 mechanism on a mainline path (grep confirms agent-common has zero identity-related code; the sub-agent middleware stack has no `IdentityConsentMiddleware`). The GP agent is fine — its catalog comes from the wrapped registry.

Suggested direction: apply the wrap where MCP tools materialize into `BaseTool` objects (the shared discovery layer in agent-common), so every consumer gets wrapped tools by construction.

Minor hardening note on the same wrap point: dict-form registry entries (`elif isinstance(tool, dict)` at line 148) are silently skipped by the wrapper's `isinstance(tool, BaseTool)` guard. Not reachable with the current discovery flow (always `List[BaseTool]`), but worth a loud log/assert if a dict schema ever contains the reserved field.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+    schema = _schema_as_dict(inner.args_schema)
+    if schema is None:  # pragma: no cover - detection guarantees a dict schema
+        return inner
+    cleaned_schema = _schema_without_identity_field(schema)
+    tool_name = inner.name
+    inner_metadata = inner.metadata if isinstance(inner.metadata, dict) else {}
+
+    async def _identity_scoped(runtime: ToolRuntime = None, **kwargs: Any) -> Any:  # type: ignore[assignment]
+        context: Any = getattr(runtime, "context", None)
+        consent = _consent_state(context, tool_name, inner_metadata)
```
alloy-cicd: 🔴 **Remembered grants are unreachable on sub-agent / GP / PTC-in-sub-agent paths.** The sub-agent runtime context (agent-common `dynamic_agent.py:1550`) is a `SimpleNamespace` carrying only `tool_bypass_rules` / `tool_risk_cache` / `_pending_bypass_rules` (+ `tool_registry`/`whitelisted_tool_names` in catalog mode) — no `identity_consent_grants` and no `email`. Wrapped tools from the registry do flow into sub-agents (`utils.py:287/403-414` → `_exposed_catalog`) and execute under that context, so `_consent_state` returns `None` here and the tool fails closed — **even after the user granted consent** — with a payload claiming "no consent has been granted yet". Since `IdentityConsentMiddleware` is installed only on the orchestrator main graph (`graph_factory.py:728`), no prompt can ever fire there either: a permanent dead end that contradicts the documented "remembered grant → pass through" contract (docstring lines 189–193).

First-use fail-closed on those paths is documented as intentional; ignoring *valid remembered grants* is the gap. Fix: thread `identity_consent_grants` and `email` into the sub-agent context namespace (the data is already loaded on `GraphRuntimeContext`).

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+                continue
+            if idx in decision_by_index:
+                decision = decision_by_index[idx]
+                tool = tool_registry.get(tool_name)
+                metadata = getattr(tool, "metadata", None)
+                decision_type = (
+                    decision.get("type") if isinstance(decision, dict) else None
+                )
+                if decision_type == "approve":
+                    _record_consent(context, tool_name, metadata, granted=True)
```
alloy-cicd: 🔴 **Last-write-wins consent when one AI message calls the same tool twice.** Consent is keyed per `(tool, server)` (`consent_key`, line 102) but decisions arrive per call (`_call_id` alignment in `executor.py:147-167` supports differing per-call answers). Approve call A + reject call B → `_record_consent` writes `granted=True` then `granted=False` into the same key; since `aafter_model` completes before any tool runs and the wrapper re-reads the live grants dict at execution time (line 204), the explicitly-approved call A is then blocked with a "user declined" payload, and the denial is durably persisted (last PUT wins in `persist_identity_consents`) with no future re-prompt.

The mirror ordering is worse: reject at index 0, approve at index 1 → the user's explicit **rejection is converted into a standing grant** and the tool runs without re-prompt from then on.

Suggested fix: dedupe ask-indices per consent key (one ActionRequest per key), or detect conflicting per-key decisions and treat any rejection as authoritative for the key while still honoring the per-call approval for that single call.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+                )
+                action_requests.append(
+                    ActionRequest(
+                        name=tool_name,
+                        # Display-only; the reserved field is injected at execution
+                        # time and is never part of the model-visible args.
+                        args={
+                            **tool_call.get("args", {}),
+                            "_call_id": tool_call["id"],
+                            "_consent_metadata": {
```
alloy-cicd: 🔴 **`_consent_metadata` renders as a raw `[object Object]` chip in every client.** All three HITL renderers hide only `['reason', '_risk_metadata']`: console `InterruptConfirmCard.tsx:20` (renders remaining args via `String(v)`), Slack `taskResponseHandler.ts:570/741/796`, Google Chat `googleChatService.ts:11`. None were updated in this PR and none contain `_consent_metadata` / `identity_consent`, so the consent prompt shows as a generic risk-approval card with a literal `_consent_metadata: [object Object]` entry — in the very dialog meant to build trust for identity disclosure. (The `_call_id` chip leak is pre-existing, shared with risk interrupts.)

Consider reusing the existing `_risk_metadata` envelope with its `source` discriminator (it exists precisely to distinguish interrupt kinds) instead of introducing a parallel key every client must now special-case — or at minimum add `_consent_metadata` to the three HIDDEN_KEYS lists and give the card consent-specific rendering.

packages/console-backend/console_backend/routers/auth_router.py
```diff
@@ -356,6 +356,60 @@ async def upsert_tool_bypass_rule(
...
+
+    Used by the orchestrator after the first-use consent gate (ADR 0006 Gate 3)
+    records the user's decision. `remove` resets a remembered answer so the
+    gate prompts again on next use. Deliberately separate from tool bypass
+    rules — identity disclosure is a different axis from action-risk tolerance.
+    """
+    user_settings_service = get_user_settings_service(request)
+    settings = await user_settings_service.get_settings(db, user.id)
+
+    grants: dict = dict(settings.identity_consent_grants)
```
alloy-cicd, line 398: 🟡 **Lost-update race on the grants dict.** `get_settings` (line 390) → `dict(...)` copy → `upsert_settings` (line 400), where the `ON CONFLICT` clause rewrites every column from `EXCLUDED` with no `SELECT ... FOR UPDATE` and no JSONB merge. Two concurrent consent PUTs for the same user (parallel sessions answering prompts) can silently drop one grant. The failure direction is fail-safe (a lost grant/denial re-prompts; nothing escalates to granted), and this faithfully copies the pre-existing `upsert_tool_bypass_rule` pattern — but consent grants are security-relevant state, and the repo already has a row-lock idiom for exactly this shape (`rate_card_repository.py:60-64`). A `FOR UPDATE` on the read, or a targeted JSONB `||` / `-` update, would close it. (The general settings PATCH cannot clobber grants — `UserSettingsUpdate` deliberately has no `identity_consent_grants` field, which is good; `tool_bypass_rules` *is* PATCH-exposed and has the wide-window variant of this race, pre-existing.)

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+            for idx in ask_indices:
+                tool_call = last_ai_msg.tool_calls[idx]
+                tool_name = tool_call["name"]
+                tool = tool_registry.get(tool_name)
+                server_slug = _resolve_server_slug(
+                    tool_name, context, getattr(tool, "metadata", None)
+                )
+                description = (
+                    f"First use of '{tool_name}' ({server_slug}): this tool will receive your "
+                    f"verified email address ({email}) to scope its access to your own records "
```
alloy-cicd: 🟡 **Empty-email edge case records a grant for a permanently broken tool.** `email` is interpolated unguarded; `GraphRuntimeContext.email` is a required `str`, so `None` can't reach here, but an empty-string claim from the IdP renders "verified email address ()" — and if the user approves, the grant is durably persisted while the wrapper's `if not email` backstop (line 214) returns `auth_required` forever, with a message claiming "no consent has been granted yet" while the middleware (consent is True → pass through) never re-prompts. Deadlock until the grant is manually removed via the console API.

Cheap fix: skip the prompt (and refuse to record a grant) when `email` is falsy, so the wrapper's backstop and the consent flow can't disagree.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+                )
+                if decision_type == "approve":
+                    _record_consent(context, tool_name, metadata, granted=True)
+                    revised_tool_calls.append(tool_call)
+                else:
+                    # Remember only explicit rejections; a defaulted/missing
+                    # decision blocks this call without recording a denial.
+                    if decision_type == "reject":
+                        _record_consent(context, tool_name, metadata, granted=False)
+                    artificial_tool_messages.append(
```
alloy-cicd: 🟡 **Misleading payload for unknown decision types.** Any non-approve decision gets `identity_auth_required_payload(tool_name, rejected=True)` — "The user declined to share their identity … blocked until the user grants consent" — but a denial is recorded only for `decision_type == "reject"`. For an unknown/malformed decision the model is told a durable denial exists when the gate will actually re-prompt on next use. The docstring documents default-reject-without-remembering as intentional (and the executor materializes genuinely *missing* decisions as explicit rejects, so this branch only fires for malformed types), but the wording still lies to the model. `rejected=(decision_type == "reject")` reuses the existing accurate wording for the defaulted case at zero cost.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+            logger.warning(
+                "Identity-scoped tool '%s' dispatched without a verified email on the "
+                "runtime context — failing closed",
+                tool_name,
+            )
+            return identity_auth_required_payload(tool_name, rejected=False)
+
+        # Force-populate, overwriting whatever the model attempted to supply.
+        kwargs[NANNOS_USER_IDENTITY_FIELD] = email
+        return await inner.arun(kwargs)
```
alloy-cicd: 🟡 **Wrapper silently drops MCP `structuredContent`.** MCP tools from langchain-mcp-adapters are `response_format="content_and_artifact"`; this wrapper is a plain `StructuredTool` (default `"content"`) and calls `inner.arun(kwargs)` without a `tool_call_id`, so langchain-core's `_format_output` returns only the content half — the artifact (`MCPToolArtifact.structured_content`) is discarded and `ToolMessage.artifact` is `None` downstream. Low severity today: with adapters 0.2.2, images/embedded resources travel in *content* (unaffected), and nothing in the repo currently reads the MCP artifact. But it's a silent behavior change for every identity-scoped tool the moment anything starts consuming `structuredContent`. Also note `inner.arun` is called without config/callbacks, detaching the inner run from the tracing chain. Fix shape: declare `response_format="content_and_artifact"` on the wrapper, pass through the tuple, and return `(payload, None)` from the auth-required early exits.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+        cleaned["required"] = [r for r in required if r != NANNOS_USER_IDENTITY_FIELD]
+    return cleaned
+
+
+def consent_key(tool_name: str, server_slug: str) -> str:
+    """Key a consent grant by tool + server (same shape as bypass-rule keys)."""
+    return f"{tool_name}::{server_slug}"
+
+
+def _resolve_server_slug(
```
alloy-cicd: 🔵 **Third hand-rolled copy of server-slug resolution — already divergent.** Canonical: `ConditionalHumanInTheLoopMiddleware._get_server_slug` (agent-common `conditional_hitl.py:452`, 4-step resolution including the middleware-level `_tool_server_map` used on sub-agent paths); second copy in `ptc_guard.py:241`; this is the third (the docstring itself says "mirrors the HITL middleware") and it drops the middleware-level fallback step. Consent keys and bypass keys share the `tool::server_slug` format — if resolution ever diverges further, a remembered grant stops matching at lookup time and the gate silently re-prompts or fails closed. Extract one shared resolver in agent-common and use it in all three.

packages/orchestrator-agent/app/core/registry.py
```diff
@@ -528,3 +531,45 @@ async def persist_bypass_rules(
...
+            return
+
+        client = await self._get_client()
+        headers = {"Authorization": f"Bearer {access_token}"}
+
+        for entry in pending_consents:
+            key: str = entry["key"]
+            grant: dict[str, Any] = entry["grant"]
+
+            parts = key.split("::", 1)
```
alloy-cicd: 🔵 **Near-verbatim copy of `persist_bypass_rules` + `::` join-then-split round trip.** Same client/headers setup, same sequential per-entry PUT loop, same warning/exception handling — only endpoint and body differ; factor a shared helper. Also, `_record_consent` joins `tool_name`/`server_slug` into a key (`identity_scoped.py:104`) only for this code to split it back apart (and `auth_router.py:393` to re-join it): queueing structured entries `{tool_name, server_slug, granted}` would drop the round trip and the latent mis-parse for tool names containing `::` (a tool `a::b` on server `s` parses as tool `a`, server `b::s` — same latent flaw as the bypass original). Sequential PUTs also mean N consent answers = N serialized round trips, each triggering its own cache invalidation; a batch endpoint (or at least `asyncio.gather`) would help.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+logger = logging.getLogger(__name__)
+
+NANNOS_USER_IDENTITY_FIELD = "nannos__user_identity"
+"""Reserved input-schema field marking a tool as identity-scoped."""
+
+IDENTITY_SCOPED_METADATA_KEY = "identity_scoped"
+"""Metadata flag stamped on wrapped identity-scoped tools."""
+
+
+def _schema_as_dict(args_schema: Any) -> dict[str, Any] | None:
```
alloy-cicd: 🔵 **Duplicates `_safe_json_schema` (agent-common `ptc_signatures.py:125`) with narrower guarding.** The existing helper wraps the `tool.args_schema` *access* in try/except (and documents why `get_input_schema()` must not be used as a fallback); this copy guards only the `model_json_schema()` call. A tool whose `args_schema` property raises would crash `is_identity_scoped_tool` → `wrap_identity_scoped_tools` → `build_runtime_context`, killing the whole turn instead of degrading like the PTC render path. Import/promote the shared helper instead.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,412 @@
...
+    return StructuredTool(
+        name=tool_name,
+        description=inner.description,
+        args_schema=cleaned_schema,
+        coroutine=_identity_scoped,
+        metadata={**inner_metadata, IDENTITY_SCOPED_METADATA_KEY: True},
+    )
+
+
+def wrap_identity_scoped_tools(tool_registry: dict[str, Any]) -> list[str]:
```
alloy-cicd: 🔵 **Re-wraps the entire registry every turn.** `build_runtime_context` runs per invocation and rebuilds `tool_registry` from the cached-but-unwrapped `user_config.tools`, so the `is_wrapped_identity_scoped_tool` short-circuit never hits across turns: every turn re-runs `is_identity_scoped_tool` → `_schema_as_dict` → `model_json_schema()` over every Pydantic-schema tool in the registry (hundreds of tools; discovery results are cached per-user precisely to avoid this class of per-turn cost) and rebuilds a fresh `StructuredTool` (including a `json.dumps`/`loads` deep copy of the schema in `_schema_without_identity_field`) for each identity-scoped tool. Cheaper: wrap once at discovery time so wrapped tools land in the per-user discovery cache (they carry no per-turn state — email/consent are read from the runtime context at call time) — which would also dovetail with fixing the sub-agent bypass at the discovery layer. Short of that: for Pydantic schemas check `NANNOS_USER_IDENTITY_FIELD in args_schema.model_fields` instead of generating the full JSON schema, and shallow-copy the top level instead of the json round trip.

packages/console-backend/console_backend/routers/auth_router.py
```diff
@@ -356,6 +356,60 @@ async def upsert_tool_bypass_rule(
...
+    else:
+        grants[key] = {"granted": body.granted}
+
+    await user_settings_service.upsert_settings(db, user.id, identity_consent_grants=grants)
+    await db.commit()
+
+    # Consent grants ride on the orchestrator's cached User (not part of its cache key), so
+    # flush this user's cache to apply the change on their next turn instead of after the TTL.
+    if user.sub:
+        schedule_orchestrator_discovery_cache_invalidation(
```
alloy-cicd, line 434: 🔵 **Every consent answer flushes the user's entire discovery cache.** A consent grant changes only a small dict riding on the cached `User`; the tools/sub-agents themselves are unchanged, yet the next turn pays the full cold-start (gatana token exchange + per-server `list_tools` handshakes, multi-seconds per `discovery_cache.py`) — and the granting session already holds the answer in its runtime context. Combined with one PUT per consent entry from the orchestrator, N answers in a turn schedule N redundant flushes. Consider a targeted update of the cached User's `identity_consent_grants` instead of a full flush, or loading grants fresh per turn outside the discovery cache.

alloy-cicd (commented):

Reviewed the identity-scoped tools + first-use consent gate (Gate 3). The core mechanism — one wrap point that strips `nannos__user_identity` from the model-visible schema and force-populates it at execution — is at the right altitude for registry-derived paths. However, verification found **5 confirmed bugs** (2 of them undermining the security property itself), 3 smaller correctness issues, and several copy-paste/efficiency cleanups.

**High severity (🔴):**
1. Denied/rejected consent calls are *dropped* from `tool_calls` while an artificial ToolMessage still answers their id → orphaned `tool_result` → provider 400 on the next model call, and the corrupted history persists in the checkpoint (`identity_scoped.py:411`).
2. Sub-agents with their own `mcp_tools` whitelist rediscover tools via `MultiServerMCPClient` and never pass through `wrap_identity_scoped_tools` — the reserved field stays model-writable with no consent gate on that mainline path (`utils.py:157`).
3. Remembered grants are unreachable on sub-agent/GP/PTC-in-sub-agent paths: the sub-agent `SimpleNamespace` context carries neither `identity_consent_grants` nor `email`, so consented tools still fail closed forever there with a misleading message (`identity_scoped.py:204`).
4. Approve+reject of two calls to the *same* tool in one AI message → last-write-wins consent; an explicit rejection can be durably converted into a standing grant (`identity_scoped.py:391`).
5. The consent card renders `_consent_metadata: [object Object]` in all three clients (console, Slack, Google Chat) and shows as a generic risk-approval card (`identity_scoped.py:340`).

**Medium (🟡):** lost-update race on the grants dict (inherited pattern, fail-safe direction); empty-email edge case records a grant for a permanently broken tool; misleading "user declined" payload for unknown decision types; wrapper silently drops MCP `structuredContent`.

**Quality (🔵):** third copy of server-slug resolution (already divergent), near-verbatim copy of `persist_bypass_rules` with a `::` join-then-split round trip, duplicated `_safe_json_schema` with narrower guarding, per-turn re-wrapping of the whole registry, and a full discovery-cache flush per consent answer.

Details inline. Submitted as COMMENT — the human reviewer decides.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — blocked calls are now **kept** in the AIMessage and paired with their artificial `auth_required` ToolMessage, mirroring the upstream `_process_decision` reject semantics. Verified the agent loop's `model_to_tools` edge filters out calls whose id already has a ToolMessage, so nothing double-executes.

packages/orchestrator-agent/app/utils.py
aartaria: ✅ Fixed in 42e17f57 — the wrapper machinery moved to `agent_common.core.identity_scoped` and now wraps where MCP tools materialize: the orchestrator's `ToolDiscoveryService._get_tools_from_server` **and** the dynamic sub-agent's `_discover_mcp_tools` (right after `_validate_tool_schema`). The `build_runtime_context` wrap remains as an idempotent backstop. Dict-form registry entries remain skipped (unreachable today, as noted).

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — `identity_consent_grants` and `email` are now threaded into the sub-agent `SimpleNamespace` context (via new `create_dynamic_local_subagent` params, passed from `build_runtime_context`), so remembered grants pass through in sub-agents/GP. First-use in a sub-agent still fails closed by design (no interrupt possible there); consent is granted via a direct orchestrator call. Covered by `test_grants_reachable_from_subagent_style_context`.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — the middleware now asks **once per tool** (`ask_calls_by_tool` grouping): one ActionRequest per unique tool, and the single decision applies to every call of that tool in the message. No conflicting per-call answers can exist for one consent key anymore. Covered by `test_same_tool_twice_asks_once_and_applies_to_both`.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — took the suggested route: the metadata now rides in the existing `_risk_metadata` envelope with `source: "identity_consent"` as the discriminator, so all three clients already hide it and the card renders cleanly. Consent-specific card styling keyed off the `source` field is left as frontend follow-up.

packages/console-backend/console_backend/routers/auth_router.py
aartaria: 🚫 Not addressed in this PR — acknowledged real, but it faithfully mirrors the pre-existing `upsert_tool_bypass_rule` pattern and the failure direction is fail-safe (a lost answer re-prompts; nothing escalates to granted). Moving both endpoints to a row-locked or targeted-JSONB write is a follow-up worth doing together rather than forking the idiom for one column here.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — when `email` is falsy the middleware now skips the prompt entirely and blocks the call for the turn **without recording anything** (with the not-yet-granted wording, not "declined"), so the consent flow and the wrapper backstop can no longer disagree. Covered by `test_missing_email_blocks_without_prompt_or_grant`.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — only an explicit `reject` records a denial and gets the "user declined" wording; malformed/unknown decisions block the call with the accurate "no consent has been granted yet" payload. Covered by `test_unknown_decision_blocks_without_remembering`.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — the wrapper now mirrors `inner.response_format`, calls the inner coroutine directly so `content_and_artifact` tuples pass through intact (same direct-coroutine precedent as `_wrap_tool_with_agent_name`), and the auth-required early exits return `(payload, None)` for artifact tools. Covered by `test_preserves_content_and_artifact_tuple` / `test_fails_closed_tuple_shape_for_artifact_tools`.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Resolved differently in 42e17f57 — consent keys no longer use server-slug resolution at all: they're keyed by **tool name alone** (which is what ADR 0006 specifies — per (user, tool)), precisely because slug resolution is path-dependent (sub-agent rediscovery doesn't tag `server_name`) and a path-dependent key would silently stop matching. The only remaining server lookup is display wording in the prompt (`_display_server`, metadata-only). Extracting one shared resolver for HITL/ptc_guard remains a pre-existing follow-up.

packages/orchestrator-agent/app/core/registry.py
aartaria: ✅ Partially fixed in 42e17f57 — pending entries are now structured `{tool_name, granted}` end-to-end: no join/split round trip, no `::` mis-parse class (the endpoint keys by plain tool name). 🚫 Kept the sequential per-entry PUT loop mirroring `persist_bypass_rules`: consent answers arrive at most a handful per turn (one per newly-consented tool), so batching isn't worth forking the established idiom here.

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — `is_identity_scoped_tool` now wraps the `args_schema` access itself in try/except and is documented as never-raising (a misbehaving tool degrades to not-identity-scoped instead of killing the turn). It also avoids full JSON-schema generation via `model_fields` for pydantic schemas. Covered by `test_detection_never_raises`. Kept it self-contained rather than importing `_safe_json_schema` (private to ptc_signatures, and the fast path no longer generates schemas at all).

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 42e17f57 — wrapping moved to discovery time (orchestrator `_get_tools_from_server` + sub-agent `_discover_mcp_tools`), so wrapped instances land in the per-user discovery cache exactly as suggested — which also closes the sub-agent bypass. The `build_runtime_context` wrap remains as an idempotent backstop that's now a cheap no-op (metadata-flag check per tool; pydantic detection uses `model_fields`, and the schema copy is shallow instead of a json round trip).

packages/console-backend/console_backend/routers/auth_router.py
aartaria: 🚫 Not addressed in this PR — the flush mirrors the bypass-rule endpoint's behavior and is what guarantees a grant recorded in session A applies to session B's next turn. A targeted cached-`User` update (or loading grants outside the discovery cache) is a real optimization but belongs in a follow-up touching both settings paths; consent answers are rare one-time events per (user, tool), so the practical cost is a handful of cold discoveries per user ever.

packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -720,6 +721,11 @@ async def _forward_to_active_subagents(context_id: str, messages: list) -> None:
             self._auth_middleware,
             ErrorClassificationMiddleware(),
             hitl_middleware,
+            # After hitl_middleware in the list so its aafter_model runs BEFORE
+            # the risk-HITL one (after_model hooks run in reverse list order):
+            # a consent-rejected identity-scoped call is dropped before it can
+            # be risk-prompted. Gate 3, ADR 0006.
+            IdentityConsentMiddleware(),
```
alloy-cicd, line 730: 🟡 **Consent-blocked calls now reach the risk-HITL — duplicate tool_result possible.** This ordering (consent `aafter_model` runs before the risk-HITL's) was correct under round 1's removal semantics, but the fix keeps blocked calls in `last_ai_msg.tool_calls` — and `ConditionalHumanInTheLoopMiddleware.aafter_model` iterates **all** tool calls with no filter for ids that already have a ToolMessage (`conditional_hitl.py:254/404`). Since the orchestrator risk-scores every non-task tool (threshold 0.8, lines 87-91 above), an identity-blocked call whose tool scores ≥ threshold triggers a second, spurious HITL prompt for a call that can never execute — and a **reject** there emits a second ToolMessage with the same `tool_call_id` (via upstream `_process_decision`) → two tool_results for one tool_use → provider 400. That's the round-1 failure class, reachable from the other direction.

Suggested fix: have `ConditionalHumanInTheLoopMiddleware.aafter_model` skip tool calls whose id already has a ToolMessage after the last AIMessage (mirroring the agent loop's own filter in langchain `factory.py:1870-1873`).

Also stale: this comment block ("is dropped before it can be risk-prompted") and `test_graph_factory.py:144-146` still document the removed dropping behavior — worth updating to the keep-and-pair invariant either way.

packages/orchestrator-agent/app/middleware/identity_scoped.py
```diff
@@ -0,0 +1,212 @@
...
+                )
+                if decision_type == "approve":
+                    record_consent(context, tool_name, granted=True)
+                    continue
+                # Remember only explicit rejections; a malformed/unknown
+                # decision blocks this turn without recording a denial —
+                # and the payload must not claim a denial exists.
+                explicit_reject = decision_type == "reject"
+                if explicit_reject:
+                    record_consent(context, tool_name, granted=False)
```
alloy-cicd: 🔵 **Executor-synthesized default rejects record durable denials.** When a client answers with id-keyed decisions and an identity action_request's `_call_id` is missing/stale, `executor._decisions_for_interrupt` substitutes a plain `{"type": "reject"}` (`executor.py:164`). This branch can't distinguish that synthesized default from the user's explicit rejection, so it records a **permanent** denial — contradicting the comment above ("a malformed/unknown decision blocks this turn without recording a denial"): the user never answered, yet a denial persists until manually cleared via the settings endpoint. Unreachable with the current frontend (which echoes `_call_id` faithfully), so low severity — but cheap to close: have the executor tag its synthesized default (e.g. `{"type": "reject", "_defaulted": true}`) and treat it here like an unknown decision, or only record denials for decisions carrying the matching id.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (981bbbe23835c80c786e559892a08b2a267b89f5..42e17f573cad3d0708958c27967fe1c3e94a025e)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | 🔴 Orphaned tool_result → provider 400 | **verified-fixed** — blocked calls kept + paired; confirmed langchain's `model_to_tools` edge filters answered ids (factory.py:1870-1873), so nothing double-executes |
| 2 | 🔴 Sub-agent MCP rediscovery bypass | **verified-fixed** — wrapping at both materialization points (orchestrator `_get_tools_from_server`, sub-agent `_discover_mcp_tools` after schema validation); wrapped instances land in the discovery cache; backstop idempotent |
| 3 | 🔴 Remembered grants unreachable in sub-agents | **verified-fixed** — grants+email threaded end-to-end into the sub-agent context; same-object grants dict means mid-turn grants apply immediately in sub-agents |
| 4 | 🔴 Last-write-wins consent (same tool twice) | **verified-fixed** — once-per-tool grouping; executor `_decisions_for_interrupt` alignment verified (count/order/legacy-blanket all consistent) |
| 5 | 🔴 `_consent_metadata: [object Object]` | **verified-fixed** — rides `_risk_metadata` with `source: identity_consent`; all three clients gate badges on `source === 'risk_score'` and hide the key cleanly |
| 6 | 🟡 Lost-update race on grants dict | **rejected-by-author** (follow-up with the bypass-rule twin; fail-safe direction) — accepted, not re-raised |
| 7 | 🟡 Empty-email grant trap | **verified-fixed** — prompt skipped, nothing recorded, calls stay paired |
| 8 | 🟡 Misleading unknown-decision payload | **verified-fixed** — `rejected` tracks explicit rejects only (but see new low-severity finding below) |
| 9 | 🟡 structuredContent dropped | **verified-fixed** — `response_format` mirrored, tuple passthrough verified end-to-end incl. blocked-path `(payload, None)` shapes |
| 10 | 🔵 Third slug-resolver copy | **resolved-differently** — consent keyed by tool name alone (ADR 0006); shared-resolver extraction stays a pre-existing follow-up |
| 11 | 🔵 persist copy + `::` round trip | **partially-fixed** — structured `{tool_name, granted}` end-to-end (mis-parse class gone); sequential PUT loop kept with reasoning — accepted |
| 12 | 🔵 `_safe_json_schema` duplication/guarding | **verified-fixed** — detection never raises, `model_fields` fast path |
| 13 | 🔵 Per-turn registry re-wrap | **verified-fixed** — discovery-time wrapping; backstop is a cheap metadata-flag check |
| 14 | 🔵 Full cache flush per consent | **rejected-by-author** (rare one-time events; follow-up) — accepted, not re-raised |

All touched test suites pass at this head (26 identity-scoped, 19 graph-factory, 6 console-backend consent-router).

**New findings in the delta: 2** (inline below). The keep-blocked-calls fix is correct against the agent loop, but the kept calls are now visible to the downstream risk-HITL middleware, which re-opens the duplicate-tool_result variant of the round-1 failure class; and the executor's synthesized default-reject is indistinguishable from an explicit rejection, so it records a durable denial the docstring says shouldn't exist.

Non-blocking observations (no inline comments): the wrapper is coroutine-only (sync `invoke` would raise `NotImplementedError` — fine for async-only MCP tools, cites existing precedent); the `model_fields` detection path would miss a pydantic field *aliased* to the reserved name (requires a hostile local tool author — out of scope); the `_call_id` chip leak in clients is pre-existing and shared with risk/PTC interrupts.

packages/orchestrator-agent/app/core/graph_factory.py
aartaria: ✅ Fixed in 9adba45d — took the suggested route: `ConditionalHumanInTheLoopMiddleware.aafter_model` now collects the ids of ToolMessages after the last AIMessage and skips those tool calls entirely (no scoring, no prompt), mirroring the agent loop's own `model_to_tools` filter. A consent-blocked call can no longer be risk-prompted or double-answered. Also updated the stale graph_factory comment and `test_graph_factory` wording to the keep-and-pair invariant. Covered by `TestAnsweredCallsSkipped` (answered call never scored; unanswered sibling in the same message still guards).

packages/orchestrator-agent/app/middleware/identity_scoped.py
aartaria: ✅ Fixed in 9adba45d — the executor now tags its synthesized safe reject as `{\"type\": \"reject\", \"_defaulted\": true}`, and the consent gate records a durable denial only for an untagged (explicit) reject; defaulted rejects block the turn with the \"no consent granted yet\" wording. Covered by `test_missing_call_id_defaults_to_tagged_safe_reject` (executor) and `test_defaulted_reject_blocks_without_remembering` (middleware).

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -857,11 +859,13 @@ async def awrap_tool_call(self, request: Any, handler: Any) -> Any:
...
-        Only engages for the ``eval`` tool when risk scoring is active; all other
-        tool calls (and the unguarded configuration) pass straight through.
+        Only engages for the ``eval`` tool; all other tool calls pass straight
+        through. It is NOT gated on a configured risk scorer: identity-scoped
+        tools record first-use consent questions on the same collector (ADR
+        0006), and those must be drained even where risk scoring is off.
         """
         tool_call = getattr(request, "tool_call", None) or {}
-        if not self._ptc_enabled or self._ptc_risk_scorer is None or tool_call.get("name") != self._tool_name:
+        if not self._ptc_enabled or tool_call.get("name") != self._tool_name:
```
alloy-cicd, line 868: 🟡 **PTC-eval consent interrupt re-runs side-effecting tools with zero dedup when risk scoring is off.** Dropping the scorer requirement from this gate is right (consent pendings must drain where scoring is off), but it newly exposes an un-deduplicated re-run: with a scorer, run-1 successes are cached in `turn.results` and skipped on the re-run (`ptc_guard.py:354-357/407-410`), and the residual replay double-execution is documented (`ptc_guard.py:134-141`). With `risk_scorer=None`, `wrap_tool_for_ptc._guarded` short-circuits before any turn/results bookkeeping (`ptc_guard.py:341-342`) — so on a scoring-off graph, a consent interrupt means every side-effecting inner tool call in the eval executes in run 1, again in the post-resume replay, and again in run 2, with nothing skipped and nothing documenting the case.

Cheap fix: populate `turn.results` in the scorer-`None` branch too — the turn state is now always present for eval, so the caching can be unconditional even when scoring is skipped.

packages/agent-common/agent_common/core/identity_scoped.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+IDENTITY_SERVER_METADATA_KEY = "server_name"
+"""Tool-metadata key MCP discovery stamps with the owning server slug."""
+
+SELF_SERVER_SLUG = "_self"
+"""Slug for tools with no resolvable MCP server (in-process platform tools)."""
+
+
+def identity_server_slug(tool_name: str, context: Any, tool: Any = None) -> str:
```
alloy-cicd, line 141: 🔵 **Slug fallback can record a grant under the shared gateway name — coarser than intended.** The `tool_server_map`-first resolution fixes the common case, but for an identity tool absent from the map (e.g. present only in a sub-agent's own rediscovery, or a registry/map name mismatch — the `dynamic_agent.py:1311-1319` map update is filtered to `name in identity_tool_registry`), the fallback is `metadata.server_name`, which on the rediscovery path is the gateway name shared by every integration behind it (the commit message itself describes this mis-tagging). A grant recorded under e.g. `gatana` then covers all identity-scoped tools of **all** integrations routed through that gateway; `_self` similarly lumps every server-less tool under one grant. Given the grant is deliberately per-server, a fallback that widens it to per-gateway deserves either a guard (refuse to record/ask under a known-gateway or `_self` slug — fail closed like the wrapper already does elsewhere) or at least a comment acknowledging the breadth.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -730,6 +730,11 @@ def _make_stream(resuming: bool):
             if pending_bypass:
                 user_config._pending_bypass_rules = pending_bypass
 
+            # Same for identity-scoped tool consent answers (Gate 3, ADR 0006).
+            pending_consents = getattr(runtime_context, "_pending_identity_consents", None)
+            if pending_consents:
+                user_config._pending_identity_consents = pending_consents
+
```
alloy-cicd, line 744: 🔵 **Assign-vs-extend can drop sub-agent pending consents from persistence.** `record_consent` appends to whichever pending list the active context carries: sub-agent contexts share `user_config._pending_identity_consents` by reference, but on the orchestrator's own graph the middleware setattrs a **fresh** list onto `GraphRuntimeContext` (the dataclass has no such field). If one turn records consents on both graphs, this assignment overwrites the user_config list with the orchestrator-graph one, discarding the sub-agent entries from durable persistence (in-session behavior stays correct — the shared grants dict has both — but the dropped answers get re-asked next session). `user_config._pending_identity_consents.extend(...)`-style merging (or seeding the runtime context with the user_config list, as the sub-agent path already does) closes it. Same latent pattern exists for `_pending_bypass_rules` two lines up — pre-existing, flagging only the new copy.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (42e17f573cad3d0708958c27967fe1c3e94a025e..e59a1d72fed7fef92a431871243ac08075f6a934)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | 🟡 Consent-blocked calls reach the risk-HITL (duplicate tool_result) | **verified-fixed** — answered-ids filter runs as step 0 before static guard and scoring, mirrors the agent loop's own filter, replay-safe; skipped calls stay paired; stale graph_factory/test comments updated |
| 2 | 🔵 Executor-synthesized default rejects recorded durable denials | **verified-fixed** — `{"type": "reject", "_defaulted": true}` tag honored by both the consent middleware and the PTC decision path; upstream `_process_decision` ignores the extra key safely |

The branch was rebased onto main (picking up #143) and gained substantial new work beyond the round-2 fixes: per-MCP-server consent keying with the middleware moved to agent-common and installed on sub-agent graphs, PTC-eval consent via the shared pending collector with a batched interrupt, catalog `call_tool` gating, consent audit logging (migration 080), a console settings section, discovery-schema stripping in console-backend, and the recursion-limit raise. All of it was re-verified, not just the fix commits:

- **Keying consistency**: every reader/writer keys by bare server slug end-to-end (wrapper, middleware, registry persist body ↔ endpoint request model, tests, frontend); wrapper and middleware share one `identity_server_slug` resolver on the same shared `tool_server_map`, so grant/enforce divergence is structurally impossible (and a hypothetical mismatch fails closed, never open).
- **Sub-agent consent**: grants dict and pending list are threaded by reference into sub-agent contexts, so in-sub-agent answers are persisted by the orchestrator's normal flush; interrupt/resume rides the same tested infrastructure as risk-HITL.
- **Audit**: same-transaction write confirmed (audit failure rolls back the grant); migrations 079/080 are collision-free post-rebase and 080 follows the established `ADD VALUE IF NOT EXISTS` enum pattern; no PII in audit rows.
- **Discovery strip**: applied at the single MCPTool construction point, covers /tools, /tools/search, and scheduler watch-params; constant matches agent-common's exactly.
- **Recursion limit**: pure superstep-ceiling raise; loop protection remains with the loop-detection middleware, which is wired and tested.
- All touched test suites pass at this head (agent-common 71, orchestrator 78, console-backend 23).

**New findings in the delta: 3** (inline below) — one medium (PTC-eval re-run double-executes side-effecting tools when risk scoring is off, a case this PR newly enables), two low (slug-fallback breadth can record a grant under the shared gateway name; assign-vs-extend can drop sub-agent pending consents from persistence).

Non-blocking follow-up notes (no inline comments): the sync `after_model` of ConditionalHITL lacks the answered-ids filter (unreachable today — only `aafter_model` sources exist); a consent granted out-of-band between interrupt and resume can shrink the replayed partition and trip the decisions-count ValueError (pre-existing pattern, very low probability); `remove` of a never-answered server audits a no-op REVOKE row; `dynamic_agent.py:113-118`'s "mirrors MAX_RECURSION_LIMIT" comment is stale now that the orchestrator default moved to 250; the test comment at agent-common `test_identity_scoped.py:588` still says "per tool" where behavior is per server; no end-to-end test covers a consent interrupt+resume raised inside a sub-agent (the generic path is tested via risk-HITL).

packages/agent-common/agent_common/core/graph_utils.py
aartaria: ✅ Fixed in 89fc1bcb — took the suggested route: the scorer-`None` branch of `wrap_tool_for_ptc._guarded` now checks and populates `turn.results` (same call-key dedup as the scored path), so a consent-interrupt replay serves run-1 results from cache instead of re-executing side-effecting calls. Covered by `test_scorer_none_caches_results_on_active_turn` (inner executes once across two calls) and `test_scorer_none_executes_directly_without_turn` (no turn → straight execution, unchanged).

packages/agent-common/agent_common/core/identity_scoped.py
aartaria: ✅ Partially fixed in 89fc1bcb, remainder documented — `may_ask_consent_for_slug` now refuses to ask/record under `_self` (a grant there would lump every server-less tool into one answer); enforcement still uses the resolved slug and simply fails closed, covered by `test_self_slug_blocks_without_prompt_or_grant` + `test_self_slug_never_queues_consent`. The gateway-name case is narrower than flagged: the sub-agent's own `tool_server_map` is overwritten with orchestrator slugs for every orchestrator-visible identity tool, so only a tool visible *exclusively* to sub-agent rediscovery keys under the gateway name — the existing per-server test suite deliberately trusts `metadata.server_name` as a slug source, and a blanket refusal broke that intended design. That residual breadth is now explicitly documented on `may_ask_consent_for_slug` as accepted.

packages/orchestrator-agent/app/core/agent.py
aartaria: ✅ Fixed in 89fc1bcb — `agent.py` now dedup-extends `user_config._pending_identity_consents` with the runtime-context list (guarding against the same-object case) instead of assigning, so consents recorded on both the orchestrator graph and a sub-agent graph in one turn all reach durable persistence. Left `_pending_bypass_rules` as-is per your note (pre-existing latent pattern, different data flow).

aartaria:
https://github.com/ringier-data/nannos/issues/140#issuecomment-5340542629

PR #143, opened 2026-08-13
feat(console): paginate conversation history with infinite scroll

### LINKED ISSUE #142: Paginate conversation message history with infinite scroll (chat is truncated at 100 messages)
## Problem

`GET /api/v1/messages/{conversation_id}` caps the result at 100 messages with no pagination, so long conversations are silently truncated in the chat UI.

Two distinct defects:

1. **No pagination.** The router hard-validates `limit` to `1..100` and the frontend (`ChatContext.loadMessages`) always requests `limit=100`. Anything older is unreachable — the console even has a `MESSAGE_FETCH_LIMIT` constant in `ChatApp.tsx` and marks exported transcripts as possibly truncated, acknowledging the gap.
2. **Wrong 100 are returned.** The PostgreSQL query in `MessagesService.get_messages_by_conversation` is `ORDER BY created_at ASC LIMIT :limit`, i.e. it returns the *oldest* 100 messages, so a long conversation opens on its beginning and the most recent turns are missing. (The in-memory implementation returns the *last* `limit`, so the two backends disagree.)

## Expected behaviour

- Opening a conversation loads the **most recent** page of messages (default 100) and the view is scrolled to the bottom.
- Scrolling **to the top** of the message list loads the previous page (infinite scrolling upwards), prepending older messages while preserving the reader's scroll position.
- Paging stops cleanly when the conversation start is reached.

## Proposed implementation

**Backend** (`console-backend`)
- Keyset (cursor) pagination on the messages endpoint: optional `before` cursor; page ordered `(created_at, id) DESC` internally, returned chronologically.
- Response gains `has_more` and `next_cursor` so the client knows whether to keep paging; `count`/`messages` shape stays backwards compatible.
- Apply the same semantics to both the PostgreSQL and the in-memory messages service so they no longer diverge.

**Frontend** (`console-frontend`)
- Track a per-conversation pagination cursor + `has_more` in `ChatContext`; `loadMessages` gains a "load older" mode that prepends into `messagesMap` (timeline aggregation must be re-run over the combined list, not per page).
- `MessageList` observes a sentinel at the top of the scroll viewport to trigger the next page, with a loading indicator, and restores scroll offset after prepend so the viewport does not jump.
- Initial open keeps the existing scroll-to-bottom behaviour; auto-scroll-to-bottom must not fire on a prepend.

## Acceptance criteria

- A conversation with >100 messages opens showing the newest messages, scrolled to the bottom.
- Scrolling to the top loads older messages repeatedly until the conversation start, without scroll jumps or duplicated messages.
- Backend unit tests cover: default page returns the newest messages, `before` cursor returns the preceding page, `has_more` is correct at the boundary, and invalid `limit`/cursor are rejected.

aartaria:
closes #142

Long conversations were silently truncated in the chat: the endpoint capped results at 100 messages with no pagination, and the PostgreSQL query ordered `created_at ASC`, so a conversation actually opened on its **oldest** 100 messages (the in-memory service disagreed and returned the last 100).

### Backend (`console-backend`)
- Keyset pagination on `(created_at, id)`: `GET /api/v1/messages/{id}` returns the newest `limit` messages, or the page preceding the optional `before` cursor (a `message_id`), always chronologically ordered.
- Response gains `has_more` and `next_cursor`; `messages`/`count` are unchanged.
- An unknown cursor is a 400 instead of a silently empty page.
- The in-memory service mirrors the same semantics, so local dev and production no longer diverge.

### Frontend (`console-frontend`)
- `ChatContext` tracks a per-conversation cursor + `hasMore`, exposes `loadOlderMessages`, `hasMoreMessages` and `isLoadingOlderMessages`, and merges older pages into the existing list. A refetch of the newest page (socket snapshot) no longer rewinds a cursor that has already been paged backwards.
- New `useChatScroll` hook: keeps the transcript pinned to the bottom on open and while streaming, fetches the next page when the reader comes within 200px of the top, and re-anchors `scrollTop` by exactly the height the prepended page added so the view doesn't jump.
- `MessageList` shows a "Loading earlier messages…" row while a page is in flight.
- `PlaygroundChatPanel` now uses the shared hook instead of its own copy of the auto-scroll effect (and gets pagination for free).
- Conversation export marks the transcript as partial based on `has_more` rather than a hard-coded 100.

### Notes
- The generated API client (`src/api/generated`) is unaffected — the chat uses a raw `fetch` for this endpoint — but `npm run gen-sdk` should be re-run against a running backend at some point to pick up `before`/`has_more`/`next_cursor`.

### Verification
- `uv run pytest tests/` in console-backend: 1365 passed, including new tests for newest-page-first ordering, the cursor keyset filter, `has_more` at the boundary and unknown-cursor rejection.
- `npx tsc -b` and `npm run build` clean; `npm run lint` reports the same pre-existing findings as `main` (no new ones).

## Checklist

### DIFF
--- a/packages/console-backend/console_backend/routers/message_router.py
+++ b/packages/console-backend/console_backend/routers/message_router.py
@@ -6,6 +6,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request
 
 from ..dependencies import require_auth
 from ..models.user import User
+from ..services.messages_service import UnknownCursorError
 
 logger = logging.getLogger(__name__)
 
@@ -15,19 +16,28 @@ router: APIRouter = APIRouter(prefix="/api/v1/messages", tags=["messages"])
 
 @router.get("/{conversation_id}")
 async def get_messages_by_conversation(
-    request: Request, conversation_id: str, user: User = Depends(require_auth), limit: int = 100
+    request: Request,
+    conversation_id: str,
+    user: User = Depends(require_auth),
+    limit: int = 100,
+    before: str | None = None,
 ) -> dict:
-    """Get all messages for a conversation.
+    """Get one page of a conversation's messages, newest page first.
 
     Args:
         conversation_id: The conversation ID
         limit: Maximum number of messages to return (default: 100, max: 100)
+        before: Cursor — the `message_id` to page back from (exclusive). Omit to
+            get the newest page; pass the previous response's `next_cursor` to
+            walk further back through the history.
 
     Returns:
         Dictionary containing:
         - conversation_id: The conversation ID
         - messages: List of messages ordered chronologically
         - count: Number of messages returned
+        - has_more: Whether older messages exist before this page
+        - next_cursor: Cursor to pass as `before` for the next (older) page, or None
     """
     try:
         # Validate limit
@@ -35,10 +45,15 @@ async def get_messages_by_conversation(
             raise HTTPException(status_code=400, detail="Limit must be between 1 and 100")
 
         messages_service = request.app.state.messages_service
-        messages = await messages_service.get_messages_by_conversation(conversation_id, user.id, limit=limit)
+        try:
+            page = await messages_service.get_messages_by_conversation(
+                conversation_id, user.id, limit=limit, before=before
+            )
+        except UnknownCursorError:
+            raise HTTPException(status_code=400, detail="Unknown pagination cursor") from None
 
         # Hydrate file parts with presigned URLs
-        messages = await messages_service.hydrate_messages_files(messages)
+        messages = await messages_service.hydrate_messages_files(page.messages)
 
         return {
             "conversation_id": conversation_id,
@@ -60,6 +75,9 @@ async def get_messages_by_conversation(
                 for msg in messages
             ],
             "count": len(messages),
+            "has_more": page.has_more,
+            # The oldest message on this page is where the next (older) page starts.
+            "next_cursor": messages[0].message_id if page.has_more and messages else None,
         }
 
     except HTTPException:
--- a/packages/console-backend/console_backend/routers/message_router.py
+++ b/packages/console-backend/console_backend/routers/message_router.py
@@ -6,7 +6,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request
 
 from ..dependencies import require_auth
 from ..models.user import User
-from ..services.messages_service import UnknownCursorError
+from ..exceptions import UnknownCursorError
 
 logger = logging.getLogger(__name__)
 
@@ -76,8 +76,7 @@ async def get_messages_by_conversation(
             ],
             "count": len(messages),
             "has_more": page.has_more,
-            # The oldest message on this page is where the next (older) page starts.
-            "next_cursor": messages[0].message_id if page.has_more and messages else None,
+            "next_cursor": page.next_cursor,
         }
 
     except HTTPException:
--- a/packages/console-backend/console_backend/services/messages_service.py
+++ b/packages/console-backend/console_backend/services/messages_service.py
@@ -11,14 +11,11 @@ from sqlalchemy import text
 from uuid6 import uuid7
 
 from ..db.connection import get_async_session_factory
+from ..exceptions import UnknownCursorError
 from ..models.message import Message, MessagePage
 from .file_storage_service import FileStorageService
 
 logger = logging.getLogger(__name__)
-
-
-class UnknownCursorError(ValueError):
-    """Raised when a pagination cursor does not name a message in the conversation."""
 
 
 EXPIRY_BUFFER_SECONDS = (
@@ -254,11 +251,14 @@ class MessagesService:
             before: Cursor — the `message_id` to page back from (exclusive)
 
         Returns:
-            A MessagePage whose messages are ordered chronologically (oldest first)
-            and whose `has_more` says whether older messages remain.
+            A MessagePage whose messages are ordered chronologically (oldest first),
+            whose `has_more` says whether older messages remain and whose
+            `next_cursor` is the `before` value that reaches them.
 
         Raises:
             UnknownCursorError: `before` does not name a message in this conversation.
+            Exception: the read failed — callers must not mistake a failure for the
+                start of the conversation.
         """
         try:
             async with self._session_factory() as db:
@@ -346,13 +346,24 @@ class MessagesService:
                     continue
 
             logger.debug(f"Retrieved {len(results)} messages for conversation: {conversation_id}")
-            return MessagePage(messages=results, has_more=has_more)
+            # Take the cursor from the RAW rows, not from `results`: a row that failed
+            # to parse is dropped from the page but still occupies its place in the
+            # history, and a cursor derived from a fully-dropped page would be null
+            # while has_more stayed true — leaving the client unable to page further.
+            return MessagePage(
+                messages=results,
+                has_more=has_more,
+                next_cursor=rows[0]["message_id"] if has_more and rows else None,
+            )
 
         except UnknownCursorError:
             raise
         except Exception as e:
+            # Surfacing the failure (the router turns it into a 500) matters more than
+            # a graceful empty page: an empty page is indistinguishable from "start of
+            # the conversation" and would end the client's pagination for good.
             logger.exception(f"Failed to get messages for conversation {conversation_id}: {e}")
-            return MessagePage(messages=[], has_more=False)
+            raise
 
     async def insert_message(
         self,
--- a/packages/console-backend/console_backend/services/messages_service.py
+++ b/packages/console-backend/console_backend/services/messages_service.py
@@ -11,10 +11,14 @@ from sqlalchemy import text
 from uuid6 import uuid7
 
 from ..db.connection import get_async_session_factory
-from ..models.message import Message
+from ..models.message import Message, MessagePage
 from .file_storage_service import FileStorageService
 
 logger = logging.getLogger(__name__)
+
+
+class UnknownCursorError(ValueError):
+    """Raised when a pagination cursor does not name a message in the conversation."""
 
 
 EXPIRY_BUFFER_SECONDS = (
@@ -228,29 +232,84 @@ class MessagesService:
         self._session_factory = get_async_session_factory()
         logger.info("MessagesService initialized (PostgreSQL)")
 
-    async def get_messages_by_conversation(self, conversation_id: str, user_id: str, limit: int = 100) -> list[Message]:
-        """Retrieve messages for a conversation.
+    async def get_messages_by_conversation(
+        self,
+        conversation_id: str,
+        user_id: str,
+        limit: int = 100,
+        before: str | None = None,
+    ) -> MessagePage:
+        """Retrieve one page of a conversation's messages, newest page first.
+
+        Pagination is keyset-based on `(created_at, id)`: a page holds the newest
+        `limit` messages older than the `before` cursor (or the newest `limit`
+        messages overall when no cursor is given). Paging backwards through the
+        history means passing the `message_id` of the page's oldest message as the
+        next `before`.
 
         Args:
             conversation_id: The conversation ID (partition key)
             user_id: The user ID (for access control)
             limit: Maximum number of messages to return (default: 100)
+            before: Cursor — the `message_id` to page back from (exclusive)
 
         Returns:
-            List of messages ordered by created_at (chronological order)
+            A MessagePage whose messages are ordered chronologically (oldest first)
+            and whose `has_more` says whether older messages remain.
+
+        Raises:
+            UnknownCursorError: `before` does not name a message in this conversation.
         """
         try:
             async with self._session_factory() as db:
+                cursor_created_at = None
+                cursor_id = None
+                if before is not None:
+                    cursor_row = (
+                        (
+                            await db.execute(
+                                text(
+                                    "SELECT created_at, id FROM messages "
+                                    "WHERE message_id = :before "
+                                    "AND conversation_id = :conversation_id AND user_id = :user_id"
+                                ),
+                                {"before": before, "conversation_id": conversation_id, "user_id": user_id},
+                            )
+                        )
+                        .mappings()
+                        .first()
+                    )
+                    if cursor_row is None:
+                        raise UnknownCursorError(f"Unknown cursor for conversation {conversation_id}")
+                    cursor_created_at = cursor_row["created_at"]
+                    cursor_id = cursor_row["id"]
+
+                # Fetch one extra row to detect whether older messages remain.
                 result = await db.execute(
                     text(
                         "SELECT * FROM messages "
                         "WHERE conversation_id = :conversation_id AND user_id = :user_id "
-                        "ORDER BY created_at ASC "
+                        + (
+                            "AND (created_at, id) < (:cursor_created_at, :cursor_id) "
+                            if before is not None
+                            else ""
+                        )
+                        + "ORDER BY created_at DESC, id DESC "
                         "LIMIT :limit"
                     ),
-                    {"conversation_id": conversation_id, "user_id": user_id, "limit": limit},
+                    {
+                        "conversation_id": conversation_id,
+                        "user_id": user_id,
+                        "cursor_created_at": cursor_created_at,
+                        "cursor_id": cursor_id,
+                        "limit": limit + 1,
+                    },
                 )
                 rows = result.mappings().all()
+
+            has_more = len(rows) > limit
+            # Rows come newest-first; drop the probe row, then flip to chronological order.
+            rows = list(reversed(rows[:limit]))
 
             results = []
             for row in rows:
@@ -287,11 +346,13 @@ class MessagesService:
                     continue
 
             logger.debug(f"Retrieved {len(results)} messages for conversation: {conversation_id}")
-            return results
+            return MessagePage(messages=results, has_more=has_more)
 
+        except UnknownCursorError:
+            raise
         except Exception as e:
             logger.exception(f"Failed to get messages for conversation {conversation_id}: {e}")
-            return []
+            return MessagePage(messages=[], has_more=False)
 
     async def insert_message(
         self,
--- a/packages/console-frontend/src/components/chat/contexts/ChatContext.tsx
+++ b/packages/console-frontend/src/components/chat/contexts/ChatContext.tsx
@@ -253,6 +253,9 @@ export function ChatProvider({ children, playgroundMode }: ChatProviderProps) {
   const [conversations, setConversations] = useState<Conversation[]>([]);
   const [activeConversationId, setActiveConversationId] = useState<string | null>(null);
   const [messagesMap, setMessagesMap] = useState<Map<string, Message[]>>(new Map());
+  // Mirror for synchronous reads inside loadMessages (which is deliberately kept
+  // dependency-free so effects don't resubscribe on every message).
+  const messagesMapRef = useRef<Map<string, Message[]>>(new Map());
   const [tasksMap, setTasksMap] = useState<Map<string, Task[]>>(new Map());
   const [contextIdsMap, setContextIdsMap] = useState<Map<string, string>>(new Map());
   const [waitingMap, setWaitingMap] = useState<Map<string, boolean>>(new Map());
@@ -303,6 +306,10 @@ export function ChatProvider({ children, playgroundMode }: ChatProviderProps) {
   useEffect(() => {
     activeConversationIdRef.current = activeConversationId;
   }, [activeConversationId]);
+
+  useEffect(() => {
+    messagesMapRef.current = messagesMap;
+  }, [messagesMap]);
 
   // Keep playground-mode ref in sync so config-hash filtering and message
   // tagging pick up changes (e.g. when the user views a different sub-agent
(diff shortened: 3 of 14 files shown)


packages/console-frontend/src/components/chat/hooks/useChatScroll.ts
```diff
@@ -0,0 +1,100 @@
...
+      const viewport = getViewport();
+      if (viewport) viewport.scrollTop = viewport.scrollHeight;
+    };
+
+    // While a page of older messages is pending or has just landed, the reader is at
+    // the top of the history and must stay where they are — never yanked to the bottom.
+    const anchor = prependAnchorRef.current;
+    const viewport = getViewport();
+    if (anchor && viewport) {
+      if (viewport.scrollHeight > anchor.scrollHeight) {
```
alloy-cicd: 🔴 **[correctness]** The anchor is consumed by ANY `scrollHeight` growth, and the first growth after `loadOlderMessages()` is not the prepended page — it's the *“Loading earlier messages…”* spinner row (`MessageList.tsx:365`), which renders as soon as `isLoadingOlderMessages` flips true (a dep of this effect, line 99).

Deterministic sequence on every older-page load:
1. Scroll handler stores the anchor and calls `loadOlderMessages` → `setLoadingOlderIds` re-renders with the spinner row.
2. This layout effect re-runs: `scrollHeight > anchor.scrollHeight` (by the spinner's height), so the anchor is **cleared** and `scrollTop` shifted by the spinner height.
3. The actual page lands; the anchor is already null, so the effect falls through to `scrollToBottom()` — the reader is yanked from the top of history to the newest message on **every** pagination attempt.

The same branch also misfires when a streaming reply grows the *bottom* of the transcript while an older page is in flight.

Possible fixes: hold the anchor while `isLoadingOlderMessages` is still true (only compensate once it flips false), render the spinner as an overlay that doesn't affect `scrollHeight`, or anchor to the previously-oldest message element (`offsetTop` delta) instead of total `scrollHeight`.

packages/console-frontend/src/components/chat/hooks/useChatScroll.ts
```diff
@@ -0,0 +1,100 @@
...
+    viewport.addEventListener('scroll', handleScroll, { passive: true });
+    return () => viewport.removeEventListener('scroll', handleScroll);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [hasMoreMessages, isLoadingOlderMessages, loadOlderMessages]);
+
+  // Drop a stale anchor when switching conversations — the next render is a fresh
+  // history that must land at the bottom, not at a previous conversation's offset.
+  useEffect(() => {
+    prependAnchorRef.current = null;
+  }, [activeConversationId]);
```
alloy-cicd: 🟠 **[correctness]** This stale-anchor reset is a passive `useEffect`, but the scroll-restore `useLayoutEffect` below runs **first** in the same commit. Switching to a cached conversation B while conversation A has an anchor pending (older page in flight) means the layout effect applies A's anchor to B: if B's transcript is taller, `scrollTop = B.scrollHeight − A.scrollHeight + A.scrollTop` — an arbitrary mid-transcript offset — and it returns without scrolling to bottom. Since a cached conversation triggers no further renders, B stays there instead of opening at its newest message.

Clearing the anchor in a `useLayoutEffect` keyed on `activeConversationId` *declared before* the scroll-restore effect (or storing the conversation id inside the anchor and ignoring mismatches) closes the gap.

packages/console-frontend/src/components/chat/hooks/useChatScroll.ts
```diff
@@ -0,0 +1,100 @@
...
+  const getViewport = () =>
+    (scrollAreaRef.current?.querySelector('[data-radix-scroll-area-viewport]') as HTMLElement | null) ?? null;
+
+  // Scrolling to the top pages in older history (infinite scroll upwards).
+  useEffect(() => {
+    const viewport = getViewport();
+    if (!viewport) return;
+
+    const handleScroll = () => {
+      if (viewport.scrollTop > LOAD_OLDER_THRESHOLD_PX) return;
```
alloy-cicd: 🟡 **[correctness/edge]** Older pages are only ever requested from a `scroll` event. When the newest page renders shorter than the viewport (plausible: most of its 100 rows can be timeline-only messages filtered out of the bubble list), there is no scrollbar, no scroll event ever fires, and older history is unreachable despite `hasMoreMessages === true`.

A fallback after each page lands — `if (viewport.scrollHeight <= viewport.clientHeight && hasMoreMessages) loadOlderMessages()` — or an explicit “load earlier messages” affordance would cover this.

packages/console-backend/console_backend/routers/message_router.py
```diff
@@ -60,6 +75,9 @@ async def get_messages_by_conversation(
                 for msg in messages
             ],
             "count": len(messages),
+            "has_more": page.has_more,
+            # The oldest message on this page is where the next (older) page starts.
+            "next_cursor": messages[0].message_id if page.has_more and messages else None,
```
alloy-cicd: 🔴 **[correctness]** `next_cursor` is derived from the **parsed** message list, but the service drops unparseable rows with `continue` (messages_service.py:341-345). Two failure modes:

1. If *every* row on a page fails parsing (e.g. a contiguous run of corrupt/legacy payloads), the response is `has_more: true, next_cursor: null`. The frontend stores `{hasMore: true, cursor: null}` and `loadOlderMessages`' guard (`!page.cursor → return`) then no-ops forever — everything older than the bad page is permanently unreachable, with no error surfaced.
2. If just the *oldest* row of a page is dropped, the cursor points at the oldest **parsed** message, so the next page re-spans rows already fetched — skip/overlap at the seam.

The keyset position belongs to the raw row set: have the service compute `next_cursor` from the raw `rows` (the oldest fetched row's `message_id`) and carry it on `MessagePage`, leaving the router a plain serializer.

packages/console-backend/console_backend/services/messages_service.py
```diff
@@ -287,11 +346,13 @@ async def get_messages_by_conversation(self, conversation_id: str, user_id: str,
...
             logger.debug(f"Retrieved {len(results)} messages for conversation: {conversation_id}")
-            return results
+            return MessagePage(messages=results, has_more=has_more)
 
+        except UnknownCursorError:
+            raise
         except Exception as e:
             logger.exception(f"Failed to get messages for conversation {conversation_id}: {e}")
-            return []
+            return MessagePage(messages=[], has_more=False)
```
alloy-cicd: 🟠 **[correctness]** The catch-all turns a transient DB error (connection blip, timeout) into HTTP 200 with `has_more=False`. New consequence with pagination: for an older-page request, `ChatContext.loadMessages` **unconditionally overwrites** the page state when `isOlderPage` (the seed-once guard at ChatContext.tsx:1077 only protects newest-page fetches), so one transient error latches `{cursor: null, hasMore: false}` — infinite scroll silently stops for the rest of the session, and the export path omits its truncation marker while the transcript is actually incomplete.

Pre-PR this swallow was self-healing (next full fetch retried); now the failure is persisted client-side. Letting the exception propagate (the router already maps unexpected errors to 500, and the client's `catch` keeps the old cursor so a later scroll retries) seems strictly better than a fabricated empty page.

packages/console-frontend/src/components/chat/contexts/ChatContext.tsx
```diff
@@ -1033,6 +1066,20 @@ export function ChatProvider({ children, playgroundMode }: ChatProviderProps) {
...
 
+      // Remember where the next (older) page starts. The newest page only seeds this
+      // when nothing is known yet: a refetch of the newest page (e.g. after a socket
+      // snapshot) must not rewind a cursor that has already been paged backwards.
+      const pageState = {
+        cursor: (data.next_cursor as string | null) ?? null,
+        hasMore: !!data.has_more,
+      };
+      setMessagePageMap((prev) => {
+        if (!isOlderPage && prev.has(conversationId)) return prev;
```
alloy-cicd: 🟠 **[correctness]** The seed-once guard assumes the loaded window always contains the newest-page refetch, but `messagesMap` is merge-only. If the conversation gains more than `MESSAGE_PAGE_SIZE` messages while the client is out of the room (long agent run, tab backgrounded, reconnect), the snapshot-reconcile refetch (line ~1322) merges only the newest 100 onto the old window, leaving a silent hole in the middle of the transcript — and this guard discards exactly the `next_cursor` that pointed into that hole. `hasMoreMessages` stays false (or the stale cursor stays put), so the gap can never be paged in, and the export claims completeness.

The cursor is a property of the *contiguous loaded window*, not of whichever fetch ran first. A newest-page refetch whose `has_more` is true and whose oldest message isn't already in the loaded window should replace the window (or at least adopt the fresh cursor) rather than merging.

packages/console-frontend/src/components/chat/contexts/ChatContext.tsx
```diff
@@ -1208,10 +1257,27 @@ export function ChatProvider({ children, playgroundMode }: ChatProviderProps) {
     } catch (e) {
       console.error('loadMessages failed', e);
```
alloy-cicd, line 1280: 🟡 **[correctness]** A 400 from the new `UnknownCursorError` path (cursor message deleted server-side, e.g. retention/cleanup) lands here, only logs, and leaves the stored `{cursor, hasMore: true}` untouched — so every subsequent scroll near the top fires the same doomed request forever, with no recovery and no user-visible signal.

On a non-404 failure of an older-page fetch, dropping the conversation's `messagePageMap` entry (or at least clearing the cursor on 400) would let the next newest-page fetch re-seed a valid cursor.

packages/console-backend/console_backend/services/messages_service.py
```diff
@@ -11,12 +11,16 @@
...
 
 from ..db.connection import get_async_session_factory
-from ..models.message import Message
+from ..models.message import Message, MessagePage
 from .file_storage_service import FileStorageService
 
 logger = logging.getLogger(__name__)
 
 
+class UnknownCursorError(ValueError):
```
alloy-cicd: 🔵 **[structure]** `UnknownCursorError` lives in the PostgreSQL-backed module, so `in_memory_messages_service.py` and `message_router.py` now import the sqlalchemy/db-loaded module just for this one class. The codebase already has `console_backend/exceptions.py` for exactly this shape of cross-service domain error (`ConversationNotFoundError`, `SessionNotFoundError`, …) — moving it there keeps the in-memory stack independent of the PG implementation.

packages/console-backend/console_backend/services/messages_service.py
```diff
@@ -228,30 +232,85 @@ def __init__(self, conversation_service=None) -> None:
...
                     text(
                         "SELECT * FROM messages "
                         "WHERE conversation_id = :conversation_id AND user_id = :user_id "
-                        "ORDER BY created_at ASC "
+                        + (
+                            "AND (created_at, id) < (:cursor_created_at, :cursor_id) "
+                            if before is not None
+                            else ""
+                        )
+                        + "ORDER BY created_at DESC, id DESC "
```
alloy-cicd, line 297: 🔵 **[perf]** `ORDER BY created_at DESC, id DESC` isn't covered by any index — `idx_messages_conversation_created` is only `(conversation_id, created_at)` (ddl/039). On long conversations every page fetch (this now runs on every conversation open and every scroll page) needs a sort step instead of a pure backward index scan of `limit+1` rows. A migration extending the index to `(conversation_id, created_at, id)` — or `(conversation_id, user_id, created_at, id)` to also cover the filter — makes the keyset query a straight range scan.

alloy-cicd (commented):

Solid pagination design overall (keyset `(created_at, id)` cursor, limit+1 probe, prepend anchor), but the scroll-anchor mechanism and a few error paths have correctness issues that defeat or wedge the feature. 9 inline comments: 7 correctness (2 high), 2 quality/perf.

**Highlights**
- 🔴 The prepend anchor is consumed by the *loading spinner's* height growth before the page arrives, so every older-page load ends with the reader yanked to the bottom (`useChatScroll.ts:65`).
- 🔴 `next_cursor` is derived from the *parsed* message list while the service drops unparseable rows, producing `has_more=true, next_cursor=null` — which the frontend can never advance past (`message_router.py:80`).
- 🟠 A transient DB error is returned as HTTP 200 `has_more=false` and latched client-side, permanently disabling pagination for the session (`messages_service.py:355`).

**Not inline (outside diff hunks / broader notes)**
- `ChatContext.tsx:1196` (pre-existing merge, now load-bearing): pages are merged as `[...existing, ...older]` then stable-sorted by millisecond timestamp only. Messages sharing the same `created_at` ms across a page boundary come back in inverted order — the SQL orders by `(created_at, id)` but the client sort loses the tiebreaker. Consider carrying `sort_key`/`id` into the client sort, or plain-prepending the deduped older page (it is strictly older than everything loaded), which is also O(n) instead of re-sorting the whole list per page.
- Timeline aggregation runs per-fetch, so an older page ending in timeline-only messages renders an orphaned synthetic timeline block instead of folding into the next page's bubble — visible artifact at the pagination seam.
- The pagination state triplet (`messagePageMap` + `loadingOlderIds` state + `loadingOlderIdsRef` mirror) re-implements what `@tanstack/react-query`'s `useInfiniteQuery` (already a dependency, used in `MessageList.tsx`) provides — `getNextPageParam`, `hasNextPage`, `isFetchingNextPage` with built-in dedupe. Also `hasMore` is fully derivable from `cursor != null` (the router returns them in lockstep), so the map could hold just the cursor.
- The endpoint still returns an untyped `-> dict` and the frontend duck-types it (`data.items ?? data.messages`, `data.next_cursor as string | null`). Declaring a `response_model` would flow `has_more`/`next_cursor`/`before` into the OpenAPI schema and the generated client, which is now stale for this endpoint.
- Nits: `MessagePage`'s `arbitrary_types_allowed` config is unnecessary (its fields are plain); `in_memory_messages_service.py:50`'s `if limit else []` guards a `limit=0` the router already rejects; `useChatScroll.ts:80` re-queries the viewport already held in `viewport`.

aartaria:
### Triage of the automated review (all findings addressed or answered) — 0575c33a

**Fixed**
1. **Anchor consumed by the indicator's height** (high) — confirmed, and deterministic: the request-time render grew `scrollHeight` by the spinner row, the layout effect took that as the prepend and cleared the anchor, so the real page always fell through to `scrollToBottom()`. Replaced the scrollHeight delta with an element anchor: the topmost `[data-message-id]` (already rendered by `MessageCard`) and its viewport offset, re-corrected on every commit and held until the request settles. Neither the indicator appearing nor unmounting can drift the view now.
2. **`next_cursor` derived from parsed messages** (high) — confirmed. It now comes from the raw rows in the service (`MessagePage.next_cursor`), so a dropped row no longer produces `has_more=true` with a null cursor. Test added: `test_get_messages_cursor_survives_unparseable_rows`.
3. **Transient DB error read as end-of-history** — the service no longer swallows it; the router turns it into a 500, so the client retries instead of latching `hasMore=false` for the session. Test added: `test_get_messages_propagates_db_failure`.
4. **Stale anchor on conversation switch** — the reset is now a `useLayoutEffect` declared before the scroll effect, so it wins the same commit.
5. **400 unknown cursor retried forever** — the dead cursor is retired client-side on a 400.
6. **Gap when >1 page arrives while the tab is away** — a newest-page refetch now adopts the fresh cursor when its oldest message is newer than everything held (i.e. the pages are disjoint), letting the reader page back across the gap. It still refuses to rewind a cursor in the normal overlapping case.
7. **No scroll event possible when content is shorter than the viewport** — the hook now pulls further pages until the content overflows or the history ends.
8. **`UnknownCursorError` placement** — moved to `console_backend/exceptions.py`; the in-memory service no longer imports the PostgreSQL module.
9. **Index** — added migration `078_index_messages_keyset_pagination.sql` extending the index to `(conversation_id, created_at, id)` to match the ORDER BY and the keyset predicate.

**Not changed (with reasons)**
- **Millisecond ties at page seams** — the backend keyset is `(created_at, id)`, so pages neither skip nor overlap. What remains is the frontend's merge sorting by timestamp alone, which can order two same-millisecond messages by which page they arrived on. That's pre-existing behaviour of the merge and a fix means threading a monotonic sort key into the frontend `Message` model — worth its own issue rather than this PR.
- **Orphaned timeline block at a seam** — an older page's trailing timeline-only events render as their own block instead of attaching to the first bubble of the newer page. It still appears in the right chronological place; purely cosmetic.
- **`useInfiniteQuery` instead of the hand-rolled state** — this context hand-rolls all of its per-conversation state (messages, tasks, streaming, waiting, interrupts) in the same idiom; introducing react-query for one slice would be inconsistent, and converting the rest is a separate refactor.
- **Untyped `dict` response vs the generated client** — matches every other endpoint in this router; the generated SDK needs a `npm run gen-sdk` pass against a running backend regardless (noted in the PR description).

**Verification after the fixes:** console-backend `pytest tests/` — 1367 passed; `tsc -b` and `npm run build` clean; `npm run lint` reports the same findings as `main`.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (42aaf68daa46b44abd4ad9a5c6e325ce0733c2d6..0575c33abebdc4540eb381d8a1e1e0e37283bd41)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Anchor consumed by spinner height (`useChatScroll.ts`) | ✅ verified-fixed — element anchor on `[data-message-id]`, re-corrected per commit, held until the request settles |
| 2 | Stale anchor on conversation switch | ✅ verified-fixed — reset is now a `useLayoutEffect` declared before the scroll-restore effect |
| 3 | `next_cursor` from parsed rows (`message_router.py`) | ✅ verified-fixed — cursor computed in the service from raw rows, carried on `MessagePage`; covered by `test_get_messages_cursor_survives_unparseable_rows` |
| 4 | Transient DB error → 200 `has_more=false` latch | ✅ verified-fixed — service re-raises, router returns 500; covered by `test_get_messages_propagates_db_failure` |
| 5 | Snapshot-refetch gap discards fresh cursor | ✅ verified-fixed — disjoint newest page (`gapAboveHeld`) adopts the fresh cursor |
| 6 | 400 unknown-cursor retried forever | ✅ verified-fixed — dead cursor retired client-side on 400 |
| 7 | Content shorter than viewport → no scroll event | ✅ verified-fixed — auto-pull until overflow or end of history (but see the one new inline comment) |
| 8 | `UnknownCursorError` module placement | ✅ verified-fixed — moved to `console_backend/exceptions.py`, all imports updated |
| 9 | Keyset index | ✅ verified-fixed — migration `078_index_messages_keyset_pagination.sql`, numbering is sequential |

**Rejected-by-author (accepted, not re-raised):** millisecond-tie ordering at page seams (pre-existing merge behaviour, needs a monotonic sort key threaded through the frontend model — follow-up issue), orphaned timeline block at a seam (cosmetic), `useInfiniteQuery` refactor (consistency with the context's idiom), typed response model / SDK regen (matches the rest of the router, regen noted in the PR description). All four rationales are sound; nothing in the delta changes them.

**Delta review:** the fixes are correctly implemented and well-tested. One new edge case introduced by the element-anchor mechanism (inline comment below). One observation, not blocking: `gapAboveHeld` compares the fetched page's server `created_at` against the held tail's client-side `timestamp`; an optimistic local message stamped by a client clock that lags the server can make an overlapping page look disjoint. The consequence is only a redundant page-back through already-held (deduped) messages, so this is fine as-is — just worth remembering if cursor rewinds ever show up in the wild.

packages/console-frontend/src/components/chat/hooks/useChatScroll.ts
```diff
@@ -0,0 +1,125 @@
...
+  // at and where it sat on screen, so it can be put back exactly there.
+  const prependAnchorRef = useRef<{ messageId: string; top: number } | null>(null);
+
+  const getViewport = () =>
+    (scrollAreaRef.current?.querySelector('[data-radix-scroll-area-viewport]') as HTMLElement | null) ?? null;
+
+  const anchorTopmostMessage = (viewport: HTMLElement) => {
+    const first = viewport.querySelector('[data-message-id]') as HTMLElement | null;
+    const messageId = first?.dataset.messageId;
+    if (!messageId) return false;
```
alloy-cicd: 🟡 **[correctness/edge — new in this round]** `requestOlderPage` now hard-depends on an anchor: when the viewport contains no `[data-message-id]` element, it returns before calling `loadOlderMessages()` — for **both** the scroll path and the new shorter-than-viewport auto-pull. `data-message-id` is only rendered by `MessageCard` (`MessageList.tsx:74`), and timeline-only entries (`showMessageCard === false`) render just a `UnifiedTimelineBlock` without it. So a loaded window whose visible rows are all timeline-only — e.g. a long agent turn that wrote >100 status updates, exactly the scenario behind the round-1 shorter-than-viewport finding — has `hasMoreMessages === true` but can never fetch: pagination is silently wedged for that conversation.

Two cheap fixes: move/duplicate `data-message-id={msg.id}` onto the per-message wrapper `<div key={msg.id}>` in `MessageList` so every entry (timeline-only included) is anchorable, or let `requestOlderPage` proceed without an anchor when none exists (the correction loop already tolerates a missing anchored element — `anchored` null just skips the `scrollTop` adjustment).

aartaria:
### Round 2 triage — c332a7a0

**Fixed:** the round-2 finding is real. `data-message-id` comes from `MessageCard`, which timeline-only entries never render, so a window made purely of status updates gave the hook nothing to anchor on — and `requestOlderPage` treated "no anchor" as "don't fetch", so neither the scroll path nor the shorter-than-viewport pull could reach that history. Both halves are addressed: `MessageList` now marks **every** entry with `data-message-anchor` (bubbles and timeline-only alike), and a request without an anchor proceeds regardless — the position correction is skipped, but the view is still held rather than snapping to the newest message.

**On the `gapAboveHeld` clock-skew note:** agreed, and the consequence is bounded as described — an optimistic client-stamped message can make an overlapping page look disjoint, costing one redundant refetch whose messages are deduped by id on merge. Not worth a server-timestamp round-trip to avoid.

**Verification:** `tsc -b` and `npm run build` clean; `npm run lint` unchanged vs `main`. Backend untouched by this commit (1367 tests still green as of 0575c33a).

PR #146, opened 2026-08-14
fix(agent-common): inline URL-sourced file attachments for the gateway

### LINKED ISSUE #145: URL-referenced file attachments crash on OpenAI Chat Completions: "OpenAI Chat Completions does not support file URLs"
## Summary

When a conversation carries a **non-image file attachment referenced by URL** (e.g. a PDF presigned URL) and the model is served through an OpenAI **Chat Completions** endpoint, the call fails client-side inside `langchain_openai` before any request is sent:

```
ValueError: OpenAI Chat Completions does not support file URLs.
  langchain_core/messages/block_translators/openai.py:129 in convert_to_openai_data_block
  langchain_openai/chat_models/base.py:314 in _format_message_content
  langchain_openai/chat_models/base.py:1709 in _get_request_payload
  agent_common/core/model_factory.py:168 in _get_request_payload
```

The turn dies with an unhandled `ValueError`; the user gets a failed run rather than a degraded-but-working answer.

## Root cause

We pass standard content blocks of the shape `{"type": "file", "url": "<https url>", "mime_type": ...}` straight through to the chat model. `langchain_core`'s OpenAI block translator accepts file blocks only as inline base64 (`file_data`) or as a previously uploaded `file_id` — a `url` source is rejected outright for the Chat Completions API. (Image blocks are fine: they map to `image_url`. This affects `file` blocks specifically.)

The passthrough is deliberate — `AgentBase._apply_provider_transforms` in `packages/agent-common/agent_common/a2a/base.py` was reduced to a no-op on the assumption that the LiteLLM gateway normalizes multimodal content for the target provider:

> Gateway-only: the app no longer knows the provider, and the LiteLLM gateway normalizes content for the target provider (image fetch/encode for Bedrock, MIME inference for Gemini). So this is a passthrough.

That assumption does not hold here: the failure happens in the LangChain client **before** the payload leaves the process, so the gateway never gets a chance to normalize it. The older Bedrock path (`preprocess_content_blocks_for_bedrock`, which fetched URLs and inlined base64) is still present but no longer invoked.

## Expected behaviour

A URL-referenced file attachment should reach OpenAI-family models successfully — or, if it genuinely cannot, fail as a handled, user-visible message rather than an unhandled `ValueError` mid-stream.

## Suggested direction (to be confirmed)

- Normalize `file` blocks with a `url` source before they reach the model: fetch the bytes and emit inline base64 (`file_data` + `filename`), reusing the existing fetch/encode helper rather than writing a new one.
- Decide where that normalization belongs — reviving a client-side transform hook vs. fixing it in the gateway — given that the client-side translator is what raises today.
- Guard the size: inlining large files as base64 is expensive; consider a size ceiling with a graceful fallback (e.g. surface the file through the attachments filesystem instead of the message content).
- Ensure the failure path is handled: an unsupported attachment should not surface as a raw `ValueError` traceback.

## Repro sketch

1. Send a message with a non-image file attachment carried as a `{"type": "file", "url": ...}` content block.
2. Route it to a model served via an OpenAI Chat Completions endpoint.
3. The run fails with the `ValueError` above.

## Pointers

- `packages/agent-common/agent_common/a2a/base.py` — `_apply_provider_transforms` (no-op passthrough), `_transform_blocks_for_bedrock`, `_infer_mime_types`
- `packages/agent-common/agent_common/core/model_factory.py` — gateway `ChatOpenAI` subclass whose `_get_request_payload` is on the failing stack
- `packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/bedrock_image_processor.py` — `preprocess_content_blocks_for_bedrock`, the existing URL→base64 helper
- `packages/agent-common/agent_common/backends/attachments_store.py` — file blocks with `url` / `base64` sources

aartaria:
closes ringier-data/nannos#145

A `file` content block carrying a `url` never reached the model. Every model is reached through the OpenAI-compatible gateway, so `langchain_openai` builds each request against the Chat Completions spec — and `langchain_core`'s block translator **raises** `ValueError: OpenAI Chat Completions does not support file URLs` while assembling the payload. That is client-side, before a request exists, so the gateway's own content normalization never got the chance to run: the turn died on an unhandled error (seen with a Slack-attached spreadsheet dispatched to a sub-agent, alongside image blocks that worked fine).

`LocalA2ARunnable._apply_provider_transforms` — a passthrough since content handling moved to the gateway — now inlines URL-sourced file blocks as base64 before the message is built.

Scope is deliberately narrow:
- **Images keep their URLs.** They map to `image_url` and work today; fetching them here would be pure cost.
- **Already-inline blocks are untouched.**
- **The filename keeps its extension**, since OpenAI identifies the file type from it. That is the reverse of the Bedrock document-name rule (which strips extensions), so the shared converter takes it as a flag rather than getting two copies.
- **A 20 MB inline ceiling.** base64 inflates by ~4/3 and the bytes then ride in every subsequent request and in the checkpoint. Over the ceiling — or on download failure, or when no MIME type can be resolved — the block degrades to a text description naming the file, so the turn continues instead of crashing. The ceiling also now covers the pre-existing Bedrock path.

Note this fixes the crash, not model support for every format: a file the target model cannot read is still a file it cannot read. It now arrives as a well-formed request rather than an unhandled `ValueError`.

## Tests

- `packages/ringier-a2a-sdk/tests/test_chat_completions_file_blocks.py` (new) — inlining, filename/extension preservation, images untouched, and each degradation path. One test asserts the regression directly: the raw block raises in `convert_to_openai_data_block`, the converted one serializes to `file_data`.
- `packages/agent-common/tests/test_local_a2a_runnable.py` — end-to-end through `_prepare_human_message_input` with a mixed image + file message.

`ringier-a2a-sdk`: 366 passed, 2 skipped. `agent-common`: 711 passed; 9 pre-existing failures in `test_thinking_blocks_round_trip.py` and 1 collection error in `test_foundry_cost_version.py`, both from optional deps missing locally (`langchain_openai`, `foundry_sdk`) and both identical on the base commit.

## Checklist


packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/bedrock_image_processor.py
```diff
@@ -131,7 +159,26 @@ async def _convert_block(block: dict) -> list[ContentBlock]:
         async with httpx.AsyncClient(follow_redirects=True) as client:
             resp = await client.get(url, timeout=60.0)
             resp.raise_for_status()
-            b64_data = b64.b64encode(resp.content).decode("utf-8")
+            raw = resp.content
```
alloy-cicd: **[major] The size ceiling is enforced only after the entire body is already in memory.** `client.get()` buffers the full response, so `len(raw) > _MAX_INLINE_BYTES` caps what gets base64'd but not what gets downloaded or held in RSS. A multi-GB presigned URL is read in full before line 163 runs — in a memory-limited pod that's an OOM kill, i.e. exactly the failure the ceiling was added to prevent. (And if a `MemoryError` fires inside the `try`, the bare `except Exception` below reports it as "could not load from URL", hiding the cause.)

The codebase already has the right primitive: `_fetch_bytes_capped` in `packages/orchestrator-agent/app/agents/file_analyzer.py:330` rejects on `Content-Length` before reading and aborts mid-stream otherwise. Promoting that into `ringier_a2a_sdk/utils` and using it here would fix this and de-duplicate the pattern.

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/bedrock_image_processor.py
```diff
@@ -131,7 +159,26 @@ async def _convert_block(block: dict) -> list[ContentBlock]:
         async with httpx.AsyncClient(follow_redirects=True) as client:
```
alloy-cicd: **[major] This fetch is now on the default path for every `LocalA2ARunnable`, with no SSRF guard.** The helper fetches an arbitrary block-supplied URL with `follow_redirects=True` and no address check. That was tolerable while it only ran for Bedrock agents; `_apply_provider_transforms` now runs it for every sub-agent dispatch carrying a file block. The repo already has the guard this needs — `_assert_public_url` + `_is_blocked_address` in `file_analyzer.py:221-268`, which also deliberately disables redirect-following so a redirect can't bounce past the check. As written, a file block pointing at `169.254.169.254` or an in-cluster service gets fetched and its bytes inlined into model context. Suggest reusing that guard here (same promotion into the SDK as `_fetch_bytes_capped`).

packages/agent-common/agent_common/a2a/base.py
```diff
@@ -580,26 +583,28 @@ def get_model_type(self) -> Optional[str]:
...
-            Transformed content blocks ready for the provider's API
+            Content blocks the OpenAI-compatible client can serialize
         """
-        # Gateway-only: the app no longer knows the provider, and the
-        # LiteLLM gateway normalizes content for the target provider (image
-        # fetch/encode for Bedrock, MIME inference for Gemini). So this is a
-        # passthrough. (Provider-specific content handling is now the gateway's
-        # responsibility — verify multimodal round-trips during rollout.)
-        return content_blocks
+        return await preprocess_file_blocks_for_chat_completions(content_blocks)
```
alloy-cicd: **[major] `audio`/`video` URL blocks hit the same class of client-side raise, and this fix doesn't cover them.** `_ensure_supported_block_types` keeps an audio block whenever the agent advertises the mode (dynamic agents return the model's declared `input_modes` verbatim), but `preprocess_file_blocks_for_chat_completions` is hardcoded to `("file",)`. Verified against the installed `langchain_core` translator: a URL-sourced `audio` block raises `ValueError: Key base64 is required for audio blocks.` at payload build, and `video` raises `Block of type video is not supported.` regardless of representation. So an audio-capable model's turn still dies client-side exactly like the file case this PR fixes. Suggest inlining audio here too (it has the same `base64` representation) and downgrading video blocks to the text description, rather than leaving both to crash.

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/bedrock_image_processor.py
```diff
@@ -148,29 +195,35 @@ async def _convert_block(block: dict) -> list[ContentBlock]:
...
         # content_builder._process_file_part and attachments_store) is a top-level
-        # `filename` key, which Bedrock Converse also reads. Keep that convention.
-        converted["filename"] = _sanitize_document_name(block.get("filename") or filename)  # type: ignore[typeddict-unknown-key]
+        # `filename` key, which Bedrock Converse and langchain_core's OpenAI
+        # translator both read. Keep that convention.
+        raw_name = block.get("filename") or filename
+        converted["filename"] = _sanitize_document_name(raw_name) if sanitize_document_name else raw_name  # type: ignore[typeddict-unknown-key]
     logger.info("Converted URL %s to inline base64 (%d chars)", block_type, len(b64_data))
     return [
         TextContentBlock(type="text", text=f"[Attached {block_type}: {filename}, URL: {url}]"),
```
alloy-cicd: **[medium] The full presigned URL (signature included) is now written into sub-agent message content on the gateway path.** Reusing `_convert_block` wholesale inherits this text block, which was written for the Bedrock path. That contradicts the deliberate policy in `content_builder._describe_file` (`content_builder.py:49`): *"The description deliberately OMITS the raw URI to prevent the orchestrator LLM from attempting to reproduce long pre-signed URLs, which leads to hallucination and corrupted security tokens."* On the new path the signed URL lands in the `HumanMessage`, persists into the LangGraph checkpoint, and is replayed to the gateway and tracing on every subsequent turn — and the model is free to echo a corrupted variant into a tool call. All three degradation branches embed the URL too. Suggest forwarding just the filename on the chat-completions path (or truncating the query string).

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/bedrock_image_processor.py
```diff
@@ -148,29 +195,35 @@ async def _convert_block(block: dict) -> list[ContentBlock]:
...
-        # Bedrock requires a (sanitized) name on the document block. ContentBlock
-        # has no dedicated name field, but the codebase convention (see
+        # Both transports want a name on the document block. ContentBlock has no
+        # dedicated name field, but the codebase convention (see
         # content_builder._process_file_part and attachments_store) is a top-level
-        # `filename` key, which Bedrock Converse also reads. Keep that convention.
-        converted["filename"] = _sanitize_document_name(block.get("filename") or filename)  # type: ignore[typeddict-unknown-key]
+        # `filename` key, which Bedrock Converse and langchain_core's OpenAI
+        # translator both read. Keep that convention.
+        raw_name = block.get("filename") or filename
```
alloy-cicd, line 251: **[medium] Filename edge cases defeat the extension-based typing this path exists for.** Two related gaps when the block has no `filename` key:

1. Attachments stored under opaque keys (`https://bucket.s3.../9f3c1a2e-...?X-Amz-...` with explicit `mime_type: application/pdf`) yield an extensionless `_filename_from_url` result. With `sanitize_document_name=False` it's forwarded as-is, and the whole point of this flag — OpenAI infers the file type from the extension — is lost; the request can 400 server-side, which none of the graceful-degradation branches catch. A `mimetypes.guess_extension(mime_type)` fallback when the name has no extension would close it.
2. A URL whose path ends in `/` makes `_filename_from_url` return `""` (the `if url` guard only catches an empty URL), so `converted["filename"] = ""` — which langchain treats as absent and replaces with `LC_AUTOGENERATED`. Without an explicit mime_type it's worse: `mimetypes.guess_type("")` → None → the fetchable file silently degrades to "(unknown type, not loaded)".

packages/agent-common/agent_common/a2a/base.py
```diff
@@ -580,26 +583,28 @@ def get_model_type(self) -> Optional[str]:
...
 
     @staticmethod
     async def _transform_blocks_for_bedrock(content_blocks: List) -> List:
```
alloy-cicd: **[minor] Dead code left behind:** with `_apply_provider_transforms` now routing to the chat-completions helper, `_transform_blocks_for_bedrock` and `_infer_mime_types` have zero callers anywhere in the repo (verified by grep — no source, no tests). `_infer_mime_types` also carries its own `_EXT_TO_MIME` table that silently competes with `_resolve_mime_type` in the SDK. Deleting both (and dropping `preprocess_content_blocks_for_bedrock` from the import) removes two live-looking entry points a future reader could wire back in — reintroducing exactly the provider branching this PR's gateway migration removed.

packages/ringier-a2a-sdk/ringier_a2a_sdk/utils/bedrock_image_processor.py
```diff
@@ -148,29 +195,35 @@ async def _convert_block(block: dict) -> list[ContentBlock]:
...
     for block in content:
-        if (
-            isinstance(block, dict)
-            and block.get("type") in _BINARY_BLOCK_TYPES
-            and "url" in block
-            and "base64" not in block
-        ):
-            new_blocks.extend(await _convert_block(block))
+        if _is_convertible(block, block_types):
+            new_blocks.extend(await _convert_block(block, sanitize_document_name=sanitize_document_name))
```
alloy-cicd: **[minor] N attachments download strictly sequentially, each with a fresh `AsyncClient`.** Every block pays its own DNS + TLS handshake and a 60 s timeout budget, in series, inside `_prepare_human_message_input` — i.e. before the sub-agent emits its first event. Six slow PDFs can add minutes of dead air to a dispatch and risk the upstream A2A timeout. One shared client passed into `_convert_block` plus `asyncio.gather` over the convertible blocks (splicing results back in order) keeps the output identical at the latency of the slowest download.

alloy-cicd (commented):

The fix itself is correct and well-tested: URL-sourced `file` blocks are inlined before `langchain_core`'s translator can raise, images stay URL-based, and every degradation path has a test. The inline comments cover three substantive gaps on the new hot path (memory-unsafe size ceiling, SSRF-unguarded fetch, audio/video blocks hitting the same class of client-side raise) plus smaller items.

Findings in files outside this diff (can't anchor inline comments there):

1. **`file_analyzer.py:634` still emits the exact block shape this PR declares unserializable.** `FileAnalyzerRunnable` overrides `_prepare_human_message_input` and never goes through `_apply_provider_transforms`; its unknown-file-type fallback appends `{"type": "file", "url": file_url}`, so an extensionless/octet-stream URL still dies with `ValueError: OpenAI Chat Completions does not support file URLs`. Pre-existing, but this PR is the natural place to route that branch through `preprocess_file_blocks_for_chat_completions`.
2. **Remote SDK agents are still unprotected.** `LangGraphAgent` builds `HumanMessage(content_blocks=a2a_parts_to_content(msg.parts))` (`langgraph.py:880`) and the base `_preprocess_input_messages` is a passthrough — only the Bedrock subclass overrides it. Any remote agent behind the gateway that receives a forwarded file URL hits the same raise. Worth a follow-up issue if out of scope here.
3. **A degraded file block also disappears from `/attachments/`.** When the download fails or the file is oversized, the replacement text block carries neither `url` nor `base64`, so `build_attachments_backend_from_blocks` skips it. Before this PR the URL block survived (passthrough) and the attachments backend could still lazily fetch it under its own 25 MB cap — now the advertised `/attachments/{name}` path 404s for the same file. Consider keeping the original block alongside the degradation text, or at least keeping `url` on it.
4. **Constant drift**: `_MAX_INLINE_BYTES` (20 MB) is the third independent inline ceiling (`MAX_DOC_FETCH_BYTES`/`MAX_AUDIO_FETCH_BYTES` in file_analyzer are also 20 MB, `attachments_store._MAX_ATTACHMENT_SIZE` is 25 MB). One exported constant would prevent them drifting apart on the next limits change.

aartaria:
Round 1 findings triaged — all seven addressed, plus the two related out-of-diff notes that fell inside the same helpers.

**Fixed**
1. **Ceiling after the download** → the cap is enforced while streaming: `Content-Length` precheck, then a running tally that abandons the stream. As suggested, `_fetch_bytes_capped` and `_assert_public_url`/`_is_blocked_address` were promoted from file_analyzer into the SDK (`ringier_a2a_sdk/utils/url_fetch.py`) and both callers now share them, rather than a second copy.
2. **SSRF** → `assert_public_url` runs before every fetch, and redirects are now **off** (a 302 would bounce a guarded URL to an unguarded one). A refusal degrades to a text description.
3. **Audio/video crash the same way** → confirmed against the installed translator and covered here rather than deferred: `audio` + `url` is inlined, `video` degrades to a text description without downloading (no video representation exists in Chat Completions, inline or not). Entry point renamed `preprocess_blocks_for_chat_completions`.
4. **Presigned URL in message content** → the URL is dropped from the text block on this path, including every degradation path. This matches `content_builder._describe_file`, which omits raw URIs because the model tries to reproduce them and corrupts the signature. The Bedrock path keeps its current text behind a `surface_url` flag. Tests assert the signature never appears in the output.
5. **Filename edge cases** → an extensionless or empty basename borrows an extension from the resolved MIME type.
6. **Dead code** → `_transform_blocks_for_bedrock` and `_infer_mime_types` removed (zero callers, as flagged — they went dead when content handling moved to the gateway, not in this PR).

**Not doing here**
7. **Serial downloads / client per block** — real but marginal: a message carries a handful of attachments, and the fetch is dominated by transfer time, not client setup. Parallelizing adds concurrency semantics to a path that just got its safety guarantees; not worth bundling into a bugfix.

Out-of-diff notes on file_analyzer's own fallback, remote SDK agents, `/attachments/` visibility for degraded blocks, and the drifting size constants are real but belong to their own change — they are not regressions from this PR.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (c91d4c0ea7c18a5bf232260c47a3874d87c85b33..81f900f4339935aec80e1b34752f0c7e92bd1b88)

**Every round-1 finding is resolved and the delta introduces nothing new. No inline comments this round.**

Ledger of round-1 findings:

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | [major] Size ceiling enforced after full download | **verified-fixed** — `fetch_bytes_capped` (new `ringier_a2a_sdk/utils/url_fetch.py`) rejects on `Content-Length` before reading and abandons the stream on a running byte tally; both cases have dedicated tests (`test_oversized_content_length_is_refused_before_reading_body`, `test_oversized_body_without_content_length_aborts_mid_stream`) |
| 2 | [major] No SSRF guard on the new default fetch path | **verified-fixed** — `assert_public_url` runs before every fetch, redirects are now off (with the correct rationale documented: a 302 would bounce past the guard), a refusal degrades to text, and the degradation text does not echo the blocked URL. The guard was promoted from file_analyzer into the SDK per the layering rule, and file_analyzer now imports the shared copy — no second implementation left behind |
| 3 | [major] `audio`/`video` URL blocks hit the same client-side raise | **verified-fixed** — audio is inlined (`_CHAT_COMPLETIONS_INLINE_TYPES`), video degrades to a text description in any form (correctly, since the spec has no video representation even inline — `test_inline_video_also_degrades` covers the case round 1 didn't mention). Both tests assert against the real `convert_to_openai_data_block` |
| 4 | [medium] Presigned URL written into message content | **verified-fixed** — `surface_url=False` on the chat-completions path drops the URL from the success text block and all degradation branches; tests assert `X-Amz-Signature` never appears. Bedrock keeps its existing behavior behind the flag |
| 5 | [medium] Filename edge cases (opaque keys, trailing slash) | **verified-fixed** — `_ensure_typed_filename` borrows the extension from the resolved MIME type, `_filename_from_url` falls back to `"unknown"` on empty basenames; both exact scenarios from round 1 are now tests (`9f3a1c.pdf`, `unknown.pdf`) |
| 6 | [minor] Dead code (`_transform_blocks_for_bedrock`, `_infer_mime_types`) | **verified-fixed** — both removed, import trimmed |
| 7 | [minor] Serial downloads, fresh client per block | **rejected-by-author** — reasoning accepted: a handful of attachments per message, transfer-dominated, and not worth adding concurrency semantics to a path that just gained its safety guarantees. Not re-raised |

The four out-of-diff notes (file_analyzer's unknown-type fallback, remote SDK agents, `/attachments/` visibility of degraded blocks, drifting size constants) were acknowledged by the author as follow-up material outside this PR — agreed, they are not regressions from this change.

Verification: checked out 81f900f4 and ran the touched suites locally — `test_chat_completions_blocks.py` + `test_langgraph_bedrock_middleware.py` (27 passed), `test_local_a2a_runnable.py` (29 passed), orchestrator file_analyzer/fetch tests (17 passed). Also confirmed `PayloadTooLarge` subclasses `ValueError`, so file_analyzer callers that catch `ValueError` for the user-facing too-large message are unaffected by the promotion.

aartaria:
## Manual QA run — 6 pass, 2 blocked, one finding outside this PR

Ran the attachment path end to end in a browser against a local stack built from this branch, with S3-backed uploads so attachments arrive as genuinely public presigned URLs. Sub-agent under test: a purpose-made local agent with **no MCP tools**, on a Gemini-family model whose gateway registration advertises `[text, image, audio, file]`.

Each fixture carries a verification phrase, so a pass means the bytes arrived — not merely that nothing crashed.

| Case | Result | Evidence |
|---|---|---|
| Spreadsheet → sub-agent | pass¹ | `Converted URL file to inline base64 (7008 chars)`; no `does not support file URLs` |
| PDF the agent must read | pass | Answered the document's phrase and headline number; 1,244 B → 1,660 chars |
| Image + document together | pass | Both canaries returned; **exactly one** conversion (the PDF), none for the 50 KB image |
| No signed URL in message text | pass | Sub-agent checkpoint holds `[Attached file: …pdf]` + `{type:file, base64:'JVBERi0xLjQK…'}` |
| Audio attachment | pass | Transcribed correctly; `Converted URL audio to inline base64 (377384 chars)` |
| file-analyzer unaffected | pass | Right answer, `Prepared 1 file content block`, **zero** conversions — the converter correctly stays out of that path |
| Oversized (>20 MB) | blocked | Browser upload tooling caps at 10 MB; stays covered by the streaming-ceiling unit tests |
| Non-public URL refused | blocked | Needs the stack restarted without cloud credentials; covered by `test_non_public_url_is_refused` |

### Against the issue's acceptance criteria

#145 asked that an attachment either reach OpenAI-family models successfully, **or** fail as a handled, user-visible message rather than an unhandled `ValueError`.

- **Handled failure: covered.** ¹The spreadsheet case turned out to be exactly that. The block is inlined correctly, then the provider rejects it — `400 … mimeType application/vnd…spreadsheetml.sheet, which is not supported`. The sub-agent errors, the orchestrator falls back, and the user reads a coherent "could not read the Excel file, try CSV". No unhandled `ValueError` in any case of the run.
- **Reaching an OpenAI-family model: not covered end to end.** Every case ran on a Gemini model. No OpenAI-family model in that gateway advertises `file`, so a sub-agent on one strips the block to a text description before this code runs. The gap is narrow — the translator that raised is invoked by `langchain_openai` for *every* model behind the gateway and builds the same payload whatever the destination, so the crash itself is verified for all of them. What is unverified is whether OpenAI/Azure *accepts* `file_data` for a PDF the way Vertex does.

### Follow-up worth filing

The gateway advertises one coarse `file` input mode, broader than what any single provider accepts per MIME type. We therefore inline an attachment the provider will refuse and spend a 400 on it. Filtering by the model's real MIME support and degrading to a text description would turn a failed turn into a graceful refusal. Out of scope here — this PR fixes the client-side crash, which it does.

Two caveats for anyone re-reading the results: a sub-agent with no MCP tools still has the built-in `ls`/`read_file`, so `/attachments` is an alternate read path — which is why the document verdicts rest on the conversion log and the checkpoint block shape rather than the answer text, and why the audio case is the confound-free one. And presigned URLs do persist in checkpoints via image blocks' own `url` field, which is by design but worth knowing.

PR #148, opened 2026-08-15
fix(client-slack): stop silently discarding a turn when the A2A stream drops

### LINKED ISSUE #147: client-slack: a dropped A2A stream silently discards the turn — widget deleted, in-flight record deleted, nothing posted
## Summary

When the A2A SSE stream to the orchestrator breaks mid-task, the Slack client **silently drops the turn**: the "thinking" widget that was already visible in the thread is deleted, the in-flight task record is deleted, no message and no error is posted, and the recovery loop finds nothing to recover. From the user's point of view the bot acknowledged the mention, showed a widget for a few seconds, and then pretended nothing happened.

A dropped stream is *recoverable* — the task is still running server-side and `tasks/get` would return the answer — so this is a pure client-side delivery bug, not a lost result.

## Reproduction / observed sequence

1. User mentions the bot; the client sends the message via `sendMessageStream`.
2. The first `kind: "task"` event arrives (state `submitted`); the in-flight record is saved and the thinking widget is posted.
3. No further SSE frames arrive; a few seconds later the connection is closed by the peer and `fetch` throws `TypeError: terminated` / `SocketError: other side closed`.
4. Log: `A2A stream error: TypeError: terminated`
5. Log: `Task state still processing, not finalizing stream: submitted`
6. Log: `Deleted in-flight task <id>`
7. Next recovery sweep: `No orphaned tasks found`. Nothing is ever posted to the thread.

## Root causes

**1. `responsePosted` is set even when nothing was posted** — `packages/client-slack/src/listeners/events/messageHandler.ts`

`finalizeStreamedTask` returns early with `{ messageTs: undefined }` whenever the task is not in an interrupted/terminal state (`taskResponseHandler.ts`, the `isInterruptedOrTerminated` guard). The caller ignores that and sets the flag unconditionally:

```ts
const result = await finalizeStreamedTask({ ... });
// We've delivered the final answer to Slack — recovery must never re-post it.
responsePosted = true;
```

The `finally` block then deletes the in-flight record on `responsePosted`, defeating the comment two blocks below it that explicitly intends to *keep* the record when "the stream dropped/errored mid-flight while still working — recovery should legitimately finish it". The record is the only handle recovery has on the task, so deleting it makes the failure permanent.

**2. The stream error is swallowed with no user-facing signal** — same file

The `for await` loop is wrapped in `try { … } catch (error) { logger.error(...) }` and execution simply falls through. Because a task *was* received, the "no task information received" guard doesn't fire either, and the generic `handleError` path (used by the outer catch) is never reached. The `finally` block then calls `streamer.discard()`, which removes the already-visible widget — actively worse than leaving it, since the user watched a message appear and vanish.

**3. Push notifications are requested but never configured** — `packages/client-slack/src/services/a2aClientService.ts`

`A2ASlackBasedRequest` carries `webhookUrl` / `webhookToken`, the caller populates them, and the token is persisted on the in-flight record for later validation — but `sendMessageStream` builds `MessageSendParams` with only `message` and `metadata`. No `configuration.pushNotificationConfig` is ever sent, so the server is never told to push. The webhook fallback that would have delivered this answer out-of-band is dead code on this path.

## Suggested fix

- Set `responsePosted` from the finalize result (`!!result.messageTs`), or have `finalizeStreamedTask` signal "not finalized" explicitly, so a non-finalized turn keeps its in-flight record and stays recoverable.
- Track whether the stream ended in error and, on a still-`working`/`submitted` task, keep the widget and update it to a visible "connection to the agent dropped — retrying/recovering" state instead of `discard()`. If recovery cannot take over, post a real error message rather than nothing.
- Populate `configuration.pushNotificationConfig` from `webhookUrl`/`webhookToken` in `sendMessageStream` so the webhook fallback actually engages.
- Consider reconnecting/resubscribing (`tasks/resubscribe`) once on a mid-stream transport error before falling back to recovery.

## Tests worth adding

- Stream throws after the initial `task` event → in-flight record survives, widget is not discarded, user sees a failure or a recovery message.
- `finalizeStreamedTask` early-return path → `responsePosted` stays false.

aartaria:
closes ringier-data/nannos#147

## Problem

When the SSE stream to the orchestrator breaks mid-task, the user is left with **nothing at all**: the "thinking" widget that was already visible is deleted, the in-flight record is deleted, no message and no error is posted, and the recovery loop then finds nothing to recover.

Observed in prod on 2026-08-15, when an orchestrator OOMKill dropped the connection 5s into a task.

## Changes

**1. `responsePosted` no longer set when nothing was posted.** `finalizeStreamedTask` returns `{ messageTs: undefined }` without posting whenever the task is not interrupted/terminal — exactly the dropped-stream case — but the caller set the flag unconditionally, and the `finally` block deletes the in-flight record on that flag. That record is the only handle recovery has.

It is now `result.messageTs !== undefined || streamer.hasAnswer`. The `hasAnswer` half matters: a stream can close cleanly right after the final artifact without a terminal status-update, in which case the user *has* the answer and keying off finalize alone would let recovery re-post it — the duplicate-post bug the original unconditional flag was guarding against.

**2. A dropped stream no longer deletes the visible widget.** Because a task *had* been received, the "no task information" guard didn't fire and `handleError` was never reached, so `finally` called `streamer.discard()` and removed a message the user had watched appear. It is now sealed with a visible notice and left in place.

**3. The recovery loop can actually finish the turn.** `recoverTask` deleted the in-flight record unconditionally after its first poll — even when `handleTask` posted nothing because the task was still non-terminal — and logged "Successfully recovered task". Recovery runs every 5 min over records idle >2 min, so any task outliving one cycle lost its handle while the user had seen nothing. It now keeps the record while the task is young and non-terminal, and past 30 minutes posts a visible "please send it again".

That give-up path is what makes the notice in (2) honest. A task killed mid-run never resumes — nothing re-invokes it when the pod restarts, so the state stays non-terminal forever — and without a give-up path those records just aged out against the 1h TTL, silently.

**4. A stream dying before the first task event now reports the failure.** No task id means nothing to recover, so the turn is definitively lost rather than pending.

## Not in this PR

Sending `configuration.pushNotificationConfig` — genuinely missing, but wiring it up is not a one-liner and sending it half-configured is worse than not sending it (the callback validates against the per-installation secret, not the per-turn UUID, so every push 403s after an uncached `listAll()` scan; and the handler drops chat tasks anyway). Filed as #150 with the full picture.

## Verification

- `npx tsc --noEmit` clean; full suite 55 tests / 6 suites pass.
- New tests cover finalize's non-terminal contract and the recovery loop's keep / give-up / deliver outcomes, **verified to fail against the unfixed code**.
- ESLint could not run locally: it needs the private `@alloy-ch/eslint-config-alloy`, resolvable only in CI.

Refs alloy-ch/rcplus-alloy-infrastructure-agents#241 (the OOM that exposed this)


packages/client-slack/src/services/a2aClientService.ts
```diff
@@ -184,6 +184,20 @@ export class A2AClientService {
...
+      // caller's webhookUrl/webhookToken were built, persisted on the in-flight
+      // record, and then never sent to the server — so the out-of-band delivery
+      // path was dead and a dropped stream had no fallback at all. The token is
+      // echoed back as X-A2A-Notification-Token and validated by the callback
+      // route (see handlers/a2aNotificationHandler).
+      ...(request.webhookUrl && {
+        configuration: {
+          pushNotificationConfig: {
+            url: request.webhookUrl,
+            ...(request.webhookToken && { token: request.webhookToken }),
```
alloy-cicd: **[major] The push fallback this enables is still dead end-to-end — and every non-local turn's callbacks are now rejected 403.**

Two gaps:

1. **The token can never validate.** What's forwarded here is the per-turn `randomUUID()` minted at `messageHandler.ts:681`. The callback route validates `X-A2A-Notification-Token` against the per-installation secret (`slackApp.ts:90-102`, `installationSecretService.get(bot.botName)`), and the `webhookToken` persisted on the in-flight record is read by nothing. So every push the orchestrator sends is 403'd — after `validateNotificationToken` has done a `storage.botInstallation.listAll()` scan. Since this registers push on **every** non-local turn, that's one-or-more wasted DB scans + a warn log per pushed event per turn.
2. **Even a valid callback can't deliver a chat turn.** `handleA2ANotification` only understands the scheduler payload shape (`{scheduler_status, agent_message, user_sub}` JSON in `status.message.parts[0].text`) and delivers by DM. An interactive turn's Task fails `getSchedulerPayload` and is dropped with a warn; nothing correlates by taskId back to the originating channel/thread.

The comment above ("The token is echoed back … and validated by the callback route") documents the intended design, not the current code — and the new tests only pin the params object, so the next person debugging a lost turn has to rediscover this across three files. Suggest either wiring the receiving side (validate against the record's `webhookToken`, deliver via taskId → in-flight record lookup, i.e. the same handle recovery uses) or holding this hunk back until it exists. Also note `client-email`'s A2A client builds this same config with `blocking: false` and a token fallback — worth one shared builder instead of a second inline literal.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -1114,13 +1127,31 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
     // than a manual flag.
     if (!isInterruptedOrTerminated(accumulatedTask?.status?.state)) {
-      await streamer.discard();
+      if (streamErrored && accumulatedTask) {
+        // The stream dropped while the agent was still working. The widget is
+        // already on screen and the user watched it appear, so deleting it (the
+        // `discard()` below) makes the bot look like it silently ignored them.
+        // Seal it with a visible notice instead and leave it in place: the task
+        // is still running server-side and the in-flight record is kept below,
+        // so the recovery loop will post the answer into this same thread.
```
alloy-cicd: **[major] "the recovery loop will post the answer into this same thread" isn't backed by the recovery loop.**

`recoverTask` deletes the in-flight record unconditionally after polling (`taskRecovery.ts:79`) — even when `handleTask` posted nothing because the task is still non-terminal (`taskResponseHandler.ts:288-291` returns `{messageTs: undefined}`) — and still logs "Successfully recovered" and returns `true`.

Concrete failure: stream drops 1 min into a 15-min run → widget sealed with this notice, record kept (correct so far). The recovery sweep fires while the task is still `working`: `handleTask` posts nothing, line 79 deletes the record anyway. The task completes later with no record, no open stream, and (see the a2aClientService comment) no working push callback — the answer is never delivered, with an explicit promise sitting on screen. That's the same silent-loss outcome this PR fixes, deferred by one recovery cycle; the window for hitting it is any task that outlives the recovery min-age.

The fix belongs in `recoverTask`: only delete when the poll returned a terminal/interrupted state (or when `handleTask` actually posted), leaving non-terminal records for the next sweep. Notably `recoverOrphanedTasks` already has `stillExists` bookkeeping for an "in progress, left in place" outcome that `recoverTask` never actually produces. There's no test covering the recovery half of this contract.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -975,6 +982,7 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
         logger.debug({ taskId: accumulatedTask?.id }, `Current state: ${accumulatedTask?.status?.state}`);
       }
     } catch (error) {
+      streamErrored = true;
```
alloy-cicd, line 985: **[medium] `streamErrored` is set by any throw in the `for await` body, not just transport drops.** This try wraps all the Slack-side work too: `streamer.appendAnswer` / `applyWorkPlan` / `applyActivity` / `appendThinking`, the HITL `chat.postMessage`, and `inFlightTaskStore.save`. A Slack 429 or `invalid_blocks` thrown mid-loop aborts consumption of a perfectly healthy A2A stream, then the finally tells the user "Lost the connection to the agent" and keeps the record — and since the answer may already have streamed, recovery re-posts it as a duplicate minutes later, with the actual root cause (a Slack API error) invisible.

Consider scoping the flag to the stream iterator itself — e.g. catch-and-log around the Slack-side calls inside the loop, or a narrower try around the iteration protocol — so client-side failures don't masquerade as orchestrator drops.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -1005,8 +1013,13 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
     });
-    // We've delivered the final answer to Slack — recovery must never re-post it.
-    responsePosted = true;
+    // Only mark the turn delivered if finalize actually delivered it. finalize
+    // returns `{ messageTs: undefined }` without posting anything when the task
+    // is still in a non-terminal state — which is exactly what happens when the
+    // stream drops mid-flight. Setting the flag unconditionally deleted the
+    // in-flight record for a turn the user never saw, so the recovery loop found
+    // nothing to recover and the request was silently lost.
+    responsePosted = result.messageTs !== undefined;
```
alloy-cicd: **[medium] `result.messageTs !== undefined` doesn't mean "finalize actually delivered it" — it's exactly `isInterruptedOrTerminated(state)`, and it's wrong in both directions:**

- **Terminal, nothing posted:** the terminal return is `streamer.answerTs || streamer.ts || statusMessageTs || messageTs` (`taskResponseHandler.ts:409`), and the last fallback is the user's own message ts — always defined. So the flag reads true even when nothing was posted. (Harmless today only because `isTerminatedState` at line 1165 would delete the record anyway, but the comment/tests enshrine a contract the function doesn't have.)
- **Non-terminal, already delivered:** a stream that ends *cleanly* right after the final artifact — no throw, no terminal status-update, the exact case the pre-existing comment at lines 500-505 warns about — leaves the state `working`. `streamErrored` is false so the widget is discarded, but `responsePosted` is false too, so the record now survives and the recovery loop re-posts the full, already-streamed answer as a fresh message minutes later. The old unconditional `responsePosted = true` existed precisely to prevent this; this change trades the silent-loss bug for a duplicate-post one in the clean-close case.

Suggest having finalize return an explicit `posted: boolean` (which is also what `recoverTask` needs — see the other comment) instead of inferring delivery from a ts whose actual contract is "what should reactions attach to". The new test file pins the inference (`const responsePosted = result.messageTs !== undefined`) rather than the contract, which keeps this coupling alive.

Nit: the next line re-tests the same predicate — `if (result.messageTs)` could be `if (responsePosted)`.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -1114,13 +1127,31 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
     // thinking-steps message is left behind. Inferred from the task state rather
     // than a manual flag.
     if (!isInterruptedOrTerminated(accumulatedTask?.status?.state)) {
-      await streamer.discard();
+      if (streamErrored && accumulatedTask) {
```
alloy-cicd, line 1137: **[medium] A stream that throws before the first `task` event still takes the old silent path.** With `accumulatedTask === null` this guard falls through to `discard()`, and the handler already returned at the `!accumulatedTask` check ("Silently failing without sending a response to the user") — the "Working…" widget vanishes with no notice, and there's no in-flight record to recover from either (`save()` only runs on the task event). An orchestrator that dies during connection setup — before acknowledging the task — reproduces the exact user-visible symptom this PR fixes, one event earlier. Nothing in `streamer.finish()` needs the task, so the seal-with-notice could run on `streamErrored` alone (with different wording when there's no record to recover, e.g. "couldn't reach the agent, please retry").

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -1114,13 +1127,31 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
+        // already on screen and the user watched it appear, so deleting it (the
+        // `discard()` below) makes the bot look like it silently ignored them.
+        // Seal it with a visible notice instead and leave it in place: the task
+        // is still running server-side and the in-flight record is kept below,
+        // so the recovery loop will post the answer into this same thread.
+        await streamer
+          .finish({
+            trailingMarkdown:
+              '\n\n⚠️ _Lost the connection to the agent while it was working. ' +
+              "It's still running — I'll post the answer here as soon as I can reach it again._",
```
alloy-cicd: **[minor] In the common drop-before-any-answer case, this notice renders as a detached message, not in the widget.** `finish()` attaches `trailingMarkdown` to the answer stream when one exists; when the drop happens before any artifact-update (`answerStreamer` undefined) it `chat.postMessage`s the raw string as its own standalone message (`thinkingStepsStreamer.ts:586-596`) — still carrying the leading `\n\n` — while the widget itself is only relabelled "Interrupted" with no explanation inline.

Also worth a deliberate choice between primitives here: `pause()` is the existing "seal but keep continuable" operation (used by the HITL path), whereas `finish()` permanently closes the stream — so "leave it in place" in the comment above means recovery can only post a brand-new message below, never complete the widget the user is watching. If that's intended (recovery goes through `handleTask`, which posts fresh messages), fine — but then the notice's leading `\n\n` and the detached-message rendering deserve a tweak.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -1114,13 +1127,31 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
+      } else {
+        await streamer.discard();
+      }
     }
 
     // Remove the in-flight record once we've delivered a response to Slack,
     // otherwise the periodic recovery loop treats the leftover record as orphaned
     // and re-posts the message ~5 min later (the recovery cadence). The signal is
-    // `responsePosted` (set right after finalize), NOT the task state: a soft
+    // `responsePosted` (set from finalize's own result — it is false when finalize
```
alloy-cicd, line 1167: **[minor] The "deliberately KEPT" bullet just below contradicts the code.** It says a true HITL approval widget "returns early before finalize, so `responsePosted` stays false and the record persists for the resume turn" — but the HITL path sets `responsePosted = true` at line 895 (deliberately, per the rationale there: resume re-enters via the button payload, and a kept record would re-post the approval prompt). So the record is deleted, not kept. The contradiction predates this PR, but since this change rewrites this comment block and makes `responsePosted` the single load-bearing retention signal, it's worth fixing the bullet now — before someone "fixes" line 895 to match the comment and reintroduces the duplicate-approval-prompt bug.

packages/client-slack/test/utils/finalizeStreamedTask.test.ts
```diff
@@ -0,0 +1,88 @@
...
+        messageContext,
+      });
+
+      // undefined is the signal the handler uses to keep the in-flight record so
+      // the recovery loop can still finish the turn.
+      expect(result.messageTs).toBeUndefined();
+    }
+  );
+
+  test('a non-terminal result must not be read as a delivered response', async () => {
```
alloy-cicd: **[nit]** This test re-runs the `submitted` case already covered by the `test.each` above and copies the production expression into the assertion — it pins nothing the parameterised test doesn't. The coupling it wants to guard (the handler keeping the in-flight record when finalize didn't post) lives in `messageHandler`/`recoverTask` and would need a test at that level. Also, the two Proxy factories could be one `forbidden(label)` helper — and a Proxy that throws on *any* property read (including `then` / `Symbol.toStringTag`) can fail on harness-internal probes rather than genuine contract violations; asserting on a recording mock's empty call log is sturdier.

alloy-cicd (commented):

The record-retention fix in `messageHandler` addresses a real silent-loss bug, but the end-to-end story doesn't hold together yet:

- **Push fallback is still dead** — the per-turn token can never pass the callback route's per-installation-secret validation, and even a valid callback lands in a handler that only understands scheduler payloads. As merged, this hunk adds a 403 + DB scan per pushed event and nothing else. *(major)*
- **The recovery loop breaks the promise the sealed widget makes** — `recoverTask` deletes the in-flight record on its first poll even when the task is still non-terminal and nothing was posted, so a turn that outlives one recovery cycle is still permanently lost. *(major)*
- `responsePosted = result.messageTs !== undefined` is really `isInterruptedOrTerminated(state)`, which introduces a duplicate-post regression in the clean-close case the old comment at lines 500-505 explicitly warned about. *(medium)*
- `streamErrored` is set by any throw in the loop body, including Slack API errors, misreporting client-side failures as lost connections. *(medium)*
- Plus: pre-first-event drops still silent, the notice renders as a detached message when nothing streamed yet, a comment/code contradiction around the HITL path, and test nits.

All findings verified against `slackApp.ts`, `taskRecovery.ts`, `taskResponseHandler.ts`, and `thinkingStepsStreamer.ts` at head `83da596`. Details inline (8 comments).

aartaria:
## Triage of the 🤖 automatic review

All 8 findings addressed. Pushed as a second round.

**Fixed**

- **(1) Push notifications dead end-to-end** — verified: `validateNotificationToken` matches the token against the per-installation secret, not the per-turn `randomUUID`, so every push would 403 after an uncached `listAll()` scan; and `handleA2ANotification` only handles scheduler payloads and DMs. Sending the config half-wired is worse than not sending it, so **the change was removed from this PR** and the full end-to-end work filed as #150. The reasoning is documented in place so nobody re-adds the one-liner.
- **(2) Recovery loop deletes the record unconditionally** — this was also spotted independently on the PR thread ("will it recover for real after an OOM?" — it would not). `recoverTask` now keeps the record while the task is young and non-terminal, and past 30 min posts a visible "please send it again" rather than letting it age out against the 1h TTL. Three tests cover keep / give-up / deliver, verified to fail against the old code.
- **(4) Duplicate-post regression** — correct and important: keying off finalize alone reintroduced exactly what the original unconditional flag guarded against. Now `result.messageTs !== undefined || streamer.hasAnswer`, which distinguishes "delivered, no terminal event" from "nothing reached the user".
- **(5) Silent path before the first task event** — no task id means nothing to recover, so the turn is definitively lost; it now reports the failure instead of returning silently.
- **(7) Stale HITL bullet** — correct, the comment contradicted the code. Rewritten: HITL is a delete, not a keep.
- **(8) Test nits** — removed the duplicated third finalize case.

**Accepted, not changed**

- **(3) `streamErrored` set by any throw, including Slack errors.** True. Narrowing it to transport-only errors needs a bigger refactor of the loop body than this PR should carry. The consequence is now bounded: with the `hasAnswer` fix, a Slack 429 *after* the answer streamed deletes the record (no duplicate), and *before* it takes the drop path, where the record is kept and recovery either delivers or gives up audibly. The notice wording is slightly off in that case ("lost the connection to the agent"), which is a cosmetic mismatch rather than a lost turn.
- **(6) Notice posts as a detached message rather than inside the widget.** Behaviour of `finish({trailingMarkdown})` in the streamer; changing it affects every finalize path, not just this one. The user sees the notice either way.
- **(8) Proxy test doubles are brittle.** Deliberate: they fail loudly if finalize ever starts touching the streamer or Slack on a non-terminal task, which is the precise contract under test.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (83da5961329da2bbc215f8a1000cabf4c3a5ebdb..8b3607c554f95d764cd10c297fd8d9da617d6c5f)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Push config dead end-to-end (token mismatch + scheduler-only handler) | **verified-fixed** — hunk removed, reasoning documented in place, follow-up filed as #150, params test deleted |
| 2 | `recoverTask` deletes the record on a non-terminal poll | **verified-fixed** — non-terminal records are now kept while young; past 30 min a visible give-up notice is posted before deletion; the three outcomes (keep / give-up / deliver) are pinned by new tests |
| 3 | `streamErrored` set by any throw incl. Slack API errors | **rejected-by-author** — accepted as known, consequence bounded by the `hasAnswer` fix; reasoning is sound, not re-raised |
| 4 | `responsePosted` wrong in both directions (terminal ts fallback / clean-close duplicate) | **verified-fixed** for both reported directions via `\|\| streamer.hasAnswer` — but the new predicate introduces one new inconsistency, see the single inline comment |
| 5 | Pre-first-task-event drop fully silent | **verified-fixed** — now reports via `handleError` |
| 6 | Seal notice renders as detached message | **rejected-by-author** — streamer-level behavior, deferred; not re-raised |
| 7 | HITL "deliberately KEPT" comment contradicts line 895 | **verified-fixed** — comment rewritten to match the code |
| 8 | Test nits (duplicated case, Proxy doubles) | **verified-fixed / rejected** — duplicated case removed; Proxy doubles kept deliberately with reasoning, fine |

**Delta review:** one new finding (inline), introduced by the fix for #4 — when the stream errors mid-answer, `hasAnswer` deletes the record while the seal notice still promises recovery-or-notification, so neither promise can be kept and the answer's lost tail is never delivered. Everything else in the delta checks out: the recovery keep/give-up logic is correct (`handleTask`'s terminal return is always truthy, so the `!result.messageTs` branch is exactly the non-terminal case), the new `taskRecovery` tests genuinely pin the contract, and the push-config removal is clean.

Minor, non-blocking: the retention comment's opening line ("set from finalize's own result") is now slightly stale since the signal is finalize-result *or* `hasAnswer`.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -1005,8 +1020,23 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
+    //
+    // But `result.messageTs` alone is not enough either: a stream can close
+    // cleanly right after the final artifact without ever delivering a terminal
+    // status-update (see the note at the top of this function). The user HAS the
+    // answer in that case, so keying off finalize alone would keep the record and
+    // let recovery re-post the same answer — the duplicate-post bug the original
+    // unconditional flag was guarding against. `streamer.hasAnswer` is true once
+    // answer text has been appended, which distinguishes "delivered, no terminal
+    // event" from "nothing reached the user".
+    responsePosted = result.messageTs !== undefined || streamer.hasAnswer;
```
alloy-cicd: **[medium — new in this round] `hasAnswer` and the seal notice now contradict each other when the stream errors mid-answer.**

`hasAnswer` is `answerStarted` — true after the *first* answer chunk (`thinkingStepsStreamer.ts:246`). So for a stream that throws while the answer is still streaming (`streamErrored && hasAnswer`, state still `working`):

1. this line sets `responsePosted = true` → the in-flight record is **deleted** at the cleanup below;
2. but the seal branch in the finally checks only `streamErrored && accumulatedTask`, so it still posts *"I'll post the answer here if it can still be recovered, and tell you if it can't."*

Recovery never sees the task (no record), so neither half of that promise can be kept — and the user may be looking at 10% of the answer that's now silently treated as delivered. The lost tail is exactly the case where recovery re-posting the full answer would have been the right outcome (a duplicate of a *partial* answer is a much better failure than a missing tail).

The two decisions should share one predicate. Options:

- **Don't count `hasAnswer` as delivered when `streamErrored`** — `responsePosted = result.messageTs !== undefined || (streamer.hasAnswer && !streamErrored)`. The clean-close duplicate-post case (this fix's target) has `streamErrored === false`, so it stays fixed; the errored-mid-answer case keeps the record and recovery either completes the turn (acceptable duplication of a partial answer) or gives up audibly.
- Or keep the delete but branch the seal wording on `responsePosted`: when the record won't survive, seal with something like *"the connection dropped — the answer above may be incomplete; please ask again if it looks cut off"* instead of promising a follow-up.

The first option seems truer to the intent — client-side it's indistinguishable whether the streamed answer was complete, and only the error case has a real risk of a missing tail.

aartaria:
## Round 3 — the new finding was right, fixed

Good catch, and it was a bug I introduced in round 2.

`streamer.hasAnswer` flips true on the **first** answer chunk, so `result.messageTs !== undefined || streamer.hasAnswer` treated a stream dying mid-answer as delivered. The in-flight record was dropped, the answer's tail silently lost, and the sealed widget left promising a recovery that could never run — a worse outcome than the duplicate it was meant to prevent, because it loses data rather than repeating it.

Took the suggested `hasAnswer && !streamErrored` rather than branching the seal wording: keeping the record means the promise on the widget stays true, and recovery either completes the answer or gives up audibly. A duplicate costs the user an extra message; a dropped tail costs them the answer.

**Made it testable.** This condition has now been wrong in both directions in three rounds — deleting a record the user never saw, then keeping one they already had, then deleting one whose answer was truncated. All three fail *silently*. It is now `isTurnDelivered()` in `taskResponseHandler.ts`, with the reasoning for each input documented and six cases pinning the truth table:

| finalize posted | answer streamed | stream errored | delivered |
|---|---|---|---|
| ✓ | – | – | **yes** (authoritative) |
| – | ✓ | – | **yes** (clean close, no terminal event) |
| – | ✓ | ✓ | **no** ← the bug you found |
| – | – | ✓ | no |
| – | – | – | no |
| ✓ | ✓ | ✓ | **yes** (finalize wins) |

Verified the mid-answer case fails against the round-2 expression and passes with the fix. Full suite: 61 tests / 6 suites.

aartaria:
## Give-up notice no longer fires at someone who moved on

Good catch — and worse than mildly confusing in practice. That notice can land up to **30 minutes** after the request that failed, in the same thread. By then the common case is that the user gave up on their own, resent, and got a good answer. Dropping "please send it again" underneath that conversation points at a request they can no longer identify and invites them to repeat work that already succeeded.

**Fix: check whether anything has succeeded since, before speaking.**

`lastProcessedTs` advances on every delivered turn for a thread, and for the stranded turn it was pinned to that turn's own `messageTs` when the task started. So a strictly-later value means another turn in this thread completed after ours was submitted — the user moved past the failure. In that case the record is cleaned up **silently**; the user is only told when nothing has succeeded since and they are plausibly still waiting.

Deliberately *not* solved by shortening the window or dropping the notice altogether: someone who is genuinely still waiting needs to be told, and that is the case this whole PR exists to fix. The check separates the two populations rather than trading one against the other.

**Also reworded**, because even in the notify case we can't detect every way a user has moved on:

> ⚠️ I couldn't finish an earlier request in this thread — the agent was interrupted while working on it and I can't recover the answer. Please send it again if you still need it.

Anchors itself ("an earlier request in this thread") instead of assuming the reader knows which request, and asks rather than demands ("if you still need it").

Two tests added — silent-when-superseded and still-notifies-when-not — with the first verified to fail against the previous behaviour. Full suite: 63 tests / 6 suites.

Rebased onto the new `main` (which now carries #149) and re-verified there: `tsc` clean, all tests green.

PR #149, opened 2026-08-15
fix(orchestrator): bound concurrent MCP tool discovery

aartaria:
## Problem

Cold discovery opens one connection per MCP server and awaits them all in a single **unbounded** `asyncio.gather`. Every server's session, response body, parsed schema and resulting tool objects are therefore alive at the same moment, so the memory peak scales with the size of the gateway catalogue rather than with anything we control.

In one deployment this reached **31 concurrent fetches** (30 gateway servers plus console) and OOMKilled the pod twice on 2026-08-15, each time within 5–6 s of a single request arriving. The container's own cgroup showed a **424 MiB idle floor of a 1 GiB limit** on a pod that had served zero requests, so the burst had roughly 600 MiB of headroom available — and consumed all of it. All 31 sessions were established inside a ~1 second window.

The connection-per-server design is deliberate (`langchain_mcp_adapters` doesn't record `server_name` in tool metadata, so one connection per server is how tool→server attribution is preserved). The missing concurrency bound is the defect.

## Change

Bound the fan-out with an `asyncio.Semaphore`, defaulting to 5 and overridable via `MCP_DISCOVERY_CONCURRENCY`.

Results stay positionally aligned with the `connections` dict, so the `zip()` that maps results back to server slugs is unaffected, as is the partial-failure handling (`return_exceptions=True`).

## Trade-off

A little cold-start latency on the first request of a conversation (discovery is cached for 300 s, so this is paid per cold request, not per turn), in exchange for a peak that no longer grows every time a server is added to the gateway.

## Verification

- New test drives 20 servers with a limit of 3 and asserts at most 3 fetches are ever in flight, and that every server is still visited exactly once. **Verified to fail against the unfixed code** (`expected at most 3 concurrent fetches, saw 20`).
- Full orchestrator unit suite: 626 tests pass.
- `ruff check` clean on the changed files. (`ruff format` reports these files as unformatted, but it does so on `origin/main` too — the repo's lint gate runs `ruff check` only, and reformatting whole files would balloon the diff.)
- One existing test fixture needed `MCP_DISCOVERY_CONCURRENCY` set: it uses `Mock(spec=AgentSettings)`, so the new setting otherwise arrives as a `Mock`. Fixed in the fixture rather than by adding defensive coercion to production code.

Refs alloy-ch/rcplus-alloy-infrastructure-agents#241

### DIFF
--- a/packages/orchestrator-agent/app/core/discovery.py
+++ b/packages/orchestrator-agent/app/core/discovery.py
@@ -32,6 +32,12 @@ from ..models.config import AgentSettings
 
 logger = logging.getLogger(__name__)
 
+# Process-wide bound on concurrent MCP tool-catalogue fetches; see
+# ToolDiscoveryService._discovery_semaphore for why this is not per-call.
+# Created lazily so the configured limit is read at first use rather than import.
+_DISCOVERY_SEMAPHORE: asyncio.Semaphore | None = None
+_DISCOVERY_SEMAPHORE_LIMIT: int | None = None
+
 
 async def _console_attribution_interceptor(request, handler):
     """Stamp the caller's cost-attribution (user_sub, conversation_id, sub_agent_id, …) on every
@@ -257,6 +263,25 @@ class ToolDiscoveryService:
                 logger.error(f"Failed to fetch MCP servers: {e}", exc_info=True)
             return []
 
+    def _discovery_semaphore(self) -> asyncio.Semaphore:
+        """Process-wide bound on concurrent MCP catalogue fetches.
+
+        Deliberately module-level, NOT per-call: discovery runs per user on cache
+        miss, so a per-invocation semaphore would still allow limit x N concurrent
+        fetches when N users hit a cold cache together — with a limit of 5, seven
+        simultaneous users already exceed the ~31 that OOMKilled the pod. What has
+        to be capped is what the *process* holds open at once.
+
+        The cost is that unrelated users' cold discoveries queue behind each other.
+        That is the intended trade: bounded memory over parallel cold starts.
+        """
+        global _DISCOVERY_SEMAPHORE, _DISCOVERY_SEMAPHORE_LIMIT
+        limit = self.config.MCP_DISCOVERY_CONCURRENCY
+        if _DISCOVERY_SEMAPHORE is None or _DISCOVERY_SEMAPHORE_LIMIT != limit:
+            _DISCOVERY_SEMAPHORE = asyncio.Semaphore(limit)
+            _DISCOVERY_SEMAPHORE_LIMIT = limit
+        return _DISCOVERY_SEMAPHORE
+
     async def _get_tools_with_retry(
         self,
         client: MultiServerMCPClient,
@@ -286,7 +311,11 @@ class ToolDiscoveryService:
 
         for attempt in range(max_retries):
             try:
-                server_tools = await client.get_tools(server_name=server_name)
+                # Hold a slot only for the fetch itself. Keeping it across the
+                # backoff sleep below would let a few flaky servers idle away the
+                # whole budget while healthy ones wait on a user-facing path.
+                async with self._discovery_semaphore():
+                    server_tools = await client.get_tools(server_name=server_name)
                 # Tag tools with server_name metadata
                 for tool in server_tools:
                     if tool.metadata is None:
@@ -433,24 +462,23 @@ class ToolDiscoveryService:
             # Gather tools from all servers with retry logic.
             # Use asyncio.gather with return_exceptions=True to handle partial failures gracefully.
             #
-            # Bounded by a semaphore: an unbounded fan-out keeps every server's session,
-            # response body, parsed schema and tool objects alive at once, so the memory
-            # peak scales with the number of servers on the gateway rather than with
-            # anything we control. On prod that reached ~31 concurrent fetches and
-            # OOMKilled the pod (2026-08-15). Results stay positionally aligned with
-            # `connections` below, so the zip() that follows is unaffected.
-            discovery_semaphore = asyncio.Semaphore(self.config.MCP_DISCOVERY_CONCURRENCY)
-
-            async def _bounded_get_tools(slug: str) -> list:
-                async with discovery_semaphore:
-                    return await self._get_tools_with_retry(client, slug)
-
+            # Each fetch takes a slot from a process-wide semaphore (see
+            # _get_tools_with_retry), so the number of MCP sessions and response
+            # bodies held open AT ONCE no longer scales with the size of the
+            # gateway catalogue. On prod an unbounded fan-out reached ~31
+            # concurrent fetches and OOMKilled the pod (2026-08-15).
+            #
+            # Note this bounds only what is IN FLIGHT. The tools that have already
+            # been fetched are retained for the whole call, so total retained
+            # memory still scales with the catalogue; it is the concurrent peak
+            # that is capped. Results stay positionally aligned with `connections`
+            # below, so the zip() that follows is unaffected.
             logger.debug(
                 f"Discovering tools from {len(connections)} MCP servers "
-                f"(max {self.config.MCP_DISCOVERY_CONCURRENCY} concurrent)"
+                f"(max {self.config.MCP_DISCOVERY_CONCURRENCY} concurrent, process-wide)"
             )
             results = await asyncio.gather(
-                *[_bounded_get_tools(slug) for slug in connections], return_exceptions=True
+                *[self._get_tools_with_retry(client, slug) for slug in connections], return_exceptions=True
             )
 
             # Process results - filter out exceptions and log failures
--- a/packages/orchestrator-agent/app/core/discovery.py
+++ b/packages/orchestrator-agent/app/core/discovery.py
@@ -430,10 +430,27 @@ class ToolDiscoveryService:
                 tool_interceptors=[_console_attribution_interceptor],
             )
 
-            # Gather tools from all servers with retry logic
-            # Use asyncio.gather with return_exceptions=True to handle partial failures gracefully
+            # Gather tools from all servers with retry logic.
+            # Use asyncio.gather with return_exceptions=True to handle partial failures gracefully.
+            #
+            # Bounded by a semaphore: an unbounded fan-out keeps every server's session,
+            # response body, parsed schema and tool objects alive at once, so the memory
+            # peak scales with the number of servers on the gateway rather than with
+            # anything we control. On prod that reached ~31 concurrent fetches and
+            # OOMKilled the pod (2026-08-15). Results stay positionally aligned with
+            # `connections` below, so the zip() that follows is unaffected.
+            discovery_semaphore = asyncio.Semaphore(self.config.MCP_DISCOVERY_CONCURRENCY)
+
+            async def _bounded_get_tools(slug: str) -> list:
+                async with discovery_semaphore:
+                    return await self._get_tools_with_retry(client, slug)
+
+            logger.debug(
+                f"Discovering tools from {len(connections)} MCP servers "
+                f"(max {self.config.MCP_DISCOVERY_CONCURRENCY} concurrent)"
+            )
             results = await asyncio.gather(
-                *[self._get_tools_with_retry(client, slug) for slug in connections], return_exceptions=True
+                *[_bounded_get_tools(slug) for slug in connections], return_exceptions=True
             )
 
             # Process results - filter out exceptions and log failures
--- a/packages/orchestrator-agent/app/models/config.py
+++ b/packages/orchestrator-agent/app/models/config.py
@@ -346,6 +346,15 @@ class AgentSettings:
     MCP_GATEWAY_URL = os.getenv("MCP_GATEWAY_URL", "https://alloych.gatana.ai/mcp")
     MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
 
+    # How many MCP servers may be queried for their tool catalogue at the same time
+    # during a cold discovery.  Discovery opens one connection per server, so an
+    # unbounded fan-out holds every server's session, response body, parsed schema
+    # and resulting tool objects in memory simultaneously — with ~31 servers on prod
+    # that spike OOMKilled the pod twice on 2026-08-15, each time within seconds of a
+    # single request.  Bounding it trades a little cold-start latency for a peak that
+    # no longer scales with the size of the gateway catalogue.
+    MCP_DISCOVERY_CONCURRENCY = max(1, int(os.getenv("MCP_DISCOVERY_CONCURRENCY", "5")))
+
     # Gatana compression: slug of the MCP server that provides compression utilities.
     # When tools from compression-enabled servers are in use, all tools from this
     # server are automatically included so agents can access compressed outputs.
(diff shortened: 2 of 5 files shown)


alloy-cicd (commented):

The fix is sound in its core mechanism — verified against `langchain_mcp_adapters`: `get_tools(server_name=...)` opens a fresh session per call, so the semaphore genuinely bounds concurrent sessions and response bodies. The findings below are about the *scope* of the bound, the config parsing, and test strength.

**Findings (most severe first):**

1. **Semaphore is per-call, not per-process** — concurrent cold discoveries multiply the bound (discovery.py:442)
2. **`MCP_DISCOVERY_CONCURRENCY` parsing bypasses `_int_env`** — a malformed env value crash-loops the pod at import (config.py:356)
3. **Semaphore slot held across retry backoff sleeps** — flaky servers block healthy ones (discovery.py:444)
4. **New knob is not declared in `.env.template` or the example k8s manifest** (config.py:356)
5. **Comment overstates what the semaphore bounds** — retained tool/schema memory still scales with catalogue size (discovery.py:436)
6. **Test pins only the upper bound** — full serialization would still pass (test_discovery.py:297)

Minor notes not worth an inline thread:
- The module-level `import asyncio` added at test_discovery.py:3 makes the function-local `import asyncio` inside `test_concurrent_discovery` (line 350) redundant.
- `test_concurrent_discovery` (line 328) does not set `MCP_DISCOVERY_CONCURRENCY` on its `Mock(spec=AgentSettings)`; it passes today only because unpatched `fetch_available_servers` bails out before the semaphore. If that path ever reaches the semaphore, `asyncio.Semaphore(<Mock>)` raises and the broad `except Exception: return []` at discovery.py:507 swallows it — the `isinstance(tools, list)` assertion can't tell success from a crash. Worth setting the attribute there too.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -430,10 +430,27 @@ async def discover_tools(
...
+            # Gather tools from all servers with retry logic.
+            # Use asyncio.gather with return_exceptions=True to handle partial failures gracefully.
+            #
+            # Bounded by a semaphore: an unbounded fan-out keeps every server's session,
+            # response body, parsed schema and tool objects alive at once, so the memory
+            # peak scales with the number of servers on the gateway rather than with
+            # anything we control. On prod that reached ~31 concurrent fetches and
+            # OOMKilled the pod (2026-08-15). Results stay positionally aligned with
+            # `connections` below, so the zip() that follows is unaffected.
+            discovery_semaphore = asyncio.Semaphore(self.config.MCP_DISCOVERY_CONCURRENCY)
```
alloy-cicd: **The bound is per-invocation, not per-process.** The semaphore is constructed inside `discover_tools()`, and this method runs once per discovery-cache miss with a per-user cache key (`executor.py:282-298`). N users cold-starting simultaneously (deploy rollout, cache TTL expiry, restart after the previous OOM) get N independent `Semaphore(5)`s → `5*N` concurrent fetches. Seven simultaneous cold discoveries already exceed the ~31 concurrent fetches that OOMKilled the pod, so the stated invariant — a memory peak that no longer scales with anything uncontrolled — holds only for a single in-flight discovery.

A module-level semaphore (or one owned by `ToolDiscoveryService`) would cap the process. Note the event-loop-binding trap with module-level `asyncio.Semaphore` that `_get_export_semaphore()` in `console_backend/catalog/adapters/google_drive.py:117-124` already works around lazily — worth reusing that pattern. (A service-level one created lazily on first use works too, since there's one `ToolDiscoveryService` per app, `app/core/agent.py:172`.)

packages/orchestrator-agent/app/models/config.py
```diff
@@ -346,6 +346,15 @@ class AgentSettings:
...
     MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
 
+    # How many MCP servers may be queried for their tool catalogue at the same time
+    # during a cold discovery.  Discovery opens one connection per server, so an
+    # unbounded fan-out holds every server's session, response body, parsed schema
+    # and resulting tool objects in memory simultaneously — with ~31 servers on prod
+    # that spike OOMKilled the pod twice on 2026-08-15, each time within seconds of a
+    # single request.  Bounding it trades a little cold-start latency for a peak that
+    # no longer scales with the size of the gateway catalogue.
+    MCP_DISCOVERY_CONCURRENCY = max(1, int(os.getenv("MCP_DISCOVERY_CONCURRENCY", "5")))
```
alloy-cicd: **Use `_int_env` here** — `int(os.getenv(...))` runs in the class body at import time, so `MCP_DISCOVERY_CONCURRENCY=""` (a common artifact of a templated-but-empty Helm value), `"5 "` with stray whitespace, or `"auto"` raises `ValueError` while importing `app.models.config` and crash-loops every pod before logging exists — a worse outage mode than the OOM this PR fixes, on the exact knob ops will be told to tune during an incident.

This file already defines `_int_env(name, default)` at line 23 for precisely this reason (its docstring: *"A misconfigured value … must not crash the process at import time"*), and `AGENT_DISCOVERY_CACHE_TTL` at line 326 uses it:

```python
MCP_DISCOVERY_CONCURRENCY = max(1, _int_env("MCP_DISCOVERY_CONCURRENCY", 5))
```

Separately: the `max(1, ...)` clamp silently rewrites `0`/negatives — an operator setting `0` intending "unlimited" silently gets fully-serialized discovery with no warning. A log line (or treating `<1` as the default) would make that visible.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -430,10 +430,27 @@ async def discover_tools(
...
+
+            async def _bounded_get_tools(slug: str) -> list:
+                async with discovery_semaphore:
+                    return await self._get_tools_with_retry(client, slug)
```
alloy-cicd: **The semaphore slot is held across the entire retry loop**, including the exponential-backoff `await asyncio.sleep(delay)` at line 322 (1s + 2s over 3 attempts). A server returning 502/503 occupies one of only 5 slots while doing nothing — each attempt opens a fresh session anyway (verified in `langchain_mcp_adapters`), so nothing is kept warm across the sleep. With a few flaky servers out of ~31, healthy servers queue behind pure sleep time and cold-discovery latency degrades from ~max(server latency) toward ceil(31/5) serialized waves — on a user-facing cache-miss path (`executor.py:290`).

Acquiring the semaphore per attempt (e.g. passing it into `_get_tools_with_retry` and wrapping only the `client.get_tools(...)` call) keeps the same in-flight bound without the idle slots.

packages/orchestrator-agent/app/models/config.py
```diff
@@ -346,6 +346,15 @@ class AgentSettings:
     MCP_GATEWAY_URL = os.getenv("MCP_GATEWAY_URL", "https://alloych.gatana.ai/mcp")
     MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
 
+    # How many MCP servers may be queried for their tool catalogue at the same time
+    # during a cold discovery.  Discovery opens one connection per server, so an
+    # unbounded fan-out holds every server's session, response body, parsed schema
+    # and resulting tool objects in memory simultaneously — with ~31 servers on prod
+    # that spike OOMKilled the pod twice on 2026-08-15, each time within seconds of a
+    # single request.  Bounding it trades a little cold-start latency for a peak that
+    # no longer scales with the size of the gateway catalogue.
```
alloy-cicd, line 355: **The knob can't be turned where it matters**: `MCP_DISCOVERY_CONCURRENCY` is not declared in `packages/orchestrator-agent/.env.template` (which lists the sibling `MCP_GATEWAY_URL`/`MCP_GATEWAY_CLIENT_ID` at lines 65-66) nor in `example-k8s-deployment/base/orchestrator.yaml` (env block ~line 62). During the next memory incident, an on-call engineer grepping the template/manifest for a tunable will conclude the concurrency is hard-coded — the mitigation lever added by this PR is invisible exactly when it's needed.

Also a nit while here: this incident narrative now appears three times nearly verbatim (here, discovery.py:436-441, and the test docstring). Keeping the full story once next to the knob and one-line cross-references elsewhere avoids the three copies drifting as the mitigation evolves.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -430,10 +430,27 @@ async def discover_tools(
...
-            # Use asyncio.gather with return_exceptions=True to handle partial failures gracefully
+            # Gather tools from all servers with retry logic.
+            # Use asyncio.gather with return_exceptions=True to handle partial failures gracefully.
+            #
+            # Bounded by a semaphore: an unbounded fan-out keeps every server's session,
+            # response body, parsed schema and tool objects alive at once, so the memory
+            # peak scales with the number of servers on the gateway rather than with
+            # anything we control. On prod that reached ~31 concurrent fetches and
+            # OOMKilled the pod (2026-08-15). Results stay positionally aligned with
+            # `connections` below, so the zip() that follows is unaffected.
```
alloy-cicd: **The comment overstates what the semaphore bounds.** It genuinely bounds concurrent sessions and response bodies (verified: `get_tools` opens a session per call). But "parsed schema and tool objects alive at once" is not bounded — `asyncio.gather` still creates all N coroutines up front, and every successful result is retained in `results` and then `tools` until the function returns, so retained tool/schema memory scales with catalogue size regardless of the limit. If the gateway grows to ~100 servers, someone trusting this comment will keep concurrency at 5 and the pod can OOM again on retained objects with the mitigation apparently in place. Scoping the claim to *in-flight sessions and response bodies* keeps the next investigation honest.

packages/orchestrator-agent/tests/test_discovery.py
```diff
@@ -239,6 +242,62 @@ async def test_discover_tools_with_white_list(self):
...
+
+        with patch("app.core.discovery.MultiServerMCPClient") as mock_client:
+            mock_client_instance = Mock()
+            mock_client_instance.get_tools = AsyncMock(side_effect=fake_get_tools)
+            mock_client.return_value = mock_client_instance
+            service.fetch_available_servers = AsyncMock(return_value=servers)
+
+            result = await service.discover_tools("test_token")
+
+        assert max_in_flight <= 3, f"expected at most 3 concurrent fetches, saw {max_in_flight}"
```
alloy-cicd: **Only the upper bound is pinned.** `max_in_flight <= 3` also passes under full serialization (`max_in_flight == 1`) — e.g. someone "simplifying" the gather into a sequential loop, or a semaphore of 1 — while cold-start latency multiplies by ~20x and no test fails. With 20 fakes each yielding at `asyncio.sleep(0)` and a limit of 3, the bound is deterministically reached, so this can be tightened to:

```python
assert max_in_flight == 3, ...
```

which pins both over-fetching and over-throttling. (A related soft spot: because the fake spans exactly one event-loop tick, a refactor that wraps only a cheap prelude in the semaphore rather than the actual fetch could still pass; having the fake await an `asyncio.Event` released after a few tasks are in flight would pin the bound harder, but `== 3` is the cheap high-value fix.)

aartaria:
## Triage of the 🤖 automatic review

All 6 inline findings plus the 2 summary notes addressed. Pushed as a second round.

**Fixed**

- **(1) Per-call semaphore scope** — the most important finding, and correct. Discovery runs per user on cache miss, so a per-invocation bound still allowed limit×N concurrent fetches; with a limit of 5, seven simultaneous cold users already exceed the ~31 that OOMKilled the pod. The semaphore is now **module-level**, built lazily so the configured limit is read at first use rather than at import. The cost — unrelated users' cold discoveries queue behind each other — is the intended trade and is documented.
- **(2) Import-time env parsing** — now uses the file's own `_int_env` helper, so an empty or malformed value falls back with a warning instead of crash-looping the pod.
- **(3) Slot held across retry backoff** — the semaphore is now acquired around the fetch itself rather than the whole retry loop, so a flaky server no longer occupies a slot through up to 3s of sleep.
- **(4) Knob undeclared** — `MCP_DISCOVERY_CONCURRENCY` added to `.env.template` and the example k8s manifest, with a note on which way to turn it.
- **(5) Comment overclaims** — corrected: only the in-flight peak is bounded; tools already fetched are retained for the whole call, so total retained memory still scales with the catalogue.
- **(6) Test pins only the upper bound** — now asserts `== 3`, and resets the module-level semaphore so a limit cached by an earlier test cannot leak in.
- **Summary note: redundant local `import asyncio`** — gone with the removal of the per-call wrapper.

**Note on the remaining summary item**

`test_concurrent_discovery`'s stale `Mock(spec=AgentSettings)` does only pass by bailing out early. Left as-is: it isn't testing the discovery fan-out, and giving it a full config would change what that test exercises. Worth a separate tidy-up rather than a drive-by here.

Thanks — finding (1) meant the original fix would not actually have capped the process.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (6318610b86858725d7a3184028925084fb60f12a..704088c481d0c214cd09b746dc202e6a6da29e5a)

1. **Per-call semaphore scope** — ✅ verified-fixed. The semaphore is now module-level, built lazily in `_discovery_semaphore()`, so the bound is process-wide. The docstring records the trade-off explicitly.
2. **Import-time `int(os.getenv(...))`** — ✅ verified-fixed. Now `max(1, _int_env("MCP_DISCOVERY_CONCURRENCY", 5))`; a malformed value warns and falls back instead of crash-looping.
3. **Slot held across retry backoff** — ✅ verified-fixed. The slot is now acquired per attempt inside `_get_tools_with_retry`, wrapping only the `client.get_tools(...)` call; backoff sleeps no longer occupy a slot.
4. **Knob undeclared in `.env.template` / example manifest** — ✅ verified-fixed. Declared in both, with accurate comments.
5. **Comment overclaimed what the semaphore bounds** — ✅ verified-fixed. The comment now scopes the claim to the in-flight peak and explicitly states retained memory still scales with the catalogue.
6. **Test pinned only the upper bound** — ✅ verified-fixed. `assert max_in_flight == 3`, and the test resets the module-level semaphore so an earlier limit can't leak in.

**Delta review: no new blocking issues.** All 14 discovery tests pass at the new head. I checked the two risks the new design could introduce: (a) a module-level `asyncio.Semaphore` binds to the first event loop that *blocks* on it — in this process discovery runs on the single server loop (the one `asyncio.run` in `dynamic_tool_dispatch.py` never reaches the discovery path), and the pattern matches the existing `_get_export_semaphore()` precedent in `console_backend/catalog/adapters/google_drive.py`; (b) the rebuild-on-limit-change check could double the bound if the limit changed mid-flight, but the limit is static per process, so this only serves tests.

Non-blocking follow-up notes (not inline threads):
- Only `test_discover_tools_bounds_concurrent_server_fetches` resets the module-level semaphore globals. If a future discovery test blocks on a semaphore cached from an earlier test's (now-closed) event loop, the resulting `RuntimeError` would be swallowed by the broad `except` into an empty tool list and fail confusingly. An autouse fixture resetting `_DISCOVERY_SEMAPHORE`/`_DISCOVERY_SEMAPHORE_LIMIT` for the discovery test module would remove the ordering sensitivity.
- Two round-1 summary nits remain open (both fine to leave): the function-local `import asyncio` in `test_concurrent_discovery` is redundant now that the module imports it, and that test's `Mock(spec=AgentSettings)` still lacks `MCP_DISCOVERY_CONCURRENCY` (it passes only because unpatched `fetch_available_servers` bails out before the semaphore).

Every previous finding is resolved and the delta introduces nothing new — no inline comments this round. Submitted as COMMENT; the human reviewer decides.

PR #153, opened 2026-08-17
fix(orchestrator): guard inbound MCP SSE events against unbounded parsing

### LINKED ISSUE #152: orchestrator: unbounded MCP SSE event parsing lets one fat server OOM the pod — add size guard + per-server size logging
## Summary

The orchestrator parses every inbound MCP SSE event with `JSONRPCMessage.model_validate_json` with **no size bound**. A single oversized event is amplified ~7× into pydantic objects, several parses run concurrently, and the process OOMs. Nothing names the offending server, so the failure surfaces as an unexplained pod kill instead of a config error.

This is the root cause of the prod OOMKills of 2026-08-15/17: one gateway server answered `tools/list` with a **21.9 MB** single SSE event. Measured on a local stack (same gateway):

| | with the fat server | without it |
|---|---|---|
| cold-cache turn, RSS peak | **2348 MB** (heavy task: 4159 MB) | **456 MB** |
| top allocation site | 663 MB / 10M blocks at `mcp/client/streamable_http.py:217` | tracemalloc never crossed 250 MB |

Attribution chain (each link measured): fat `tools/list` → whole-event string buffering in `httpx_sse` (2 copies) → unbounded pydantic parse → concurrent parses → 2.3–4.2 GB transient peak → OOMKill. Disabling the server at the gateway removed ~80% of the peak — this issue is about making that protection automatic instead of forensic.

## What to build

1. **Size guard before the parse.** Cap the size of a single MCP SSE event (env `MCP_SSE_MAX_EVENT_BYTES`, sensible default ~10 MB). An oversized event is rejected **before** `model_validate_json`, with an ERROR log naming the server slug and size. Discovery already degrades gracefully per server (`_get_tools_with_retry` catches and logs per-slug failures), so a rejected catalogue costs one server's tools, not the pod.
2. **Permanent per-server size observability.** Log catalogue/event sizes per server slug at discovery (WARNING above ~1 MB), so catalogue growth is a visible trend instead of a silent time bomb. This is what turned "some server is fat" into a name in minutes once instrumented.

The MCP SDK offers no hook at the parse site, so the guard wraps `StreamableHTTPTransport._handle_sse_event` process-wide — covering discovery, tool calls, and the code-mode/quickjs sessions in one place.

## Explicitly out of scope (follow-ups)

- **Lazy catalogue loading** (precedent: GP sub-agent, #120) — reduces aggregate parse cost and cold-start latency, but does not prevent a single fat payload from OOMing the pod; it only reschedules the parse. Separate issue.
- Gateway-side budgets (defense in depth, different repo).
- `agent-runner` (separate process; same guard applies, separate change).
- Upstream: `mcp`'s client could bound or stream-parse SSE events; worth an upstream issue.

## Why a cap is the right fix

A 5–10 MB cap would have prevented every OOM in this incident, converting each into a logged, attributable per-server failure. Trimming what is *cached* does not help (the transient parse **is** the spike, and trimming happens after parsing); lazy loading defers the same parse.

aartaria:
closes ringier-data/nannos#152

## Problem

Every MCP message the orchestrator receives — tool catalogues at discovery, tool results mid-turn, code-mode sessions — arrives as an SSE event the SDK parses with `JSONRPCMessage.model_validate_json` and **no size bound**. The parse amplifies wire size ~7× into live pydantic objects, and several parses run concurrently.

Measured consequence (local stack, same gateway as prod): one server answering `tools/list` with a **21.9 MB** event drove a cold-cache turn from a 56 MB baseline to a **2348 MB** peak (**4159 MB** on a heavy task); top allocation site was 663 MB in 10M blocks at `mcp/client/streamable_http.py:217`. This is what OOMKilled prod on 2026-08-15/17. Nothing named the server, so attribution took a tracemalloc hunt; disabling the one server dropped the same turn's peak to **456 MB**.

## Change

`app/core/mcp_guard.py`, installed at lifespan startup, wraps `StreamableHTTPTransport._handle_sse_event` process-wide (the SDK offers no hook at the parse site; one wrap covers discovery, tool dispatch and code-mode):

- **Reject before the parse**: events over `MCP_SSE_MAX_EVENT_BYTES` (default 10 MB) raise `McpEventTooLargeError` naming the server slug and sizes. Discovery already degrades gracefully per server, so a rejected catalogue costs one server's tools, not the pod; a rejected tool result surfaces as that call's error.
- **Warn with attribution**: events over `MCP_SSE_WARN_EVENT_BYTES` (default 1 MB) log the server slug, making catalogue growth a visible trend. Yesterday this line alone would have named the culprit weeks earlier.

Knobs declared in `.env.template` and the example k8s manifest.

## Deliberately out of scope

- **Lazy catalogue loading** — reduces aggregate parse cost, but only *reschedules* the fatal parse; a cap prevents it. Follow-up per #152.
- **Cache trimming** — the transient parse *is* the spike; trimming happens after parsing and fixes nothing.
- `agent-runner` (separate process), gateway-side budgets, and an upstream `mcp` issue for bounded/streaming SSE parsing.

## Verification

- 7 new unit tests pin the contract: pass-through under cap, warn-with-slug, **reject-before-parse** (asserts the original handler never sees the event), empty priming events unaffected, non-gateway URL slug fallback, idempotent install.
- Full orchestrator suite: **633 passed**.
- `ruff check` clean on the new files; `main.py`/`config.py` error counts identical to baseline (32/1 — pre-existing).

Refs the production incident in our internal tracker.

### DIFF
--- a/packages/orchestrator-agent/app/core/mcp_guard.py
+++ b/packages/orchestrator-agent/app/core/mcp_guard.py
@@ -1,0 +1,118 @@
+"""Size guard for inbound MCP SSE events.
+
+Every MCP message the orchestrator receives — tool catalogues at discovery,
+tool results mid-turn, code-mode sessions — arrives as a Server-Sent Event that
+the MCP SDK parses with ``JSONRPCMessage.model_validate_json`` and NO size
+bound (``mcp/client/streamable_http.py``). Parsing amplifies the wire size
+roughly 7x into live pydantic objects, and several parses run concurrently, so
+a single oversized payload can OOM the pod.
+
+This is not hypothetical: in prod, one gateway server answered ``tools/list``
+with a 21.9 MB single event; cold-cache turns peaked at 2.3-4.2 GB RSS and the
+pod was OOMKilled repeatedly (2026-08-15/17, see
+alloy-ch/rcplus-alloy-infrastructure-agents#241). Nothing named the offending
+server, so the failure surfaced as an unexplained pod kill instead of a config
+error.
+
+The guard wraps ``StreamableHTTPTransport._handle_sse_event`` process-wide —
+the SDK offers no hook at the parse site — and does two things:
+
+* **rejects** any event over ``MCP_SSE_MAX_EVENT_BYTES`` *before* it is
+  parsed, raising :class:`McpEventTooLargeError` that names the server slug
+  and size. Discovery already degrades gracefully per server
+  (``_get_tools_with_retry`` catches and logs per-slug failures), so a
+  rejected catalogue costs one server's tools, not the process. A rejected
+  tool result surfaces as that tool call's error.
+* **logs** every event over ``MCP_SSE_WARN_EVENT_BYTES`` with the server slug,
+  so catalogue growth is a visible trend long before it reaches the cap.
+
+Sizes are measured on the raw SSE ``data`` string, before any parsing.
+"""
+
+from __future__ import annotations
+
+import logging
+from urllib.parse import parse_qs, urlparse
+
+logger = logging.getLogger(__name__)
+
+_installed = False
+
+
+class McpEventTooLargeError(RuntimeError):
+    """An inbound MCP SSE event exceeded the configured size cap."""
+
+    def __init__(self, server: str, size_bytes: int, max_bytes: int) -> None:
+        self.server = server
+        self.size_bytes = size_bytes
+        self.max_bytes = max_bytes
+        super().__init__(
+            f"MCP SSE event from server '{server}' is {size_bytes / 2**20:.1f}MB, "
+            f"exceeding the {max_bytes / 2**20:.1f}MB cap (MCP_SSE_MAX_EVENT_BYTES). "
+            f"Refusing to parse it — an unbounded parse of this payload is what "
+            f"OOMKilled the orchestrator (see ringier-data/nannos#152). "
+            f"Trim this server's catalogue/results at the gateway, or raise the cap "
+            f"only with a matching memory-limit review."
+        )
+
+
+def _server_slug(url: str) -> str:
+    """Best-effort server identity for logs: the gateway slug, else the host.
+
+    Gateway connections carry ``includeOnlyServerSlugs=<slug>``; other MCP
+    endpoints (e.g. the console backend) are identified by their netloc.
+    """
+    try:
+        parsed = urlparse(url)
+        slugs = parse_qs(parsed.query).get("includeOnlyServerSlugs")
+        return slugs[0] if slugs else (parsed.netloc or url)
+    except Exception:
+        return url
+
+
+def install_mcp_size_guard(max_event_bytes: int, warn_event_bytes: int) -> None:
+    """Wrap the MCP transport's SSE handler with the size guard. Idempotent.
+
+    Must be called once at startup, before any MCP session is opened. Patching
+    the class method covers every ``StreamableHTTPTransport`` in the process —
+    discovery, tool dispatch and code-mode sessions alike — which is the point:
+    the failure mode does not care which path the fat payload arrives on.
+    """
+    global _installed
+    if _installed:
+        return
+
+    from mcp.client.streamable_http import StreamableHTTPTransport
+
+    original = StreamableHTTPTransport._handle_sse_event
+
+    async def _guarded_handle_sse_event(self, sse, *args, **kwargs):
+        data = getattr(sse, "data", None)
+        if data:
+            size = len(data)
+            if size > max_event_bytes:
+                server = _server_slug(self.url)
+                logger.error(
+                    "[MCP-GUARD] rejecting %.1fMB SSE event from server=%s (cap %.1fMB)",
+                    size / 2**20,
+                    server,
+                    max_event_bytes / 2**20,
+                )
+                raise McpEventTooLargeError(server, size, max_event_bytes)
+            if size > warn_event_bytes:
+                logger.warning(
+                    "[MCP-GUARD] large SSE event: %.1fMB from server=%s "
+                    "(cap %.1fMB — growth here eventually becomes an outage)",
+                    size / 2**20,
+                    _server_slug(self.url),
+                    max_event_bytes / 2**20,
+                )
+        return await original(self, sse, *args, **kwargs)
+
+    StreamableHTTPTransport._handle_sse_event = _guarded_handle_sse_event
+    _installed = True
+    logger.info(
+        "MCP SSE size guard installed (cap=%.1fMB, warn=%.1fMB)",
+        max_event_bytes / 2**20,
+        warn_event_bytes / 2**20,
+    )
--- a/packages/orchestrator-agent/app/core/mcp_guard.py
+++ b/packages/orchestrator-agent/app/core/mcp_guard.py
@@ -7,12 +7,11 @@ tool results mid-turn, code-mode sessions — is parsed by the MCP SDK with
 7x into live pydantic objects, and several parses run concurrently, so a
 single oversized payload can OOM the pod.
 
-This is not hypothetical: in prod, one gateway server answered ``tools/list``
-with a 21.9 MB single event; cold-cache turns peaked at 2.3-4.2 GB RSS and the
-pod was OOMKilled repeatedly (2026-08-15/17, see
-alloy-ch/rcplus-alloy-infrastructure-agents#241). Nothing named the offending
-server, so the failure surfaced as an unexplained pod kill instead of a config
-error.
+This is not hypothetical: one gateway server answered ``tools/list`` with a
+21.9 MB single event; cold-cache turns peaked at 2.3-4.2 GB RSS and the pod was
+OOMKilled repeatedly. Nothing named the offending server, so the failure
+surfaced as an unexplained pod kill instead of a config error. The full
+write-up is in ringier-data/nannos#152.
 
 The guard wraps the transport's two inbound parse paths process-wide — the SDK
 offers no hook at the parse site — and does two things:
--- a/packages/orchestrator-agent/app/core/mcp_guard.py
+++ b/packages/orchestrator-agent/app/core/mcp_guard.py
@@ -1,11 +1,11 @@
-"""Size guard for inbound MCP SSE events.
+"""Size guard for inbound MCP messages.
 
 Every MCP message the orchestrator receives — tool catalogues at discovery,
-tool results mid-turn, code-mode sessions — arrives as a Server-Sent Event that
-the MCP SDK parses with ``JSONRPCMessage.model_validate_json`` and NO size
-bound (``mcp/client/streamable_http.py``). Parsing amplifies the wire size
-roughly 7x into live pydantic objects, and several parses run concurrently, so
-a single oversized payload can OOM the pod.
+tool results mid-turn, code-mode sessions — is parsed by the MCP SDK with
+``JSONRPCMessage.model_validate_json`` and NO size bound
+(``mcp/client/streamable_http.py``). Parsing amplifies the wire size roughly
+7x into live pydantic objects, and several parses run concurrently, so a
+single oversized payload can OOM the pod.
 
 This is not hypothetical: in prod, one gateway server answered ``tools/list``
 with a 21.9 MB single event; cold-cache turns peaked at 2.3-4.2 GB RSS and the
@@ -14,19 +14,29 @@ alloy-ch/rcplus-alloy-infrastructure-agents#241). Nothing named the offending
 server, so the failure surfaced as an unexplained pod kill instead of a config
 error.
 
-The guard wraps ``StreamableHTTPTransport._handle_sse_event`` process-wide —
-the SDK offers no hook at the parse site — and does two things:
+The guard wraps the transport's two inbound parse paths process-wide — the SDK
+offers no hook at the parse site — and does two things:
 
-* **rejects** any event over ``MCP_SSE_MAX_EVENT_BYTES`` *before* it is
-  parsed, raising :class:`McpEventTooLargeError` that names the server slug
-  and size. Discovery already degrades gracefully per server
-  (``_get_tools_with_retry`` catches and logs per-slug failures), so a
-  rejected catalogue costs one server's tools, not the process. A rejected
-  tool result surfaces as that tool call's error.
-* **logs** every event over ``MCP_SSE_WARN_EVENT_BYTES`` with the server slug,
-  so catalogue growth is a visible trend long before it reaches the cap.
+* **rejects** any payload over ``MCP_SSE_MAX_EVENT_BYTES`` *before* it is
+  parsed. Rejection follows the SDK's own parse-failure contract: the error is
+  delivered through ``read_stream_writer`` so the pending request FAILS with
+  :class:`McpEventTooLargeError` naming the server slug and size. It must not
+  be raised out of the handler instead — every SDK call site swallows
+  exceptions (``except Exception: logger.debug``) and then *reconnects with
+  Last-Event-ID*, which would replay the same oversized payload and leave the
+  pending request hanging on a permanently-held semaphore slot. Discovery
+  already degrades gracefully per server (``_get_tools_with_retry``), so a
+  rejected catalogue costs one server's tools, not the process; a rejected
+  tool result surfaces as that call's error.
+* **logs** every payload over ``MCP_SSE_WARN_EVENT_BYTES`` with the server
+  slug, so catalogue growth is a visible trend long before it reaches the cap.
 
-Sizes are measured on the raw SSE ``data`` string, before any parsing.
+Sizes are measured in UTF-8 **bytes** (what the parser and the allocator see),
+not characters: SSE ``data`` is already str-decoded, and a multibyte-heavy
+catalogue can weigh up to 4x its character count. The exact encode is only
+paid when the cheap character-count bounds cannot decide.
+
+Setting a threshold to ``0`` disables that threshold.
 """
 
 from __future__ import annotations
@@ -36,18 +46,18 @@ from urllib.parse import parse_qs, urlparse
 
 logger = logging.getLogger(__name__)
 
-_installed = False
+_WRAP_MARKER = "_nannos_mcp_guard"
 
 
 class McpEventTooLargeError(RuntimeError):
-    """An inbound MCP SSE event exceeded the configured size cap."""
+    """An inbound MCP message exceeded the configured size cap."""
 
     def __init__(self, server: str, size_bytes: int, max_bytes: int) -> None:
         self.server = server
         self.size_bytes = size_bytes
         self.max_bytes = max_bytes
         super().__init__(
-            f"MCP SSE event from server '{server}' is {size_bytes / 2**20:.1f}MB, "
+            f"MCP message from server '{server}' is {size_bytes / 2**20:.1f}MB, "
             f"exceeding the {max_bytes / 2**20:.1f}MB cap (MCP_SSE_MAX_EVENT_BYTES). "
             f"Refusing to parse it — an unbounded parse of this payload is what "
             f"OOMKilled the orchestrator (see ringier-data/nannos#152). "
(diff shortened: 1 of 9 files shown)


packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,118 @@
...
+            size = len(data)
+            if size > max_event_bytes:
+                server = _server_slug(self.url)
+                logger.error(
+                    "[MCP-GUARD] rejecting %.1fMB SSE event from server=%s (cap %.1fMB)",
+                    size / 2**20,
+                    server,
+                    max_event_bytes / 2**20,
+                )
+                raise McpEventTooLargeError(server, size, max_event_bytes)
```
alloy-cicd: **[critical] The raise never reaches the awaiting request — it is swallowed by the SDK, and the reconnect path then replays the same oversized event.**

In `mcp` 1.27.1 (the version in `uv.lock`), every call site of `_handle_sse_event` catches broadly:

- `_handle_sse_response` (streamable_http.py:429): `except Exception as e: logger.debug(f"SSE stream ended: {e}")` — the exception is logged at DEBUG and dropped. Nothing is ever written to `read_stream_writer`, so the pending JSON-RPC request gets neither a response nor an error: `client.get_tools()` inside `_get_tools_with_retry` just hangs, holding an `MCP_DISCOVERY_CONCURRENCY` semaphore slot. And because no layer sets an MCP read timeout (discovery passes no `session_kwargs`, so `BaseSession` awaits with `timeout=None`), the hang is *permanent*: after `MCP_DISCOVERY_CONCURRENCY` (default 5) hits on the oversized server, the module-global discovery semaphore is exhausted and every subsequent discovery for every user blocks forever — the pod stays green to the liveness probe while serving nothing, strictly harder to diagnose than the OOMKill it replaces. The docstring's claim that "a rejected catalogue costs one server's tools" and "a rejected tool result surfaces as that tool call's error" does not hold in practice.
- Worse: `last_event_id` is recorded from `sse.id` **before** the handler is called, so after the swallowed raise `_handle_sse_response` falls through to `_handle_reconnection(ctx, last_event_id, ...)` — the transport reconnects with `Last-Event-ID` pointing *before* the fat event and the server replays the same multi-MB payload, which is rejected again, recursing through the reconnect attempts. The guard converts one memory spike into several re-downloads plus a hang, instead of a fail-fast.

The SDK's own parse-failure path in the same method shows the intended error-delivery contract (streamable_http.py:241): `await read_stream_writer.send(exc)` — the session then fails the pending request promptly.

Suggested fix: don't raise. The wrapper already receives the writer as the second positional arg:

```python
async def _guarded_handle_sse_event(self, sse, read_stream_writer, *args, **kwargs):
    ...
    if size > max_event_bytes:
        await read_stream_writer.send(McpEventTooLargeError(server, size, max_event_bytes))
        return True  # complete: stops the SSE loop cleanly, no reconnect replay
    ...
    return await original(self, sse, read_stream_writer, *args, **kwargs)
```

Returning `True` (rather than the SDK parse path's `False`) also closes the response and skips the `last_event_id` reconnect, avoiding the replay loop. Verifiable later by: a test driving the real `_handle_sse_response` with an oversized event and asserting the error object arrives on the read stream and no reconnect is attempted. (The GET stream for server-initiated messages has the mirror problem today: two oversized pushed events silently kill server-push for the session.) Independent of the fix here, consider passing `session_kwargs={'read_timeout_seconds': timedelta(...)}` in discovery so no future transport bug can wedge the semaphore permanently.

packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,118 @@
...
+        return
+
+    from mcp.client.streamable_http import StreamableHTTPTransport
+
+    original = StreamableHTTPTransport._handle_sse_event
+
+    async def _guarded_handle_sse_event(self, sse, *args, **kwargs):
+        data = getattr(sse, "data", None)
+        if data:
+            size = len(data)
```
alloy-cicd: **[major] `len(data)` counts str characters, not bytes — the cap under-measures UTF-8 payloads by up to 4×.**

`sse.data` is already decoded by httpx-sse, so `len(data)` is Unicode code points. The knobs are named `*_BYTES`, and the log/exception report "MB". A catalogue heavy in non-ASCII (German umlauts, CJK, emoji in tool descriptions) that is e.g. 18 MB on the wire can measure ~9.5 M "bytes" here, sail under the 10 MiB cap, and hit the exact unbounded parse this guard exists to prevent — while logging a reassuringly small size.

Use `size = len(data.encode())` (compute once, reuse for both thresholds). The UTF-8 encode of a multi-MB string is cheap relative to the parse being guarded, and it makes the measured number match the wire/memory reality the env-var docs describe.

alloy-cicd (commented):

The problem analysis and the intent of this PR are solid, and the incident write-up in the module docstring is genuinely useful. But the central mechanism doesn't deliver the promised contract: **raising out of `_handle_sse_event` is swallowed by every SDK call site** (`except Exception → logger.debug`), so a rejected event produces a hung request plus a `Last-Event-ID` reconnect loop that re-downloads the same oversized payload — not the clean per-server/per-call error the docstring and PR description claim. The tests don't catch this because they exercise a stub transport whose caller is the test itself. Details and a concrete fix (deliver the error via `read_stream_writer`, return `True`) are in the inline comments.

Findings (7 inline):
1. **critical** — guard's raise is swallowed by the SDK; request hangs and the oversized event is replayed on reconnect (`mcp_guard.py:101`)
2. **major** — `len(data)` measures characters, not bytes; UTF-8-heavy payloads under-measured up to 4× (`mcp_guard.py:92`)
3. **major** — `max(1, ...)` makes `0` a reject-everything cap (no way to disable); warn/max relationship unvalidated (`config.py:353`)
4. **major** — tests never exercise the real SDK call site, so the headline contract is unverified (`test_mcp_guard.py:52`)
5. **medium** — `application/json` response path runs the same unbounded parse, unguarded (`mcp_guard.py:73`)
6. **minor** — guard fires on non-`message` events the SDK never parses (`mcp_guard.py:90`)
7. **minor** — `_installed` global vs. marking the wrapper; no CI defence against an `mcp` bump moving/renaming the private method (`mcp_guard.py:39`)

Nits not worth their own threads: the reject/warn branches duplicate `_server_slug` + MB formatting (compute once, then `if/elif`); the `logger.error` before the raise duplicates what the exception message already carries and is re-logged by discovery's handlers (2–3 near-identical ERROR lines per event); and since the process-wide monkeypatch of a private SDK method is invisible from any call site, it belongs in `packages/orchestrator-agent/AGENTS.md` under Critical Design Decisions per that file's own "Maintaining These Instructions" rule.

Scope observation, not a defect in this diff: the PR deliberately scopes out `agent-runner`, but `voice-agent` (raw `streamablehttp_client` + `ClientSession`) consumes the same gateway catalogue in its own process and isn't mentioned — if the fat server pushes it over, that pod OOMs identically while this fix is believed shipped. Worth a line in the follow-up issue.

Submitted as COMMENT; verified against `mcp` 1.27.1 as pinned in `uv.lock`. The human reviewer decides.

packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,118 @@
...
+    global _installed
+    if _installed:
+        return
+
+    from mcp.client.streamable_http import StreamableHTTPTransport
+
+    original = StreamableHTTPTransport._handle_sse_event
+
+    async def _guarded_handle_sse_event(self, sse, *args, **kwargs):
+        data = getattr(sse, "data", None)
```
alloy-cicd: **[minor] Guard fires on non-`message` events the SDK never parses.**

The original `_handle_sse_event` parses only `sse.event == "message"`; anything else is logged (`Unknown SSE event`) and ignored with `return False`. The wrapper sizes-and-rejects *all* events with data, so a large non-message event (vendor keep-alive/progress) that the SDK would have skipped now tears down the stream — with the swallow-and-reconnect consequences noted above — for a payload that was never going to be parsed.

Gate the check: `if data and getattr(sse, "event", None) == "message":`.

packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    try:
+        parsed = urlparse(url)
+        slugs = parse_qs(parsed.query).get("includeOnlyServerSlugs")
+        return slugs[0] if slugs else (parsed.netloc or url)
+    except Exception:
+        return url
+
+
+def install_mcp_size_guard(max_event_bytes: int, warn_event_bytes: int) -> None:
```
alloy-cicd, line 194: **[medium] The `application/json` response path is unguarded — the module docstring overclaims coverage.**

Streamable HTTP servers may answer a POST with `Content-Type: application/json` instead of an SSE stream. The SDK then takes `_handle_json_response` (streamable_http.py:376): `content = await response.aread()` + `JSONRPCMessage.model_validate_json(content)` — the same unbounded parse, not covered by this wrap. If the gateway (or one upstream server) is ever reconfigured to reply with plain JSON, the 22 MB `tools/list` parses unbounded again and the guard logs nothing, while the team believes this failure class is fixed.

Either wrap `_handle_json_response` with the same cap (it's a two-line addition to the same install function), or state the gap explicitly in the docstring so the next incident isn't diagnosed from scratch.

Related sizing note: by the time this wrapper sees `sse.data`, httpx-sse has already materialised the full decoded event in memory, so the guard prevents the ~7× parse amplification (the dominant term) but not the wire-size allocation itself. Worth softening "BEFORE parsing" comments in `.env.template`/k8s so memory-limit sizing isn't done on the assumption that oversized events never land in RAM.

packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,118 @@
...
+"""
+
+from __future__ import annotations
+
+import logging
+from urllib.parse import parse_qs, urlparse
+
+logger = logging.getLogger(__name__)
+
+_installed = False
```
alloy-cicd: **[minor] Module-global `_installed` tracks state derivable from the patched object — and there's no defence against SDK drift.**

Two related points:

1. The flag can disagree with reality (exactly what the tests exploit by resetting `mcp_guard._installed`). Marking the wrapper itself is self-consistent and removes the need for tests to poke module internals:

```python
if getattr(StreamableHTTPTransport._handle_sse_event, "_mcp_guard", False):
    return
...
_guarded_handle_sse_event._mcp_guard = True
```

Also add `functools.wraps(original)` so the patched method keeps its name in tracebacks.

2. `mcp` is only a *transitive* dependency (via `langchain-mcp-adapters`), pinned solely by `uv.lock`, and `_handle_sse_event` is a private method. On a routine lock upgrade this either crashes the lifespan at boot (`AttributeError` → prod crashloop) or — if the parse moves elsewhere — silently becomes a no-op while CI stays green (all tests use a stand-in class). One cheap test asserting the real `StreamableHTTPTransport._handle_sse_event` exists with the expected leading parameters (`self, sse, read_stream_writer`) turns both outcomes into a red build instead.

packages/orchestrator-agent/app/models/config.py
```diff
@@ -346,6 +346,13 @@ class AgentSettings:
     MCP_GATEWAY_URL = os.getenv("MCP_GATEWAY_URL", "https://alloych.gatana.ai/mcp")
     MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
 
+    # Inbound MCP SSE event size guard (see app/core/mcp_guard.py). Events over
+    # the cap are rejected BEFORE the pydantic parse whose ~7x amplification
+    # OOMKilled the pod (ringier-data/nannos#152); events over the warn level are
+    # logged with their server slug so catalogue growth is visible early.
+    MCP_SSE_MAX_EVENT_BYTES = max(1, _int_env("MCP_SSE_MAX_EVENT_BYTES", 10 * 1024 * 1024))
```
alloy-cicd: **[major] `max(1, ...)` turns the natural "disable" value into a reject-everything cap, and warn/max are never validated against each other.**

- `MCP_SSE_MAX_EVENT_BYTES=0` — the value an operator will reach for to switch the guard off during an incident — becomes a 1-byte cap that rejects every SSE event including the `initialize` response: a self-inflicted, process-wide MCP outage, with the guard's own error message telling them to raise the cap. The `max(1, ...)` idiom is right for `MCP_DISCOVERY_CONCURRENCY` below, wrong for a size cap. Suggested: keep the raw `_int_env(...)` value and have `install_mcp_size_guard` treat `<= 0` as "guard disabled" (skip the wrap, log once).
- Nothing enforces `WARN <= MAX`. If someone raises the warn level above the cap, the warn branch is dead code and the early-warning telemetry — the PR's stated second goal — silently disappears with no boot-time signal. A one-line check (log or raise) at install time closes it.
- Cosmetic but related: the warn log in `mcp_guard.py` prints the *cap* (`max_event_bytes`), not the warn threshold that actually tripped, so the message never states its own trigger level.

packages/orchestrator-agent/tests/test_mcp_guard.py
```diff
@@ -0,0 +1,122 @@
...
+        async def _handle_sse_event(self, sse, *args, **kwargs):
+            calls.append(sse.data)
+            return True
+
+    monkeypatch.setattr(mcp_guard, "_installed", False)
+
+    import mcp.client.streamable_http as real_module
+
+    monkeypatch.setattr(real_module, "StreamableHTTPTransport", FakeTransport)
+    install_mcp_size_guard(max_event_bytes=CAP, warn_event_bytes=WARN)
```
alloy-cicd: **[major] Every test runs against `FakeTransport`, whose caller is the test itself — the asserted contract is an artifact of the stub.**

The suite pins "oversized event raises before the original handler", but in production the raise happens inside the real SDK's `_handle_sse_response`, which catches it with a bare `except Exception` and drops it (see the comment on `mcp_guard.py:101`). So the suite is green while the actual behaviour is a silent hang plus a reconnect replay loop. The one test that would have caught this drives the real `StreamableHTTPTransport._handle_sse_response` (or at minimum the real `_handle_sse_event` signature) with an oversized event and asserts what arrives on a `read_stream_writer` memory stream.

Smaller test nits, no need for their own threads:
- `FakeTransport.handled = calls  # type: ignore` staples mutable state onto the class; returning `(FakeTransport, calls)` from the fixture drops the `type: ignore` and the hidden contract.
- `test_install_is_idempotent` re-declares the whole fixture harness; asserting idempotency against the existing `transport_cls` fixture removes ~15 duplicated lines that can drift.

aartaria:
## Triage of the 🤖 automatic review — all 7 findings addressed

The critical finding was right, and worth being explicit about: **round 1's mechanism would not have worked.** I verified it against the SDK source before changing anything — every call site of `_handle_sse_event` swallows exceptions (`except Exception: logger.debug`) and then reconnects with `Last-Event-ID`, replaying the same payload. So a raise meant: pending request hangs forever, oversized payload replayed on reconnect, and each hang pins a discovery-semaphore slot — five fat events and discovery deadlocks process-wide while the pod stays green to probes. Strictly worse than the OOM. This is also the second time in this incident chain that a reviewer caught me raising into a swallowing caller; the lesson is taken.

**Fixed (1, critical):** rejection now follows the SDK's own parse-failure contract — the error is delivered via `read_stream_writer` (pending request fails immediately, attributably) and the handler reports the stream complete so the SDK closes the response instead of reconnecting. Pinned by a test named for the failure mode: `test_oversized_event_fails_request_without_reconnect_replay`.

**Fixed (2):** sizes measured in UTF-8 **bytes**; the exact encode is paid only when the cheap char-count bounds (chars ≤ bytes ≤ 4·chars) can't decide. Test: a 4-byte-emoji payload whose char count is under the cap and byte size over it.

**Fixed (3):** `0` now disables a threshold (the `max(1, …)` clamp indeed turned `cap=0` into reject-everything); warn > cap is clamped with a log line; the warn message prints the warn threshold, not the cap.

**Fixed (4):** the test suite was rewritten to drive the **real** `StreamableHTTPTransport` with real anyio memory streams — the round-1 suite only exercised its own stub, which is exactly how finding 1 stayed hidden. 12 cases now, both parse paths.

**Fixed (5):** `_handle_json_response` (the `application/json` body path, same unbounded parse) is now guarded; httpx caches `aread()` so the original's own read is a no-op replay.

**Fixed (6):** only `message` events are size-checked.

**Fixed (7 + nits):** idempotency via a marker attribute on the wrapper; warn/error formatting deduped through one `_check` helper; AGENTS.md documents the process-wide monkeypatch and flags `mcp` version bumps as re-test triggers (a rename fails loudly at startup — `install_mcp_size_guard` touches the attributes directly).

**Acknowledged, out of scope:** `voice-agent` consumes the same catalogue unguarded — separate process, separate change; noted on #152 as a follow-up.

Full suite: **638 passed**. `ruff` clean on all touched files.

packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,196 @@
...
+        # Only "message" events are ever parsed by the SDK; priming/unknown
+        # events pass through untouched.
+        if getattr(sse, "event", None) == "message" and getattr(sse, "data", None):
+            verdict = _check(sse.data, self.url, max_event_bytes, warn_event_bytes)
+            if verdict.error is not None:
+                logger.error("[MCP-GUARD] %s", verdict.error)
+                try:
+                    # The SDK's own parse-failure contract: deliver the error to
+                    # the session so the pending request fails immediately...
+                    await read_stream_writer.send(verdict.error)
```
alloy-cicd: **[critical, still-open] The delivered exception is dropped one layer up: `ClientSession`'s default message handler is a no-op, so the pending request still hangs forever — the reconnect replay is fixed, the deadlock is not.**

Verified against `mcp` 1.27.1. The routing for what `read_stream_writer` receives:

- `BaseSession._receive_loop` (shared/session.py:357-359): `if isinstance(message, Exception): await self._handle_incoming(message)`.
- `ClientSession._handle_incoming` forwards to `self._message_handler`, which defaults to `_default_message_handler` = `await anyio.lowlevel.checkpoint()` — a no-op (client/session.py:59-62). langchain-mcp-adapters constructs `ClientSession(read, write, **(session_kwargs or {}))` and discovery passes no `session_kwargs`, so the default applies.
- The *only* thing that completes `send_request` is a `JSONRPCResponse`/`JSONRPCError` routed via `_response_streams[request_id]` (shared/session.py:260, 291), awaited with `timeout=None` since no read timeout is configured anywhere in this path.

So on rejection: the ERROR log fires, the error object is checkpointed into oblivion, `return True` closes the response (good — no replay, no re-download), and `client.get_tools()` still awaits forever holding its `MCP_DISCOVERY_CONCURRENCY` semaphore slot. The round-1 deadlock scenario — 5 hits on the fat server → discovery frozen process-wide, pod green to probes — is unchanged. The commit message's "the pending request fails immediately, attributably" and the docstring's equivalent claim don't hold. (To be fair: my round-1 comment asserted the SDK's parse-failure path fails the request promptly — that was wrong too; in 1.27.1 that path has the same dead end for clients using the default message handler.)

Two fix directions, ideally both:

1. **Fail the request attributably**: instead of the bare exception, send a synthetic protocol error that `_receive_loop` routes to the pending request:
```python
m = re.search(r'"id"\s*:\s*("(?:[^"\\]|\\.){1,200}"|-?\d{1,20})', data if isinstance(data, str) else data.decode(errors="replace"))
if m:
    await read_stream_writer.send(SessionMessage(JSONRPCMessage(JSONRPCError(
        jsonrpc="2.0", id=json.loads(m.group(1)),
        error=ErrorData(code=INTERNAL_ERROR, message=str(verdict.error)),
    ))))
```
The regex scan is O(n) with no object graph — none of the parse amplification the guard exists to prevent — and `send_request` then raises `McpError` immediately with the slug-and-size message. Fall back to the current exception-send (plus the ERROR log) if no id is found (notifications don't have a pending request anyway).

2. **Backstop regardless of mechanism**: pass `session_kwargs={"read_timeout_seconds": timedelta(seconds=...)}` where discovery/tool sessions are created, so *no* transport-level bug — this one or the next — can hold a semaphore slot forever. This was suggested in round 1 and is still absent.

Verifiable next round by: a test that runs a real `ClientSession` over anyio memory streams, feeds the oversized payload through the guarded handler, and asserts `session.send_request(...)`/`list_tools()` raises promptly rather than timing out the test.

packages/orchestrator-agent/tests/test_mcp_guard.py
```diff
@@ -0,0 +1,220 @@
...
+    assert WARN < len(padded) <= CAP
+    with caplog.at_level(logging.WARNING, logger="app.core.mcp_guard"):
+        complete = await t._handle_sse_event(sse(padded), send)
+    delivered = await drain_one(recv)
+    assert complete is True
+    assert not isinstance(delivered, Exception)
+    assert any("fat-server" in r.getMessage() for r in caplog.records)
+
+
+async def test_oversized_event_fails_request_without_reconnect_replay(real_transport, caplog):
```
alloy-cicd, line 99: **[major, follow-on] The test name asserts "fails the pending request", but no test puts a `ClientSession` in the loop — which is exactly where the delivered error is dropped.**

This suite is a big improvement (real transport class, real anyio streams, replay contract pinned). But it stops at the transport boundary: it asserts `McpEventTooLargeError` arrives on the raw memory stream, and the round-1 lesson repeats one layer up — the session's `_receive_loop` hands that object to a no-op default message handler and the awaiting `send_request` never completes (see the comment on `mcp_guard.py:162`). A test like:

```python
async def test_oversized_event_fails_pending_session_request():
    # real ClientSession over memory streams; guarded handler feeds the read side
    # assert list_tools()/send_request raises McpError/McpEventTooLargeError within 2s
```

would fail on this head, and is the assertion the name of this test currently promises.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (cc4c7d5287cb9117b6181e9bfe3c148e6c266486..4d0525c6f6746dedd62e89c559f10e7182afb6ac)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Raise swallowed by SDK; hang + Last-Event-ID replay | **partially fixed / still-open** — replay and re-download are gone (`return True` closes the response, correct), but the error object sent to `read_stream_writer` is dropped by `ClientSession`'s no-op default message handler, so the pending request still hangs forever and the discovery-semaphore deadlock scenario is unchanged. See inline comment. |
| 2 | `len(data)` chars vs bytes | **verified-fixed** — `_utf8_size` with the cheap-bound/exact-encode split is correct (bounds math checks out), emoji test pins it |
| 3 | `max(1, …)` no-disable; warn/max unvalidated; warn log printed cap | **verified-fixed** — `0` disables, warn>cap clamped with a boot log, warn message prints its own threshold |
| 4 | Tests only exercised a stub | **improved, one gap left** — suite now drives the real `StreamableHTTPTransport`, but no test crosses the session boundary, which is where the remaining bug lives. See inline comment. |
| 5 | `application/json` path unguarded | **verified-fixed** — `_handle_json_response` wrapped; the httpx `aread()` caching claim is correct |
| 6 | Guard fired on non-`message` events | **verified-fixed** |
| 7 | `_installed` global; no SDK-drift defence | **verified-fixed** — marker attribute on the wrapper; real-class tests + AGENTS.md note make an `mcp` bump loud |

Round-1 nits (duplicate slug/formatting, AGENTS.md documentation) were also addressed. The author's commit message is accurate about everything except the headline: "the pending request fails immediately, attributably" is not yet true — my own round-1 comment shared the same wrong assumption about the SDK's parse-failure contract, so this is a correction to both of us, with the routing verified line-by-line this round.

**Delta verdict:** the fix commit is high quality and resolves 6 of 7 findings outright, but the critical one is only half-closed: the failure mode changed from *hang + replay loop* to *clean, attributed ERROR log + silent permanent hang*. The semaphore-poisoning deadlock that motivated the critical rating is still reachable. Two inline comments: the remaining routing gap with a concrete fix (synthetic `JSONRPCError` via a cheap id scan, plus a `read_timeout_seconds` backstop), and the session-level test that would pin it.

aartaria:
## Round 3 — the routing gap was real; fixed and pinned across the session boundary

Verified the review's analysis against `mcp` 1.27.1 before changing anything, and it holds: `isinstance(message, Exception) → _handle_incoming` (a no-op for pending requests), and `send_request` awaits with `timeout=None`. Round 2 killed the reconnect-replay but left the hang — the deadlock scenario survived with a nicer log line. The review also gets credit for explicitly correcting its own round-1 assumption about the parse-failure contract.

**Both suggested fixes taken:**

1. **Routable rejection.** The guard now synthesizes a `JSONRPCError` carrying the request id recovered from the raw payload by a bounded O(n) regex scan (4KB head first, full-string fallback — no parse amplification). `ErrorData(code=INTERNAL_ERROR, message=…)` carries the server slug and sizes, so `send_request` raises `McpError` immediately and attributably. A wrong or unrecoverable id degrades to the round-2 bare-exception path — never worse than before.
2. **`MCP_DISCOVERY_TIMEOUT_S` backstop** (default 120 s, `0` disables) around each `tools/list` fetch, guaranteeing the semaphore slot is released even for stalls the guard can't attribute (or ids it can't recover). Declared in `.env.template` and the example manifest.

**And the test the review said would fail on the round-2 head — written, and it does:** a real `ClientSession` issues a real `list_tools` request over anyio streams, the guarded real transport rejects the oversized response, and the test asserts the pending request **fails** with the guard's message. Against the round-2 head it dies in `anyio.fail_after` — a TimeoutError, i.e. precisely the hang. A second new test asserts the synthetic `JSONRPCError` carries the recovered id.

Suite: **640 passed** (two `Mock(spec=AgentSettings)` discovery fixtures needed the new setting — same pattern as #149's round 2). `ruff` clean.

packages/orchestrator-agent/app/core/mcp_guard.py
```diff
@@ -0,0 +1,253 @@
...
+# "result" follows it. A bounded head-scan plus full-scan fallback is O(n) on the
+# raw string with NO parse amplification — the entire point of rejecting here.
+_ID_RE = re.compile(r'"id"\s*:\s*(\d+|"(?:[^"\\]|\\.){1,256}")')
+
+
+def _extract_request_id(data: str | bytes):
+    """Best-effort top-level ``id`` of a JSONRPC payload, without parsing it.
+
+    Returns an ``int`` or ``str``, or ``None`` when nothing safe was found.
+    Wrong-id extraction is tolerable: the synthetic error then matches no
```
alloy-cicd: **[minor, non-blocking] "Wrong-id extraction is tolerable: the synthetic error then matches no pending request" is incomplete — a wrong id *can* match a different pending request on the same session, failing an innocent in-flight call with a misattributed error.**

Session request ids are small sequential ints (`_request_id = request_id + 1`), so an `"id"` picked out of the *body* of a payload — reachable via the full-string fallback `_ID_RE.search(data)` when the head scan misses, e.g. a server that serializes `result` before `id`, or any oversized payload whose first 4096 chars contain a body-level `"id"` — has a realistic chance of colliding with another in-flight request when a session multiplexes (code-mode/tool dispatch; discovery's `tools/list` is effectively single-request, so it's safe). The blast radius is bounded and arguably acceptable: the innocent request fails fast with an error correctly naming the offending server, and the actual stalled request is reaped by the new `MCP_DISCOVERY_TIMEOUT_S`/caller timeout — but that trade-off deserves to be stated in the docstring instead of "matches no pending request", so a future debugging session isn't misled by it.

Small adjacent nits in the same function, all degrading safely to the fallback path (fine to fix or ignore):
- `_ID_RE` doesn't match negative integer ids (`-?\d+`) or `null` (JSONRPC allows a null id on some errors; `null` correctly shouldn't route anyway).
- The escaped-string branch returns `raw[1:-1]` without unescaping, so an id like `"a\"b"` produces a non-matching string (→ fallback, harmless).
- The bytes path scans only the first 4096 bytes with no full-scan fallback, while the str path scans the whole document — deliberate asymmetry or not, worth one line saying which is intended, since the JSON-response path (bytes) is the one where a late top-level `id` silently downgrades to the bare-exception + timeout path.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (4d0525c6f6746dedd62e89c559f10e7182afb6ac..53e3d22624cb1395b2e92c260506b9068c783c79)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Rejected error dropped by `ClientSession`'s no-op default message handler → pending request still hung, discovery-semaphore deadlock survived | **verified-fixed** — rejection now synthesizes a `JSONRPCError` routed by a request id recovered via a bounded O(n) regex scan (no parse amplification), with the bare-exception path as explicit fallback; and `MCP_DISCOVERY_TIMEOUT_S` (default 120 s, wrapped around the semaphore-held fetch with `asyncio.timeout`) guarantees the slot is released even when no id is recoverable. Verified: `asyncio` is imported, the `TimeoutError` lands in the existing per-server `except Exception` degradation path, and the routing matches the `_response_streams` mechanics checked in round 2. |
| 2 | No test crossed the session boundary — exactly where the round-2 bug lived | **verified-fixed** — `test_session_pending_request_actually_fails` runs a real `ClientSession` over memory streams, captures the outgoing request id, feeds the oversized response through the real guarded handler, and asserts the pending request fails with the guard's message. I ran the suite at this head: **14/14 pass**, including this test and the recovered-id assertion. |

**Delta verdict:** both round-2 findings are resolved, and the fix is the right shape — attributable fast-fail where the id is recoverable, a hard timeout backstop where it isn't. One new minor, non-blocking observation on the id-extraction edge cases (inline): a wrongly extracted id can collide with a *different* pending request on a multiplexed session, which is a bounded and arguably acceptable trade-off, but the docstring currently claims the opposite ("matches no pending request"), plus three regex/asymmetry nits that all degrade safely. Nothing blocking; docstring-level follow-up material.

aartaria:
Round-3 follow-up addressed in `d6d27a6a`: the `_extract_request_id` docstring now states the id-collision trade-off honestly (bounded by the fail-fast attributed error on the innocent request plus the `MCP_DISCOVERY_TIMEOUT_S` reap of the real one), negative ids match, the bytes path scans the full document like the str path, and escaped-string ids are documented as degrading to the fallback. Extractor variants pinned by a new test — suite at 15/15, 641 total.

From my side the PR is complete: all three review rounds' findings are either fixed and verified or explicitly documented as accepted trade-offs. Over to the human reviewer.

PR #156, opened 2026-08-18
fix(sdk): share the MCP size guard with voice-agent and agent-runner

### LINKED ISSUE #155: voice-agent, agent-runner, console-backend: still exposed to unbounded MCP SSE parsing (guard is orchestrator-only)
## Summary

The MCP SSE size guard added in #153 protects the **orchestrator** process only. Three other packages open MCP sessions against the same gateway and remain exposed to the failure mode that OOMKilled prod:

| package | entry point |
|---|---|
| `voice-agent` | `voice_agent/agent.py` — `from mcp.client.streamable_http import streamablehttp_client` |
| `agent-runner` | `agent/core.py` — `MultiServerMCPClient` with `transport="streamable_http"` |
| `console-backend` | uses the MCP client (its `/mcp` endpoint and gateway sub-calls) |

## Why it matters

The root cause is in the SDK, not in the orchestrator: `JSONRPCMessage.model_validate_json` parses inbound SSE events with no size bound, amplifying wire size ~7× into pydantic objects. One gateway server's 21.9 MB `tools/list` produced 663 MB of live objects in a single parse and drove peaks of 2.3–4.2 GB, repeatedly OOMKilling the orchestrator (#152).

Any process that loads the same catalogue can hit the same wall. `voice-agent` and `agent-runner` both read `MCP_GATEWAY_URL` and talk to the same gateway, so the exposure is not theoretical — it simply hasn't been triggered yet, and their memory limits are smaller than the orchestrator's.

## What to do

Apply the same guard. `app/core/mcp_guard.py` (orchestrator) is deliberately self-contained — no orchestrator imports — so the options are:

1. **Preferred:** move it to a shared package (`agent-common` or `ringier-a2a-sdk`, whichever these three already depend on) and call `install_mcp_size_guard(...)` once at each service's startup.
2. Copy it per service (fast, but three copies of a monkeypatch against private SDK internals is the wrong end state).

Each service needs its own `MCP_SSE_MAX_EVENT_BYTES` / `MCP_SSE_WARN_EVENT_BYTES` (or shared defaults), and the same guarantee the orchestrator got: the rejection must fail the pending request rather than hang it — see #153 for why a raised exception is swallowed by the SDK's call sites and why the rejection is delivered as a routable `JSONRPCError`.

Also worth carrying over: the per-server-timeout backstop (`MCP_DISCOVERY_TIMEOUT_S` equivalent) wherever these services await `tools/list`, since `send_request` waits with `timeout=None`.

## Note

Once #154 lands (raw-bytes catalogue ingest), the catalogue path stops paying the parse for whichever services adopt it — but tool *call results* still arrive through the session and its unbounded parse, so the guard remains necessary everywhere regardless.

aartaria:
closes #155

> **Stacked on #153** — base branch is `fix/mcp-sse-size-guard`, not `main`. `mcp_guard.py` does not exist on `main` yet, so this PR must merge after #153 (GitHub will retarget it to `main` automatically once #153 lands).

## Problem

The guard added in #153 protects the orchestrator process only, but the unbounded `JSONRPCMessage.model_validate_json` parse it defends against lives in the **MCP SDK's client transport**, not in the orchestrator. Every service that opens a session against the same gateway is exposed to the failure mode that OOMKilled prod (#152) — and `voice-agent` / `agent-runner` have *smaller* memory limits than the orchestrator, so they hit the wall sooner, not later.

## Where it went, and why

`orchestrator-agent/app/core/mcp_guard.py` → `ringier_a2a_sdk/utils/mcp_guard.py`.

The issue proposed "`agent-common` or `ringier-a2a-sdk`, whichever these three already depend on". It resolves to **`ringier-a2a-sdk`**: `voice-agent` has no `agent-common` dependency, so the SDK is the only package all of them share.

## Changes

- **`install_mcp_size_guard_from_env()`** — the entry point for services with no typed settings object of their own, so thresholds can't silently drift per service. The orchestrator keeps reading its own `AgentSettings` and calls `install_mcp_size_guard` directly.
- **voice-agent** installs in the lifespan *before* the warm-up task — that task opens a gateway session immediately, and its `tools/list` is exactly the call that blew up.
- **agent-runner** installs at the top of its lifespan.
- **`mcp` declared as a core SDK dependency, pinned `>=1.25.0,<2.0.0`.** The SDK now ships a module whose job is patching it, and `middleware/credential_injector.py` already imported `mcp.types` at module scope while `mcp` came in only transitively via the langgraph extras. On the bounds: **mcp 2.0.0 has just been released**, and this module wraps *private* transport internals a major bump is free to rename — an install that silently no-ops is the outage the guard exists to prevent. The old `>=1.0.0` floor predated `StreamableHTTPTransport` entirely, which would turn installation into a startup `ImportError` crash-loop.
- **Discovery timeout backstop** (`MCP_DISCOVERY_TIMEOUT_S`, default 120s, `<= 0` disables) for **both** voice-agent and agent-runner — see the correction below.
- Knobs declared in both services' `.env.template` and example k8s manifests.

## `console-backend` is not exposed — deliberately untouched

The issue lists it as "uses the MCP client (its `/mcp` endpoint and gateway sub-calls)". It doesn't:

- Its `/mcp` endpoint is an MCP **server** (`mcp.server.streamable_http_manager`, `@mcp.server.list_tools()` handlers). It has no `mcp.client` import anywhere, and its only `mcp` dependency is `fastapi-mcp`.
- Its gateway sub-calls are plain `httpx` REST against `/api/v1/mcp-servers`, not MCP sessions.

The guard patches `StreamableHTTPTransport` — a class this process never instantiates — so installing it there would be dead code. **Worth a separate look:** that httpx path does `response.json()` on a gateway response with no size bound. Same *class* of exposure, different mechanism, and out of scope here.

## Correction: agent-runner *does* need the backstop

An earlier revision of this PR claimed agent-runner was already covered because both `get_tools()` sites pass `sse_read_timeout`. **That was wrong**, and the automated review caught it. Verified against the SDK source:

- `sse_read_timeout` is **deprecated and no longer used** by the MCP SDK (`streamable_http.py:85` — `# Deprecated - no longer used`), so it bounded nothing.
- `ClientSession.send_request` waits with `timeout=None`.
- On the guard's id-less rejection path the handler returns `True`, closing the stream as *complete* rather than stalled — so no read timeout could fire even if the parameter still worked.

Both `get_tools()` call sites now go through `_get_tools_bounded`.

The voice-agent backstop stays scoped to discovery rather than applied session-wide on purpose: a `read_timeout_seconds` on the `ClientSession` would also cap legitimate slow tool calls mid-conversation, turning a fix into a regression.

## Also fixed from the automated review

- **Fail-open config parse.** voice-agent read `MCP_DISCOVERY_TIMEOUT_S` with a bare `int(os.getenv(...))`. A typo raised inside `_init_mcp_tools`, got swallowed by `_run_impl`'s broad `except Exception`, and dropped the call to a **tool-less session with risk checks silently skipped**. Now uses the guard's own fail-safe `int_env` (promoted from private `_int_env`).
- **`original_request_id` was dropped.** The guard ignored the SDK's 3rd positional and routed its synthesized `JSONRPCError` by the id scraped from the payload. On the resumption path the SDK *rewrites* a response's id to `original_request_id`, so the scraped id names the pre-resumption request, matches nothing, and hangs the request the rejection exists to fail. Pre-existing from #153, but shared by three services as of this PR.
- Guard docstring no longer promises a reaper only one service implements; `orchestrator-agent/AGENTS.md` no longer points at the pre-move paths.

## Verification

| package | result |
|---|---|
| `ringier-a2a-sdk` | 392 passed, 2 skipped |
| `voice-agent` | 11 passed |
| `agent-runner` | 14 passed |
| `orchestrator-agent` | 626 passed (633 minus the 7 guard tests that moved to the SDK) |

- New coverage: `install_mcp_size_guard_from_env` (defaults, overrides, `0` disables, unparseable values **failing safe rather than open**), `original_request_id` routing on the reject / fallback / pass-through paths, the `int_env` fail-safe table, and the discovery backstop in both services. The id-routing test was verified to **fail against the pre-fix code**, so it is not vacuous.
- `uv lock --check` passes for all six packages that depend on the SDK — including `agent-common` and `console-backend`, whose lockfiles the new core dependency would otherwise have left stale.
- `ruff` clean on all touched files; orchestrator `main.py`/`config.py` error lists byte-identical to the base branch.

> **Note for the reviewer:** PyPI was unreachable from this environment, so the six `uv.lock` files were updated by hand rather than regenerated. The edits are metadata-only — the resolved `mcp` version (1.25–1.27, already present in every lock) does not change — and `uv lock --check` validates all six. Still worth a `uv lock` run on a networked machine to confirm.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no. The guard is idempotent and process-wide; `ringier-a2a-sdk` gains a declared dependency it was already importing.


packages/voice-agent/voice_agent/agent.py
```diff
@@ -462,7 +462,19 @@ async def _init_mcp_tools(
...
         """
-        tools_result = await mcp_session.list_tools()
+        # Hard backstop on the catalogue fetch. ClientSession.send_request awaits
+        # with timeout=None, and this session sets no sse_read_timeout, so without
+        # this a stalled — or guard-rejected-without-a-recoverable-id (see
+        # ringier_a2a_sdk.utils.mcp_guard) — tools/list hangs the whole call setup
+        # with the caller hearing dead air. Scoped to discovery on purpose: a
+        # session-wide read_timeout_seconds would also cap legitimate slow tool
+        # calls mid-conversation. 0 disables.
+        discovery_timeout = int(os.getenv("MCP_DISCOVERY_TIMEOUT_S", "120"))
```
alloy-cicd: 🤖 **Unparseable `MCP_DISCOVERY_TIMEOUT_S` fails open — every call runs tool-less**

This `int()` raises `ValueError` on a blank or non-numeric value, and it runs *inside* `_init_mcp_tools`, whose caller catches `except Exception` with `session_started=False`. So `MCP_DISCOVERY_TIMEOUT_S: "120s"` (or a k8s `value: ""`) makes every voice call log `MCP gateway connection FAILED` and continue **without tools** — write-tool risk checks silently skipped, indistinguishable from a real gateway outage.

The same PR argues the opposite policy for the sibling knobs: `_int_env` in `mcp_guard.py` logs and falls back precisely so "a typo in a manifest must not leave a pod unguarded". Worth reusing it here. Also note a negative value silently disables the backstop even though only `0` is documented as the opt-out.

```suggestion
        discovery_timeout = _int_env("MCP_DISCOVERY_TIMEOUT_S", 120)
```
(with `from ringier_a2a_sdk.utils.mcp_guard import _int_env`, or a small local helper if you'd rather not import a private name.)

packages/agent-runner/main.py
```diff
@@ -63,6 +64,12 @@
 @asynccontextmanager
 async def lifespan(app) -> AsyncIterator[None]:
     """Lifespan context manager for the FastAPI application."""
+    # Guard inbound MCP messages before any agent run can open a session against
+    # the shared gateway: the unbounded SDK parse that OOMKilled the orchestrator
+    # (ringier-data/nannos#152, #155) is transport-level, and this pod's memory
+    # limit is smaller. Process-wide and idempotent.
+    install_mcp_size_guard_from_env()
```
alloy-cicd, line 71: 🤖 **`sse_read_timeout` does not cover the rejection path this guard creates — agent-runner can hang**

The PR reasons agent-runner needs no discovery backstop because both `get_tools()` call sites pass `sse_read_timeout=MCP_TIMEOUT_SECONDS`, "which reaps a stalled stream". That is true for a *stalled* stream, but the guard's id-less rejection path does not stall — it completes:

1. Oversized payload whose top-level `id` `_extract_request_id` cannot recover (non-JSON body, or an id the bounded regex misses) → `_reject` sends a **bare** `McpEventTooLargeError` on the read stream.
2. `BaseSession._receive_loop` routes `isinstance(message, Exception)` to `_handle_incoming`, which is a no-op — nothing fails the pending request (the guard's own docstring says this).
3. The guarded handler returns `True`, so the SDK closes the response as *complete*. `sse_read_timeout` never fires: the stream ended normally.
4. `send_request` awaits with `timeout=None` → `MultiServerMCPClient.get_tools()` hangs indefinitely.

The orchestrator survives this via `asyncio.timeout` in `app/core/discovery.py:327` and voice-agent now has `MCP_DISCOVERY_TIMEOUT_S`; agent-runner has neither. Either wrap the two `get_tools()` awaits in `asyncio.timeout`, or pass `read_timeout_seconds` via `session_kwargs` on the discovery connections.

packages/ringier-a2a-sdk/pyproject.toml
```diff
@@ -13,6 +13,10 @@ dependencies = [
     "cryptography>=43.0.0",
     "authlib>=1.6.5",
     "sqlalchemy>=2.0.51",
+    # Core, not an extra: utils/mcp_guard.py patches the MCP client transport and
+    # middleware/credential_injector.py imports mcp.types at module scope. Until
+    # now this came in only transitively via the langgraph extras' langchain-mcp-adapters.
+    "mcp>=1.0.0",
```
alloy-cicd: 🤖 **`mcp>=1.0.0` is below the version that has the class this SDK patches**

`install_mcp_size_guard` does `from mcp.client.streamable_http import StreamableHTTPTransport`, and that module/class only appeared with the streamable-HTTP client (mcp ~1.8), well above this floor. A consumer resolving without this repo's lockfile — or after a downgrade pin — can install mcp 1.6 and get an `ImportError` inside the service lifespan, i.e. a crash-loop at startup rather than a degraded run.

Since the declared dependency now exists *because* of this module, the floor should be the first version shipping `StreamableHTTPTransport._handle_sse_event` / `_handle_json_response` (matching what the lock resolves today). Copying voice-agent's existing `mcp>=1.0.0` propagates the same too-loose bound.

alloy-cicd:
🤖 Three more findings that could not be anchored inline (lines outside this diff's hunks):

**1. `ringier_a2a_sdk/utils/mcp_guard.py:232` — the rejection ignores the SDK's `original_request_id`.**
`*args` swallows `original_request_id`, which the SDK passes from its reconnect path (`_handle_reconnection`) and uses to *rewrite* the response id before delivery. If the replayed payload is oversized, the guard routes its synthesized `JSONRPCError` with the id scraped from the server's payload instead. When those differ, no pending request matches and the caller hangs — the exact failure the id-recovery logic exists to prevent. Pre-existing from #153; fix is to prefer `args[0]` when not `None`.

**2. `ringier_a2a_sdk/utils/mcp_guard.py:149` — the shared module cites a knob only one service has.**
`_extract_request_id`'s docstring says a mis-routed rejection is "reaped by the `MCP_DISCOVERY_TIMEOUT_S` backstop". That variable exists only in voice-agent. The orchestrator relies on its own `asyncio.timeout` (`app/core/discovery.py:327`) and agent-runner has nothing (see the inline comment on `agent-runner/main.py`). Now that the module is shared by three services, this reads as a global guarantee it does not have — and is what makes the "agent-runner needs no backstop" conclusion look safe.

**3. `packages/orchestrator-agent/AGENTS.md:126` and `:136` — stale paths after the move.**
The "MCP Message Size Guard" section still names `app/core/mcp_guard.py` and instructs "treat any `mcp` version bump as a signal to re-run `tests/test_mcp_guard.py`" — both moved to `ringier-a2a-sdk` by this PR. On the next SDK bump an engineer finds no such file in orchestrator-agent, and the suite that actually catches a renamed private SDK method goes unrun.

---
Verified locally: `ringier-a2a-sdk` guard suite 22 passed, `voice-agent` 11 passed, `agent-runner` 11 passed. Both new k8s manifest env blocks land in the correct container. No remaining references to `app.core.mcp_guard` in code.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (05d7e5ebb56c747c18243347b12b6a9736578eb7..3b0cddfdcbe8c2969484cf6f3a1a2d1ed7a35d26)

**Every previous finding is resolved and the delta introduces nothing new.** No inline comments this round.

## Ledger

| # | Round-1 finding | Disposition |
|---|---|---|
| 1 | `voice_agent/agent.py` — bare `int(os.getenv("MCP_DISCOVERY_TIMEOUT_S"))` fails open into a tool-less session | **verified-fixed** — now `int_env(...)` (`agent.py:476`); comment records the fail-open reasoning; `<= 0 disables` documented, closing the undocumented-negative gap |
| 2 | agent-runner has no discovery backstop; `sse_read_timeout` does not cover the guard's id-less rejection | **verified-fixed** — `_get_tools_bounded` (`agent/core.py:87`) wraps **both** `get_tools()` sites (`core.py:788`, `core.py:1308`); each is inside a `try` that surfaces the `TimeoutError` rather than hanging |
| 3 | Guard swallowed the SDK's `original_request_id`, routing the rejection by the payload-scraped id | **verified-fixed** — `_guarded_handle_sse_event` names it explicitly and prefers it in `_reject` (`mcp_guard.py:190`, `:241`, `:255`, `:257`) |
| 4 | Guard docstring promised a reaper only one service implemented | **verified-fixed** — `mcp_guard.py:148-151` now says it is the caller's job and names all three services |
| 5 | `orchestrator-agent/AGENTS.md` pointed at pre-move paths | **verified-fixed** — `AGENTS.md:126,138` retargeted at `packages/ringier-a2a-sdk/tests/test_mcp_guard.py` |
| 6 | `mcp>=1.0.0` floor below the version shipping `StreamableHTTPTransport` | **verified-fixed** — `>=1.25.0,<2.0.0` in both `ringier-a2a-sdk` and `voice-agent`, with the reason for each bound in a comment |

Nothing was rejected by the author; all six were accepted and acted on.

## Verification of the delta

Checked against the real MCP SDK source (1.25.0 / 1.27.0 / 1.27.1 in this repo's locks), not from memory:

- **`original_request_id` really is the 3rd positional** of `StreamableHTTPTransport._handle_sse_event` (`streamable_http.py:200-206`), and `_handle_resumption_request` / `_handle_reconnection` pass it **positionally** — so the wrapper's `(self, sse, read_stream_writer, original_request_id=None, *args, **kwargs)` captures it on exactly the paths that matter. `_handle_sse_response` (the normal POST path) passes `resumption_callback` / `is_initialization` by **keyword** and omits the id, so `None` is the correct capture there and the pass-through forward is a no-op. Confirmed the SDK's rewrite the fix is chasing: `message.root.id = original_request_id` at `:225`, guarded to `JSONRPCResponse | JSONRPCError`.
- **`_handle_json_response` has no `original_request_id` parameter** in any 1.25–1.27, so leaving that wrapper on the scrape-only path is right, not an oversight.
- **`sse_read_timeout: float | None = None  # Deprecated - no longer used`** at `streamable_http.py:85` — the correction in the PR description is accurate.
- **`asyncio.timeout` needs 3.11+**; agent-runner is `requires-python >=3.12`. OK.
- Locks: all six resolve `mcp` to 1.25.0/1.27.0/1.27.1, inside the new bounds. The `{ name = "mcp" }` edges added to `agent-common` / `console-backend` both have a matching `[[package]]` stanza, so the hand edits are complete rather than dangling. `uv lock --check` passes for all six (`agent-common` 131, `console-backend` 178, `orchestrator-agent` 147, `ringier-a2a-sdk` 116, `agent-runner` 142, `voice-agent` 122).
- Manifests/templates: `MCP_DISCOVERY_TIMEOUT_S` present in all three of `orchestrator.yaml:81`, `voice-agent.yaml:78`, `agent-runner.yaml:66` and both `.env.template`s — so the docstring's claim that all three services supply the knob is true, and `orchestrator` does read that exact name (`app/models/config.py:362`).

Tests re-run locally on `3b0cddfd`:

| suite | result |
|---|---|
| `ringier-a2a-sdk/tests/test_mcp_guard.py` | 33 passed (22 → 33) |
| `agent-runner` (full) | 14 passed (11 → 14) |
| `voice-agent` (full) | 11 passed |

No new lines over the 120-char limit in any touched file.

## Follow-up material only — not blocking

- `_get_tools_bounded` was inserted *between* module constants in `agent/core.py`, leaving `_DOCUMENT_STORE_S3_BUCKET` / `_MAX_RECURSION_LIMIT` stranded below a function definition (`core.py:87-94`). Legal and ruff-clean under this repo's rule set; purely a reading-order nit.
- `orchestrator-agent/app/models/config.py:23` still carries its own private `_int_env`, now a duplicate of the SDK's promoted `int_env`. Untouched by this PR; worth folding together whenever that file is next opened.
- `_MCP_DISCOVERY_TIMEOUT_S` is resolved at *import* time in agent-runner, so `int_env`'s fail-safe warning can land before logging is configured. The value still fails safe — only the log line is at risk — and the neighbouring bare `int(os.getenv(...))` constants are strictly worse on the same axis, so this is not worth changing in isolation.
- Still outstanding from the PR's own text: `console-backend`'s httpx path does `response.json()` on a gateway response with no size bound. Same class of exposure, different mechanism, correctly scoped out of this PR.

packages/voice-agent/voice_agent/agent.py
alloy-cicd: 🤖 **Round 2: verified-fixed.** `agent.py:476` now reads `int_env("MCP_DISCOVERY_TIMEOUT_S", 120)`, so a `"120s"` or empty manifest value falls back to 120 instead of raising into `_run_impl`'s `except Exception` and dropping the call to a tool-less session. The undocumented-negative gap is closed too — the comment and both templates now say `<= 0` disables, matching `if discovery_timeout > 0`. `int_env` was promoted out of `_int_env`, so the private-name concern is moot. voice-agent: 11 passed.

packages/agent-runner/main.py
alloy-cicd: 🤖 **Round 2: verified-fixed.** The backstop landed in `agent/core.py` (where the `get_tools()` calls actually live) rather than `main.py`: `_get_tools_bounded` at `core.py:87` wraps **both** sites — `core.py:788` (watch check_tool) and `core.py:1308` (scheduled-job tool load) — and each sits inside a `try` that turns the `TimeoutError` into a logged failure instead of a hang.

Confirmed the author's stronger correction against the SDK source: `sse_read_timeout: float | None = None  # Deprecated - no longer used` at `streamable_http.py:85`, so it bounded nothing even before the rejection-path argument. New coverage in `tests/test_mcp_discovery_timeout.py` (reaped / `<= 0` disables / fast path unaffected); agent-runner 14 passed.

packages/ringier-a2a-sdk/pyproject.toml
alloy-cicd: 🤖 **Round 2: verified-fixed, and tightened further than asked.** Now `mcp>=1.25.0,<2.0.0` here and in `voice-agent/pyproject.toml`, each bound explained inline. The floor matches what the guard actually needs — verified against the SDK source that 1.25 has both `StreamableHTTPTransport._handle_sse_event`/`_handle_json_response` and the `original_request_id` parameter the routing fix depends on. The `<2.0.0` cap is well argued for a module that wraps private transport internals, where a silent no-op is the outage the guard exists to prevent.

Lockfiles check out: all six resolve `mcp` to 1.25.0/1.27.0/1.27.1, inside the bounds, and the `{ name = "mcp" }` edges hand-added to `agent-common`/`console-backend` each have a matching `[[package]]` stanza. `uv lock --check` passes for all six.

PR #158, opened 2026-08-20
feat(console): allow a watch to trigger a sub-agent

### LINKED ISSUE #157: Console UI: allow a watch to trigger a sub-agent (backend already supports it)
## Problem

The backend fully supports a **watch** schedule triggering a sub-agent workflow, but the console UI never exposes it — so from the UI a watch can only "poll condition then notify".

- A watch is two-phase (`agent-runner/agent/core.py`, ~`_run_watch`): one MCP `check_tool` call + condition eval (JSONPath `condition_expr`, plus `llm_condition` / `expected_value` / non-null default). If the condition is met and `sub_agent_id` is set, it runs a **full sub-agent conversation** — the same code path as a `task` job.
- The data model states it: `sub_agent_id` is "Required for job_type='task'; **optional for 'watch'**" (`console-backend/models/scheduled_job.py`), and the scheduler router docstring says "poll a condition before optionally invoking an agent".
- The console UI guards the sub-agent block with `form.job_type === 'task'` (`console-frontend/src/pages/SchedulerPage.tsx`), never sets `sub_agent_id` in the watch payload, and the dropdown reads "Watch – poll condition then notify".
- Service-level nuance: inline sub-agent creation (`sub_agent_parameters`) is honored only for `job_type == 'task'` (`scheduler_service.py`) — a watch must reference a **pre-existing** `sub_agent_id`.

## Proposed change

1. In `SchedulerPage.tsx`, allow attaching an existing `sub_agent_id` to a watch (selector for pre-existing sub-agents only; keep inline creation task-only — forcing pre-reviewed agents on watches is the safer default). Include it in the watch create/update payload.
2. Update the watch copy ("poll condition then notify" → reflect the optional agent trigger).
3. Keep the notify-only behavior as the default when no sub-agent is selected.

## Notes

- **Sequencing:** should land *after* the scheduled-run approval-handling (HITL fail-open) stopgap — exposing watch→agent in the UI without approval handling widens the fail-open surface.
- Out of scope: lifting the task-only restriction on inline `sub_agent_parameters` for watches (deliberately kept as-is).

aartaria:
closes #157

The backend has always supported an optional `sub_agent_id` on watch jobs (condition met → full sub-agent conversation, same code path as a task), but the console UI only exposed sub-agents for task jobs, so watches were stuck at notify-only.

### Changes

- **SchedulerPage (create dialog):** optional "Sub-agent" selector on the watch form (existing sub-agents only — inline `sub_agent_parameters` stay task-only by design, so watches can only reference pre-existing, reviewed agents). Included in the create payload; the notification-message field is hidden while a sub-agent is selected, since the agent's response is what gets delivered (`agent-runner/agent/core.py`, watch branch). Watch dropdown copy updated to "poll condition, then notify or run agent".
- **SchedulerJobDetailPage (edit form):** same selector on the watch edit form, with a "None (notify only)" option to clear it; the notification-text helper explains it is unused while a sub-agent is set.
- **console-backend:** `sub_agent_id` moved from the ignore-`None` update path to the `_UNSET` pattern (router passes it via `model_fields_set`), so an explicit `null` clears it on a watch. Clearing is rejected with a 400 for task jobs (they require a sub-agent), and the existing sub-agent access check still applies when setting one.

### Verification

- `console-backend`: `pytest tests/test_scheduler_service.py` → 33 passed (3 new tests: set-on-watch with access check, clear-on-watch skips access check, clear-on-task rejected); `ruff` clean; `mypy` adds no errors on touched files.
- `console-frontend`: `tsc -b` clean; `eslint` on both touched pages shows the same 5 pre-existing findings as `origin/main` — none added.

### Note on sequencing

Per the issue: this widens the watch→agent surface in the UI, and scheduled sub-agent runs currently handle HITL approvals fail-open. If the approval-handling stopgap hasn't landed yet, consider holding the merge until it does.

## Checklist

### DIFF
--- a/packages/console-backend/console_backend/services/scheduler_service.py
+++ b/packages/console-backend/console_backend/services/scheduler_service.py
@@ -11,6 +11,7 @@ from console_backend.services.user_settings_service import UserSettingsService
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from ..models.scheduled_job import (
+    JobType,
     ScheduledJob,
     ScheduledJobCreate,
     ScheduledJobRun,
@@ -219,6 +220,7 @@ class SchedulerService:
         destroy_after_trigger: bool | None = _UNSET,
         check_args: dict | None = _UNSET,
         delivery_channel_id: int | None = _UNSET,
+        sub_agent_id: int | None = _UNSET,
         **kwargs,
     ) -> ScheduledJob | None:
         job = await self.repo.get_job(db, job_id)
@@ -250,9 +252,13 @@ class SchedulerService:
             if delivery_channel_id is not None:
                 await self._validate_delivery_channel(db, delivery_channel_id)
             fields["delivery_channel_id"] = delivery_channel_id
+        if sub_agent_id is not _UNSET:
+            if sub_agent_id is None and job.job_type == JobType.TASK:
+                raise ValueError("sub_agent_id cannot be cleared on a task job")
+            fields["sub_agent_id"] = sub_agent_id
 
         # Handle fields that still use old pattern (from kwargs/data)
-        for attr in ("enabled", "max_failures", "sub_agent_id", "voice_call"):
+        for attr in ("enabled", "max_failures", "voice_call"):
             val = getattr(data, attr, None)
             if val is not None:
                 fields[attr] = val
--- a/packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
+++ b/packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
@@ -330,6 +330,9 @@ function EditForm({ job }: { job: ScheduledJob }) {
   );
   const initialSubAgentId = job.sub_agent_id != null ? String(job.sub_agent_id) : '';
   const [subAgentId, setSubAgentId] = useState(initialSubAgentId);
+  // Watch jobs: instruction for the sub-agent triggered by the condition
+  // (task jobs edit their prompt through `message` instead).
+  const [watchPrompt, setWatchPrompt] = useState(job.prompt ?? '');
   const [checkTool, setCheckTool] = useState(job.check_tool ?? '');
   const [checkArgsText, setCheckArgsText] = useState(
     job.check_args ? JSON.stringify(job.check_args, null, 2) : '',
@@ -385,6 +388,7 @@ function EditForm({ job }: { job: ScheduledJob }) {
     setRunAt(initialRunAt);
     setMessage(job.job_type === 'task' ? (job.prompt ?? '') : (job.notification_message ?? ''));
     setSubAgentId(initialSubAgentId);
+    setWatchPrompt(job.prompt ?? '');
     setCheckTool(job.check_tool ?? '');
     setCheckArgsText(job.check_args ? JSON.stringify(job.check_args, null, 2) : '');
     setConditionExpr(job.condition_expr ?? '');
@@ -480,6 +484,7 @@ function EditForm({ job }: { job: ScheduledJob }) {
       }),
       ...(job.job_type === 'watch' && {
         notification_message: message.trim() ? message.trim() : null, // Watch jobs use notification_message field
+        prompt: watchPrompt.trim() ? watchPrompt.trim() : null, // Instruction for the triggered sub-agent
         check_tool: checkTool || undefined,
         check_args,
         condition_expr: conditionExpr || undefined,
@@ -754,11 +759,31 @@ function EditForm({ job }: { job: ScheduledJob }) {
               includeNone
             />
             <p className="text-xs text-muted-foreground">
-              When the condition is met, this sub-agent is invoked with the check tool's result as its input
-              ("Watch condition triggered. Take appropriate action based on: &lt;check result&gt;") and acts per
-              its own system prompt. Its response is delivered instead of the notification message.
+              When the condition is met, this sub-agent is invoked with the check tool's result as its
+              input, plus the instruction below. Its response is delivered instead of the notification
+              message.
             </p>
           </div>
+
+          {/* Instruction for the triggered sub-agent */}
+          {subAgentId && (
+            <div className="grid gap-1.5">
+              <Label>
+                Sub-agent instruction{' '}
+                <span className="text-muted-foreground text-xs">(optional)</span>
+              </Label>
+              <Textarea
+                rows={3}
+                value={watchPrompt}
+                onChange={(e) => { setWatchPrompt(e.target.value); touch(); }}
+                placeholder="e.g. Summarize the result and email it to the account owner…"
+              />
+              <p className="text-xs text-muted-foreground">
+                Sent to the sub-agent together with the check result when the condition triggers. If
+                empty, the agent is asked to "take appropriate action based on the check result".
+              </p>
+            </div>
+          )}
 
         </>
       )}
--- a/packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
+++ b/packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
@@ -779,7 +779,9 @@ function EditForm({ job }: { job: ScheduledJob }) {
               </SelectContent>
             </Select>
             <p className="text-xs text-muted-foreground">
-              Run this sub-agent when the condition is met. Its response is delivered instead of the notification message.
+              When the condition is met, this sub-agent is invoked with the check tool's result as its input
+              ("Watch condition triggered. Take appropriate action based on: &lt;check result&gt;") and acts per
+              its own system prompt. Its response is delivered instead of the notification message.
             </p>
           </div>
 
--- a/packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
+++ b/packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
@@ -49,6 +49,7 @@ import {
   SelectValue,
 } from '@/components/ui/select';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
+import { SubAgentSelect } from '@/components/SubAgentSelect';
 import { config } from '@/config';
 import {
   type JobRunStatus,
@@ -327,9 +328,8 @@ function EditForm({ job }: { job: ScheduledJob }) {
   const [message, setMessage] = useState(
     job.job_type === 'task' ? (job.prompt ?? '') : (job.notification_message ?? '')
   );
-  const [subAgentId, setSubAgentId] = useState(
-    job.sub_agent_id != null ? String(job.sub_agent_id) : '',
-  );
+  const initialSubAgentId = job.sub_agent_id != null ? String(job.sub_agent_id) : '';
+  const [subAgentId, setSubAgentId] = useState(initialSubAgentId);
   const [checkTool, setCheckTool] = useState(job.check_tool ?? '');
   const [checkArgsText, setCheckArgsText] = useState(
     job.check_args ? JSON.stringify(job.check_args, null, 2) : '',
@@ -384,7 +384,7 @@ function EditForm({ job }: { job: ScheduledJob }) {
     setIntervalSeconds(job.interval_seconds != null ? String(job.interval_seconds) : '');
     setRunAt(initialRunAt);
     setMessage(job.job_type === 'task' ? (job.prompt ?? '') : (job.notification_message ?? ''));
-    setSubAgentId(job.sub_agent_id != null ? String(job.sub_agent_id) : '');
+    setSubAgentId(initialSubAgentId);
     setCheckTool(job.check_tool ?? '');
     setCheckArgsText(job.check_args ? JSON.stringify(job.check_args, null, 2) : '');
     setConditionExpr(job.condition_expr ?? '');
@@ -463,14 +463,23 @@ function EditForm({ job }: { job: ScheduledJob }) {
       // resending the prefill on an unrelated edit would needlessly re-touch
       // the schedule.
       ...(job.schedule_kind === 'once' && runAt !== initialRunAt && { run_at: runAt || undefined }),
+      // Only send sub_agent_id when it actually changed: an unchanged value
+      // would still re-run the backend's access check on every save, and would
+      // reject unrelated edits outright once the agent is no longer accessible.
+      // For watches an explicit null clears it back to notify-only; task jobs
+      // require one, so an emptied value is simply not sent.
+      ...(subAgentId !== initialSubAgentId && {
+        sub_agent_id: subAgentId
+          ? parseInt(subAgentId)
+          : job.job_type === 'watch'
+            ? null
+            : undefined,
+      }),
       ...(job.job_type === 'task' && {
-        sub_agent_id: subAgentId ? parseInt(subAgentId) : undefined,
         prompt: message.trim() ? message.trim() : null, // Task jobs use prompt field
       }),
       ...(job.job_type === 'watch' && {
         notification_message: message.trim() ? message.trim() : null, // Watch jobs use notification_message field
-        // Optional for watches: null clears it back to notify-only.
-        sub_agent_id: subAgentId ? parseInt(subAgentId) : null,
         check_tool: checkTool || undefined,
         check_args,
         condition_expr: conditionExpr || undefined,
@@ -588,33 +597,14 @@ function EditForm({ job }: { job: ScheduledJob }) {
         <>
           <div className="grid gap-1.5">
             <Label>Sub-agent</Label>
-            <Select
+            <SubAgentSelect
               value={subAgentId}
-              onValueChange={(v) => { setSubAgentId(v); touch(); }}
+              onChange={(v) => { setSubAgentId(v); touch(); }}
+              subAgents={subAgents}
               disabled={!editing}
-            >
-              <SelectTrigger>
-                <SelectValue placeholder="Select a sub-agent…" />
-              </SelectTrigger>
-              <SelectContent>
-                {subAgents.length === 0 ? (
-                  <div className="px-3 py-2 text-sm text-muted-foreground">
-                    No sub-agents found
-                  </div>
-                ) : (
-                  subAgents.filter((sa) => sa.name !== 'voice-agent').map((sa) => (
-                    <SelectItem key={sa.id} value={String(sa.id)}>
-                      <span>{sa.name}</span>
-                      {sa.type === 'automated' && (
-                        <span className="ml-2 text-xs text-muted-foreground">(automated)</span>
-                      )}
-                    </SelectItem>
-                  ))
-                )}
-              </SelectContent>
-            </Select>
+            />
             <p className="text-xs text-muted-foreground">
-              {subAgents.find(sa => sa.id === parseInt(subAgentId))?.type === 'automated' 
+              {subAgents.find(sa => sa.id === parseInt(subAgentId))?.type === 'automated'
                 ? 'This automated sub-agent has a predefined system prompt.'
                 : 'Select a sub-agent to execute for this scheduled job.'}
             </p>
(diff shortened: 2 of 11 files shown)


packages/console-frontend/src/pages/SchedulerPage.tsx
```diff
@@ -412,6 +412,9 @@ function CreateJobDialog({
       body.llm_condition = form.llm_condition.trim() || undefined;
       body.destroy_after_trigger = form.destroy_after_trigger;
       body.notification_message = form.notification_message.trim();
+      // Optional: run an existing sub-agent when the condition is met (inline
+      // sub_agent_parameters stay task-only by design).
+      if (form.sub_agent_id) body.sub_agent_id = parseInt(form.sub_agent_id);
```
alloy-cicd: **[bug] Stale task-mode selection leaks into the watch.** `form.sub_agent_id` is shared between the task picker (~line 555) and the new watch picker below, and the `update('job_type', …)` helper never resets it. The dialog opens in the default `task` mode, so: pick a sub-agent → switch type to `watch` → the agent is silently attached to the watch, and (because of the `!form.sub_agent_id` gate at line 981) the notification-message field disappears as well, hiding the evidence. `form.voice_call` leaks the same way (line 381 sends it; the engine silently ignores it for watches).

Suggest clearing `sub_agent_id` (and `voice_call`) when `job_type` changes.

Repro: open Create, keep type = task, select any sub-agent, switch type to watch — the watch picker is pre-populated.

packages/console-frontend/src/pages/SchedulerPage.tsx
```diff
@@ -939,11 +942,43 @@ function CreateJobDialog({
...
+                  Run this sub-agent when the condition is met. Its response is delivered instead of the notification message.
+                </p>
+              </div>
             </>
           )}
 
-          {/* Notification message for watch jobs */}
-          {form.job_type === 'watch' && (
+          {/* Notification message for watch jobs (superseded by the sub-agent's response when one is set) */}
+          {form.job_type === 'watch' && !form.sub_agent_id && (
```
alloy-cicd: **[bug] The hidden field is still submitted.** Unmounting the textarea doesn't clear its state: line 414 still sends `body.notification_message = form.notification_message.trim()` unconditionally. A user who types a message and *then* picks a sub-agent has that text persisted invisibly; it resurfaces as the live notification if the sub-agent is later cleared on the detail page.

Note the edit page made the opposite choice for the same situation — field stays visible with a "Not used while a sub-agent is set" caption (SchedulerJobDetailPage.tsx:797-806). Reusing that pattern here (drop the `&& !form.sub_agent_id`, swap the caption) fixes both the stale submit and the create/edit inconsistency.

packages/console-frontend/src/pages/SchedulerPage.tsx
```diff
@@ -939,11 +942,43 @@ function CreateJobDialog({
               <p className="text-xs text-muted-foreground -mt-1 ml-6">
                 When enabled (default), the watch will automatically be disabled after the condition is met once. Disable this to keep the watch running indefinitely.
               </p>
+
+              {/* Optional sub-agent trigger */}
```
alloy-cicd, line 941: **[dup] Fourth copy of the sub-agent picker.** This `<Select>` (and its twin in SchedulerJobDetailPage.tsx:753-783) repeats the task picker item-for-item, so the `sa.name !== 'voice-agent'` exclusion and the `(automated)` badge now live in four pickers (plus the same filter in SkillsPage.tsx:128), each with slightly different empty-state behavior. Consider extracting a `SubAgentSelect` (`value`, `onChange`, `allowNone`, `disabled`) under `components/subagents/` so "which agents are selectable" has one home — especially since a hard-coded name match is the only thing keeping the system agent out of these lists.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -469,6 +469,8 @@ function EditForm({ job }: { job: ScheduledJob }) {
       }),
       ...(job.job_type === 'watch' && {
         notification_message: message.trim() ? message.trim() : null, // Watch jobs use notification_message field
+        // Optional for watches: null clears it back to notify-only.
+        sub_agent_id: subAgentId ? parseInt(subAgentId) : null,
```
alloy-cicd: **[behavior] `sub_agent_id` is sent on every watch save, unlike its neighbours.** The other optional fields in this literal follow the send-only-when-meaningful convention (`run_at` a few lines up is guarded by `runAt !== initialRunAt`). Sending this unconditionally means (a) every save — even a rename — re-runs the server-side `get_accessible_sub_agents` check (which fully hydrates every accessible agent), and (b) if the attached agent has since been unshared or deleted, *any* unrelated edit is rejected with 400 "Access denied … sub-agent N" until the user notices the picker and resets it to None.

Suggest including the key only when `subAgentId` differs from the loaded `job.sub_agent_id`, mirroring the `run_at` guard.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -748,6 +750,39 @@ function EditForm({ job }: { job: ScheduledJob }) {
...
           </div>
 
+          {/* Optional sub-agent trigger */}
+          <div className="grid gap-1.5">
+            <Label>
+              Sub-agent{' '}
+              <span className="text-muted-foreground text-xs">(optional)</span>
+            </Label>
+            <Select
+              value={subAgentId || '_none'}
```
alloy-cicd: **[bug] Accessible-but-not-owned agents render as "None (notify only)".** The options come from `consoleListSubAgentsOptions({ query: { owned_only: true } })` (line 352), but the backend accepts any agent from `get_accessible_sub_agents` (owned ∪ public ∪ group-shared) — and `voice-agent` is filtered out besides. A watch attached to a shared agent (e.g. via the MCP scheduler tool) has a `value` with no matching `SelectItem`, so Radix falls back to the placeholder: the page reads as notify-only while the agent keeps executing on every trigger. Worse, if the user opens the picker and clicks "None" to confirm what they see, the save sends `sub_agent_id: null` and silently detaches a working trigger.

Suggest populating from the accessible list, or at minimum rendering a fallback `SelectItem` for an id not in the list.

packages/console-backend/console_backend/services/scheduler_service.py
```diff
@@ -250,9 +252,13 @@ async def update_job(
             if delivery_channel_id is not None:
                 await self._validate_delivery_channel(db, delivery_channel_id)
             fields["delivery_channel_id"] = delivery_channel_id
+        if sub_agent_id is not _UNSET:
+            if sub_agent_id is None and job.job_type == JobType.TASK:
+                raise ValueError("sub_agent_id cannot be cleared on a task job")
```
alloy-cicd, line 257: **[behavior change] Explicit `sub_agent_id: null` on a task PATCH used to be a silent no-op; now the whole update 400s.** The old `if val is not None` loop dropped the null, so a client that round-trips the full job representation (read → mutate one field → PATCH everything back — a common MCP/LLM pattern) previously had e.g. `{"enabled": false, "sub_agent_id": null}` apply the pause; now it's rejected with "sub_agent_id cannot be cleared on a task job" and *nothing* is applied.

Surfacing the invariant is defensible — but if that's the intent, it's worth a test for this round-trip shape (the three new tests cover the direct kwarg call only), and possibly a mention in the endpoint description. If not intended, ignore the null for task jobs instead of raising.

Minor: the comment on the loop below ("fields that still use old pattern") now covers only `enabled`/`max_failures`/`voice_call` — worth either migrating those three to the `_UNSET` pattern (small and mechanical, and would fix `enabled: false` being un-settable through the loop) or noting why they stay.

alloy-cicd (commented):

The backend change itself is sound: the `_UNSET` sentinel path is correctly wired through the router, the pre-existing access check at `scheduler_service.py:267-273` still covers the new branch (verified against the new tests), the `JobType.TASK` comparison is safe, and the DB CHECK constraint (ddl/034) already permits `sub_agent_id` on watch rows. The engine/runner support this UI exposes is real (`scheduler_engine.py:236-238`, `agent-runner/agent/core.py:618-660`).

**6 inline comments below** — the two most important are the stale `form.sub_agent_id` leaking from task mode into a watch, and the hidden-but-still-submitted notification message.

Additionally, three observations on files this PR doesn't touch but whose behavior it newly exposes (no inline anchor possible):

1. **Wasted LLM call per trigger** — `agent-runner/agent/core.py:636` computes `agent_message = message_text or await self._generate_watch_message(...)` *before* the `if sub_agent_id:` block at 640, which unconditionally overwrites it. Since this UI hides the message field whenever a sub-agent is set, `message_text` is empty for exactly the configuration this PR enables — so every trigger of a watch-with-sub-agent pays a full LLM generation whose output is discarded one statement later. Cheap fix in the runner: skip generation when `sub_agent_id` is set.
2. **No way to steer the sub-agent** — `core.py:643` hard-codes the watch prompt (`"Watch condition triggered. Take appropriate action based on: {last_check_result}"`) and ignores `message_text`. The create dialog hides the only free-text field a watch has, so user intent ("open a Jira ticket for the failing host") is not conveyable. Consider repurposing/relabelling the message field as the sub-agent instruction rather than hiding it — the engine already transports it.
3. **API-layer accept-and-ignore** — a watch created via MCP/API with both `sub_agent_id` and a `notification_message` gets a 200, but the message is never delivered; and `sub_agent_parameters` on a `job_type='watch'` create is silently dropped (`create_job` gates the inline branch on task). Both deserve either a validator rejection or a docstring note on `ScheduledJobCreate`, since LLM callers (e.g. the task-scheduler sub-agent) can't self-correct from a silent success. Related test gap: no create-side test for watch + `sub_agent_id` (happy or access-denied path).

aartaria:
Triage of the 🤖 automatic review (round 1), addressed in 7f2f3b55:

1. **Stale `sub_agent_id`/`voice_call` across job-type switch** — ✅ fixed: switching job type now resets both fields.
2. **Hidden notification field still submitting stale text** — ✅ fixed as suggested: the field stays visible with a "not used while a sub-agent is set" caption (matching the edit page), so nothing is submitted invisibly.
3. **Duplicated sub-agent picker** — ✅ fixed: extracted `SubAgentSelect`, now used by all four scheduler forms (task/watch × create/edit). Other pages' pickers left alone to keep the PR scoped.
4. **`sub_agent_id` sent on every watch save** — ✅ fixed: it is only included when actually changed (applied to task saves too, which had the same pre-existing behavior), so unrelated edits no longer re-run the access check or fail when the referenced agent became inaccessible.
5. **`owned_only` picker hiding shared agents → silent detach** — ✅ fixed via 3+4: `SubAgentSelect` renders a fallback "Agent #N (not in your list)" item when the current value isn't in the owned list, and an unchanged value is never resubmitted, so saving other fields can't detach it.
6. **Explicit `null` on a task PATCH now 400s instead of no-op** — ❌ not changed, working as intended: an explicit `null` on a task is an impossible request (tasks require a sub-agent), and silently ignoring it was the worse failure mode. Round-trip clients echoing a task job back send its actual id, never `null`; the behavior is covered by `test_update_clearing_sub_agent_on_task_is_rejected`.

The three out-of-diff notes (wasted LLM message generation before a sub-agent run, the hard-coded watch→agent prompt, and accept-and-ignore of `sub_agent_parameters` on watches) are real but out of scope here — they belong to the scheduled-runs follow-up work tracked from the gap analysis.

Also since round 1: helper text under the picker now spells out exactly what input the sub-agent receives (65d5ac3e).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (84c985aa30c1121182686c975907146f0239b329..7f2f3b557ee1056529ddbffd3001829406d625b1)

All previous findings are resolved and the delta introduces nothing new — no inline comments this round. TypeScript check on the new head passes (`tsc --noEmit`, console-frontend).

1. Stale `sub_agent_id`/`voice_call` leaking across job-type switch — **verified-fixed**: the job-type `onValueChange` now resets both fields in one `setForm` (SchedulerPage.tsx:461-470).
2. Hidden notification-message field still submitting stale text — **verified-fixed**: the field stays visible for watches with the "not used while a sub-agent is set" caption, matching the edit page; nothing is submitted invisibly.
3. Duplicated sub-agent picker — **verified-fixed**: extracted `SubAgentSelect`, used by all four scheduler forms; the `voice-agent` filter and `(automated)` badge now have one home. Leaving SkillsPage's copy out of scope is reasonable.
4. `sub_agent_id` sent on every watch save — **verified-fixed**: the changed-only guard (`subAgentId !== initialSubAgentId`) now covers both job types, so unrelated edits no longer re-run the access check or fail on an inaccessible agent.
5. Accessible-but-not-owned agent rendering as "None (notify only)" → silent detach — **verified-fixed**: the `unlisted` fallback item ("Agent #N (not in your list)") displays the reference, and the changed-only guard means an untouched value is never resubmitted.
6. Explicit `sub_agent_id: null` on a task PATCH 400s instead of no-op — **rejected-by-author** with reasoning (impossible request, silent-ignore is the worse failure mode, covered by `test_update_clearing_sub_agent_on_task_is_rejected`). Accepted; not re-raised.

The three out-of-diff runner/API notes (wasted LLM message generation before a sub-agent run, hard-coded watch→agent prompt, accept-and-ignore of `sub_agent_parameters` on watch creates) are acknowledged by the author as follow-up work — fine. The helper-text commit (65d5ac3e) documenting the exact sub-agent input is a nice touch.

Two cosmetic observations for whenever `SubAgentSelect` is next touched, explicitly non-blocking: with zero owned agents and `includeNone`, the dropdown shows only "None (notify only)" with no "No sub-agents found" hint; and a job that references the filtered `voice-agent` id would render as "Agent #N (not in your list)" rather than by name.

aartaria:
Scope addition per Andrea (no separate issue): the watch→sub-agent instruction is now customizable (7e4f71f7). The job's existing `prompt` column — previously task-only in practice and silently ignored on watches — carries an optional instruction: `scheduler_engine` forwards it in the watch dispatch metadata, `agent-runner` prepends it to the check result (falling back to the previous generic wording when unset), and both console forms show a "Sub-agent instruction" field once a sub-agent is selected on a watch. Engine tests cover the metadata shape. This also resolves the round-1 out-of-diff notes about the hard-coded prompt and the accept-and-ignore of `prompt` on watches.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (7f2f3b557ee1056529ddbffd3001829406d625b1..7e4f71f78cab28ee89719252a94775ba0a6a1bc6)

**Ledger of round-2 findings:** round 2 had no blocking findings — only two explicitly non-blocking cosmetic notes on `SubAgentSelect` (no empty-list hint when `includeNone`; unlisted agent shown by number). `SubAgentSelect` is untouched this round, so both remain open as non-blocking follow-up material; nothing to re-raise.

**Delta review** (the customizable watch→sub-agent instruction, 7e4f71f7): the design is sound — reusing the previously-inert `prompt` column avoids a migration, `scheduler_engine` forwards `job.prompt or None` in the watch metadata, and the runner's `(watch or {}).get("prompt")` fallback preserves the old generic wording. Verified locally on the new head: the two new engine tests pass (23 passed in `test_scheduler_engine.py`) and `tsc --noEmit` is clean. This also genuinely resolves two of the round-1 out-of-diff notes (hard-coded prompt; accept-and-ignore of `prompt` on watches). The remaining one — the wasted LLM notification-message generation before a sub-agent run (`agent-runner/agent/core.py:636`) — is still deferred to the scheduled-runs follow-up, which stands.

**One new minor finding, inline below:** `form.prompt` is now shared between the task prompt field and the watch instruction field, and the round-2 job-type reset doesn't clear it.

One non-blocking observation, not worth an inline comment: on the detail page the watch spread sends `prompt` unconditionally (unlike the changed-only guard adopted for `sub_agent_id`), so clearing the sub-agent silently retains the now-hidden instruction. Since a prompt on an agent-less watch is inert in the runner and re-attaching an agent restores it, this reads as acceptable — just noting it's a deliberate semantic, not an oversight, in case anyone expects clearing the agent to also clear the instruction.

packages/console-frontend/src/pages/SchedulerPage.tsx
```diff
@@ -939,10 +936,50 @@ function CreateJobDialog({
...
+              {form.sub_agent_id && (
+                <div className="grid gap-1.5">
+                  <Label htmlFor="watch_prompt">
+                    Sub-agent instruction{' '}
+                    <span className="text-muted-foreground text-xs">(optional)</span>
+                  </Label>
+                  <Textarea
+                    id="watch_prompt"
+                    rows={3}
+                    value={form.prompt}
```
alloy-cicd, line 970: **[minor] Task-mode prompt pre-fills the watch instruction.** `form.prompt` is now shared between the task prompt textarea (~line 781) and this watch instruction field, but the job-type reset added in round 2 clears only `sub_agent_id` and `voice_call`. Type a task prompt → switch type to watch → pick a sub-agent: the instruction field appears pre-filled with the task prompt, and submit sends it (`body.prompt`, line 420). Unlike the round-1 leak this is visible and editable, so it's minor — but the two fields have unrelated semantics ("what the task agent should do" vs "what to do when the condition fires"), and the fix is one line: add `prompt: ''` to the reset in the job-type `onValueChange`.

aartaria:
Round 3 triage: the one inline finding (task-mode prompt pre-filling the watch instruction after a job-type switch) is fixed in e09c4160 — `prompt: ''` added to the existing job-type reset, as suggested.

On the non-blocking note: yes, intentional — the detail page keeps `prompt` when the sub-agent is cleared, since it's inert without an agent and restores on re-attach, same as the notification message surviving in the other direction.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (7e4f71f78cab28ee89719252a94775ba0a6a1bc6..e09c4160820e6af5ec6438935d67c62055d245e0)

**Ledger of round-3 findings:**

1. Task-mode prompt pre-filling the watch instruction after a job-type switch — **verified-fixed**: `prompt: ''` added to the existing job-type reset (e09c4160), exactly as suggested. The delta is that single line; `tsc --noEmit` on the new head is clean.
2. Detail page retaining `prompt` when the sub-agent is cleared — **confirmed intentional** by the author (inert without an agent, restores on re-attach, symmetric with the notification message). Settled; no change needed.

Still-open non-blocking follow-up material, unchanged and already acknowledged: the two `SubAgentSelect` cosmetic notes from round 2, and the wasted LLM notification-message generation before a sub-agent run in `agent-runner/agent/core.py` (deferred to the scheduled-runs follow-up). None of these block this PR.

From this reviewer's perspective the PR has no remaining findings across all four rounds; final judgment rests with the human reviewer.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (e09c4160820e6af5ec6438935d67c62055d245e0..8b1217374d8c266ee6003f6cc147a963e06393d9)

Every previous finding is resolved and the delta introduces nothing new at a blocking level — no inline comments this round.

**Ledger:** round 4 was a clean round with zero findings, so there is nothing to classify from it. The long-running non-blocking follow-ups remain unchanged: the two `SubAgentSelect` cosmetic notes (round 2), and the wasted LLM notification-message generation before a sub-agent run — this round touches exactly that block in `core.py` (adding the timezone parameter) but the `message_text or await self._generate_watch_message(...)` still runs before the sub-agent branch overwrites its result, so that item stays deferred to the scheduled-runs follow-up as agreed.

**Delta review** (8b121737 — four latent watch-path bugs surfaced by QA over gRPC). All four fixes verified on the head:

1. **Struct metadata extraction** — `MessageToDict` is the right tool: Struct keys are map keys, not proto field names, so no camelCase mangling, and nested `watch`/`check_args` come out as plain dicts. The regression test constructs the Task the same way the gRPC path does and exercises the exact accesses that crashed; 3/3 tests pass locally.
2. **Float id coercion** — `_meta_int` correctly handles Struct doubles; the `_meta_int("scheduled_job_run_id") or ""` fallback preserves the previous default semantics.
3. **MCP client** — confirmed `langchain-mcp-adapters>=0.1.11` in agent-runner's pyproject/lock (the removed `async with` was indeed dead code walking), and the plain `get_tools()` pattern matches the existing usage in `agent_common/agents/dynamic_agent.py` and `ringier_a2a_sdk/agent/langgraph.py`.
4. **Time anchoring** — `_current_time_context` swallows unresolvable zones (covered by test), the engine now ships `job.timezone` in dispatch metadata for the tool-less LLM calls, and leaving sub-agent prompts un-anchored is the right call since agents have date tools. Engine tests: 23/23 pass.

One non-blocking observation for the follow-up list: `MessageToDict` turns *all* Struct numbers into floats, not just ids — so numeric values inside `check_args` (e.g. a `limit: 10`) reach the check tool as `10.0` over gRPC. Most MCP tool schemas coerce, but a strictly-validated integer parameter would reject; if a watch's check tool ever fails with a type-validation error, this is where to look. Not blocking — before this fix the same path crashed outright.

From this reviewer's perspective the PR remains without blocking findings; final judgment rests with the human reviewer.

Issue #150, opened 2026-08-15
client-slack: push-notification fallback is not wired end-to-end (config unsent, token mismatch, chat tasks dropped)


aartaria:
## Summary

The Slack client builds a push-notification callback for every turn — `webhookUrl`, a per-turn `webhookToken` — and persists the token on the in-flight task record for later validation. None of it works end-to-end, so the out-of-band delivery path that should rescue a turn whose SSE stream dropped does not exist in practice.

This was originally going to be fixed as a one-liner in the dropped-stream PR (send `configuration.pushNotificationConfig` in `MessageSendParams`, which is genuinely missing). A code review showed that sending it alone is *worse* than not sending it, so the one-liner was pulled back out and the work tracked here instead.

## The three gaps

1. **The config is never sent.** `A2AClientService.sendMessageStream` builds `MessageSendParams` with only `message` and `metadata`. The server is therefore never told to push, whatever the caller supplies.

2. **The token would never validate.** The caller generates a per-turn `randomUUID()` as `webhookToken`, but the callback route validates the incoming `X-A2A-Notification-Token` against the **per-installation secret** (`validateNotificationToken`, which walks `botInstallation.listAll()` and compares against each installation's stored secret). A per-turn UUID cannot match, so every push would 403 — and each rejected push costs an uncached scan of all installations.

3. **A validated chat push would still be dropped.** `handleA2ANotification` understands scheduler payloads and DM delivery. A chat turn's `Task` envelope isn't handled, so even a correctly authenticated callback for a normal conversation turn ends in a warning and no Slack message.

## What a real fix needs

- Send `configuration.pushNotificationConfig` with a token the callback route will actually accept — either use the installation secret, or extend validation to recognise per-task tokens (the in-flight record already stores one, so a per-task lookup is available and avoids the `listAll()` scan).
- Teach the notification handler to deliver a chat turn's task result into the originating thread, reusing the same delivery path as recovery so a push and a recovery poll cannot double-post.
- Decide precedence between the push callback and the polling recovery loop: both would be racing to deliver the same answer.

## Why it matters

Right now a dropped stream has exactly one safety net — the recovery loop, which polls every 5 minutes. A working push callback would turn a multi-minute silence into near-immediate delivery, and would cover the case the poller cannot: a task that completes normally while the client is disconnected.

[...]

PR #161, opened 2026-08-20
feat: correlate scheduled-run thread replies and reconstruct the run as context

### LINKED ISSUE #160: Scheduled-run thread replies: correlate to the run and adopt its conversation as a sub-conversation
## Problem

When a scheduled job run delivers its output to a chat client (Slack, Google Chat), a user reply in that thread is forwarded to the orchestrator as a **brand-new conversation** with no correlation to the job, the run, or the sub-agent that produced the output. The user cannot follow up on what the scheduled run did.

All the correlation ingredients exist but are not wired:

- The delivery path posts bare text and **discards the returned `ts` and the task's `contextId`** — `client-slack/src/handlers/a2aNotificationHandler.ts:101-104` never reads `task.contextId` and writes nothing to `context_store`.
- A thread reply finds no `context_store` row, so no `contextId` is sent and the orchestrator mints a fresh conversation (`client-slack/src/listeners/events/messageHandler.ts:634-636, :749`).
- Server-side, the run's conversation is already persisted and indexed: `scheduled_job_runs.conversation_id` (`console-backend/sqlmigrations/ddl/034_create_scheduler_tables.sql:75-92`), along with `scheduled_jobs.prompt` and `scheduled_job_runs.result_summary`.

## Why the naive fix is a trap

The scheduler dispatches **directly to agent-runner**, never through the orchestrator (`console_backend/services/scheduler_engine.py:177-183`), with a fresh contextId per run. Agent-runner checkpoints that conversation under the **bare `thread_id = contextId`, empty `checkpoint_ns`** (`agent-runner/agent/core.py:1101-1163`) — and the orchestrator checkpoints its own conversations under exactly the same key convention, in the same Postgres schema (`orchestrator-agent/app/core/executor.py:664-668`). ContextIds are accepted without ownership validation (`ringier-a2a-sdk/.../server/context_builder.py:44-56`).

⚠️ Therefore, forwarding the run's contextId to the orchestrator **as the A2A `contextId`** would make the orchestrator load agent-runner's graph state into its own graph and overwrite it — a checkpoint collision, not a continuation. The run's contextId must travel as **metadata/DataPart only**.

## Design: adopt the run as a sub-conversation of a new parent

The run's conversation is already shaped like a sub-agent conversation (bare contextId on agent-runner) — it is merely orphaned. The orchestrator's existing delegation model provides the parent/child relationship: the sub-agent contextId waterfall reuses a persisted per-agent contextId when one exists (`agent-common/agent_common/a2a/base.py:320-369`, tracked in `a2a_tracking` state, `orchestrator-agent/app/middleware/a2a_tracking.py:36-58`).

So: mint a **new parent conversation** for the thread, and reconstruct the delegation inside it as a **synthetic first turn**, so the model sees exactly what it would have seen had it dispatched the run itself:

```python
input_data = {
    "messages": [
        # role-ordering: first non-system message must be human (Bedrock/Gemini
        # reject a leading assistant tool-call; no alternation-repair in the stack).
        # The job prompt IS the human message, thinly marked as scheduled so the
        # model doesn't attribute it to the live user:
        HumanMessage(f'<scheduled_run job="{job_name}" ran_at="{ts}">{job_prompt}</scheduled_run>'),
        AIMessage(content="", tool_calls=[{"id": cid, "name": "task", "type": "tool_call",
                  "args": {"subagent_type": agent_name, "description": job_prompt}}]),
        ToolMessage(content=run_result_summary, tool_call_id=cid,
                    additional_kwargs={"a2a_metadata": {"context_id": run_context_id,
                        "is_complete": True, "state": "TASK_STATE_COMPLETED",
                        "agent_name": agent_name, "sub_agent_id": sub_agent_id}}),
        current_msg,  # the user's actual reply — must come LAST (see gotchas)
    ],
    # seed explicitly: A2ATaskTrackingMiddleware only fires when messages[-1] is a
    # ToolMessage (a2a_tracking.py:112-114), and the reply is last
    "a2a_tracking": {agent_name: {"context_id": run_context_id, "is_complete": True,
                                  "state": "TASK_STATE_COMPLETED"}},
}
```

Effects: the model sees the delegation and the prompt used; any follow-up dispatch to that sub-agent **resumes the actual agent-runner checkpoint** (full tool state, not a text replay) via the existing waterfall.

Verified gotchas:
1. **Order** — the stream handler treats everything after the last HumanMessage as "this turn" (`app/handlers/stream_handler.py:40-57`); the synthetic pair after the reply would trigger blocked-agent/phantom-completion checks. Pair goes before the reply.
2. Delegation is one shared `task` tool; the prompt arg is `description` (`app/middleware/dynamic_tool_dispatch.py:625-700, :1483-1486`).
3. Nothing re-executes historical tool calls (dispatch only in live `wrap_tool_call`), and tool-call-bearing AIMessages survive the empty-content filter — no replay hazard.

## Implementation plan

1. **agent-runner** (`agent/core.py:676-681` and failure/condition branches): add `scheduled_job_id`, `scheduled_job_run_id`, `sub_agent_id` to `result_meta` — all already in scope; `_parse_result` tolerates unknown keys.
2. **client-slack** `a2aNotificationHandler.ts`: keep the `chat.postMessage` `ts`; persist `{contextId, job_id, run_id, sub_agent_id}` keyed `{teamId}:{channelId}:{ts}` (context_store row with a kind marker, and/or Slack message metadata).
3. **client-slack** `messageHandler.ts`: on a thread reply that resolves to a scheduled-run row, send the provenance as a **DataPart/metadata** (pattern: `console_backend/services/debug_agent_service.py:144-155`) — never as the request `contextId`. Store the orchestrator's returned contextId as usual for subsequent turns (already works, `:784`).
4. **console-backend / payload**: the orchestrator has no DB access to the scheduler tables (different schema) and `list_runs` is not MCP-exposed, so `prompt` + `result_summary` must ride along — simplest is to include them in the push payload so client-slack can forward them, or expose a scoped lookup.
5. **orchestrator-agent** (`app/core/agent.py` ~`:423`): when the incoming message carries scheduled-run provenance, build the synthetic-turn `input_data` above.
6. Optional v1 fallback: with steps 1–3 only, the existing `<thread_context>` replay (`messageHandler.ts:661-676`) already gives the new conversation the run's output text — machine correlation without checkpoint adoption.

## Caveats

- DM threads only for Slack (channel replies without @mention fire no listener, `listeners/events/directMessage.ts:40-42`).
- Overlaps with #150 (callback handler rejects non-scheduler payloads) — coordinate.
- Google Chat client has the identical gap; same design applies.
- Per-run-fresh contextId is a scheduler choice, not a constraint — `dispatch_streaming` accepts a caller-supplied `context_id`; a stable per-job conversation is an alternative worth considering.

aartaria:
closes #160

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — No. All new payload fields, deps, and the DataPart are optional; the new `scheduled_run_store` table ships as migration 008.

## What

A reply under a delivered scheduled-run notification previously reached the orchestrator as a brand-new conversation with no link to the job, run, or sub-agent that produced the output (#160).

**agent-runner** — `result_meta` now carries `scheduled_job_id`, `scheduled_job_run_id`, `sub_agent_id`, the sub-agent's name, and the dispatched prompt, so the push payload the delivery channel receives has full run provenance. (`_parse_result` in console-backend reads only known keys, so no change needed there.)

**client-slack / client-google-chat** — the notification handlers persist that provenance (including `scheduler_status`/`error_message`) keyed by the delivered message (Slack: `{channelId}:{messageTs}`, migration 008; Google Chat: the thread resource name, migration 006; both 30-day TTL). Thread replies under a notification forward it as a conversation-origin DataPart. ⚠️ The run's contextId is deliberately **never** sent as the request `contextId`: it names the sub-agent's own conversation on agent-runner, and both services checkpoint under the same bare-thread-id convention in the same schema, so reusing it would load/overwrite the sub-agent's graph state.

**orchestrator-agent** — on the first turn of a conversation opened by such a reply (empty checkpoint), it prepends synthetic history: the job prompt as the human request, the `task` tool call (raw sub-agent name), and the run's output — or the failure — as the tool result. Watch-only runs inject the delivered notification text. The model answers the reply knowing exactly what ran, what it was asked, and what it produced.

**Scope note (from review round 1):** an earlier revision also seeded `a2a_tracking` with the run's contextId to make follow-up delegations resume the run's sub-agent conversation. That seed was dropped — the run's checkpoint is not reachable from any of the orchestrator's delegation paths (local delegations run in-process on a different thread key; remote/foundry never received the run's contextId), and the seed desynchronized the HITL resume probe. True cross-service conversation adoption is tracked separately in the follow-up issue.

## Testing

- client-slack: `tsc --noEmit` clean; 67 jest tests pass (4 for the notification handler, via a small `slackClientFactory` test seam).
- agent-runner: 14 pytest pass (success-path test asserts the provenance fields).
- orchestrator-agent: 635 pytest pass (9 new in `test_scheduled_run_adoption.py`, incl. failure rendering, tag-escaping, watch-only injection).
- client-google-chat: `tsc --noEmit` clean; 22 jest tests pass (5 new for the notification handler).
- eslint could not run locally (`@alloy-ch/eslint-config-alloy` needs the private registry) — environment limitation, CI covers it.

## Out of scope

Slack channel replies without @mention fire no listener; Google Chat correlation requires an actual in-thread reply (flat DM messages open new threads). Coordinate with #150 (callback handler payload types). Cross-service conversation adoption (resuming the run's agent-runner checkpoint from orchestrator delegations) is a follow-up.


packages/orchestrator-agent/app/core/agent.py
```diff
@@ -84,6 +84,92 @@
...
+        tool_calls=[
+            {
+                "id": tool_call_id,
+                "name": "task",
+                "type": "tool_call",
+                "args": {"subagent_type": sub_agent_name, "description": prompt},
+            }
+        ],
+    )
+    tracking_entry = {
```
alloy-cicd: **[critical] The `a2a_tracking` seed cannot resume the run's conversation for any sub-agent type, and it breaks HITL resume.**

The docstring says the seed makes the next delegation "resume the run's own agent-runner conversation". Tracing the actual paths:

- **Local/automated sub-agents:** agent-runner checkpointed the scheduled run under `thread_id = <raw context_id>` in *agent-runner's own* Postgres (`packages/agent-runner/agent/core.py:1168-1171`). On the follow-up delegation, `LocalA2ARunnable._instrument` consumes the seed (`agent_common/a2a/base.py:904-922`) and runs the dynamic agent on `"{run_context_id}::dynamic-{name}"` in the *orchestrator's* checkpointer (`dynamic_agent.py:376-385`). Different database **and** different thread key → the sub-agent starts blank.
- **Remote sub-agents:** agent-runner never propagates the run task's `context_id` to them — `_run_remote_agent` (`core.py:1089-1096, 1535-1539`) builds `SubAgentInput` without `a2a_tracking`/`context_id`, so the remote agent mints its own contextId which is discarded. The `context_id` stored in provenance is the *scheduler task's* context, which names no conversation on the remote agent; the seed then sends this alien id as `Message.context_id`.
- **Foundry sub-agents:** receive no context either, and the seed carries no `foundry_session_rid`.

**Worse, the seed introduces a HITL regression for local agents:** the dispatch middleware's HITL checkpoint probe and Command-resume path use the *orchestrator conversation id* thread (`dynamic_tool_dispatch.py:1688-1690, 1726-1744`; `base.py:959-963` skips `_instrument` on the Command path), while the graph now *runs* on the seeded `"{run_ctx}::dynamic-{name}"` thread. An interrupt raised in an adopted conversation is checkpointed under the seeded thread, the probe looks under the orchestrator thread, finds nothing, and re-invokes the sub-agent fresh — the user's approval is lost. Without the seed both sides agree via the waterfall fallback, so this divergence is introduced by this PR.

The synthetic-history injection alone (which genuinely gives the model the run's prompt/output) does the useful work here. Suggest dropping the `a2a_tracking` seed entirely until there is a real cross-service resume story, or scoping it to a path that provably resumes something.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -84,6 +84,92 @@
...
+        )
+    )
+    ai_msg = AIMessage(
+        content="",
+        tool_calls=[
+            {
+                "id": tool_call_id,
+                "name": "task",
+                "type": "tool_call",
+                "args": {"subagent_type": sub_agent_name, "description": prompt},
```
alloy-cicd: **[bug] Space-stripped `subagent_type` in the synthetic tool call is not a valid registry key for local/foundry sub-agents.**

`subagent_registry` is keyed by the *raw* config name for local and foundry agents (`app/utils.py:305, 499` — `subagent_registry[config.name]`); only remote agents are registered stripped (`discovery.py:166`). For a sub-agent named `"Report Agent"` the injected history shows `task(subagent_type="ReportAgent")`; when the model mimics that on the next turn, `subagent_registry.get("ReportAgent")` misses and **silently falls back to the general-purpose handler** (`dynamic_tool_dispatch.py:1489-1497, 1606-1616`) instead of delegating to the intended agent.

Note the stripped form *is* correct for the tracking-seed key (that's what `base.py:341` reads), so the fix is to use the raw name in the tool-call `args` while keeping the stripped name for the seed — or better, resolve the registry key from `sub_agent_id`, which is already carried in the provenance but currently unread.

(Related pre-existing wrinkle noticed while verifying: `foundry_agent.py:323` reads `a2a_tracking.get(self.name)` *unstripped* for `foundry_session_rid`, inconsistent with `base.py:341`.)

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -416,6 +502,31 @@ async def stream(
...
+            # trailing synthetic pair would trip its blocked-agent heuristics.
+            # Never inject into a conversation that already has history — the
+            # provenance is only meaningful for the reply that opened it.
+            scheduled_run = _extract_scheduled_run_provenance(message_parts)
+            if scheduled_run and not checkpoint_msgs:
+                synthetic = _build_scheduled_run_history(scheduled_run)
+                if synthetic:
+                    synthetic_msgs, tracking_seed = synthetic
+                    input_data = {
+                        "messages": [*synthetic_msgs, current_msg],
```
alloy-cicd: **[bug] The injected run output can never be surfaced via `include_subagent_output`, and trying costs a wasted corrective round.**

The comment above correctly orders the synthetic turn before `current_msg` to avoid the blocked-agent heuristics — but the same ordering makes the synthetic `task` ToolMessage invisible to `_extract_recently_called_subagents`, which scans only messages *after the last HumanMessage* (`stream_handler.py:41-57, 73-93`).

`FinalResponseSchema` coaches the model to set `include_subagent_output=true` with `message=""` when "the most recent sub-agent response should be appended" (`app/models/schemas.py:66-92`) — which is exactly what a reply like "send me the full report" under a notification invites, with the synthetic ToolMessage as the most recent sub-agent response. The phantom-completion check then fires (`stream_handler.py:128-137`), the response is discarded, and the graph is re-entered with the delegation nudge (`executor.py:820-838`) — typically forcing a real re-delegation that (per the seed issue) starts a blank sub-agent conversation.

Either exempt the synthetic tool_call_id (`scheduled_run_*` prefix is recognizable) from the phantom check, or instruct via the synthetic content that its output must be restated, not included by reference.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -84,6 +84,92 @@
...
+
+    Returns None when the provenance lacks a sub-agent or contextId (e.g. a
+    watch notification that ran no sub-agent) — there is no delegation to
+    reconstruct in that case.
+    """
+    # Same normalization as the delegation waterfall's tracking key
+    # (agent_common.a2a.base: ``self.name.replace(" ", "")``).
+    sub_agent_name = (run.get("sub_agent_name") or "").replace(" ", "")
+    run_context_id = run.get("context_id")
+    if not sub_agent_name or not run_context_id:
```
alloy-cicd: **[bug] Watch-only runs (no sub-agent) silently lose the one piece of context the user is replying to.**

For a watch job with no `sub_agent_id`, agent-runner sends `sub_agent_name=None, prompt=None` but a real `agent_message` (`core.py:646-649, 681-683`); client-slack persists and forwards it as `result_summary`. This guard then returns `None` (the failing condition is `sub_agent_name` — `context_id` is present, it's the scheduler dispatch context), the DataPart is never rendered into text (`content_builder.py:222-224` skips data parts at debug level), and the model answers the reply blind to the notification it sits under.

`result_summary` alone is enough to inject a synthetic context note (no delegation turn needed) — worth handling this branch rather than dropping it, since "why did this trigger?" under a watch notification is presumably the most common follow-up.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -84,6 +84,92 @@
...
+        # MessageToDict renders protobuf numbers as floats; ids must not print as "42.0".
+        return str(int(value)) if isinstance(value, (int, float)) else ""
+
+    job_id = _fmt_id(run.get("scheduled_job_id"))
+    run_id = _fmt_id(run.get("scheduled_job_run_id"))
+    tool_call_id = f"scheduled_run_{run_id or run_context_id}"
+
+    human_msg = HumanMessage(
+        content=(
+            f'<scheduled_run source="scheduler" job_id="{job_id}" run_id="{run_id}">'
```
alloy-cicd: **[security/robustness] `prompt` and `result_summary` are interpolated unescaped, and the whole `scheduled_run` dict is trusted from the client.**

Two layers:

1. Sub-agent output routinely summarizes untrusted content (watch jobs even fold `json.dumps(last_check_result)` into the prompt — `agent-runner/core.py:691-693`). Text containing `</scheduled_run>` followed by instructions escapes the frame and reads as first-person user input on turn 1 — a promotion from tool-channel to human-channel text. At minimum escape/strip the closing tag in `prompt`, and consider the same for `result_summary` in the ToolMessage.
2. Nothing in `stream()` validates the DataPart: any authenticated A2A client can fabricate `scheduled_run` with an arbitrary `context_id`, which gets seeded into `a2a_tracking` and forwarded as `Message.context_id` on the next dispatch — and agent-runner keys checkpoint threads by `context_id` alone with no per-user scoping (`core.py:1166-1171`). Requires a valid JWT and a leaked/guessed UUID, so moderate severity, but a server-side ownership check (resolve job/run id for the authenticated `user_sub`) would close it.

Minor, same function chain: `_extract_scheduled_run_provenance` runs on every turn and `MessageToDict`-converts every DataPart, though the result is only usable when `not checkpoint_msgs` — swap the two checks at line 517-518 to make it first-turn-only.

alloy-cicd (commented):

The Slack-side plumbing (store, migration, DataPart forwarding) is solid and well-commented, but the orchestrator-side adoption mechanism has a fundamental problem: **the `a2a_tracking` seed does not resume the run's conversation for any sub-agent type, and it actively breaks HITL resume in adopted conversations** (details in the inline comment on `agent.py:160`). The synthetic-history injection itself is a nice approach; several edge cases around it need attention (space-stripped names, watch-only runs, phantom-delegation interaction, unescaped interpolation).

**Findings not anchorable to the diff:**

- `packages/client-google-chat/src/handlers/a2aNotificationHandler.ts` still parses the old 3-field `SchedulerPayload`, so correlation is Slack-only even though agent-runner now emits the fields for every channel. If that's intentional scope, consider a follow-up ticket; the payload mirrors (one Python dict, N hand-copied TS interfaces) will keep drifting otherwise.
- Pre-existing (not introduced here, but adjacent to the changed lines): a failed non-watch run yields `agent_message: null`, so `chat.postMessage({text: null})` throws and is swallowed by the outer catch — the user never learns the job failed. Worth fixing while in this file.
- `scheduledRunStore` is threaded as optional (`?`) through `registerListeners`/`registerMessageListeners`/`HandlerDependencies` with silent `&& scheduledRunStore` guards, although `StorageProvider.scheduledRun` is abstract-required and the only caller always supplies it. A forgotten wire-up disables the feature silently instead of failing to compile.
- Both root and orchestrator `AGENTS.md` ask for non-obvious cross-cutting design decisions to be documented. This mechanism spans three packages and has two invariants stated only in inline comments (run contextId must never become the request contextId; synthetic turn must precede `current_msg`). A short section in `packages/orchestrator-agent/AGENTS.md` would save the next reader a three-package traversal.

All findings were adversarially verified against the checked-out PR head; two initial candidate findings were refuted during verification and dropped.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -635,6 +643,37 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
     const existingContextId = existingContext?.contextId;
 
+    // Reply under a delivered scheduled-run notification: on the FIRST reply
+    // (no orchestrator conversation yet for this thread) forward the run's
+    // provenance as a DataPart so the orchestrator can reconstruct the
+    // delegation and adopt the run's sub-agent conversation. The run's
+    // contextId is deliberately NOT sent as the request contextId — it names
+    // the sub-agent's own conversation, not an orchestrator one.
+    let scheduledRunDataPart: Record<string, unknown> | undefined;
+    if (!existingContextId && threadTs !== messageTs && scheduledRunStore) {
```
alloy-cicd: **[bug] A failed first reply permanently loses the correlation.**

`contextStore.set(contextKey, contextId)` runs as soon as the A2A `task` event arrives (line 823), and the orchestrator emits that event *before* running the graph (`executor.py:357-358`). If the first reply's turn then dies before any checkpoint is written (model error, stream abort, restart), the retry sees `existingContextId` truthy → no store lookup → no DataPart; the orchestrator side (`agent.py:518`, gated on `not checkpoint_msgs`) would still have accepted it, but never gets it. No log, no recovery path for that thread.

The client gate and the orchestrator gate are trying to express the same condition ("first real turn") from different vantage points. Safer: always attach the DataPart for thread replies that have a provenance row, and let the orchestrator's `not checkpoint_msgs` gate decide — it's the side that actually knows whether history exists.

packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -635,6 +643,37 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
+      try {
+        const runRecord = await scheduledRunStore.get(scheduledRunStore.buildKey(teamId, channelId, threadTs));
```
alloy-cicd: **[nit]** `scheduledRunStore.buildKey(teamId, channelId, threadTs)` is byte-identical to the `contextKey` computed at line 642 (both stores use the `{teamId}:{channelId}:{ts}` template). Passing `contextKey` here — or delegating one `buildKey` to the other / a shared helper — keeps the two key formats from drifting apart; a silent divergence would just make every lookup miss with no error.

packages/client-slack/src/handlers/a2aNotificationHandler.ts
```diff
@@ -98,14 +110,38 @@ export async function handleA2ANotification(
...
       `[A2ACallback] Sent notification to user ${userAuth.userId} in team ${botInstallation.teamId}`
     );
+
+    // Persist the run's provenance keyed by the delivered message, so a thread
+    // reply under it can be correlated to the scheduled job/run and forwarded
+    // to the orchestrator as structured data (see messageHandler).
+    if (scheduledRunStore && postResult.ts && task.contextId) {
+      try {
+        await scheduledRunStore.set({
+          contextKey: scheduledRunStore.buildKey(botInstallation.teamId, dmResult.channel.id, postResult.ts),
```
alloy-cicd: **[bug] Writer and reader key on different team-id sources.**

The write key uses `botInstallation.teamId` (installation record); the read key in `messageHandler.ts:653-655` uses the event's team id (`ev.team || body.team_id || ''` — `directMessage.ts:75`). Every other store is internally consistent on event-side ids; this is the only one mixing sources. Under Enterprise Grid / org-wide installs (or when `ev.team` is absent, yielding an empty segment) the two differ and `get()` silently returns null — the reply just loses its scheduled-run context with nothing logged.

Since the DM channel id + message ts already uniquely identify the message, the safest fix is to key both sides off the same source (e.g. use the event's team on neither side, or store under `botInstallation.teamId` and have the reader resolve the installation the same way this handler does).

packages/client-slack/src/handlers/a2aNotificationHandler.ts
```diff
@@ -98,14 +110,38 @@ export async function handleA2ANotification(
...
     });
 
     logger.info(
       `[A2ACallback] Sent notification to user ${userAuth.userId} in team ${botInstallation.teamId}`
     );
+
+    // Persist the run's provenance keyed by the delivered message, so a thread
+    // reply under it can be correlated to the scheduled job/run and forwarded
+    // to the orchestrator as structured data (see messageHandler).
+    if (scheduledRunStore && postResult.ts && task.contextId) {
```
alloy-cicd, line 132: **[bug] Failed runs are stored as provenance and later presented to the model as successful completions.**

Only `condition_not_met` is filtered (line 75); `scheduler_status` is available here but never checked or stored, and `error_message` from the failed payload (`agent-runner/core.py:714`) is dropped — `ScheduledRunRecord`/the DDL have no status or error column. The reachable case is a **watch job whose sub-agent fails**: `agent_message` (the watch text) is set before the sub-agent runs, so the DM posts and provenance is stored; a thread reply then makes `_build_scheduled_run_history` emit a synthetic delegation hardcoded to `TASK_STATE_COMPLETED`/`is_complete: True` with the watch text as the "tool result" — the orchestrator asserts the run succeeded and answers from a run that errored.

Suggest storing `scheduler_status` (and `error_message`) in the record and having the orchestrator reflect failure in the synthetic ToolMessage/state.

packages/client-slack/sqlmigrations/ddl/008_scheduled_run_store.sql
```diff
@@ -0,0 +1,25 @@
+-- rambler up
+
+-- Provenance of scheduled-run notifications delivered to Slack.
+-- Keyed by the delivered message ({teamId}:{channelId}:{messageTs}), so a
+-- thread reply under a notification can be correlated back to the scheduled
+-- job, run, and sub-agent conversation that produced it. The run's A2A
+-- context_id is the sub-agent's own conversation and is forwarded to the
+-- orchestrator as structured data only — never as the request contextId.
+create table scheduled_run_store (
```
alloy-cicd, line 15: **[retention] Unbounded growth: one row per delivered notification, holding the full `prompt` and the entire delivered agent output, forever.**

Rows are only ever read on the *first* thread reply, typically within hours of delivery, yet there's no `expires_at`, no index for cleanup, and `cleanup_expired_records()` (001) isn't extended. With per-user recurring jobs this grows monotonically with full-text blobs of user-visible content. (`context_store` also lacks TTL, so the codebase pattern is mixed — but those rows are small; these carry the whole agent message.)

Suggest `expires_at timestamptz` (e.g. `now() + interval '30 days'`), an index on it, and extending `cleanup_expired_records()` as 001 does for `oauth_state`/`inflight_tasks`.

packages/agent-runner/agent/core.py
```diff
@@ -704,6 +715,9 @@ def _meta_int(key: str) -> int | None:
                     "last_check_result": last_check_result,
                     "agent_message": agent_message,
                     "user_sub": user_config.user_sub,
+                    "sub_agent_name": sub_agent_name,
```
alloy-cicd, line 739: **[dead field]** `sub_agent_name` is provably always `None` here: it is initialized `None` at line 648 and only assigned by the `_execute_sub_agent` tuple unpack — the single statement inside the `try` whose failure lands in this `except`. So every failed payload serializes `"sub_agent_name": null`, and downstream consumers may build on a field that's never populated. Either drop it from the failure dict (it already has `error_message` + the correlation ids) or resolve the name before the `try` if failed runs genuinely need it.

packages/client-slack/src/storage/types.ts
```diff
@@ -117,6 +117,34 @@ export interface IContextStore {
...
+ * orchestrator as structured data only, NEVER as the request contextId
+ * (the orchestrator checkpoints under the same key convention, so reusing it
+ * as a conversation id would collide with the sub-agent's state).
+ */
+export interface ScheduledRunRecord {
+  contextKey: string;
+  contextId: string;
+  scheduledJobId?: number;
+  scheduledJobRunId?: number;
+  subAgentId?: number;
```
alloy-cicd, line 135: **[nit]** `subAgentId` travels payload → column → record → DataPart, but nothing reads it: `_build_scheduled_run_history` uses only `sub_agent_name`, `context_id`, `prompt`, `result_summary` and the job/run ids. Either drop it end-to-end, or — better, given the registry-key comment on `agent.py:156` — make it the field the orchestrator actually resolves the sub-agent by, which would give it a purpose and fix the name-matching fragility at once.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 2237e094 by **dropping the seed entirely** — confirmed against the code: agent-runner passes the run contextId only to `_run_langgraph_agent` (never remote/foundry), and local delegations run in-process on `{ctx}::dynamic-{name}` in the orchestrator's checkpointer, so the seed resumed nothing anywhere and the HITL probe divergence was a real regression. The synthetic-history injection is kept as the mechanism that does the useful work. True cross-service adoption is now tracked as a follow-up issue.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 2237e094: the synthetic tool call now uses the **raw config name**, matching the local/foundry registry keys. With the tracking seed dropped there is no stripped-key consumer left. Resolving via `sub_agent_id` would need an id→name lookup against console-backend that the orchestrator doesn't have today — noted in the follow-up issue.

packages/orchestrator-agent/app/core/agent.py
aartaria: Addressed in 2237e094 via coaching rather than a structural exemption: the synthetic frame now instructs the model to restate the run output and not use `include_subagent_output`. A phantom-check exemption alone would make it worse — `parse_agent_response`'s include-extraction is also current-turn-only, so an exempted completion would surface an *empty* response instead of the corrective nudge. The nudge stays as the backstop; worst case remains one corrective round.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 2237e094: a provenance record without a sub-agent now injects a single framing HumanMessage carrying the delivered notification text (`result_summary`), so the model knows what the user is replying to; the delegation reconstruction is skipped since none happened.

packages/orchestrator-agent/app/core/agent.py
aartaria: Partially fixed in 2237e094: `</scheduled_run` is neutralized in all text interpolated into the frame (prompt and watch-notification summary; the ToolMessage content is tool-channel and stays raw, as with any sub-agent output). Extraction is now gated behind `not checkpoint_msgs` so it runs on first turns only. Server-side ownership validation is deliberately deferred: with the tracking seed dropped, a fabricated DataPart can now only inject *text* into the fabricator's own conversation — the same trust level as the client-composed `<thread_context>` replay that already exists. Flagged in the follow-up issue for when adoption makes contextId meaningful again.

packages/client-slack/src/listeners/events/messageHandler.ts
aartaria: Fixed in 2237e094 exactly as suggested: the client attaches the DataPart on every thread reply that has a provenance row, and the orchestrator's `not checkpoint_msgs` gate — the side that knows whether history exists — decides whether to inject.

packages/client-slack/src/listeners/events/messageHandler.ts
aartaria: Resolved by the team-id fix (2237e094): `scheduled_run_store` is now keyed `{channelId}:{messageTs}` — intentionally a different format from `context_store`'s team-scoped keys, so the two no longer look interchangeable.

packages/client-slack/src/handlers/a2aNotificationHandler.ts
aartaria: Fixed in 2237e094: the key no longer contains a team id at all — `{channelId}:{messageTs}`, one id source on both sides (the DM channel id + ts already uniquely identify the message).

packages/client-slack/src/handlers/a2aNotificationHandler.ts
aartaria: Fixed in 2237e094: `scheduler_status` and `error_message` are persisted and forwarded, and the orchestrator renders failed runs with `status=\"failed\"` in the frame and a `TASK_STATE_FAILED` ToolMessage that carries the error and marks any delivered watch text as partial output.

packages/client-slack/sqlmigrations/ddl/008_scheduled_run_store.sql
aartaria: Fixed in 2237e094: `expires_at` (30 days) with an index, `cleanup_expired_records()` extended (and restored in the down-migration), and `get()` filters expired rows so correctness doesn't depend on the cleanup cadence.

packages/agent-runner/agent/core.py
aartaria: Fixed in 2237e094: `_fetch_sub_agent_config` is hoisted into `_stream_impl` (inside the try), so the failure payload carries the real `sub_agent_name` whenever the config fetch succeeded — which also serves the failed-watch provenance case from the client-slack finding.

packages/client-slack/src/storage/types.ts
aartaria: Kept deliberately, with a purpose now documented: it is persisted as correlation metadata and is the natural resolver key for the cross-service adoption follow-up (the orchestrator's registry is name-keyed today, so resolving by id would require a console-backend lookup this PR shouldn't grow). If you'd rather trim it until then, happy to drop it end-to-end.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (02d9c9a72b5bfcd4e3d745fd0c83b838c839c610..1f027a5ca17ca790db707192300cdf72218e5016)

**Ledger of round-1 findings** (each re-verified against the current head, not taken on faith):

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | `a2a_tracking` seed (critical) | **verified-fixed** — seed fully removed; injection is messages-only; no tracking consumer can act on the synthetic ToolMessage's remaining `a2a_metadata` (no `context_id`/`task_id` in it) |
| 2 | Space-stripped `subagent_type` | **verified-fixed** — raw config name used; no `.replace(" ", "")` remains in `agent.py` |
| 3 | `include_subagent_output` phantom round | **verified-fixed** — coaching sentence present in the synthetic frame; the author's reasoning against a phantom-check exemption is correct (the include-extraction is also current-turn-only, so an exemption alone would surface an empty response) |
| 4 | Watch-only runs lose context | **verified-fixed** — framing HumanMessage with `result_summary`; the empty-summary edge is unreachable in practice since provenance is only written after a successful post |
| 5 | Unescaped interpolation / trusted DataPart | **verified-fixed** (escaping + first-turn gating); server-side ownership validation **rejected-by-author with sound reasoning** — with the seed gone a fabricated DataPart only injects text into the fabricator's own conversation; accepted, tracked in the follow-up issue |
| 6 | Failed first reply loses correlation | **verified-fixed** — DataPart attached on every thread reply with a row; orchestrator's `not checkpoint_msgs` gate decides |
| 7 | Duplicated `buildKey` | **verified-fixed** — store now keyed `{channelId}:{messageTs}`, deliberately distinct format |
| 8 | Writer/reader team-id asymmetry | **verified-fixed** — no team id in the key; single id source both sides; interface signature change propagated to all call sites |
| 9 | Failed runs presented as success | **verified-fixed** — `scheduler_status`/`error_message` persisted, forwarded, and rendered as `status="failed"` frame + `TASK_STATE_FAILED` ToolMessage |
| 10 | No TTL on `scheduled_run_store` | **verified-fixed** — `expires_at` + index, `cleanup_expired_records()` extended; down-migration correctly restores the 001 version of the function; `get()` filters expired rows |
| 11 | Always-null `sub_agent_name` in failure payload | **verified-fixed** — config fetch hoisted before execution; agent-runner test suite passes (11 tests) |
| 12 | Unread `subAgentId` | **rejected-by-author** — kept deliberately as the resolver key for the cross-service adoption follow-up, with documentation; accepted |

**Delta review:** the generalization to `urn:nannos:a2a:conversation-origin:1.0` is internally consistent — writer/reader wire shape agrees field-for-field (verified through the actual proto `Value` round-trip, no camelCase mangling), the kind registry handles unknown kinds gracefully, and `test_conversation_origin.py` exercises the real deserialization path (12 tests pass). No bug-severity findings; four minor inline comments below.

**Non-blocking observations** (no inline comments): (1) the origin DataPart is honored without `X-A2A-Extensions` activation, unlike the feedback/activity-log extensions which are gated on `requested_extensions` — defensible leniency, but an inconsistency with the repo's own extension handling, and client-slack doesn't advertise the new URI alongside the five it already sends; (2) the TS↔Python wire shape is pinned only by convention plus the `a2a_extensions.py` docstring — a future key rename on either side degrades silently to the builder's fallback strings; (3) the store upsert doesn't refresh `expires_at` on conflict (practically irrelevant for ts-keyed rows).

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -416,6 +571,28 @@ async def stream(
...
+            # HumanMessage as "this turn", and a trailing synthetic pair would
+            # trip its blocked-agent heuristics. Never inject into a
+            # conversation that already has history — the origin is only
+            # meaningful for the message that opened it (clients attach it on
+            # every message of its context precisely so a first turn that failed
+            # before any checkpoint was written still gets it on retry).
+            if not checkpoint_msgs:
+                origin = _extract_conversation_origin(message_parts)
+                if origin:
+                    synthetic_msgs = _build_origin_history(origin)
```
alloy-cicd: **[robustness]** The builder dispatch runs outside the `try:` below with no exception guard. A malformed origin descriptor — e.g. a buggy client sending `prompt` or `result_summary` as a number (MessageToDict yields a float, which is truthy, so the `or`-fallback doesn't save it) — raises `AttributeError` in `_scheduled_run_frame_text` and fails the entire turn. That contradicts the contract `_build_origin_history`'s own docstring states (optional enrichment, skipped rather than an error) and which unknown kinds already honor. A `try/except` around `_build_origin_history(origin)` that logs and skips injection would make malformed descriptors degrade the same way unknown kinds do.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -84,6 +84,161 @@
...
+
+
+def _scheduled_run_frame_text(text: str) -> str:
+    """Neutralize a closing tag inside text interpolated into the <scheduled_run> frame.
+
+    Run prompts/outputs routinely contain untrusted content; a literal
+    ``</scheduled_run>`` would escape the frame and read as first-person user
+    input on the conversation's first turn.
+    """
+    return text.replace("</scheduled_run", "<\\/scheduled_run")
```
alloy-cicd: **[nit]** The escape is case-sensitive: `</SCHEDULED_RUN>` (or `</Scheduled_Run>`) in untrusted run output passes through verbatim, and models treat XML-ish tags case-insensitively, so the frame can still be visually closed. A case-insensitive replace (e.g. `re.sub(r'(?i)</(scheduled_run)', r'<\\/\1', text)`) closes the remaining gap.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -84,6 +84,161 @@
...
+                    "they are now replying to it."
+                )
+            )
+        ]
+
+    prompt = prompt or "Execute your configured task."
+    if failed:
+        tool_content = f"The scheduled run FAILED: {error_message or 'unknown error'}"
+        if result_summary:
+            tool_content += f"\nPartial output delivered to the user: {result_summary}"
```
alloy-cicd: **[nit]** For a failed run, `result_summary` is what client-slack stored from `agent_message` — the user-facing failure *notification* text, not partial task output. Labeling it "Partial output delivered to the user" invites the model to restate notification prose as if it were actual run output. "Notification delivered to the user:" (or similar) would be accurate.

packages/client-slack/sqlmigrations/ddl/008_scheduled_run_store.sql
```diff
@@ -0,0 +1,60 @@
...
+-- under Enterprise Grid), so a thread reply under a notification can be
+-- correlated back to the scheduled job, run, and sub-agent conversation that
+-- produced it. The run's A2A context_id is the sub-agent's own conversation
+-- and is forwarded to the orchestrator as structured data only — never as the
+-- request contextId.
+--
+-- Rows are only read by the first replies under a notification, so they expire:
+-- cleanup_expired_records() removes them, and reads filter on expires_at.
+create table scheduled_run_store (
+    context_key text not null primary key,  -- {channelId}:{messageTs}
```
alloy-cicd, line 16: **[observation]** This migration was created earlier in this same PR and is now edited in place (key format change, three new columns) rather than adding an 009 — fine as long as no environment applied the round-1 version of 008. If any dev/preview deployment ran the branch mid-PR, the migration runner will see a changed applied migration and rows written under the old `{teamId}:{channelId}:{ts}` key become unreachable under the new two-part `buildKey`. Worth a quick check before merge; if nothing deployed the branch, ignore.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 861ac33b: `_build_origin_history` wraps the builder in try/except and degrades to no-injection with a warning log — malformed descriptors now behave exactly like unknown kinds, per the docstring's contract. Covered by a test (numeric `prompt`).

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 861ac33b: the escape is now `re.sub(r'(?i)</(scheduled_run)', ...)` — case-insensitive, with a test asserting `</SCHEDULED_RUN>` is neutralized.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 861ac33b: the label is now "Notification delivered to the user:" with a comment explaining that a failed run's result_summary is the failure notification text, plus a test pinning the wording.

packages/client-slack/sqlmigrations/ddl/008_scheduled_run_store.sql
aartaria: Checked and discarded: the branch was never deployed anywhere — migration 008 has only ever existed on this PR branch and no environment ran the round-1 version, so editing it in place is safe. Agreed that after merge any change would need an 009.

packages/client-google-chat/src/handlers/a2aNotificationHandler.ts
```diff
@@ -81,11 +89,39 @@ export async function handleA2ANotification(
...
 
     logger.info(
       `[A2ACallback] Sent notification to user ${userAuth.userId} in space ${dmSpace.name}`
     );
+
+    // Persist the run's provenance keyed by the delivered message's thread, so
+    // a thread reply under it can be correlated to the scheduled job/run and
+    // forwarded to the orchestrator as a conversation-origin DataPart (see
+    // messageHandler).
+    const threadName = sentMessage.thread?.name;
```
alloy-cicd, line 108: **[observation — please verify live before merge]** The code-level identity is consistent (writer keys by the create-response `thread.name`, reader keys by the reply event's `message.thread?.name`, `buildKey` is the identity on both ends), but the whole feature rests on a Google Chat platform behavior the repo can't prove: that a user's reply in a **1:1 DM space** arrives carrying the *notification message's* thread resource name. Slack's equivalent works because `thread_ts == root ts` is a documented invariant; Google Chat 1:1 DMs are historically unthreaded (no reply-in-thread UI; new messages can get fresh thread resources), and DMs are the only surface notifications are delivered to (`findDirectMessage` above). If replies arrive with a different `thread.name`, every lookup silently misses and the feature is dead on its only delivery surface — no error anywhere.

A single manual smoke test settles it: reply under a delivered notification and compare the `[A2ACallback] Stored... thread=` log value with the reply event's `threadId`. Worth doing once before merge (and perhaps noting the result in the PR description).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (1f027a5ca17ca790db707192300cdf72218e5016..de7a59d311e46517b49b723e4bb4577c0780317b)

**Ledger of round-2 findings** (each re-verified against the current head; `test_conversation_origin.py`: 15 passed):

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Unguarded origin-builder dispatch | **verified-fixed** — `except Exception` (correctly not bare) → warn + no-injection; the new malformed-descriptor test genuinely exercises the except path |
| 2 | Case-sensitive frame escape | **verified-fixed** — verified empirically against `</SCHEDULED_RUN>` and `</Scheduled_Run>`; backreference preserves casing while breaking the closing form; no double-escaping. (Only the theoretical `</ scheduled_run` whitespace variant remains, which models don't parse as a closing tag — fine to leave) |
| 3 | "Partial output" mislabel on failed runs | **verified-fixed** — relabeled with explanatory comment; test pins the wording |
| 4 | Migration 008 edited in place | **resolved** — author confirmed the branch was never deployed anywhere; accepted |

Round-2 non-blocking observations (extension activation gating, convention-pinned wire contract, upsert `expires_at`) remain as stated — no change expected this round.

**Delta review — `de7a59d3` (Google Chat mirror):** this also closes the round-1 observation that correlation was Slack-only. The mirror is faithful and internally consistent: migration `006` is verified against google-chat's *own* baseline (its `001` defines `cleanup_expired_records()` and `trigger_set_updated_at()`; the down-migration restores google-chat's two-delete function body, not Slack's), the origin DataPart is byte-for-byte the orchestrator contract, gating matches the Slack final design (attach on every threaded reply with a row; orchestrator's empty-history gate decides), failed-run status/error parity is present, store failures are best-effort on both sides, and the required (non-optional) `scheduledRunStore` wiring is complete through all five handler call sites. One inline observation below — a platform-behavior assumption that can't be proven from the repo.

Minor notes, not worth inline comments: (1) the new tests pin the writer key via mocks but no test constructs a realistic Chat reply event to assert the reader derives the same key — a shared gap with the Slack package, so noted once here for a potential shared follow-up; (2) `isInThread` (`threadId !== messageId`) is effectively always true in Google Chat since a thread resource name never equals a message resource name, so the provenance lookup runs one indexed PK SELECT on essentially every message — harmless, just nominal gating.

packages/client-google-chat/src/handlers/a2aNotificationHandler.ts
aartaria: Acknowledged — this can't be proven from the repo and I can't run it locally (needs a live Google Chat deployment). Flagged to the maintainer as a pre-merge manual smoke test: deliver a scheduled-run notification in a DM, reply in-thread, and compare the stored provenance key (logged as `Stored scheduled-run provenance for thread=...`) with the reply event's `threadId` (the `Thread reply correlates to scheduled run` log line firing is the pass signal). If the invariant doesn't hold for 1:1 DM spaces, the fallback fix is keying by message name and resolving the thread's first message on the read side.

Issue #159, opened 2026-08-20
Auto-mode: headless tool approval for scheduled runs


aartaria:
## Problem

Scheduled runs execute tools with **no approval layer at all**. The HITL machinery exists and is substantial (`ConditionalHumanInTheLoopMiddleware` with DB-backed risk scoring, PTC approval batching), and the interactive path opts in — but the scheduled path never installs it:

- `agent-runner/agent/core.py` builds the sub-agent graph with no `risk_scorer` (`extra_middlewares` is always `None`), so the HITL conditional in `agent_common/core/graph_utils.py` is false for every scheduled run.
- No `interrupt` handling in agent-runner, no "awaiting approval" run status, no tests.
- The only substitute is design-time (auto-approval size caps for `AUTOMATED` agents); a TODO in the scheduled-job model acknowledges the gap.

**Consequence:** a scheduled agent handed a risky tool executes it silently, unattended, under the user's fully-privileged exchanged token. "Read-only scheduling" is currently enforced by nothing except which tools the agent happens to be given.

## Design: auto-mode (headless resolution strategy)

Not new machinery — the existing approval layer installed in the scheduled path with a headless resolution strategy (deny instead of `interrupt()`). Inspired by Claude Code's auto mode (deterministic rules first, per-call classifier second, denials recorded).

1. **Mode selection** — keyed to the execution path: agent-runner always builds graphs headless (thread `risk_scorer`/config through `_run_langgraph_agent`); other callers may opt in via an explicit flag. No reachability sensing.
2. **Decision model per tool call:**
   - **Bypasses carry unchanged** (checked first, as today): a user's "don't ask again" clears any score in auto-mode too — uniform semantics across modes.
   - **Risk bands** on the cached deterministic score (system-level defaults only; deliberately *not* configurable per sub-agent/job — the guarded workload must not configure its own guard): `< 0.3` auto-allow · `≥ 0.8` deny unless covered by a standing authorization · `0.3–0.8` → per-call classifier. Former "static guards" are score-1.0 rows in `tool_risk_scores`; no separate tier.
[...]

PR #163, opened 2026-08-21
feat: adopt a scheduled run's remote sub-agent conversation on delegation

### LINKED ISSUE #162: Cross-service conversation adoption: let orchestrator delegations resume a scheduled run's sub-agent conversation
Follow-up to #160 / #161 (see the round-1 review finding on PR #161 for the full trace).

#161 gives the orchestrator a synthetic reconstruction of a scheduled run (prompt + output as history), but a follow-up delegation to the same sub-agent still starts a **blank** sub-agent conversation. The run's own conversation (tool state, intermediate results) is checkpointed by agent-runner and is not reachable from any orchestrator delegation path:

- **Local/automated** sub-agents delegated by the orchestrator run in-process on thread key `{ctx}::dynamic-{name}` in the orchestrator's checkpointer — a different key (and checkpointer wiring) than agent-runner's bare `thread_id = contextId`.
- **Remote** and **foundry** sub-agents never receive the run task's contextId from agent-runner in the first place (`_run_remote_agent` / `_run_foundry_agent` build their input without it), so `scheduled_job_runs.conversation_id` names no conversation on their side.
- Naively seeding `a2a_tracking` with the run's contextId also desynchronizes the HITL checkpoint probe (dispatch middleware probes the orchestrator-conversation thread while the graph runs on the seeded thread), silently breaking tool-approval resume.

Designing real adoption needs one of:
1. agent-runner propagating a stable contextId into remote/foundry sub-agent dispatches (so the stored conversation_id names something on the executing side), plus the orchestrator's remote-delegation path (`A2AClientRunnable`) resuming it — the near-feasible case, since remote agents key checkpoints by the A2A contextId they receive;
2. for local agents, a thread-key/namespace convention shared between agent-runner's scheduled executions and the orchestrator's in-process dynamic agents;
3. HITL probe/resume paths made aware of per-delegation thread overrides.

Also from the review: when adoption makes the forwarded contextId meaningful again, add server-side ownership validation (resolve job/run for the authenticated `user_sub`) rather than trusting the client-supplied DataPart; `sub_agent_id` is already persisted end-to-end as the natural resolver key.

aartaria:
closes #162

**Stacked on #161** (base: `feat/scheduled-run-thread-correlation`) — rebase onto `main` after #161 merges.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — No. New fields are optional/additive (`interactive` defaults True, dispatch field optional, adoption seed additive); one new user-scoped read endpoint.

## What

#161 reconstructs a scheduled run as synthetic history, but a follow-up delegation to the same sub-agent started a **blank** conversation (#162). This PR makes the delegation **continue the run's own conversation** — for every adoptable agent kind. One contract, two continuity mechanisms:

- **Remote agents** — agent-runner now sends the run task's contextId on remote dispatches, so the executing server checkpoints the run under the id stored as `scheduled_job_runs.conversation_id`; the orchestrator seeds `a2a_tracking[key].context_id` and the existing waterfall resumes it on the wire.
- **Local/automated agents** (the common case for scheduled jobs) — the run's checkpoint lives at bare `thread_id = run ctx` in the checkpoint tables **both services share** (same DB/schema, same `build_sub_agent_graph`); the orchestrator seeds `adopt_thread_from` and the dispatch middleware **forks** that checkpoint onto the conversation's own `{ctx}::dynamic-{name}` thread on first delegation (blank target only). Never a raw `context_id` seed on a local runnable — that reintroduces the round-1 HITL probe/execution desync. Fork, not re-point: the run's thread stays an untouched record and parallel conversations adopting the same run stay independent.
- **Foundry** — not adoptable (continuity is a session rid the provenance doesn't carry).

### The five load-bearing pieces

1. `_validate_scheduled_run_origin` (orchestrator `agent.py`) — the security boundary: job + run re-resolved via console-backend under the **authenticated user's** token (unowned job 404s), server-side sub-agent binding checked, **server-stored** `conversation_id` used; the DataPart's `context_id` plays no part. Uses the new user-scoped single-run endpoint `GET /api/v1/scheduler/jobs/{id}/runs/{run_id}` (the run listing is capped to the newest 50).
2. `_build_adoption_seed` — the policy switch mapping the validated run onto the runnable's continuity mechanism (`context_id` / `adopt_thread_from` / none), keyed by `runnable.tracking_key`.
3. Automated-agent exposure — `type="automated"` configs flow through the registry as `interactive=False` and are registered into **exactly the conversation that adopted one of their runs** (`adopted_sub_agent_ids`, resolved before the runtime context is built); delegable there for the rest of the conversation, invisible everywhere else.
4. `_maybe_adopt_run_thread` + `_seal_dangling_tool_calls` (dispatch middleware) — the fork: latest checkpoint only, `pending_writes` never copied (no interrupt can teleport, aligned with #159's no-interrupt-at-rest auto-mode design), and a run that died mid-tool gets its unanswered tool_calls sealed with explicit error ToolMessages **in the copy** so the history stays provider-valid.
5. agent-runner's `_run_remote_agent` one-liner (`orchestrator_conversation_id=context_id`) — what makes the remote half real.

### task_state-aware framing (from live testing)

Live testing showed the fork working while the model sabotaged it in front: told nothing about the run being unfinished or resumable, it role-played the sub-agent instead of delegating. agent-runner now reports the sub-agent's terminal A2A task state (`completed | input_required | failed`) in the run provenance; the clients persist and forward it (`task_state` column on `scheduled_run_store` — dev DBs that already ran migrations 006/008 need `alter table scheduled_run_store add column task_state text;`); and the reconstruction frames accordingly: `input_required` + adoption ⇒ "the run is waiting for the user's answer — forward it via `task`, don't answer on the sub-agent's behalf"; `input_required` without adoption ⇒ honest "the exchange cannot be resumed"; any adoption ⇒ an explicit note that delegating RESUMES the run's conversation, countering the model's stateless-sub-agent prior.

The synthetic delegation turn renders the resolved dispatchable label (`delegation_label` through the origin-builder dispatch) instead of the provenance name, so the model re-delegates with a label that actually dispatches.

## Testing

- **Real-Postgres integration** (console-backend's container pattern; skips without Docker): an agent-runner-shaped graph (no HITL middleware) writes the run thread, a real `AsyncPostgresSaver` over `build_checkpointer_pool` (schema `docstore`) forks it, and an **orchestrator-shaped graph (HITL middleware attached)** resumes it — the exact cross-variant load production performs; plus the sealed dangling-tool-call fork through real serialization, no-refork on second delegation, and source-thread immutability.
- orchestrator-agent: 672 pytest pass (validation, seed policy per kind, fork unit tests, adopted-id re-derivation, delegation-label rendering, task_state framing).
- agent-runner: 16 pytest pass (contextId propagation, task_state reporting).
- client-slack: tsc clean, 62 jest pass (one pre-existing unrelated suite failure in this env: taskRecovery TS config drift, reproduces on the base branch). client-google-chat: tsc clean, 22 jest pass.
- console-backend: 1376 pytest pass (2 new `get_run` service tests: owned resolves, unowned is None).
- ruff clean on all touched files; mypy not runnable locally — CI covers it.

## Design notes

- **Open alternative (unmeasured)**: routing local/automated delegations to agent-runner as an A2A server would collapse adoption to the remote mechanism — parked pending a real latency benchmark (per-turn double harness build vs in-process warm-path reuse) and an interrupt-capable approval mode on agent-runner. Documented in orchestrator `AGENTS.md`.
- **#159 interplay**: auto-mode preserves the no-interrupt-at-rest invariant the fork relies on, and adoption becomes the interactive on-ramp for auto-mode denials (the model re-attempts the denied call under the orchestrator's interactive HITL). The later "real async approval" stage should coordinate with adoption (skip/defer the fork for an `awaiting_approval` run to avoid double execution).
- Round-1 review fixes are in `013e7dbb` (round 2 clean); round 3's finding — adopted automated agents deregistered after turn one, dropping HITL approvals — is fixed in `27e35bcf` by stamping `sub_agent_id` into the persisted adoption record and re-deriving `adopted_sub_agent_ids` from checkpoint state on every turn (round 4 verified it clean).

## Out of scope

Foundry adoption; resuming a still-open remote task by taskId (we adopt the conversation, not the task); the agent-runner-as-A2A-server unification (pending measurement); #159's async-approval stage.


packages/orchestrator-agent/app/core/agent.py
```diff
@@ -249,6 +254,116 @@ def _build_origin_history(origin: dict[str, Any]) -> list[Any] | None:
...
+                f"(HTTP {job_resp.status_code})"
+            )
+            return None
+        if _origin_int(job_resp.json().get("sub_agent_id")) != sub_agent_id:
+            logger.warning(
+                f"Conversation adoption skipped: origin sub_agent_id {sub_agent_id} does not match "
+                f"job {job_id}'s server-side sub-agent binding"
+            )
+            return None
+        runs_resp = await client.get(f"/api/v1/scheduler/jobs/{job_id}/runs", headers=headers)
```
alloy-cicd: **Correctness — adoption silently unreachable for runs older than the newest 50.** `GET /api/v1/scheduler/jobs/{id}/runs` is hard-capped at the 50 most recent runs (`scheduler_router.py:346` → `scheduler_service.list_runs(limit=50)` → `ORDER BY started_at DESC LIMIT :limit`), with no `limit` param and no single-run endpoint. A job on a 5-minute cron accumulates 50 runs in ~4 h, so a user replying under yesterday's notification gets `run is None` → adoption skips, and the log says `run 42 not found on job 7` — indistinguishable from a forged run id, and unreproducible on any fresh test job. The clean fix is a `GET /jobs/{job_id}/runs/{run_id}` endpoint (ownership-scoped like `get_job`), which also collapses this function to one HTTP call — `list_runs` already does the same `repo.get_job` ownership check, so the separate job GET would only be needed for the `sub_agent_id` binding, which a run endpoint could return.

Minor, same block: the non-200 branch at line 348 is the only skip path with no log line at all.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -249,6 +254,116 @@ def _build_origin_history(origin: dict[str, Any]) -> list[Any] | None:
         return None
 
 
+# Timeout for the console-backend ownership lookups behind conversation
+# adoption. Adoption is an enrichment on the conversation's first turn only;
+# a slow backend must degrade to "no adoption", not stall the turn.
+_ADOPTION_LOOKUP_TIMEOUT_S = 5.0
```
alloy-cicd: **The stated latency contract isn't what the code enforces.** The comment says a slow backend "must degrade to 'no adoption', not stall the turn", but 5.0 s is an httpx *per-request* timeout applied to two *sequential* GETs awaited inline before `graph.astream()` — worst case ~10 s of dead air on the first token of every reply under a scheduled-run notification (and the lookup runs even when the reply never delegates, e.g. "thanks"). Suggest one overall deadline around the whole resolution (`asyncio.timeout(2–3s)`), and optionally `asyncio.gather` for the two GETs (the binding check can post-filter) — or drop to a single request per the runs-endpoint suggestion above.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -249,6 +254,116 @@ def _build_origin_history(origin: dict[str, Any]) -> list[Any] | None:
...
+            if isinstance(candidate, A2AClientRunnable):
+                runnable = candidate
+            break
+    if runnable is None:
+        logger.info(
+            f"Conversation adoption skipped: sub-agent {sub_agent_id} is not a registered remote agent"
+        )
+        return None
+
+    backend_url = os.getenv("CONSOLE_BACKEND_URL", "http://localhost:5001")
```
alloy-cicd: **Reuse: this duplicates existing console-backend client infrastructure.** `os.getenv("CONSOLE_BACKEND_URL", "http://localhost:5001")` is now the 4th copy of this env-var+default in the package (`RegistryConfig.console_backend_url` at `registry.py:117` is the declared owner; also `main.py:110/182` and `AgentSettings.CONSOLE_BACKEND_URL` in `models/config.py:364`), and the throwaway `httpx.AsyncClient` per adoption attempt sits alongside three established pooled-client patterns for this same base URL (`registry.py:163`, `risk_score_api_client.py:45`, `budget_guard.py:175`). A `resolve_scheduled_run(...)` method on `RegistryService` (or a sibling client) would inherit config, pooling, retry and auth handling — and make the tests assert URL/auth construction instead of globally patching `httpx.AsyncClient`, which breaks the moment any other httpx call enters this code path. As written, a transient 502 permanently costs the adoption for that conversation (first-turn-only, no retry).

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -249,6 +254,116 @@ def _build_origin_history(origin: dict[str, Any]) -> list[Any] | None:
...
+        logger.info(f"Conversation adoption skipped: run {run_id} not found on job {job_id}")
+        return None
+    conversation_id = run.get("conversation_id")
+    if not isinstance(conversation_id, str) or not conversation_id:
+        return None
+
+    # Key must match A2AClientRunnable._extract_tracking_ids' reader key
+    # (self.name with spaces stripped — for remote agents this also equals the
+    # registry key the tracking middleware later writes under).
+    tracking_key = str(getattr(runnable, "name", "")).replace(" ", "")
```
alloy-cicd: **The tracking-key convention now has three independent copies and no code home.** This line duplicates the reader (`A2ABaseRunnable._extract_tracking_ids`, `agent_common/a2a/base.py:341`: `self.name.replace(" ", "")`) and the registry key derivation (`discovery.py:166`) — the comment documents the coupling instead of removing it, and `test_tracking_key_strips_spaces_from_runnable_name` pins the *duplicated* form, so it keeps passing if the shared convention ever changes. I verified the three coincide today (seed key = registry key = reader key, since `A2AClientRunnable.name` is the card name). Suggest a public `tracking_key` property on `A2AClientRunnable` (or a helper in `agent_common/a2a/base.py`) used by all three sites; divergence here fails silently — the seed lands under a key nobody reads while the log claims adoption succeeded.

Also simplifiable: the loop at line 319 already has the registry key in hand via `subagent_registry.items()`; returning it directly would remove this re-derivation and the `getattr`/empty-string guards entirely.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -126,9 +129,11 @@ def _build_scheduled_run_history(run: dict[str, Any]) -> list[Any] | None:
     one — some providers reject a leading assistant tool-call), the ``task``
     tool call to the sub-agent, and the run's output as the tool result. This
     gives the model the run's prompt and output as real context; it does NOT
-    resume the run's checkpoint — the run executed on agent-runner, whose
-    conversation state is not reachable from the orchestrator's delegation
-    paths (cross-service adoption is a separate follow-up).
+    by itself resume the run's checkpoint. For runs executed by a REMOTE
+    sub-agent, _resolve_scheduled_run_adoption separately seeds a2a_tracking
+    so a follow-up delegation resumes the run's conversation on the executing
```
alloy-cicd: **Non-blocking, pre-existing, but it can nullify this PR's payoff:** the synthetic history emits `subagent_type = origin['sub_agent_name']` (line 210), which is the *console config* name (`agent-runner core.py:701`), while the dispatch enum and the new adoption seed key use the *agent-card* name with spaces stripped (`discovery.py:166`). For a remote agent whose console name differs from the card name — or whose card name merely contains a space — the model is primed with a non-enum agent label; if it echoes it, dispatch falls through and the seeded conversation is never read (feature silently no-ops on the first attempt). Consider emitting the registry-resolved key in the synthetic tool call when the sub-agent resolves to a remote registry entry.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -603,6 +718,36 @@ async def stream(
...
+                    # any failure degrades to the synthetic history alone.
+                    try:
+                        adoption = await _resolve_scheduled_run_adoption(
+                            origin,
+                            runtime_context.subagent_registry,
+                            user_config.access_token.get_secret_value(),
+                        )
+                    except Exception:
+                        logger.warning(
+                            "Conversation adoption lookup failed; continuing without it", exc_info=True
```
alloy-cicd: **The blanket `except Exception` here also swallows programming errors** (a typo'd attribute on the registry entry, a `SecretStr` API change) into a WARNING — and the resulting symptom, sub-agents starting blank, is exactly the pre-PR behavior, so a real bug in the resolver would be invisible in production. The resolver already implements the degrade-to-None policy for every expected failure; consider moving the try/except inside `_resolve_scheduled_run_adoption`, wrapping only the httpx block (the sole part that legitimately raises — including `.json()` on a non-JSON 200 from an auth proxy, and the list-shape assumption at line 350, both of which currently surface only as this generic warning). The call site then shrinks to `if adoption := await _resolve_...(...)` at one nesting level less.

Relatedly, consider having the resolver return the seed dict itself rather than a tuple the caller reassembles — the `a2a_tracking` payload shape (`is_complete`) then lives next to the code that reasons about it instead of mid-`stream`.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -249,6 +254,116 @@ def _build_origin_history(origin: dict[str, Any]) -> list[Any] | None:
         return None
 
 
+# Timeout for the console-backend ownership lookups behind conversation
+# adoption. Adoption is an enrichment on the conversation's first turn only;
+# a slow backend must degrade to "no adoption", not stall the turn.
+_ADOPTION_LOOKUP_TIMEOUT_S = 5.0
+
+
+def _origin_int(value: Any) -> int | None:
```
alloy-cicd: Minor: `_origin_int` is a second coercion helper for the same "MessageToDict renders numbers as floats" problem as `_fmt_id` inside `_build_scheduled_run_history` (line ~154), with different edge-case behavior (bools, digit strings). One module-level helper used by both keeps the two readings of the same DataPart in agreement — today a string-serialized id would render in the synthetic history header but fail adoption, or vice versa.

packages/orchestrator-agent/tests/test_scheduled_run_adoption.py
```diff
@@ -0,0 +1,175 @@
...
+        return response
+
+    mock_client = AsyncMock()
+    mock_client.get = AsyncMock(side_effect=fake_get)
+
+    patcher = patch("httpx.AsyncClient")
+    mock_cls = patcher.start()
+    mock_cls.return_value.__aenter__ = AsyncMock(return_value=mock_client)
+    mock_cls.return_value.__aexit__ = AsyncMock(return_value=False)
+    return patcher, mock_client
```
alloy-cicd: Test hygiene: `_patched_backend` returns a live `patcher` that every test must remember to stop, and the `patcher, _ = ...` / `try` / `finally: patcher.stop()` triple is repeated six times. If a future test drops the `finally`, `httpx.AsyncClient` stays patched globally and unrelated tests in other files start failing with MagicMock responses. A `@contextlib.contextmanager` (yield the mock client, `patcher.stop()` in its own `finally`) makes each test a one-line `with _patched_backend(job_status=404) as client:` with no per-test cleanup.

alloy-cicd (commented):

The adoption design itself checks out: I traced the contextId round trip (agent-runner → wire → remote checkpoint key → `scheduled_job_runs.conversation_id`) and the seed shape end-to-end, and they are consistent. One real correctness gap (the 50-run listing window, first comment) and a set of quality findings, all inline below.

**Claims checked and cleared** (so later rounds don't re-litigate them):
- *Input-required runs*: seeding `is_complete: True` without a `task_id` is safe — the SDK's remote executor has no interrupt-resume branch; a taskId-less message on an existing contextId starts a new task that processes the reply as a normal user message on the accumulated thread state (`ringier_a2a_sdk/agent/langgraph.py:898`, langgraph `_loop.py` input branch).
- *Job re-pointing*: the origin embeds the run-time `sub_agent_id`, so after re-pointing a job the binding check at `agent.py:341` mismatches and adoption skips cleanly; the seed can structurally only pair an agent with its own conversation. (Residual: an old run of the *previous* agent is silently non-adoptable — a limitation, not a hole.)
- *Failed runs*: agent-level failed runs do get adopted (no status check), but on a reply under that run's notification, continuity for "what went wrong / retry" is arguably the desired behavior. Transport-level failures persist no `conversation_id` and skip.
- *Notification-vs-persist race*: the push does precede the `_finalize` write, but the window is sub-second and the fallback is exactly pre-PR behavior.

**Non-inline note — AGENTS.md**: this PR inverts a previously documented position (the orchestrator now *does* resume a scheduled run's remote conversation) and introduces a cross-service invariant (seed key must equal `_extract_tracking_ids`' reader key; local runs must never be seeded or the HITL checkpoint probe desynchronizes) that spans four packages. Both root and orchestrator `AGENTS.md` carry "Key/Critical Design Decisions" entries of exactly this shape, and their "Maintaining These Instructions" rule asks for one here. Consider adding a short entry.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 013e7dbb: added a user-scoped single-run endpoint (GET /api/v1/scheduler/jobs/{id}/runs/{run_id}, +service/repo methods with ownership check) and the resolver now uses it — runs older than the listing cap stay adoptable, and the lookup no longer scans a list. Service-level tests cover owned/unowned.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 013e7dbb: one overall asyncio.timeout(3.0) wraps the whole resolution, and the two GETs run under asyncio.gather (the binding check post-filters, as you suggested). Worst case is now ~3s total, not 2×5s. The lookup still runs on every origin-bearing first turn even if the reply never delegates — accepted for now: it's one turn per conversation and the seed must be in the initial state update.

packages/orchestrator-agent/app/core/agent.py
aartaria: Partially fixed in 013e7dbb: the URL now comes from AgentSettings.CONSOLE_BACKEND_URL (the copy agent.py's own class already owns) instead of a 4th getenv. Kept the short-lived client: the pooled patterns live in service objects (registry/budget/risk-score) not importable here without wiring a new dependency through OrchestratorDeepAgent, and this path runs at most once per conversation. Happy to lift it into a shared console client in a follow-up if one grows more callers.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 013e7dbb: added BaseA2ARunnable.tracking_key as the single home of the convention; _extract_tracking_ids, discovery's registry-key derivation, and the adoption seed all read it now, and the test asserts the seed equals runnable.tracking_key rather than pinning a re-derived copy.

packages/orchestrator-agent/app/core/agent.py
aartaria: Addressed for the case where it bites this PR (013e7dbb): when adoption resolves, the synthetic task tool call's subagent_type is rewritten to the dispatchable registry label (tracking_key), so the model is primed with a name that both dispatches and hits the seed. The general non-adopted mismatch (console name vs card name) remains pre-existing #161 behavior — the model still has the real enum in its tool description.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed as suggested in 013e7dbb: the try/except moved inside the resolver and wraps only the lookup block, catching (httpx.HTTPError, TimeoutError, ValueError) — the last covering .json() on a non-JSON 200. The blanket except at the call site is gone; programming errors now propagate.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 013e7dbb: _origin_int is the module-level helper and _fmt_id delegates to it, so both readings of the DataPart agree (a string-serialized id now renders and adopts consistently).

packages/orchestrator-agent/tests/test_scheduled_run_adoption.py
aartaria: Fixed in 013e7dbb: _patched_backend is a @contextlib.contextmanager and every test uses a `with` block; no live patcher can leak past a test body.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (5bca194cd11cdfaf52d3c1a4a7ac9c275a63046e..013e7dbb58b8072d61f954285f5f68d4c7174fb2)

**Disposition ledger (round 1 → now):**
1. **50-run listing cap** — ✅ verified-fixed. `GET /jobs/{id}/runs/{run_id}` added, user-scoped in `SchedulerService.get_run` (ownership check before the repo call, `mock_repo.get_run.assert_not_awaited()` on the unowned path), resolver now does a single-run lookup, and the previously silent non-200 branch logs.
2. **Latency contract** — ✅ verified-fixed. One `asyncio.timeout(3.0)` wraps both lookups, which run under `asyncio.gather`; `TimeoutError` is in the caught set. The still-eager lookup on non-delegating replies is accepted-by-author with reasoning (once per conversation, seed must be in the initial state update) — settled.
3. **Console-backend client reuse** — ✅ URL now from `AgentSettings.CONSOLE_BACKEND_URL`; the short-lived client is rejected-by-author with reasoning (no importable shared client here without new wiring; at most once per conversation; follow-up if a shared console client grows more callers) — accepted, not re-raised.
4. **Tracking-key convention** — ✅ verified-fixed. `BaseA2ARunnable.tracking_key` is the single home; `_extract_tracking_ids`, discovery's registry key, and the adoption seed all read it, and the test now asserts the seed equals `runnable.tracking_key` instead of pinning a re-derived copy.
5. **History label vs registry key** — ✅ addressed for the adoption case: the synthetic `task` tool call is rewritten to the dispatchable `tracking_key` when adoption resolves (mutation happens on the same message objects already placed in `input_data`, so it is effective; correctly guarded for the no-synthetic-history case). The general non-adopted mismatch remains pre-existing behavior per author — settled.
6. **Blanket `except Exception`** — ✅ verified-fixed. try/except moved inside the resolver around the lookup block only, catching `(httpx.HTTPError, TimeoutError, ValueError)` (the last covers `.json()` on a non-JSON 200); programming errors now propagate. New degradation test covers backend-down.
7. **`_origin_int`/`_fmt_id` duplication** — ✅ verified-fixed; `_fmt_id` delegates to the module-level `_origin_int`.
8. **Test patcher leak** — ✅ verified-fixed; `_patched_backend` is a `@contextlib.contextmanager`, all tests use `with` blocks.

The AGENTS.md note from the round-1 summary is also addressed (root entry + orchestrator "Scheduled-Run Conversation Adoption" section documenting both invariants).

Ran `tests/test_scheduled_run_adoption.py` + `tests/test_discovery.py` at head: 25 passed.

Two below-threshold observations, follow-up material only (no action needed on this PR): `asyncio.gather` without `return_exceptions` can leave the second request as an abandoned task logging "exception was never retrieved" if exactly one GET fails with a transport error (both target the same host, so rare); and the `or "http://localhost:5001"` at the call site keeps one literal copy of the default alongside the `AgentSettings` one.

packages/orchestrator-agent/app/core/agent.py
```diff
@@ -502,9 +700,33 @@ async def stream(
...
+                        self.config.CONSOLE_BACKEND_URL or "http://localhost:5001",
+                    )
+
         # Build GraphRuntimeContext for runtime injection (personalizes system prompt, etc.)
         # UserConfig should already have tools/agents discovered by executor via discover_capabilities()
-        runtime_context = self.build_runtime_context(user_config, sandbox_pool=self.sandbox_pool)
+        runtime_context = self.build_runtime_context(
+            user_config,
+            sandbox_pool=self.sandbox_pool,
+            adopted_sub_agent_ids={validated_origin["sub_agent_id"]} if validated_origin else None,
```
alloy-cicd: **Correctness — adopted automated agents are registered on the first turn only; multi-turn delegation and HITL resume break.** `validated_origin` is computed only when `resume is None` and the checkpoint is empty, so every later `stream()` call passes `adopted_sub_agent_ids=None` and `build_runtime_context` (utils.py:303-307) drops the `interactive=False` config again. Two concrete failures, both independently traced:

1. **Turn 2+**: the adopted agent vanishes from the system-prompt roster and the `task` tool enum (both rebuilt per call from `subagent_registry`, dynamic_tool_dispatch.py:685/787-789); if the model names it anyway from turn-1 history, the registry lookup at :1718 returns None and dispatch falls through to `SubAgentMiddleware`, which doesn't know it. This contradicts the commit message and the new AGENTS.md line ("for the rest of that conversation they behave like any local agent").
2. **HITL resume — breaks even the first delegation**: local sub-agent graphs get `ConditionalHumanInTheLoopMiddleware` on the orchestrator side, so the adopted agent's first delegation can interrupt for tool approval. The resume call (`resume is not None`) skips the pre-pass entirely, rebuilds the registry without the agent, and LangGraph replays the interrupted `task` tool from the top — where the registry lookup at :1718 precedes every `interrupt()` re-entry (:1908, :2196). The lookup misses, the fallback path never re-reaches the interrupt, and the user's approval is silently discarded, orphaning the sub-agent's pending `__interrupt__` checkpoint.

There is no compensating path: the only producer of `adopted_sub_agent_ids` is this pre-pass, the only consumer is the utils.py gate, and origin re-extraction on later turns is (correctly) blocked by the empty-checkpoint gate even though clients re-attach the DataPart.

Fix shape: the durable marker already exists — the seeded `a2a_tracking[tracking_key]` record survives in checkpoint state. Run the pre-pass `aget_state` unconditionally (including the resume branch, which already loads state for attachments anyway) and union adopted ids from persisted adoption records — either store `sub_agent_id` in the seeded record, or map tracking keys back through `user_config.local_subagents` (tracking_key == config name for local agents). A dedicated persisted `adopted_sub_agent_ids` state channel seeded on turn 1 would be equivalent and more explicit.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (013e7dbb58b8072d61f954285f5f68d4c7174fb2..673dc951bafce2fc79cba05c7cb256c613389627)

**Ledger from round 2** (which had no inline findings, only two follow-up notes): the `gather` "exception never retrieved" note is ✅ verified-fixed by 0bce1c61 (`return_exceptions=True` + uniform re-raise); the residual `or "http://localhost:5001"` literal is still present — still follow-up material, not re-raised. All round-1 findings remain settled.

**Delta review** — the fork-on-adopt feature is substantially sound. I verified the mechanics in depth before flagging anything, and the following all check out (recorded so later rounds don't re-litigate):
- *Thread/namespace identity*: blank-probe, fork write, HITL probe, and actual execution all resolve to `({ctx}::dynamic-{name}, ns="")` — including `_astream_impl`'s force-override of the instrument-derived ns back to `""`.
- *Source key*: agent-runner checkpoints scheduled local/automated runs at bare `thread_id = task contextId` with ns `""`, and `scheduled_job_runs.conversation_id` is exactly that contextId, so the fork reads the key the run wrote.
- *Copy semantics*: on langgraph-checkpoint-postgres 3.1.0, `aput` with the source's full `channel_versions` re-serializes every loadable channel onto the target; checkpoint-id reuse across threads is safe (PK includes thread_id) and the fork lands as a clean parentless root.
- *Sealing*: only the trailing AIMessage can carry unanswered tool_calls given each run thread holds exactly one graph execution (scheduler dispatches without a context_id), so append-at-end sealing is valid; `pending_writes` are never copied, so no phantom `__interrupt__` can reach the probe.
- *Seed-record interactions*: `_extract_tracking_ids` treats an `adopt_thread_from`-only record exactly like an absent one (clean fallback to the conversation-derived thread); `A2ATrackingMiddleware` merges per-key so the record survives until a late delegation; `_check_all_agents_blocked` never consults never-delegated agents.
- *Exposure*: no leak from routing automated configs through the registry — every prompt/enum/dispatch/cost surface reads the post-gate `subagent_registry`, discovery caches never hold `local_subagents`, and registration requires both run ownership (server-validated) and the user's own agent entitlement (`/activated` under their token).

One confirmed correctness bug in the delta, inline below (the first-turn-only registration gate). Everything else is follow-up material at most: the consumed `adopt_thread_from` key is never dropped, costing one extra `aget_tuple` per later delegation to that agent (harmless); a user who owns a job's runs but doesn't have the automated agent in `/activated` gets a silent adoption no-op (entitlement-correct, just unlogged as such); and `registry.py` now DEBUG-logs the full automated config including its system prompt, matching the pre-existing local-agent pattern.

Ran the orchestrator suite at head: 662 passed.

packages/orchestrator-agent/app/core/agent.py
aartaria: Fixed in 27e35bcf: _build_adoption_seed stamps sub_agent_id into the a2a_tracking record (persisted in the checkpoint from turn one), and the astream pre-pass now re-derives adopted_sub_agent_ids from that state on EVERY turn — HITL resumes included — via _adopted_sub_agent_ids_from_tracking, before building the runtime context. The server-side validation still runs only on the blank first turn, and the turn still performs exactly one aget_state (the resume branch reuses the pre-pass fetch instead of its own). Covered by TestAdoptedIdsFromTracking (recovery, ordinary-records-yield-none, malformed state) plus the updated seed-record assertions.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (673dc951bafce2fc79cba05c7cb256c613389627..27e35bcf7776f1af67354355785d9ca0c6d451e6)

**Ledger:**
- **Round-3 finding (adopted automated agents registered on the first turn only; HITL resume dropped the approval)** — ✅ verified-fixed, in exactly the shape suggested. `_build_adoption_seed` stamps `sub_agent_id` into the persisted a2a_tracking record, and the pre-pass now runs `aget_state` unconditionally and re-derives `adopted_sub_agent_ids` via `_adopted_sub_agent_ids_from_tracking` on every turn, resume branch included (which now reuses the pre-pass fetch — still exactly one `aget_state` per turn). I verified the load-bearing assumption directly: `A2ATrackingMiddleware` mutates fields on the existing per-agent record and only ever deletes `task_id`, so the stamped `sub_agent_id` survives post-delegation updates for the conversation's lifetime. The derivation is also injection-safe — the seed path (server-validated) is the only producer of `sub_agent_id` in tracking records; ordinary middleware-written records yield None, as the new test pins. AGENTS.md documents the invariant.
- Round-2/round-1 items — all remain settled; the `localhost:5001` literal note stays follow-up material.

Ran the orchestrator suite at head: 665 passed (3 new tests covering recovery, ordinary-records-yield-none, and malformed state).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (27e35bcf7776f1af67354355785d9ca0c6d451e6..7dcc4a02fd077fa486662cecde4d30b3e831dcd0)

**Ledger:** round 4 was already clean (the round-3 registration fix verified then); rounds 1–3 remain settled. Nothing was re-raised.

**Delta review** — the task_state plumb-through and adoption framing check out end to end; what I verified before calling it clean:
- **State propagation**: `SubAgentResponseSchema.task_state` literals (`completed | input_required | failed`) match the scheduler-facing strings exactly, `_collect_stream_text`'s terminal-state map ignores non-terminal states, the `ErrorEvent` path reports `failed`, and the single `_execute_sub_agent` caller plus all three execution paths were updated to the tuple return consistently (both Python suites green: 672 orchestrator — including the 7 new framing tests — and 16 agent-runner).
- **Framing correctness**: the origin's `task_state` is untrusted and allowlisted before use; the "forward the user's reply" advice renders only when adoption validated (no forwarding-into-a-void), the no-adoption `input_required` variant is honest about unrecoverable state, `scheduler_status=failed` trumps a stale `input_required`, and unknown states degrade to the default framing — each pinned by a test.
- **Synthetic metadata safety**: `is_complete: False` / `TASK_STATE_INPUT_REQUIRED` on the synthetic ToolMessage cannot leak into tracking state (`A2ATrackingMiddleware` skips records without a task/context id) and cannot trip the blocked-agent heuristics (synthetic messages precede the current turn's HumanMessage, which is the boundary those heuristics scan from). The a2a_tracking continuity seed still carries `is_complete: True` independently, as the inline comment notes.
- **Clients/migrations**: both chat clients persist and forward `task_state` symmetrically (types, insert/select, DataPart passthrough, tests). Amending `006`/`008` in place is fine here — neither file exists on `origin/main`, so only branch-deployed dev DBs are affected, exactly what the commit's manual-ALTER note covers. Worth keeping that note in the deploy checklist when this merges.

Follow-up material only (unchanged from earlier rounds): the `localhost:5001` literal at the resolver call site, and the consumed `adopt_thread_from` key never being dropped from tracking state.

PR #166, opened 2026-08-24
feat(scheduler): CEL watch conditions

### LINKED ISSUE #165: Watch conditions: CEL expressions instead of JSONPath + operator
Watch jobs currently express their trigger as a JSONPath extraction plus an operator and an expected value, optionally with an LLM judgement on the whole tool response. That shape has three problems:

- **It can't express real conditions.** Anything involving date math ("starts within the hour"), comparing against the previous check ("the price dropped"), or looking across a list needs either a model call or nothing.
- **Every quiet poll costs an LLM call.** A condition whose mechanical part is decidable ("is this list non-empty") still goes to a model on every tick.
- **Extraction and gating are two disconnected concepts**, so what a run recorded as evidence and what it decided are not the same thing, and a triggered run can't be re-derived from what it stored.

## Proposal

Make a watch condition a **CEL expression**, a **model judgement**, or both stacked.

One expression serves both purposes: it extracts the evidence a run records and hands onward, and it gates the trigger — a boolean result gates directly, anything else gates on non-empty. It evaluates against:

- `result` — the tool response
- `now` — the current time in the job's timezone, so date math stops being a model guess about the clock
- `prev` — the previous check result, for change detection

With a model judgement stacked on top, the gate runs first and the model only judges what the expression returned. Judge-only conditions stay ungated, so absence-conditions ("nothing scheduled tomorrow") still reach the model instead of short-circuiting to not-met.

### Scope

- Retire JSONPath + operator + expected-value from the API, the schema and the UI. Migrate stored rows onto CEL, preserving the old semantics exactly so no live job changes its verdict.
- Move condition evaluation out of the agent runner and into the scheduler, so the scheduler knows the outcome before dispatching: a quiet poll dispatches nothing, and a trigger can choose its target. Record per run how the condition was decided — a mechanical rule can be re-derived from a stored response, a model's reasoning only exists while the run happens.
- Fail closed: compile-check expressions at write time, cap their length, and evaluate under a timeout. A condition that cannot see its subject should fail the run, not read as "not met".
- **Dynamic check arguments.** Some checks need an argument computed per run (today's date, a window end). Allow a CEL expression per argument, resolved fresh each run and merged over the static arguments.
- **Authoring assist**, because non-trivial CEL is hard to write by hand: an endpoint that writes or refines just the condition, with the model seeing the real tool response and every candidate compiled and evaluated against it before it is offered.
- **UI**: an expression editor with highlighting rather than a bare textarea, the payload's keys one click away, a describe-or-refine box, click-a-value-to-seed, and a tester that distinguishes "would not trigger" from "fails against this payload". Run history should explain two-stage outcomes — which stage said no, and that a false gate never asked the model.

aartaria:
closes #165

A watch condition is now a **CEL expression**, a **model judgement**, or both stacked. The JSONPath + operator + expected-value trio is gone from the API, the schema and the UI.

## Backend — `feat(scheduler)`

One expression serves two purposes at once: it extracts the evidence (what the run records and hands to the model or agent) and it gates the trigger — a boolean result gates directly, anything else gates on non-empty. It is evaluated against `result` (the tool response), `now` (the current time in the job's timezone, so "starts within the hour" is date math instead of a model guessing the clock) and `prev` (the previous check result, for change detection). With `llm_condition` stacked on top, the gate runs first and the model judges only what the expression returned — the mechanical part of a condition no longer costs an LLM call on every quiet poll.

- **Migration, semantics-preserving.** `083_cel_watch_conditions.sql` rewrites stored rows onto two CEL extension functions kept for exactly that purpose: `jsonpath()` with the old extractor's contract and `eq_ci()` with the old comparison's case-insensitive semantics. No live job changes its verdict. Judge-driven rows deliberately stay **ungated** — their extraction becoming a gate would short-circuit absence-conditions to not-met without ever asking the model.
- **Evaluation moved into the scheduler** (`082_run_condition_evaluation.sql`), so the outcome is known *before* dispatch: a quiet poll dispatches nothing, and a trigger can choose its target (an agent, or a phone call). Every run records how its condition was decided — an expression can be re-derived from the stored response, a model's reasoning only exists while the run happens. The judge prompt now states the current time; it previously inferred the date from payload timestamps.
- **Fails closed.** Expressions are compile-checked at write time, capped in length, and evaluated under a timeout. A condition that cannot see its subject fails the run rather than reading as "not met".
- **Dynamic check arguments** (`084_dynamic_check_args.sql`): `check_args_exprs` maps an argument name to a CEL expression over `now`/`prev`, resolved fresh per run and merged over the static `check_args`. A resolution failure fails the run without calling the tool. Guards a celpy quirk where unbound variables become lazy error values that `string()` would coerce into garbage.
- **`POST /scheduler/generate-condition`** writes or refines just the condition: the model sees the real tool response, and every candidate is compiled, evaluated against it, and repaired with the error fed back — so a wrong-but-compiling field path is caught before it can fail every scheduled run.
- Also folded in: stuck `running` runs are swept on **every** tick, not just at startup, excluding runs this process is still dispatching.

## Frontend — `feat(console)`

The trigger section is two optional halves of one condition, at least one required: an **Expression** and an **AI-judges** field (alone it reads the whole response; with an expression it judges only what the expression returned).

- The expression gets a real authoring stage instead of a bare textarea: a CodeMirror editor with highlighting and bracket matching, the payload's top-level keys one click away, and a describe-or-refine box backed by `generate-condition`. Clicking a value in the check result seeds the expression in CEL's spelling.
- Dynamic arguments are **spreadsheet-style** rather than a separate box: type `= strftime(now, '%Y-%m-%d')` with a leading `=` directly into the argument's own field. `validate-args-expr` previews the resolution, and "Run check now" applies it.
- The condition tester evaluates the same expression the scheduler will run and distinguishes "would not trigger" from "compiles but fails against this payload". The run history panel explains two-stage outcomes — which stage said no, and that a false gate never asked the model.
- `WatchFields` is shared by the create dialog and the job detail page so the two cannot drift, and the create form is rebuilt around what it is actually asking: what to check, when to trigger, what happens then.
- Array-of-enum tool arguments now render as multi-select chips instead of falling back to "complex → edit as JSON".

## Verification

- `console-backend`: 207 tests pass across `test_cel_condition`, `test_watch_evaluator`, `test_generate_condition`, `test_generated_expression_repair`, `test_scheduler_engine`, `test_scheduler_service`, `test_mcp_tool_invoke`, `test_mcp_tool_client`, `test_generate_job_draft`.
- `agent-runner`: 17 tests pass (`test_watch_condition_ops`, `test_agent_runner_security`).
- `console-frontend`: `pnpm build` (tsc + vite) clean. `pnpm lint` reports 205 pre-existing problems repo-wide, **none** in any file this branch touches.

## Checklist
- [x] Does this PR introduce a breaking change? 🚨 — **API-breaking**: the JSONPath/operator/expected-value fields are removed from the scheduler API and schema. Stored rows are migrated onto equivalent CEL, so no existing job changes behaviour, but any out-of-tree caller sending the old fields must move to `cel_expr`.


packages/console-backend/console_backend/services/mcp_tool_client.py
```diff
@@ -0,0 +1,206 @@
...
+    except httpx.HTTPStatusError as exc:
+        raise GatewayError(
+            f"{url} returned HTTP {exc.response.status_code} for '{tool_name}'"
+        ) from exc
+    except httpx.RequestError as exc:
+        raise GatewayError(f"Cannot reach {url}: {type(exc).__name__}") from exc
+
+    elapsed_ms = int((time.perf_counter() - started) * 1000)
+
+    if "error" in data:
```
alloy-cicd: **Bug — key-presence check treats `"error": null` as a failure.** `if "error" in data` fires on a JSON-RPC reply that carries an explicit `"error": null` alongside a valid `"result"` (some server frameworks always emit both members). `(data.get("error") or {}).get("message")` then yields `None`, and line 199 raises `GatewayError("... failed: unknown gateway error")` — so every watch run against such a server fails and the job auto-pauses at `max_failures`, even though the tool answered correctly. `extract_result`'s docstring in this same file explains why key-presence (not truthiness) is right for *content*; here it's the inverse mistake. Suggest `if data.get("error") is not None:`.

alloy-cicd (commented):

Reviewed `feat/scheduler-watch-conditions` at 5c2acf2e (52 files). Findings were generated by multiple independent passes (correctness, removed behavior, cross-file tracing, reuse/simplification, efficiency) and each was adversarially verified against the checked-out head before posting; several candidates were refuted and dropped (e.g. celpy raising RecursionError through the validator, and unbound CEL references firing a watch — both disproven empirically: undeclared references raise `CelEvaluationError`, fail-closed ✅).

**Correctness (12)** — most impactful first:
- Watch conditions can be stripped via PATCH (backend accepts it; the edit page sends it) → job burns tool calls and fails every poll until auto-pause.
- Inline ('automated') sub-agent config is silently discarded on both create and edit — job downgrades to notify-only.
- `{}` check results never persist to `prev`, so `result != prev` fires forever.
- Per-tick stuck-run healer can falsely fail another replica's live >30min run (in-process `_in_flight` set is per-replica).
- JSON-RPC `"error": null` / stray SSE events mis-handled in the new gateway client.
- Typing `=` for a dynamic argument is swallowed — the feature is paste-only.
- Pasted-payload mode loops `/validate-condition` POSTs every ~500ms.
- Stale check results survive dynamic-arg edits (`callSignature` gap); JsonPathPicker emits invalid CEL for non-identifier keys; draft prompt asks for `message` which the model-field filter drops; migration 083 leaves `condition_expr=''` rows conditionless.

**Ecosystem/migration (2)**: the seeded task-scheduler agent prompt (075) still teaches `condition_expr`/`expected_value` — conversational watch creation now fails validation with no reseed in this PR; and `ruff check` fails on agent-runner `core.py` (2× F401) as committed.

**Test coverage (1)**: remote sub-agent `context_id` propagation lost its only test.

**Reuse/simplification/efficiency (10)**: triplicated LLM JSON-extraction parser, two divergent CEL repair loops, duplicated Pydantic fields, hand-rolled API wrappers bypassing the regenerated SDK (AGENTS.md convention), `resolve_timezone` bypass, per-eval CEL recompilation, per-call `httpx.AsyncClient`, full tool re-discovery per unscored invoke (+ `readOnlyHint` as the structural signal), per-poll `model_defaults` reads, CEL timeout not bounding CPU.

Optional (no inline comment): consider a line in root/console-backend AGENTS.md noting that watch evaluation now lives in console-backend's `WatchEvaluator` and agent-runner is watch-agnostic — it's the kind of non-obvious placement the "Maintaining These Instructions" rule covers.

packages/console-backend/console_backend/services/mcp_tool_client.py
```diff
@@ -0,0 +1,206 @@
...
+        message = (data.get("error") or {}).get("message") or "unknown gateway error"
+        # An unknown tool is worth distinguishing: it means the job references something
+        # the gateway no longer exposes, which is a configuration problem, not an outage.
+        if "not found" in message.lower() or "unknown tool" in message.lower():
```
alloy-cicd: **Fragile error classification by English substrings.** `ToolNotFound` is decided by `"not found" / "unknown tool"` appearing in the gateway's message rather than the JSON-RPC error code (e.g. `-32602`), so a tool that legitimately reports "resource not found" is misclassified and surfaces as HTTP 404 from `mcp_router.py:569`. The same pattern repeats at `mcp_router.py:574`, where `504 if "did not respond" in str(exc) else 502` matches the English wording this file composes at line 182 — rewording that timeout message silently turns every check-tool timeout into a 502 with no test failing. Suggest typed subclasses (`GatewayTimeout(GatewayError)`, code-based `ToolNotFound`) so both mapping sites are structural.

packages/console-backend/console_backend/services/mcp_tool_client.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def parse_envelope(response: httpx.Response) -> dict[str, Any]:
+    """Decode a JSON-RPC reply, which arrives as JSON or as a single SSE event."""
+    content_type = response.headers.get("content-type", "")
+
+    if "text/event-stream" not in content_type:
+        return response.json()  # type: ignore[no-any-return]
+
+    for line in response.text.strip().split("\n"):
```
alloy-cicd, line 127: **Bug — SSE envelope parsing takes the first `data:` line, whatever it is.** The loop returns the first parseable `data: ` event with no check of the JSON-RPC `id` (or `event:` field). A server that streams a progress notification or ping before the response makes `parse_envelope` return that object; `call_tool` then sees neither `error` nor `result`, so `rpc_result = {}` (line 201) and the watch condition is silently evaluated against `{}`. Also, `line.startswith("data: ")` misses the SSE-legal `data:` form without a space, which falls through to `GatewayError("Invalid SSE response")`. Suggest skipping envelopes whose `id` doesn't match the request and accepting `data:` with optional space.

packages/console-backend/console_backend/services/mcp_tool_client.py
```diff
@@ -0,0 +1,206 @@
...
+    outright.
+
+    Raises:
+        ToolNotFound: the server rejected the tool name.
+        GatewayError: the server was unreachable, timed out, or errored.
+    """
+    url = _console_mcp_url() if is_console_tool(tool_name) else config.mcp_gateway.url
+    started = time.perf_counter()
+    try:
+        async with httpx.AsyncClient(timeout=timeout) as client:
```
alloy-cicd: **Efficiency — new `httpx.AsyncClient` per tool call.** Every `call_tool` (the scheduler's per-poll hot path, plus every UI "Run check now") pays a fresh DNS lookup, TCP connect and TLS handshake to the same gateway host, with no keep-alive reuse. `app.py:1366` already builds long-lived pooled `AsyncClient`s with `httpx.Limits` — same pattern applies here: one shared client (module-level or on app state, closed at shutdown) with `timeout=` passed per request. `mcp_router._list_mcp_tools` (mcp_router.py:312) has the same per-call client.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -169,8 +195,41 @@ async def _dispatch_job(self, job: ScheduledJob, run_id: int | None = None) -> N
...
+                    watch_outcome = await self._watch_evaluator.evaluate(db, job, access_token)
+
+                    if watch_outcome.error:
+                        await self._finalize(
+                            run_id=run_id,
+                            job=job,
+                            status=JobRunStatus.FAILED,
+                            error_message=watch_outcome.error,
+                            delivered=False,
+                            last_check_result=watch_outcome.check_result,
```
alloy-cicd, line 213: **Bug — an empty-dict check result never becomes `prev`.** `complete_job` (scheduled_job_repository.py:255) persists `last_check_result` via `json.dumps(last_check_result) if last_check_result else None`, so a `{}` result (which `extract_result` produces for a no-content response) maps to NULL and the `COALESCE` at repository line 245 keeps the *old* payload forever. That line predates this PR, but this PR makes it load-bearing: with `cel_expr: "result != prev"`, `prev` can never catch up to `{}`, so the condition is true on every poll and the watch notifies/dispatches every tick until someone disables it. The sibling `fields` dict at repository line 208 already uses `is not None` — the persist path should too.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -128,6 +151,8 @@ async def _loop(self) -> None:
             await asyncio.sleep(self._tick_interval)
 
     async def _tick(self) -> None:
+        await self._heal_stuck_runs()
```
alloy-cicd, line 154: **Bug (multi-replica) — the per-tick healer can fail another replica's live run.** `_heal_stuck_runs` now runs on every tick, and its only exclusion is the in-process `_in_flight` set; there is no lease/heartbeat/owner column on `scheduled_job_runs`. Root AGENTS.md describes console-backend as stateless and horizontally scaled — with two replicas, B's healer sees A's 31-minute-old dispatch (a2a_dispatch has only a 300s per-SSE-event timeout, no total cap, so >30min runs are legitimate) and marks it `failed` with "Run was interrupted…" while the agent is still working. `complete_run` later overwrites the row, but the false failure is user-visible meanwhile, and stays if the finalize write fails. A restart of a single replica has the same effect on its own long run. Secondary: this is also a write transaction (UPDATE…RETURNING + COMMIT) at the head of every 30s tick, serially before any job is claimed — consider sweeping every Nth tick and/or a heartbeat-based liveness check instead of a fixed 30-minute guess.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -292,6 +400,44 @@ async def _build_message_args(
...
+
+        Falls back to reporting the raw result. A watch that triggered has something to
+        say, so an unreachable model must not turn that into silence.
+        """
+        check_result = outcome.check_result if outcome else None
+        if not check_result:
+            return f"The watch '{job.name}' triggered."
+
+        async with self._db_session_factory() as db:
+            defaults = await ModelDefaultsRepository().get_all(db)
```
alloy-cicd, line 418: **Efficiency — near-static `model_defaults` re-read per event, on an extra connection.** `_write_notification` opens a second DB session solely for `ModelDefaultsRepository().get_all(db)` while the dispatch's own session is still open (two pool connections per notification; a saturated pool can wedge this path), and `WatchEvaluator._judge` (watch_evaluator.py:261) repeats the same SELECT once per judged evaluation per poll. This table is near-static configuration — a short-TTL cache (or one read per tick passed down) removes a per-poll query, and `_write_notification` can reuse the caller's session.

packages/console-backend/console_backend/services/cel_condition.py
```diff
@@ -0,0 +1,358 @@
...
+    Raises:
+        CelSyntaxError: the expression does not compile.
+        CelEvaluationError: evaluation failed against this payload, or timed out.
+    """
+    if len(expr) > MAX_CEL_EXPR_LENGTH:
+        raise CelSyntaxError(
+            f"CEL expression is {len(expr)} characters; the maximum is {MAX_CEL_EXPR_LENGTH}."
+        )
+    try:
+        return await asyncio.wait_for(
```
alloy-cicd, line 618: **The 2s timeout doesn't actually bound CPU.** `asyncio.wait_for` only cancels the awaiting coroutine; the `asyncio.to_thread` worker keeps interpreting the expression to completion — there's no stop flag or bounded executor. Since `/validate-condition` accepts an arbitrary payload + expression from any authenticated user, repeated pathological requests (e.g. nested `exists()` comprehensions over a large `result`) saturate the default executor (capped at `min(32, cpu+4)` threads, shared with every other `to_thread` user) and pin the GIL, degrading the event loop and the scheduler tick — the module docstring says the timeout caps exactly this. Worth at least a dedicated small executor for CEL evaluation, or an interpreter-level step/size limit.

packages/console-backend/console_backend/services/cel_condition.py
```diff
@@ -0,0 +1,358 @@
...
+            f"CEL expression is {len(expr)} characters; the maximum is {MAX_CEL_EXPR_LENGTH}."
+        )
+    try:
+        celpy.Environment(annotations=_EXTENSION_ANNOTATIONS).compile(expr)
+    except CELParseError as exc:
+        raise CelSyntaxError(str(exc)) from exc
+
+
+def _evaluate_sync(expr: str, result: Any, now: datetime, prev: Any) -> CelEvaluation:
+    env = celpy.Environment(annotations=_EXTENSION_ANNOTATIONS)
```
alloy-cicd: **Efficiency — fresh `celpy.Environment` + recompile on every evaluation.** `_evaluate_sync` (here), `_evaluate_arg_exprs_sync` (line 260) and `validate_cel_expression` (line 227) each construct a new Environment and re-run the Lark parse — per scheduler poll, per 500ms-debounced `/validate-condition` call, and once per dynamic argument per run; for small expressions the parse *is* the dominant cost. One module-level Environment plus an `lru_cache`d compile keyed on the expression string (bounded, since expressions are capped at 2000 chars) removes it. Related dedup: the `MAX_CEL_EXPR_LENGTH` guard is spelled out three times (lines 222, 313, 345) with three message variants — have the evaluators call `validate_cel_expression` so the rule lives once.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -33,46 +56,292 @@
...
+        "call that tool — no dates, per the rules at the top.\n"
+        "3. `check_args_exprs`: argument name → CEL expression over `now`/`prev`, for "
+        "arguments that must move with time. MANDATORY when the tool has required "
+        "date/time arguments — they must come from here, not from check_args and not "
+        "left empty. Omit it only when no argument involves time.\n"
+        "4. `cel_expr`: the CEL condition, per the rules at the top. Prefer an expression "
+        "that returns the matching items.\n"
+        "5. `llm_condition`: only when part of the condition is genuinely semantic; it "
+        "judges what cel_expr returned. Omit it otherwise.\n"
+        "6. `message`: a concise notification text that will be sent to the user when "
```
alloy-cicd: **Bug — the prompt asks for `message`, but the filter drops it.** Instruction 6 tells the model to emit a field named `message`, and the dict comprehension at ~line 405 keeps only keys in `ScheduledJobDraft.model_fields` — which has `notification_message`, not `message`. A model that follows the numbered instruction (rather than the few-shot example at line 344, which uses the correct key) gets its generated notification text silently discarded; the frontend then reads `result.notification_message` as undefined and leaves the field empty, so the eventual trigger falls back to an extra LLM call in `_write_notification` or the generic "The watch '<name>' triggered." Rename instruction 6 to `notification_message`.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -33,46 +56,292 @@
...
+
+    # The model is told the time, in the user's timezone. Without it, any request that
+    # implies a date ("this week's meetings") gets one invented from training data —
+    # which is how generated check_args ended up watching windows in the past.
+    try:
+        user_tz = await _get_scheduler_service(request)._resolve_timezone(db, None, current_user.id)
+    except (ValueError, RuntimeError):
+        user_tz = "UTC"
+    try:
+        local_now = datetime.now(timezone.utc).astimezone(ZoneInfo(user_tz))
```
alloy-cicd: **Reuse — bypasses `resolve_timezone`.** This hand-rolls `ZoneInfo(user_tz)` + try/except + UTC fallback, but `console_backend.utils.timezones.resolve_timezone` already handles blank/None via `default_timezone_name()` (the `DEFAULT_TIMEZONE` env fallback) and normalises `ZoneInfoNotFoundError` — and `watch_evaluator.py:82` uses it. As written, a deployment whose `DEFAULT_TIMEZONE` isn't UTC gets UTC in the generation prompt here while the scheduler evaluates the job in the deployment zone — the prompt and the run disagree on the clock.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -88,30 +357,345 @@ async def generate_watch_params(
...
+            instruction,
             model=model,
             max_tokens=1024,
             metadata={"user_sub": current_user.sub},  # OIDC subject — the gateway/proxy attributes by sub, not internal id
         )
         # Strip optional markdown fences and extract the JSON object.
-        text = re.sub(r"```(?:json)?\s*", "", text).strip().rstrip("`")
-        match = re.search(r"\{.*\}", text, re.DOTALL)
-        result = json.loads(match.group()) if match else {}
+        cleaned = re.sub(r"```(?:json)?\s*", "", text).strip().rstrip("`")
```
alloy-cicd: **Duplication — this fence-strip + `re.search(r"\{.*\}", …, re.DOTALL)` JSON extractor exists three times.** Here, again at lines 526-528 (`generate_condition._generate`), and in `watch_evaluator._judge` (watch_evaluator.py:291-293) — character-for-character the same logic. Any robustness fix (new fence style, prose-wrapped output) has to be made three times, and a divergence means a draft generation can succeed while the watch judge silently reads `{}`. Suggest one helper next to `gateway_chat` (e.g. `gateway_chat_json(...)` in services/llm_gateway.py) used by all three.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -33,46 +56,292 @@
...
+    "Reserve `llm_condition` for judgement that is genuinely semantic (tone, intent, "
+    "\"looks like a company address\") — never for arithmetic, counting, time windows "
+    "or string matching. The two COMPOSE: when both are set, the CEL gate runs first "
+    "and the model judges only what the expression returned. Use that split whenever a "
+    "condition has a mechanical part and a semantic part.\n"
+    "Every watch needs cel_expr, llm_condition, or both.\n\n"
+)
+
+
+async def _repair_cel(
```
alloy-cicd, line 143: **Duplication — two independent generate-validate-repair loops for LLM-written CEL.** `_repair_cel` (compile-only, one retry, then falls back to dropping the expression in favour of `llm_condition`) serves generate-job-draft, while `generate_condition` has its own `_verify_candidate` loop (lines ~434-559: compiles *and* evaluates against the sample payload, feeds the error back, `_GENERATE_CONDITION_RETRIES` attempts, ends with a note). They already disagree on retry count, verification depth and failure behaviour, so improvements to one path silently don't reach the other — e.g. a drafted `cel_expr` referencing a field the real response lacks ships into the form, though the condition path would have caught it. Suggest generate_job_draft reuse `_verify_candidate` (with `payload=None` when no sample exists).

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -215,8 +801,8 @@ async def update_job(
                 data.notification_message if "notification_message" in data.model_fields_set else _UNSET
             ),
             check_tool=data.check_tool if "check_tool" in data.model_fields_set else _UNSET,
-            condition_expr=data.condition_expr if "condition_expr" in data.model_fields_set else _UNSET,
-            expected_value=data.expected_value if "expected_value" in data.model_fields_set else _UNSET,
+            check_args_exprs=data.check_args_exprs if "check_args_exprs" in data.model_fields_set else _UNSET,
+            cel_expr=data.cel_expr if "cel_expr" in data.model_fields_set else _UNSET,
```
alloy-cicd, line 821: **Bug — PATCH can strip a watch of both conditions.** `ScheduledJobUpdate` has per-field CEL validators but no cross-field rule, `update_job` (scheduler_service.py:259) writes explicit nulls straight through, and migration 083's CHECK only guards `check_tool` — so `{"cel_expr": null, "llm_condition": null}` on a watch returns 200 and produces exactly the state the create validator rejects and `watch_evaluator.py:167` calls impossible ("reaching here means the row predates the CEL migration"). The frontend hits this for real: SchedulerJobDetailPage sends `trim() || null` for both fields with no both-empty guard, so emptying the two inputs and saving yields a job that calls its check tool every poll (billable, side-effecting for non-read tools) and records a failed run each time until auto-pause. The invariant needs enforcing where writes converge — post-merge in `update_job`, or a DB CHECK mirroring the `check_tool` one.

packages/console-backend/console_backend/models/scheduled_job.py
```diff
@@ -32,7 +35,42 @@ class JobRunStatus(str, Enum):
...
+    #: What the CEL gate decided, recorded separately from `met` because in
+    #: "cel+judge" mode the gate can pass while the model still says no — and a run
+    #: must show which stage made the call. CEL modes only.
+    gate_met: bool | None = None
+    #: Historical "rule" runs only, so those records read on their own terms.
+    operator: str | None = None
+    expected_value: str | None = None
+    #: Rule mode only, so a run can be read without going back to the job — which may
+    #: have been edited since.
+    operator: str | None = None
```
alloy-cicd: **Bug (latent) — `operator` and `expected_value` are declared twice.** Lines 68-69 and 72-73 declare the same two fields, each pair with its own comment; Pydantic silently keeps the last declaration. An edit to the first pair's type or default would have no effect. Delete one pair and keep a single comment.

packages/console-backend/console_backend/routers/mcp_router.py
```diff
@@ -423,6 +444,157 @@ async def list_mcp_tools(
...
+        known=worst is not None,
+    )
+
+    if base_score is not None:
+        needs_ack = base_score >= _READ_ONLY_RISK_CEILING
+        reason = f"is scored {base_score:.2f}, above the read-only threshold of {_READ_ONLY_RISK_CEILING}"
+    else:
+        # Nothing has scored this tool: the table is filled lazily by agent runs, so
+        # most tools are absent. Fall back to what the gateway says the endpoint does.
+        needs_ack = not await _describes_a_read(request, user, tool_name)
```
alloy-cicd, line 537: **Efficiency + design — unscored tools trigger a full gateway re-discovery, keyed on description prose.** When `base_score` is None (most tools, since the risk table fills lazily), `_describes_a_read` re-runs the entire `_list_mcp_tools` — token exchange, full `tools/list` over hundreds of tools, `clean_gemini_schema` on every schema — to read one tool's description and prefix-match `("GET ", "HEAD ", "OPTIONS ")`. Every "Run check now" pays that round trip, and a user iterating on arguments pays it per attempt; there is no caching anywhere in this router. Two improvements: (1) cache the listing behind a short TTL or look the tool up from an already-fetched copy; (2) the read-only signal already exists structurally — MCP `tools/list` carries `annotations.readOnlyHint` — which avoids both the false 428 on every search/read tool whose description doesn't start with an HTTP verb (training users to always tick `acknowledge_risk`) and the unconfirmed invoke of any destructive tool whose description happens to open with "GET ".

packages/console-backend/sqlmigrations/ddl/083_cel_watch_conditions.sql
```diff
@@ -0,0 +1,74 @@
...
+-- Pre-operator rows without an expected value were "the value is not null"; the
+-- non-boolean gate (non-empty) is that exact check.
+UPDATE scheduled_jobs
+SET cel_expr = 'jsonpath(result, "'
+        || replace(replace(condition_expr, '\', '\\'), '"', '\"')
+        || '")'
+WHERE job_type = 'watch'
+  AND cel_expr IS NULL
+  AND (llm_condition IS NULL OR llm_condition = '')
+  AND condition_expr IS NOT NULL AND condition_expr <> '';
```
alloy-cicd, line 44: **Migration gap — rows with `condition_expr = ''` end up with no condition at all.** Both backfill UPDATEs require `condition_expr <> ''`, but the old CHECK (034) only required NOT NULL and the old PATCH path stored `''` unvalidated, so such rows are reachable. Before this PR, agent-runner treated an empty `condition_expr` as condition-met on every poll (old `agent/core.py:887`); after the migration the same row has neither `cel_expr` nor `llm_condition`, so `WatchEvaluator` fails it on every poll until auto-pause — a working (if noisy) job silently becomes a failing one. Consider a third backfill for the `= ''` case (e.g. `cel_expr = 'true'` to preserve old behaviour, or at least a migration-time notice).

packages/console-backend/sqlmigrations/ddl/083_cel_watch_conditions.sql
```diff
@@ -0,0 +1,74 @@
...
+
+-- A watch needs a tool and a condition (cel_expr, llm_condition, or both); the API
+-- enforces the condition half, the constraint keeps requiring the tool.
+ALTER TABLE scheduled_jobs DROP CONSTRAINT IF EXISTS scheduled_jobs_watch_requires_check;
+ALTER TABLE scheduled_jobs
+    ADD CONSTRAINT scheduled_jobs_watch_requires_check
+    CHECK (job_type = 'task' OR check_tool IS NOT NULL);
+
+ALTER TABLE scheduled_jobs DROP COLUMN IF EXISTS condition_expr;
```
alloy-cicd, line 65: **Missing reseed — the task-scheduler sub-agent's prompt still teaches the removed API.** The seeded prompt from 075 (lines 85, 120-131) instructs the model to produce `condition_expr` (JSONPath) + `expected_value` for watch jobs; this PR drops that column and `ScheduledJobCreate.validate_consistency` now rejects a watch without `cel_expr`/`llm_condition`. The tool *schema* the agent sees was updated, but an agent following its seeded step-by-step prompt emits the old fields (silently ignored) and no `cel_expr`, so conversational watch creation ("tell me when PR #123 is merged") fails validation with no prompt guidance on how to recover. This PR needs a reseed migration updating that prompt.

packages/agent-runner/agent/core.py
```diff
@@ -12,7 +12,8 @@
 
 Execution flow per call:
 1. Extract scheduler metadata from the A2A message (task.history)
-2. For watch jobs: call the check_tool via MCP and evaluate the JSONPath condition
+2. Nothing watch-specific: the scheduler decides whether a watch acts and what it says,
```
alloy-cicd, line 15: **Lint break + dead code left by the watch-eval removal.** `ruff check agent/core.py` fails as committed: `F401 get_default_fast_model imported but unused` (line 47) and `F401 SystemMessage imported but unused` (line 59) — pyproject selects "F", so the lint gate on this package is red. Also now unreferenced in production code: `ConditionEvaluationResult` (line 83), `GeneratedMessage` (line 90), and `_current_time_context` (line 179 — note this last one is still imported by `tests/test_metadata_extraction.py`, so removing it means removing that test too).

packages/agent-runner/tests/test_agent_runner_security.py
```diff
@@ -182,176 +182,77 @@ def test_watch_metadata_included(self):
         assert result["watch"] == watch_cfg
 
 
-class TestWatchConditionNotMet:
-    """When watch condition is not met, _stream_impl returns early with condition_not_met status."""
+class TestDispatchShapes:
```
alloy-cicd, line 185: **Test coverage lost — context-id propagation for remote sub-agents is now untested.** The rewrite deleted `TestRemoteAgentContextPropagation` (asserting `input_data.orchestrator_conversation_id` is carried and `_execute_sub_agent` forwards `context_id=` to `_run_remote_agent`), and no remaining test in the package mentions `context_id`/`orchestrator_conversation_id`. That behaviour is unchanged by this PR but load-bearing: a future edit dropping the kwarg would pass CI, and at runtime the remote agent would checkpoint under a different id than `scheduled_job_runs.conversation_id`, orphaning the run's thread. Please re-add equivalent coverage.

packages/console-frontend/src/components/McpToolArgsFields.tsx
```diff
@@ -0,0 +1,218 @@
...
+                    aria-invalid={invalid || undefined}
+                    className={cn('bg-background', expr !== undefined && 'font-mono text-xs')}
+                    inputMode={param.type === 'string' || expr !== undefined ? undefined : 'decimal'}
+                    value={shown}
+                    placeholder={param.placeholder ?? param.description ?? ''}
+                    onChange={(e) => {
+                      const text = e.target.value;
+                      if (text.startsWith('=')) {
+                        set(param.key, undefined);
+                        setExpr(param.key, text.slice(1).trimStart());
```
alloy-cicd, line 210: **Bug — dynamic-argument mode is unreachable by typing.** Typing `=` as the first character: onChange sees `"="`, takes the expression branch, calls `set(key, undefined)` (deletes the static value) and `setExpr(key, "".trimStart())` — and `setExpr` treats `''` as delete (line 68). State is unchanged, so the controlled input re-renders with `shown === ''` and the `=` disappears; continuing to type produces a literal string value instead of a CEL expression. The `= expr` syntax — the whole per-argument CEL feature in the fields editor — only works by pasting the complete string in one input event. `setExpr` needs to keep the empty-string state (or the input needs local text state) so the user can type past the `=`.

packages/console-frontend/src/components/ConditionTester.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+  })();
+
+  const payload = source === 'mock' ? mock.value : liveResult;
+  const hasPayload = payload !== undefined;
+  // A compile error does not depend on the data, so an expression can be checked
+  // before there is any: validating against {} reports whether it is even legal
+  // syntax, which is the failure that otherwise surfaces only when the job silently
+  // never fires.
+  const subject = useMemo(() => (hasPayload ? payload : {}), [hasPayload, payload]);
```
alloy-cicd, line 66: **Bug — infinite `/validate-condition` request loop in "Pasted payload" mode.** The pasted payload is produced by a bare IIFE running `JSON.parse(mockText)` on every render (lines 46-56), so `payload` — and therefore this `subject` memo — gets a new identity each render. The validate effect's deps (line 99) include `subject`, so: effect fires → 500ms debounce → POST → `setOutcome(new object)` → re-render → new parsed payload → new `subject` → effect re-fires — one backend call (an LLM call when `llm_condition` is set) every ~500ms+RTT, forever, until the mode is switched or the component unmounts. `payloadKey` being in the deps doesn't save it because `subject` defeats it. Fix: memoize the parse on `[mode, mockText]` and key the effect on `payloadKey` + the primitive inputs only. While here: `payloadKey = JSON.stringify(payload)` itself (line 68) runs unmemoized in the render body over a payload that can be 64KB — wrap it in `useMemo` so typing elsewhere in the watch form doesn't re-serialize it per keystroke.

packages/console-frontend/src/components/ConditionTester.tsx
```diff
@@ -0,0 +1,240 @@
...
+      {shown && !shown.valid && (
+        <div className="grid gap-1.5">
+          <span className="text-destructive flex items-start gap-1.5 text-xs">
+            <X className="mt-0.5 size-3.5 shrink-0" />
+            <span>
+              This expression cannot be parsed, so the job would never trigger.
+              <span className="mt-0.5 block font-mono text-[11px] opacity-80">{shown.error}</span>
+            </span>
+          </span>
+          {(shown.notes ?? []).map((note) => (
```
alloy-cicd: **Minor — triplicated notes block.** This `{(shown.notes ?? []).map(...)}` block is byte-identical at lines 185-190, 203-208 and 230-235, one per mutually-exclusive outcome branch — an icon/markup tweak needs three edits 20+ lines apart. Hoist it into one block rendered after the outcome branches. Related: the `tone` union has three members but `'unmet'` and `'muted'` map to the same classes (lines 219-220), so the union can shrink to `'met' | 'muted'`.

packages/console-frontend/src/pages/SchedulerPage.tsx
```diff
@@ -408,16 +550,22 @@ function CreateJobDialog({
...
-      if (form.sub_agent_id) {
+      // The two halves of the condition: the expression gates deterministically, the
+      // judgement is the semantic stage on what it returned. At least one is set —
+      // validated above and by the API.
+      body.cel_expr = form.cel_expr.trim() || undefined;
+      body.llm_condition = form.llm_condition.trim() || undefined;
+
+      // The two outcomes are exclusive: a sub-agent's reply replaces the
+      // notification, so sending both would leave one of them dead.
+      if (form.outcome === 'agent' && form.sub_agent_id) {
```
alloy-cicd: **Bug — inline sub-agent configuration is silently discarded on watch creation.** WatchFields offers `sub_agent_mode: 'automated'` (define an agent inline: name, model, system prompt, tools), but the watch branch of handleSubmit only checks `form.outcome === 'agent' && form.sub_agent_id` — `sub_agent_parameters` is only ever built in the `job_type === 'task'` branch (line 531). With mode `automated`, `sub_agent_id` is `''`, so the else branch sends `notification_message` and the whole inline agent definition is dropped with no error; the backend supports it (`ScheduledJobCreate.sub_agent_id` docs: "…through sub_agent_parameters, for either job type"). Also, unlike the task branch (line 469), the watch validations never reject `outcome === 'agent'` with an empty `sub_agent_id`, so "existing" mode with no agent picked silently downgrades to a notification too.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -472,23 +465,33 @@ function EditForm({ job }: { job: ScheduledJob }) {
...
-        sub_agent_id: subAgentId
-          ? parseInt(subAgentId)
-          : job.job_type === 'watch'
-            ? null
-            : undefined,
-      }),
+      ...(job.job_type === 'watch'
+        ? watch.sub_agent_id !== initialSubAgentId && {
+            sub_agent_id:
+              watch.outcome === 'agent' && watch.sub_agent_id ? parseInt(watch.sub_agent_id) : null,
```
alloy-cicd: **Bug — same inline-agent drop on the edit path.** handleSave's watch branch sends only `sub_agent_id` (null when `watch.sub_agent_id` is empty) and never `sub_agent_parameters`, so switching an existing watch's outcome to "agent" with an inline ('automated') definition PATCHes `sub_agent_id: null` and the job stays notify-only — the configuration is lost without an error. (The companion problem — this same save path sending `cel_expr: null, llm_condition: null` with no both-empty guard — is noted on the backend PATCH endpoint, which accepts it.)

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -407,18 +386,32 @@ function EditForm({ job }: { job: ScheduledJob }) {
...
-      if (result.expected_value) { setExpectedValue(result.expected_value); touch(); }
-      if (result.notification_message) { setMessage(result.notification_message); touch(); }
+      const draft = await generateJobDraft(mcpTools as unknown as Record<string, unknown>[], aiQuery);
+      setWatch((w) => {
+        const next = { ...w };
+        if (draft.check_tool) next.check_tool = draft.check_tool;
+        if (draft.check_args) {
+          next.check_args = draft.check_args as Record<string, unknown>;
+          next.check_args_text = JSON.stringify(draft.check_args, null, 2);
+          next.args_mode = 'fields';
```
alloy-cicd: **Duplication with divergence — `handleAiGenerate` re-implements the create dialog's draft-application and loses its guard.** SchedulerPage's version computes `renderable` from `parseToolSchema(...).params` and picks `args_mode: 'fields' | 'json'` so every generated argument stays visible — with a comment explaining that an arg outside the schema-declared scalars "would be submitted without ever being visible". This copy hardcodes `next.args_mode = 'fields'`, reintroducing exactly that failure for a draft with a nested/undeclared arg. Every future draft field will have to be wired twice. Suggest one shared `applyWatchDraft(draft, current, tools)` in `src/lib/watchArgs.ts` (next to `resolveArgs`), with the create dialog layering its name/schedule handling on top.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -1146,35 +1023,42 @@ export function SchedulerJobDetailPage() {
...
+                )
               )}
             </div>
           )}
 
+          {/* Why the last run did what it did. Above the configuration because on this
+              page that is the question being asked, and the answer is already stored. */}
+          {job.job_type === 'watch' && (
+            <LastCheckPanel
+              run={runs.find((r) => r.condition_evaluation) ?? runs[0]}
```
alloy-cicd, line 1081: **Minor — verdict and response can come from different runs.** `run={runs.find(r => r.condition_evaluation) ?? runs[0]}` picks the newest run *with an evaluation*, while `result={job.last_check_result}` is always the job's current payload — so an older run's verdict can be displayed against a newer run's response. Consider showing the payload only when the selected run is the latest, or storing/reading the payload per run.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -407,18 +386,32 @@ function EditForm({ job }: { job: ScheduledJob }) {
...
+        }
+        if (draft.destroy_after_trigger != null) {
+          next.destroy_after_trigger = draft.destroy_after_trigger;
+        }
+        return next;
+      });
+      touch();
     } catch {
-      setError('AI generation failed. Please fill in the fields manually.');
+      setError('AI generation failed. Please edit the fields manually.');
```
alloy-cicd: **Minor — the real error is swallowed.** Bare `catch {` reduces every generate-draft failure (403, timeout, validation) to the fixed "AI generation failed" string. `formatApiError` from `src/api/scheduler.ts` is the established way to surface these; `catch (e)` + including the message would make failures diagnosable. (Same pattern exists in SchedulerPage.tsx:429-430, pre-existing.)

packages/console-frontend/src/components/WatchFields.tsx
```diff
@@ -0,0 +1,643 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [argsExprsKey, hasArgExprs, staticArgsKey]);
+
+  const selectedTool = mcpTools.find((t) => t.name === value.check_tool);
+  const toolSchema = useMemo(() => parseToolSchema(selectedTool), [selectedTool]);
+  const missingArgs = submitMissingArgs;
+
+  // A result describes one specific call. When the tool or its arguments change it stops
+  // describing anything, so it is dropped here rather than every caller remembering to.
+  const callSignature = `${value.check_tool}|${JSON.stringify(value.check_args)}|${value.check_args_text}`;
```
alloy-cicd: **Bug — `callSignature` omits `check_args_exprs`, so a stale check result survives edits to dynamic arguments.** `runCheck` resolves the expressions into `callArgs` (lines 190-200), so the result genuinely depends on them, but the signature is only `check_tool | check_args | check_args_text`. Editing an `= …` expression after a run leaves `check.signature === callSignature`, so `liveResult` keeps returning the old payload — which feeds `testable`, JsonPathPicker, the CEL editor payload and ConditionTester, i.e. the condition is tested against a payload the job would no longer produce. `argsExprsKey` is already computed at line 107; add it to the signature. Related inconsistencies: the status line at 347-348 reads raw `check.result`/`check.elapsedMs` (bypassing the signature entirely, so "Responded in Xms" survives even tool changes that *do* invalidate `liveResult`), and line 140 `const missingArgs = submitMissingArgs;` is a pure alias that can go.

packages/console-frontend/src/components/JsonPathPicker.tsx
```diff
@@ -0,0 +1,265 @@
...
+          depth,
+          label,
+          value: '{',
+          path: allowObjects ? path : null,
+          tone: 'punctuation',
+        })
+      )
+        return;
+      entries.forEach(([k, v], i) =>
+        child(v, `${path}.${k}`, `"${k}": `, i < entries.length - 1 ? ',' : '', depth + 1)
```
alloy-cicd: **Bug — dot-only path building produces invalid/wrong CEL for non-identifier keys.** Paths are built as `${path}.${k}` (here) and `$.${k}` (line 167) with no identifier check, and `jsonPathToCel` (src/lib/watchCondition.ts:16) is a pure prefix swap. So clicking a key like `content-type` seeds `result.content-type` (CEL parses it as subtraction or errors), `a.b` becomes the wrong two-segment path, and a numeric key `2` becomes unparseable `result.2` — the click-to-build affordance emits a condition that never compiles, with no hint that `result["content-type"]` was needed. Bracket-quote any segment that isn't a bare identifier (and keep `jsonPathToCel` in sync).

packages/console-frontend/src/components/JsonPathPicker.tsx
```diff
@@ -0,0 +1,265 @@
...
+      path: null,
+      tone: 'punctuation',
+    });
+  }
+  return lines;
+}
+
+export function JsonPathPicker({
+  value,
+  allowObjects,
```
alloy-cicd: **Minor — both configuration props are dead.** `allowObjects` is literal `true` at both call sites (WatchFields.tsx:384, LastCheckPanel.tsx:110), so the `: null` branches (lines 77/88) never run — and its docstring rationale (JSONPath equality semantics) was obsoleted by CEL in this same PR. `selectedPath` is `""` or omitted everywhere, so `isSelected` (line 225) is always false and the highlight classes at 247 are unreachable. Suggest dropping both props until a caller actually needs them.

packages/console-frontend/src/components/LastCheckPanel.tsx
```diff
@@ -0,0 +1,194 @@
...
+                </span>
+              </div>
+            )}
+
+            {evaluation && !judged && !cel && (
+              <div className="grid gap-1">
+                <span className="text-muted-foreground text-[11px]">Compared with</span>
+                <span className="font-mono text-xs">
+                  {evaluation.operator}
+                  {evaluation.expected_value ? ` ${evaluation.expected_value}` : ' (no value needed)'}
```
alloy-cicd: **Minor — truthiness check misrenders an empty-string `expected_value`.** `expected_value` is `string | null`, so a legacy rule-mode run whose operator required a value that was recorded as `""` renders as `(no value needed)` — stating the operator took no operand when the run did compare. `expected_value != null ? ... : ...` says what's meant.

packages/console-frontend/src/api/scheduler.ts
```diff
@@ -88,30 +132,165 @@ export async function deleteDeliveryChannel(id: number): Promise<void> {
...
+  evaluation?: { gate: boolean; extracted: unknown } | null;
+  notes: string[];
+}
+
+/**
+ * Write or refine just the condition of a watch. Narrower than generateJobDraft: it
+ * sees the real response shape, so it writes field paths that exist — and the backend
+ * compiles, evaluates and repairs every candidate before returning it.
+ */
+export async function generateCondition(body: {
```
alloy-cicd: **Convention — these wrappers bypass the generated SDK that this same PR regenerated.** `generateCondition` (here), `validateArgsExpr` (203), `invokeMcpTool` (240), `validateCondition` (279) — and `generateJobDraft` above — are hand-rolled `(client as any).post({url: ...})` calls with hand-written result types, while `sdk.gen.ts` now exports typed operations for all of these routes (`generateConditionApiV1SchedulerGenerateConditionPost` etc., lines ~473/1803-1845) with matching types in `types.gen.ts`. The file header's rationale ("endpoints whose bodies the generated SDK cannot yet describe") is no longer true for any of them, and `packages/console-frontend/AGENTS.md` says to use the generated SDK for all backend calls. As written, a backend field rename regenerates the types without a single TS error at these call sites, and the URL strings duplicate the router. Only the 428→`McpToolRiskError` mapping and `formatApiError` need to stay hand-written — wrap the generated calls instead.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (5c2acf2ea7949fbb2fefbf74122694c220df3c7e..74008a1ad738117d7b96b392b13a855664c21536)

Each round-1 finding was re-verified against the new head (fix ledger below); the delta itself was then reviewed for regressions. Touched backend tests pass locally (165 in console-backend, 14 in agent-runner), `ruff check` on agent-runner is clean, and frontend `tsc`/eslint are clean on the changed files.

**Verified-fixed (22 of 26):**
- MCP client: `"error": null` handling, id-matched SSE parsing (+ `data:` without space), typed `GatewayTimeout`/code-based `ToolNotFound`, pooled httpx client — all with new tests.
- `{}` check results now persist (`is not None`), so `result != prev` behaves.
- PATCH can no longer strip both watch conditions (effective-pair guard in `scheduler_service`, `_UNSET`-aware, tested).
- Migration 083 backfills `condition_expr = ''` rows to `cel_expr = 'true'` (old behavior preserved); new migration 085 reseeds the task-scheduler prompt for the CEL API, with a test pinning the replace() targets to 075's literal.
- Draft prompt asks for `notification_message`; `resolve_timezone` reused; `gateway_chat_json` replaces all three inline JSON extractors; the two CEL repair loops share `_verify_candidate` + retry count; one module-level CEL Environment with cached compile and a deduplicated length guard.
- CEL resource bounding reworked as upfront constraints (1 MiB payload caps on `result`/`prev`, dedicated 4-thread executor) rather than the timeout — a sound alternative to round 1's concern, with `TestResourceCeilings` covering it.
- Rule mode dropped entirely (also resolves the duplicated `operator`/`expected_value` fields and the `(no value needed)` rendering); generated types regenerated consistently.
- agent-runner: F401s and dead classes gone; `TestRemoteAgentContextPropagation` coverage restored.
- Frontend: `=` typing works via a drafts layer (but see new finding), pasted-payload validate loop fixed, inline sub-agent sent on create with validation, shared `argsModeFor` in `lib/watchArgs.ts`, bracket-quoted JSON paths end-to-end (`result["content-type"]`), dead JsonPathPicker props removed, notes block deduplicated, `formatApiError` surfaced in the generate catch, all five API wrappers on the generated SDK with the 428→`McpToolRiskError` mapping intact.

**Partial (2)** — inline comments on this round:
- Detail-page edit path: `sub_agent_parameters` now sent, but the create dialog's `agentActionError` validation is missing, so the silent notify-only downgrade persists on edit.
- `WatchFields.callSignature` now includes expressions, but the status line still reads raw `check.result`.

**Deferred by author (1):** multi-replica stuck-run healer → #167 with the analysis attached. Agreed — not re-raised.

**Still-open, non-blocking (3, follow-up material):** `_describes_a_read` still re-runs full tool discovery per unscored invoke and classifies by description prefix rather than `readOnlyHint`; `model_defaults` still re-read per notification/judged evaluation; `LastCheckPanel` run/result pairing on the detail page (~line 1070) unchanged. All efficiency/display-level; fine as follow-ups alongside #167.

**New findings in the delta (2, inline):** the drafts layer can display an expression that won't be submitted (stale after AI-generate/tool switch, or a lingering bare `=`), and `_repair_cel`'s candidate merge resurrects an omitted `cel_expr`, wasting a retry.

packages/console-frontend/src/components/McpToolArgsFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+                </SelectContent>
+              </Select>
+            ) : (
+              (() => {
+                // Spreadsheet convention: a value starting with `=` is a CEL
+                // expression over `now` and `prev`, resolved fresh on every run —
+                // how a rolling date window lives in a date argument.
+                const expr = exprs[param.key];
+                const shown =
+                  drafts[param.key] ??
```
alloy-cicd: **New (introduced by the `=`-typing fix) — drafts are never invalidated by external state changes, so the field can display an expression that will not be submitted.** `drafts[param.key] ?? …` shadows the real state, and a draft is only cleared by typing non-`=` text *in that same field*. Two concrete desyncs: (1) type `= now` (draft `"= now"`, expr `now`), then apply an AI-generated draft or switch the check tool — `check_args_exprs`/`check_args` are replaced externally while `drafts` survives, so the input keeps showing `= now` while the stored (and submitted) expression is the generated one, or nothing. (2) Leave a field at bare `=` or `= ` and click away: nothing is stored (`setExpr('')` deletes the key), but the draft renders `=` indefinitely, so the field looks configured while the argument is unset. A `useEffect` clearing `drafts` when `tool`/`exprs` change identity externally (or deriving the draft's validity from `exprs[key]`) closes both.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -472,23 +478,40 @@ function EditForm({ job }: { job: ScheduledJob }) {
...
-          : job.job_type === 'watch'
-            ? null
-            : undefined,
-      }),
+      ...(job.job_type === 'watch'
+        ? watch.outcome === 'agent' && watch.sub_agent_mode === 'automated'
+          ? // An inline definition is a valid outcome for a watch; it used to be dropped
+            // here, so switching to "agent" with one silently PATCHed sub_agent_id: null
+            // and left the job notify-only.
+            { sub_agent_parameters: automatedSubAgentParameters(watch) }
```
alloy-cicd, line 497: **Partial fix — the edit path now sends `sub_agent_parameters`, but still lacks the create dialog's validation.** `agentActionError` is only called on SchedulerPage; here, outcome "agent" in `existing` mode with no sub-agent selected still saves silently — `sub_agent_id` is omitted/null and line ~512 sends `notification_message: null`, so the job quietly stays (or becomes) notify-only, which is the same silent downgrade round 1 flagged on create. An incomplete inline definition (empty name/system prompt) also reaches the backend as a raw 422 instead of the inline message the create dialog shows. Calling `agentActionError(watch)` in handleSave closes both.

packages/console-frontend/src/components/WatchFields.tsx
```diff
@@ -0,0 +1,645 @@
...
+  const toolSchema = useMemo(() => parseToolSchema(selectedTool), [selectedTool]);
+  const missingArgs = submitMissingArgs;
+
+  // A result describes one specific call. When the tool or its arguments change it stops
+  // describing anything, so it is dropped here rather than every caller remembering to.
+  // argsExprsKey belongs in it: runCheck resolves the `= …` expressions into the call, so
+  // editing one changes what the tool is called with. Without it the old payload survived
+  // the edit and went on feeding the picker, the CEL editor and the tester — testing the
+  // condition against a response the job would no longer produce.
+  const callSignature = `${value.check_tool}|${JSON.stringify(value.check_args)}|${value.check_args_text}|${argsExprsKey}`;
```
alloy-cicd, line 148: **Partial fix — signature now includes `argsExprsKey` (good), but the status line still reads raw `check.result`.** Lines ~350-354 render "Responded in Nms" from `check.result`/`check.elapsedMs` without the signature check, so after editing an argument the result panel correctly disappears while the status line keeps describing the stale call next to a "Run check now" button. Reading `liveResult` there (as the panel and button label do) finishes the round-1 finding.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -6,73 +6,386 @@
...
+            "Return the whole JSON object again with a cel_expr that is valid CEL over "
+            "`result`, `now` and `prev`. If the condition cannot be expressed in CEL, omit "
+            "cel_expr and put the judgement in `llm_condition`."
+        )
+        try:
+            retried = await generate(retry_prompt)
+        except Exception:
+            logger.warning("Retry for an unusable cel_expr failed", exc_info=True)
+            break
+        candidate = {**candidate, **retried}
```
alloy-cicd, line 203: **Minor (new, in the unified repair loop) — `{**candidate, **retried}` resurrects the abandoned `cel_expr`.** The retry prompt invites the model to omit `cel_expr` when the condition can't be expressed in CEL, but merging the retry over the old candidate keeps the stale uncompilable expression, so `_verify_candidate` re-fails on the identical error and the remaining retry is burned before the judged fallback is reached — one wasted gateway call per affected draft. `generate_condition` replaces the candidate wholesale each retry; doing the same here (or dropping `cel_expr` from the base when the retry omits it) makes the two paths behave alike on the omit case.

packages/console-backend/console_backend/services/cel_condition.py
```diff
@@ -0,0 +1,594 @@
...
+#: prices the iteration itself. See MeteredEvaluator.
+#:
+#: Sized from measurement (2026-08-25, this machine, ~2.2us per step): the realistic
+#: worst case is a filter over a large tool response, which costs ~65 steps per item —
+#: 32.5k steps (61ms) over 500 events, 130k (245ms) over 2000, the most a response under
+#: MAX_CEL_PAYLOAD_BYTES is likely to carry. 300k leaves that headroom while stopping a
+#: quadratic comprehension after ~650ms, comfortably inside CEL_EVAL_TIMEOUT_SECONDS so
+#: the caller gets this specific error rather than an unhelpful timeout. Re-tune from the
+#: warnings CEL_STEP_WARN_RATIO emits, not from this comment.
+MAX_CEL_STEPS = 300_000
```
alloy-cicd, line 147: **Low / non-blocking — under load the timeout fires before the step budget, which defeats the meter's stated purpose.** The comment sizes 300k steps at ~2.2µs/step so the quadratic shape trips at ~650ms, "comfortably inside CEL_EVAL_TIMEOUT_SECONDS". Re-measured here (machine load avg ~22, so partly environmental): the quadratic `result.items.filter(i, result.items.exists(j, j == i))` over 2000 items cost ~7-10µs/step — 4 runs gave 1× `CelBudgetExceededError` at 1.9s and 3× the generic timeout `CelEvaluationError` at 2.02-2.04s; without the timeout the budget fired at 2.9s. So on a busy host the author sees "exceeded 2.0s — simplify…" instead of the specific step-budget error, and the executor thread keeps running ~1s past the timeout. Work is still bounded (the meter always fires eventually), so this is tuning/observability, not a correctness bug — but the two ceilings are within a factor of ~3 of each other, which the comment's "comfortably" doesn't survive. Consider a lower `MAX_CEL_STEPS` (~100k still leaves 2000-event filters headroom, measured ~45 steps/item here) or a timeout with more margin, so the meter is reliably the first ceiling hit; `CEL_STEP_WARN_RATIO` logs will tell you which.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (74008a1ad738117d7b96b392b13a855664c21536..24bab42a979f7122f5216b7c77c5512a9d74b416)

**Round-2 ledger — all 4 verified-fixed:**
- Detail-page edit path now calls `agentActionError(watch)` before building the PATCH body — empty "existing" selection and incomplete inline definitions are blocked with the same messages as the create dialog; task branch untouched.
- `WatchFields` status line now branches on signature-gated `liveResult`; `isError`/`elapsedMs` are read only inside that branch and are set atomically with `result`/`signature`.
- `McpToolArgsFields` drafts are now *derived*: shown only while `draft.slice(1).trimStart() === (expr ?? '')`. Traced char-by-char: `=`, `= `, `= n…now` all survive (no round-1 regression); external replacement (AI-generate/tool switch) shows the new expression; external deletion falls back; bare `=` clears on blur; `= now` does not. Both sides use identical normalisation so `=now`/`= now`/`=  now` agree.
- `_repair_cel` pops `cel_expr` when a retry omits it — new test pins `await_count == 1` and the judged fallback.

**Round-2 still-open non-blocking (unchanged, follow-up material):** `_describes_a_read` full re-discovery per unscored invoke; `model_defaults` re-read per notification/judged evaluation; detail-page `LastCheckPanel` run/result pairing. Multi-replica healer remains deferred to #167.

**Delta review (6 commits).** The metered CEL evaluator (24bab42a) was the main new surface and was exercised adversarially against the installed celpy 0.5.0 / lark 1.3.1, not just read: budgets are per-evaluation (fresh `_StepMeter`/`MeteredEvaluator` per `MeteredRunner.evaluate`, so cache hits are metered and a heavy expression running concurrently with 8 trivial ones failed alone); `CelBudgetExceededError` propagated out of `false || <heavy>`, `true ? <heavy> : …`, `true && <heavy>`, `has(...) && <heavy>` and a macro body — none of celpy's `except CELEvalError` swallow sites converted it to a value; `_visit_tree`/`sub_evaluator` are the real single-dispatch hooks in the pinned versions and the two new tests would catch an upgrade renaming them; both `watch_evaluator` catch sites and `_validate_cel_condition` map the subclass to run-level/validation errors; arg-exprs get a fresh budget per argument. Writing `ConditionEvaluation` through the model (0d0b8697) keeps a single JSON encoding and unchanged read shape; `jsonpath()`/`eq_ci()` stay registered in the shared `_ENV` with new pinning tests; `gatana_auth.py` and e53a85d1 are comment-only. Touched suites: 145 backend tests pass; frontend `tsc -b` and eslint clean on the four changed files.

**New findings: 1 (low, inline)** — the step budget and the 2s timeout are within ~3× of each other, so under load the generic timeout fires first and the meter's specific error is not the one the author sees. Tuning, not correctness.

Otherwise every previous finding is resolved and the delta introduces nothing else. Submitted as COMMENT — the human reviewer decides.

packages/console-backend/console_backend/services/cel_condition.py
aartaria: Fixed in 845ebbf4 — the measurement stands, the suggested remedy does not, so this went a different way.

Re-measured here: the realistic worst case (filter over a 2000-event response) is ~65 steps/item = 130k steps, so a 100k ceiling would reject a condition that should pass. Your ~45 steps/item was probably a simpler expression than `WITHIN_THE_HOUR`.

The real defect is that a step count and a number of seconds answer different questions, and only one was being asked. Both now are:

- `MAX_CEL_STEPS` stays 300k as the *load-independent* limit — a busy host must not reject expressions an idle one accepts, which is exactly what tuning it against wall-clock would cause.
- `CEL_EVAL_DEADLINE_SECONDS` (1.5s) is the wall-clock bound, checked **inside** the interpreter and set below `CEL_EVAL_TIMEOUT_SECONDS` so it fires first. The thread now stops itself rather than outliving the failed request — which was the more interesting half of your finding.

Verified with the step budget lifted so the clock is what trips: a quadratic over 4000 items raises `CelDeadlineExceededError` at 1.51s, where it previously hit the generic timeout at ~2.02s and kept running behind it. The clock is read every 1024 nodes (a `monotonic()` call dwarfs the meter's two integer ops); worst-case overshoot ~10ms.

The "comfortably" comment is gone, replaced by the per-step range under load and why the two ceilings are independent.

PR #169, opened 2026-08-25
fix(console): enforce user status on the paths that pick a user

### LINKED ISSUE #168: user status is not enforced consistently: suspended users are still selectable in the member picker
## Summary

`users.status` (`active` / `suspended` / `deleted`) is enforced inconsistently across the console backend and frontend. Authentication does block non-active users (`console_backend/dependencies.py:187`), but everything downstream disagrees about what a suspended user is: some queries filter to `status = 'active'`, some only exclude `'deleted'`, and some ignore `status` entirely. The user-visible symptom is that a **suspended user can still be picked from the member picker** and added to a group.

_Corrected after implementation: two names in the original body were wrong. There is no `get_members_by_ids` — the unfiltered query at `user_group_service.py:702` is `_add_members`' own "return the added members" fetch. The approver function is `_get_eligible_approvers_for_sub_agent`, not `get_sub_agent_approvers`._

## Symptom

Group detail → *Add member* offers suspended users.

- `packages/console-frontend/src/pages/admin/GroupDetailPage.tsx:288` — `availableUsers` filters only on "not already a member", not on status.
- It is fed by `list_users` (`packages/console-backend/console_backend/services/user_service.py:276`), which excludes only `deleted`. That is right for the admin user list, but wrong as a picker source.
- There is **no server-side guard**: `UserGroupService._add_members` (`user_group_service.py:573`) validates the group exists and nothing about the users, so a suspended (or, given the id, a soft-deleted) user can be added via the API regardless of what the UI shows.

## Second-order effect: suspended members become invisible but still exist

Membership *reads* filter `u.status = 'active'`, so once added (or once an existing member is suspended) the row exists in `user_group_members` and in Keycloak, but disappears from every list:

- `get_group_with_members` (`user_group_service.py:168`), `list_members` (`:743`, `:752`), `get_group_member_subs` (`:798`) — all `status = 'active'`.
- `_add_members`' own return fetch (`:702`) has **no** status filter, so it reports back members that the very next read hides.

Consequences: the suspended member is absent from the picker's `memberUserIds` set and can be "added" again; group member counts and the discovery-cache invalidation fan-out (`get_group_member_subs`) silently skip them; and re-activating a user resurrects memberships nobody could see.

## Other places `status` is ignored

- `SubAgentService._get_eligible_approvers_for_sub_agent` (`sub_agent_service.py:1726`) selects approvers from `users` with **no** `status` or `deleted_at` filter — suspended and soft-deleted users are returned as eligible approvers and notified.
- `sub_agent_service.py:1426`, `:1505`, `secrets_service.py:118`, `user_group_service.py:1234` fetch role/admin flags with no status check (lower risk, since the request already passed the auth dependency, but it makes the intended invariant unclear).
- Owner joins (`sub_agent_service.py` `JOIN users u ON sa.owner_user_id`, `catalog_repository.py`, `skill_registry_service.py`) never consider status — probably correct (an owner should keep being displayed) but currently by accident, not by decision.

## What is needed

1. **Decide and document the invariant** for `suspended`: who may be *selected*, who stays *displayed*, who counts as a *member*. Suggested rule: suspended users are never selectable/assignable and never notified, but remain visible where they are already referenced (existing member, owner, audit).
2. **Server-side enforcement** at the write path (`_add_members`, and any other place a user id is accepted from a client) — reject non-active user ids rather than relying on the UI.
3. **Distinguish "list users" from "pick a user"** — either a `status` filter parameter on `list_users` or a dedicated selectable-users endpoint, and use it in `GroupDetailPage`.
4. **Make membership reads consistent**: either show suspended members marked as suspended, or exclude them everywhere, so add/list can't disagree.
5. **Filter approver discovery** by active status.
6. Tests covering: suspended user not offered by the picker, add-member rejected server-side, suspended member handling in list/count, approvers exclude suspended.

## Notes

Code lives in the `nannos` monorepo (`packages/console-backend`, `packages/console-frontend`). Line numbers are as of this filing; search for the symbol rather than trusting them later.

PR #169 covers 2, 3, 5 and 6, and picks "never selectable, never notified, nothing else pruned" for 1. Point 4 — whether a suspended member should be *badged* in the member list rather than silently filtered out — is left open there as a product/UX decision.

aartaria:
closes ringier-data/nannos#168

## What was wrong

A suspended user could be picked from the group *Add member* dialog and added:

- the picker filtered only on "not already a member" (`GroupDetailPage.tsx`),
- its source (`list_users`) excludes only `deleted` — correct for the admin user table, wrong as a picker feed,
- and `UserGroupService._add_members` validated the group but nothing about the user ids, so the API accepted them regardless of the UI.

Once added, the member was invisible: every membership read is scoped to `status = 'active'`, while the `user_group_members` row and the Keycloak membership stayed. So the member could not be seen or removed, could be "added" again, and reappeared on re-activation.

Separately, `_get_eligible_approvers_for_sub_agent` filtered on neither `status` nor `deleted_at`, so suspended and soft-deleted users were returned as eligible approvers and notified.

## The rule this implements

**A non-active user is never selectable and never notified; nothing else about them is pruned.** Their memberships and ownership stay in place, and sign-in already blocks them (`dependencies.py:187`).

## Changes

- `UserGroupService`: `_reject_inactive_users` rejects unknown, soft-deleted and suspended user ids in `_add_members`; `group_router.add_members` maps that to 400. The admin "set user groups" path already maps `ValueError` to 409 with the message.
- `ScimGroupService._add_member` passes `require_active=False`. The IdP owns the group graph and keeps suspended users (`active=false` → `suspended` here) in their groups, so mirroring it must not silently drop memberships that would be needed again on reactivation.
- `GroupDetailPage`: the picker offers active users only.
- `_get_eligible_approvers_for_sub_agent`: `status = 'active' AND deleted_at IS NULL`.

## Deliberately not in this PR

- **Whether a suspended member should be *visible* (badged) in the group member list** instead of silently filtered out. That is a product/UX call, not a correctness one — with the server-side guard in place the invisible-member bug can no longer be created, but a member suspended *after* joining still disappears from the list. Worth deciding separately; it needs a `MemberInfo.status` field, an SDK regeneration and UI work.
- `memberUserIds` in the picker is built from the *paginated* member list (limit 20), so members on page 2+ still show as available. Pre-existing and unrelated to status.
- Two names in the issue body were wrong and have been corrected there: there is no `get_members_by_ids` (the unfiltered query at that line is `_add_members`' own "return the added members" fetch, which is now covered by the guard), and the approver function is `_get_eligible_approvers_for_sub_agent`.

## Verification

- `pytest tests/test_group_router.py tests/test_admin_group_router.py tests/test_scim_router.py tests/test_user_management.py tests/test_admin_user_router.py tests/test_user_group_keycloak_transactions.py tests/test_group_default_agents.py` → 122 passed, plus the sub-agent approval/submission suites (23 passed).
- New tests: non-active and unknown ids rejected with no membership row created; SCIM still mirrors a suspended member; a suspended admin drops out of the eligible approvers.
- `tsc -b` clean; `eslint GroupDetailPage.tsx` reports the same 12 pre-existing `no-explicit-any` errors as on `main`; `ruff check` adds no new findings (the repo is not `ruff format`-clean on `main`).

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — the add-member API now rejects non-active user ids with 400 (previously accepted silently); no schema change.

### DIFF
--- a/packages/console-backend/console_backend/routers/group_router.py
+++ b/packages/console-backend/console_backend/routers/group_router.py
@@ -145,13 +145,19 @@ async def add_members(
             detail="Group not found",
         )
 
-    members = await user_group_service.add_members(
-        db,
-        actor=user,
-        group_id=group_id,
-        user_ids=request_body.user_ids,
-        role=request_body.role,
-    )
+    try:
+        members = await user_group_service.add_members(
+            db,
+            actor=user,
+            group_id=group_id,
+            user_ids=request_body.user_ids,
+            role=request_body.role,
+        )
+    except ValueError as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(e),
+        )
     await db.commit()
 
     return GroupMemberListResponse(
--- a/packages/console-backend/console_backend/routers/group_router.py
+++ b/packages/console-backend/console_backend/routers/group_router.py
@@ -22,7 +22,7 @@ from ..models.user_group import (
     UserGroupWithMembers,
 )
 from ..services.orchestrator_cache import schedule_orchestrator_discovery_cache_invalidation
-from ..services.user_group_service import UserGroupService
+from ..services.user_group_service import InactiveUserError, UserGroupService
 
 router = APIRouter(prefix="/api/v1/groups", tags=["groups"])
 
@@ -153,7 +153,7 @@ async def add_members(
             user_ids=request_body.user_ids,
             role=request_body.role,
         )
-    except ValueError as e:
+    except InactiveUserError as e:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(e),
--- a/packages/console-backend/console_backend/services/user_group_service.py
+++ b/packages/console-backend/console_backend/services/user_group_service.py
@@ -32,6 +32,14 @@ from ..services.sub_agent_service import SubAgentService
 logger = logging.getLogger(__name__)
 
 
+class InactiveUserError(ValueError):
+    """Raised when a group membership change targets a user that is not active.
+
+    A ValueError subclass so that callers which already map ValueError to a client error keep
+    working, while routers that need to answer this case specifically can catch just this.
+    """
+
+
 class UserGroupService:
     """Service for managing user groups and memberships."""
 
@@ -578,7 +586,7 @@ class UserGroupService:
             user_ids: User IDs to validate
 
         Raises:
-            ValueError: If at least one ID does not belong to an active user
+            InactiveUserError: If at least one ID does not belong to an active user
         """
         result = await db.execute(
             text("SELECT id, status FROM users WHERE id = ANY(:user_ids) AND deleted_at IS NULL"),
@@ -587,7 +595,7 @@ class UserGroupService:
         status_by_id = {row[0]: row[1] for row in result.fetchall()}
         not_active = sorted({uid for uid in user_ids if status_by_id.get(uid) != UserStatus.ACTIVE.value})
         if not_active:
-            raise ValueError(f"Cannot add users that are not active: {', '.join(not_active)}")
+            raise InactiveUserError(f"Cannot add users that are not active: {', '.join(not_active)}")
 
     async def _add_members(
         self,
--- a/packages/console-backend/console_backend/services/user_group_service.py
+++ b/packages/console-backend/console_backend/services/user_group_service.py
@@ -14,7 +14,7 @@ from ..authorization import (
 )
 from ..models.notification import NotificationData, NotificationType
 from ..models.sub_agent import ActivationSource
-from ..models.user import User
+from ..models.user import User, UserStatus
 from ..models.user_group import (
     BulkDeleteResult,
     MemberInfo,
@@ -570,6 +570,25 @@ class UserGroupService:
 
     # Member management
 
+    async def _reject_inactive_users(self, db: AsyncSession, user_ids: list[str]) -> None:
+        """Raise if any of the given user IDs is unknown, soft-deleted or not active.
+
+        Args:
+            db: Database session
+            user_ids: User IDs to validate
+
+        Raises:
+            ValueError: If at least one ID does not belong to an active user
+        """
+        result = await db.execute(
+            text("SELECT id, status FROM users WHERE id = ANY(:user_ids) AND deleted_at IS NULL"),
+            {"user_ids": user_ids},
+        )
+        status_by_id = {row[0]: row[1] for row in result.fetchall()}
+        not_active = sorted({uid for uid in user_ids if status_by_id.get(uid) != UserStatus.ACTIVE.value})
+        if not_active:
+            raise ValueError(f"Cannot add users that are not active: {', '.join(not_active)}")
+
     async def _add_members(
         self,
         db: AsyncSession,
@@ -577,6 +596,7 @@ class UserGroupService:
         group_id: int,
         user_ids: list[str],
         role: Literal["read", "write", "manager"] = "read",
+        require_active: bool = True,
     ) -> list[MemberInfo]:
         """Internal helper to add members to a group.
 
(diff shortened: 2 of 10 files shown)


packages/console-backend/console_backend/services/user_group_service.py
```diff
@@ -597,6 +619,12 @@ async def _add_members(
         if existing is None:
             raise ValueError("Group not found")
 
+        # Only active users may be added. Every membership read is scoped to active users, so accepting
+        # a suspended or soft-deleted one would create a member nobody can see or remove, and a client
+        # picking from an unfiltered user list must not be able to slip one through.
+        if require_active:
+            await self._reject_inactive_users(db, user_ids)
```
alloy-cicd, line 634: 🤖 **Admin "set groups" now aborts half-way, leaving Keycloak diverged from the DB**

`admin_user_router.update_user_groups` (operation=`set`) removes first and adds second: for every `groups_to_remove` it calls `remove_members`, which writes the rows *and* immediately calls `keycloak_service.remove_user_from_group` (`user_group_service.py:1045-1071`). Only afterwards does it loop over `groups_to_add`.

With this guard, that second loop now raises `ValueError` for a suspended user → the router maps it to 409, `db.commit()` is never reached and `get_db_session` rolls the transaction back — but the Keycloak removals already happened and are not rolled back.

Concrete scenario: an admin re-arranges the groups of a suspended user (e.g. tidying up before reactivating them). They get a 409, the console DB still shows the old memberships, and Keycloak has already dropped the user from them. The divergence only surfaces later, when the user is reactivated and their token claims/group-derived access are wrong.

Either validate the user id once *before* the removal loop (fail fast with nothing applied), or have this admin path pass `require_active=False` — the rule this PR states is that a non-active user is never *picked*, not that their existing group graph can no longer be maintained (which is exactly the argument made for the SCIM path).

packages/console-backend/console_backend/routers/group_router.py
```diff
@@ -145,13 +145,19 @@ async def add_members(
...
-        user_ids=request_body.user_ids,
-        role=request_body.role,
-    )
+    try:
+        members = await user_group_service.add_members(
+            db,
+            actor=user,
+            group_id=group_id,
+            user_ids=request_body.user_ids,
+            role=request_body.role,
```
alloy-cicd, line 154: 🤖 **The `try` covers the whole mutation, so unrelated `ValueError`s become 400s**

Minor, but the mapping is broader than the new validation:

- if the group is deleted between this handler's own 404 pre-check and the service call, `_add_members` raises `ValueError("Group not found")` and the client gets **400** instead of 404;
- `pydantic.ValidationError` is a subclass of `ValueError`, so a `MemberInfo` construction failure inside `list_members` — which runs *after* the members were already written to the DB and pushed to Keycloak — is reported as a client-side 400 carrying an internal error string, masking a real server failure.

Raising a dedicated exception type from `_reject_inactive_users` (or wrapping only that call) keeps the 400 to the case it is meant for.

packages/console-backend/console_backend/services/user_group_service.py
aartaria: Fixed in 04530176 — real bug. The refusal now happens in `update_user_groups` before the removal loop, so nothing is touched when the target is not active. Kept removals allowed for a suspended user (suspension blocks additions; it does not freeze existing memberships), and added two tests: `set` on a suspended user returns 409 with the membership it would have removed still in place, and `remove` still works.

packages/console-backend/console_backend/routers/group_router.py
aartaria: Fixed in 04530176 — agreed, and the pydantic angle is the sharp end of it (ValidationError is a ValueError subclass, so an internal message could have surfaced as a 400 detail). Added `InactiveUserError(ValueError)` in `user_group_service` and narrowed the router's except to it; every other ValueError propagates exactly as it did before this PR. Subclassing ValueError keeps the admin path's existing ValueError→409 mapping working.

PR #170, opened 2026-08-26
feat: raw-bytes tool catalogue — stateless MCP listing, shared with sub-agents, PTC eval self-correction

### LINKED ISSUE #154: orchestrator: raw-bytes tool-catalogue ingest via gateway REST — cards + lazy decode, no pydantic on the catalogue path
## Summary

Replace the MCP-session tool-catalogue ingest with a **raw-bytes REST fetch + card index + lazy decode**: keep each tool's schema as undecoded JSON (`dict[str, bytes]`), render listings from a lightweight card index extracted in one streaming pass, and decode/compile a schema only when a tool is described or first invoked.

Unlike plain "lazy loading", this attacks the measured killer directly: the catalogue **never meets pydantic at all**. It supersedes the generic lazy-catalogue follow-up promised in #152.

## Why (measured, not hypothesized)

The OOM forensics (#152, alloy-ch/rcplus-alloy-infrastructure-agents#241) established:

- The GB-scale cost is the **transient parse on receive** — `JSONRPCMessage.model_validate_json` amplifies wire size ~7× into pydantic objects (21.9 MB `tools/list` → 663 MB / 10M live blocks at the parse site). Retention of the parsed catalogue was only ~1–2 MB — trimming *after* parsing fixes nothing, and lazy-loading through the SDK only reschedules the same parse.
- The size guard (#153) caps the blast radius; this issue removes the cost for the catalogue path entirely.
- The parsed schemas are barely used afterwards: `langchain_mcp_adapters` stores `args_schema` as a raw JSON-schema **dict**, and LangChain does not validate dict schemas on invoke — so **the PTC path performs no client-side validation today**. The MCP server validates. Materializing every schema buys nothing on the call path.

## Enabling fact: the gateway serves catalogues over plain REST

`GET /mcp-servers/{serverSlug}/tools` (Bearer auth — same token/audience discovery already uses for `GET /api/v1/mcp-servers` at `discovery.py:239`) returns the full tool list as plain JSON: `toolName`, `description`, `schema`, `outputSchema`, `annotations`, `isEnabled`, override fields (`toolNameOverride`/`descriptionOverride`/`inputSchemaOverride` + boolean flags), `universalName`. No MCP handshake, no SSE, no SDK parse. Docs: https://docs.gatana.ai/routes/listMcpServerTools/

## Design

1. **Ingest**: fetch each server's catalogue with plain `httpx` as bytes. One streaming pass (ijson or a bounded scanner) splits per-tool byte slices and extracts a **card**: name, one-line description, param names, flags. Cards for 500 tools ≈ hundreds of KB; heavy schema bodies stay as bytes.
2. **Consumers**:
   - *sandbox/prompt listing + ToolsetSelector*: render from cards only (the sandbox bridges are name-keyed; the interface text needs name/description/params — verified in `langchain_quickjs`).
   - *describe tool*: `json.loads` of that one tool's bytes on demand.
   - *grep/search tools*: regex over cards (or raw bytes for full-text).
   - *first invocation*: decode the schema; optionally `fastjsonschema.compile` and validate args before dispatch — **net-new** client-side validation (better local errors, fewer wasted round trips), cached per tool. Only a handful of tools are ever called per turn, so a handful of compiles.
   - *model-bound tools* (~5): decode lazily for the API request as today.
3. **Cache by `(server, interface_hash)`, not per user.** Bytes are immutable; the SDK already tracks `_mcp_interface_hashes`. Today every user re-fetches and re-parses an identical catalogue every 5 min (discovery cache TTL) — this removes that churn class entirely.

## Open questions (first investigation steps)

- **Override semantics**: the MCP layer applies `*Override` fields when serving `tools/list`; a REST consumer must apply them itself (and filter `isEnabled=false`, which the REST route includes).
- **Dispatch naming**: confirm whether `call_tool` expects `toolName` or `universalName` (no current reference to `universalName` in our code).
- **Auth audience**: confirm the exchanged gatana token accepted by `/api/v1/mcp-servers` also covers `/mcp-servers/{slug}/tools`.
- **Tool-call execution path**: without `StructuredTool` closures from `langchain_mcp_adapters`, dispatch needs its own `session.call_tool(name, args)` against the right server connection — likely small since `DynamicToolDispatchMiddleware` already tracks `server_name` in tool metadata.

## Out of scope

- Tool **call results** still arrive via the MCP session and its pydantic parse — the #153 size guard remains the protection there.
- `voice-agent` (separate process; tracked separately).

## Expected effect

Cold discovery stops paying the parse amplification entirely (the #153 guard becomes a true last-resort), cold-start latency drops (no per-server MCP handshakes for catalogues), steady-state memory shrinks (bytes shared across users instead of per-user object graphs), and the PTC path gains argument validation it never had.

aartaria:
closes #154

Eleven commits (six feature/fix commits, then review-driven fixes and the call-time token provider), each reviewable alone. Measured locally on 34 servers / ~514 tools: **cold discovery 3.2–4.6 s → ~1.5 s**, **sub-agent delegation overhead 2.4 s + 3 token exchanges → 0**; the remaining time-to-first-token is ~1 s auth/registry + the model's own ~4 s.

## 1. `feat(agent-common): raw-bytes tool catalogue with lazy schema decode`
The catalogue is held as per-tool canonical JSON **bytes** + a lightweight card (name, description, parameter names, server). Tools are `LazyMcpTool` — a real `BaseTool` whose `args_schema` decodes from those bytes on first access — so `convert_to_openai_tool`, `tool.args`, `describe`, PTC signature rendering, ToolsetSelector and dispatch are unchanged, and only the tools actually bound/described/invoked ever decode. Invocation delegates to `langchain_mcp_adapters` for that single tool (interceptors, progress callbacks, result conversion identical).

Ingest is a strategy behind that representation:
- **Stateless JSON-RPC `tools/list`** over plain httpx to the server's own MCP URL with the caller's token — the standard MCP method/result shape, so the gateway applies overrides, disabled flags and **per-user entitlements** exactly as for the SDK — but no `initialize`/session handshake and no pydantic parse (bounded `raw_decode` scan, `nextCursor` paging, JSON or SSE frame). Same pattern console-backend runs in production (`mcp_tool_client`). Probed once per URL; refusal → SDK for that URL, any other failure → SDK for that server.
- **SDK session** as universal fallback, each `mcp.types.Tool` flattened to bytes and dropped.

`CatalogueStore` interns catalogues by `(server, interface_hash)` (hash on bytes) so identical catalogues share bytes across users, and resolves names across everything interned.

*Rejected on the way:* the Gatana REST route `GET /api/v1/mcp-servers/{slug}/tools` — vendor-specific, **server-level** (683 tools vs 373 over MCP for the same user: it ignores per-profile entitlements), and named by raw `toolName` rather than the dispatchable `universalName`. Also tried and dropped: an SDK-level exchanged-token cache (a second token lifecycle the discovery cache's invalidation doesn't know about) — superseded by commit 3.

## 2. `feat(orchestrator): discover tools through the shared catalogue`
`ToolDiscoveryService` uses the ingest above (same semaphore bound), interns, and wraps per-user `LazyMcpTool`s bound to the user's connection — tokens stay on connections, never on shared bytes. Return type/metadata unchanged. Logs which path each server took. `MCP_CATALOGUE_STATELESS_LIST` (default true) forces the SDK when false. Console-backend's own MCP server is session-based and correctly takes the SDK path.

## 3. `feat: delegated sub-agents reuse the orchestrator's authenticated tools`
`build_runtime_context` hands each dynamic sub-agent the registry tools for its whitelist + the console self-improvement tools (`pre_resolved_tools`); the runnable takes private `model_copy`s and only discovers names it wasn't handed (store first, then its own `tools/list`). No token copied or cached anywhere new.

## 4. `fix(orchestrator): stop auto-including the MCP catalogue listers`
`console_grep_mcp_tools` / `console_list_mcp_servers` on the orchestrator were a leftover from pre-#108 in-orchestrator scheduling; task-scheduler and agent-creator carry them in their own seeds. On the orchestrator they only led the model to discover tools it cannot call before delegating.

## 5. `fix(agent-common): make PTC eval self-correcting`
Two errors that cost up to 50 graph steps on flash-lite: top-level `return` (prompt rule + one-shot async-IIFE retry on exactly that parse error) and `tools.github_get_me` / `tools.search_tools` (`TypeError: not a function` without the member name → a hint derived from what the sandbox actually exposes: camelCase spelling, `tools.search`/`describe`, or in inline mode the callable list + "regular tool call or `task`"). No aliases; raw catalogue listers are kept out of the eval namespace in every mode.

## 6. `fix(orchestrator): never classify a successful eval result as an error`
`ErrorClassificationMiddleware` stamped `[ERROR_TYPE: user_fixable]` on a successful `<result>` whose schemas contain `"required"`. Structured results without an `<error type=…>` block are no longer classified.

## 7. `feat: mint MCP bearer tokens at call time through a per-user provider` (+ 3 review fixes before it)
Discovery used to bake an exchanged bearer into every tool's connection, tying token freshness to the discovery-cache TTL and to how long a tool object lived (sub-agent hand-over). Now a per-user `UserTokenProvider` (`agent_common/core/token_provider.py`) is the only place that mints exchanged tokens — memoised per audience, re-minted within 90 s of `exp` (bounded by the user token's `exp`), one exchange under concurrency, refreshed with the user's current token every turn. Tools are built on **token-free** connections and a `bearer_interceptor` injects `Authorization` per call (the same `request.override(headers=…)` hook the console attribution uses). Sub-agents share the provider. Consequences: no credential older than the leeway however long a tool lives; a mid-turn exchanged-token expiry re-mints transparently (a mid-turn *user*-token expiry still surfaces as auth, correctly); `AGENT_DISCOVERY_CACHE_TTL` is now purely a catalogue-freshness knob. Review-driven fixes in between: scanner tolerates `{"result": {}}`; sub-agents get the *unwrapped* skill tools (their own `agent_name` default applies); sub-agent connections are picked by key, not position.

## Also found, not in this PR
- `AGENT_DISCOVERY_CACHE_TTL` defaults to **60 s** (the issue says 5 min) and no environment overrides it — in low-traffic dev every turn is cold; a TTL bump is the cheapest remaining TTFT win.
- The UI is blank for the whole pre-model window; a "working" status emitted before discovery would fix perceived TTFT.

## Tests
agent-common 730 · orchestrator 684 · sdk 359 (baseline) — `ruff` and `mypy` add no findings over `main`. Automatic review: round 1 → #1 #2 #3 #5 fixed, #4/#6/#7 deferred with reasons, #9–11 rejected with reasons; round 2 → 2 findings fixed; rounds 3–5 clean.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no: `discover_tools` keeps its signature and return type; tool names/metadata identical; `MCP_CATALOGUE_STATELESS_LIST=false` restores the SDK-only listing; the cache entry tuple gained a third element (`token_provider`) with a single consumer.

### DIFF
--- a/packages/agent-common/agent_common/core/catalogue_ingest.py
+++ b/packages/agent-common/agent_common/core/catalogue_ingest.py
@@ -114,6 +114,7 @@ def _walk_object(
 ) -> Iterator[Any]:
     pos = _skip_ws(text, pos + 1)  # past '{'
     if text[pos] == "}":
+        siblings["__end__"] = pos + 1  # an empty object still has to report where it ends
         return
     while True:
         key, pos = dec.raw_decode(text, pos)
@@ -272,8 +273,10 @@ async def fetch_catalogue_stateless(
         try:
             body = rpc_body_from_response(response)
             page, cursor = parse_tools_list_reply(server_slug, body)
-        except (ValueError, IndexError) as e:
-            # IndexError: the bounded scanner ran off the end of a truncated body.
+        except (ValueError, LookupError) as e:
+            # LookupError: the bounded scanner ran off the end of a truncated body (IndexError)
+            # or met a shape it does not handle (KeyError) — never let a parser bug take the
+            # server down when the SDK path can still list it.
             raise StatelessListError(f"{url}: unparseable tools/list reply: {e}") from e
         tools.extend(page)
         if not cursor:
--- a/packages/agent-common/agent_common/core/catalogue_ingest.py
+++ b/packages/agent-common/agent_common/core/catalogue_ingest.py
@@ -1,0 +1,327 @@
+"""Ingest strategies that turn a server's tool list into a :class:`ServerCatalogue`.
+
+Two paths, one representation (see :mod:`tool_catalogue`):
+
+* :func:`fetch_catalogue_stateless` — **fast path, feature-detected, never assumed.** A
+  single stateless JSON-RPC ``tools/list`` POST over plain ``httpx`` to the server's MCP
+  URL — the standard MCP method and result shape, but without the SDK: no
+  ``initialize``, no session id, no protocol negotiation, no resumability to own. The
+  reply (JSON, or a single SSE frame) is scanned one tool at a time with a bounded
+  ``raw_decode`` walk, so at no point is the catalogue materialised as Python objects and
+  pydantic is never involved. Because it is the MCP listing itself, made with the
+  caller's own token, the gateway applies overrides, disabled flags and per-user
+  entitlements exactly as it does for the SDK path. A server that refuses a stateless
+  request (a JSON-RPC error or 4xx on the probe) marks the path unsupported for that URL
+  and everything falls back; any other failure falls back for that server only.
+  This is the same pattern console-backend uses in production for its own tool
+  listing and tool calls (``console_backend.services.mcp_tool_client``).
+* :func:`fetch_catalogue_mcp` — **universal fallback**: the MCP ``tools/list`` through
+  the SDK session exactly as before. It still pays the SDK's parse on receive, but
+  each ``mcp.types.Tool`` is immediately flattened to bytes + card and the pydantic
+  objects dropped, recovering the steady-state and cross-user sharing wins; only the
+  transient parse peak remains (bounded by the inbound size guard once #153 lands).
+
+Which path each server took is reported to the caller so discovery can log it —
+a silent fall-back to the slow path must be visible, not mysterious.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+from collections.abc import Callable, Iterator, Mapping
+from contextlib import AbstractAsyncContextManager
+from typing import Any
+
+import httpx
+
+from agent_common.core.tool_catalogue import (
+    CatalogueTool,
+    ServerCatalogue,
+    build_server_catalogue,
+    make_catalogue_tool,
+)
+
+logger = logging.getLogger(__name__)
+
+# Default per-request timeout for a stateless catalogue fetch. Catalogues can be tens of MB.
+STATELESS_TIMEOUT_S = 30.0
+
+# One JSON-RPC id per request so a notification or ping streamed ahead of the reply on an
+# SSE response can never be mistaken for it.
+_LIST_ID = "catalogue-tools-list"
+
+# Never follow more than this many `nextCursor` pages for one server.
+_MAX_LIST_PAGES = 1000
+
+# JSON-RPC error codes that mean "this server will not serve a stateless request", as
+# opposed to a transient failure. -32600 invalid request (e.g. "session required"),
+# -32601 method not found, -32602 invalid params.
+_UNSUPPORTED_RPC_CODES = {-32600, -32601, -32602}
+
+
+class StatelessListUnsupported(Exception):
+    """The MCP endpoint does not serve ``tools/list`` without a session — fall back for good."""
+
+
+class StatelessListError(Exception):
+    """The endpoint serves stateless requests but this fetch failed — fall back for this server only."""
+
+
+# --------------------------------------------------------------------------------------
+# Bounded scanner: walk a JSON object by key path and yield one array element at a time
+# --------------------------------------------------------------------------------------
+
+_WS = re.compile(r"\s*")
+
+
+def _skip_ws(text: str, pos: int) -> int:
+    return _WS.match(text, pos).end()  # type: ignore[union-attr]
+
+
+def iter_array_at(
+    text: str,
+    path: tuple[str, ...],
+    siblings: dict[str, Any] | None = None,
+    *,
+    keep: frozenset[str] = frozenset(),
+) -> Iterator[Any]:
+    """Yield the elements of the array found at ``path`` in a JSON object, one at a time.
+
+    Walks objects key by key with ``JSONDecoder.raw_decode`` so only one element of the
+    target array is alive at once; every other member along the way is decoded and
+    dropped, except keys named in ``keep`` (small scalars such as ``nextCursor`` or
+    ``error``) which are stored in ``siblings`` keyed by ``"<depth>.<key>"`` → value.
+    Locating the array by *parsing* (rather than searching for ``"tools": [``) means a
+    tool whose schema contains a ``tools`` property cannot derail the scan.
+    """
+    dec = json.JSONDecoder()
+    pos = _skip_ws(text, 0)
+    if pos >= len(text) or text[pos] != "{":
+        raise ValueError("payload is not a JSON object")
+    yield from _walk_object(text, pos, path, 0, dec, siblings if siblings is not None else {}, keep)
+
+
+def _walk_object(
+    text: str,
+    pos: int,
+    path: tuple[str, ...],
+    depth: int,
+    dec: json.JSONDecoder,
+    siblings: dict[str, Any],
+    keep: frozenset[str],
+) -> Iterator[Any]:
+    pos = _skip_ws(text, pos + 1)  # past '{'
+    if text[pos] == "}":
+        return
+    while True:
+        key, pos = dec.raw_decode(text, pos)
+        pos = _skip_ws(text, pos)
+        if text[pos] != ":":
+            raise ValueError("malformed payload (expected ':')")
+        pos = _skip_ws(text, pos + 1)
+        if key == path[depth]:
+            if depth == len(path) - 1:
+                if text[pos] != "[":
+                    raise ValueError(f"'{key}' is not an array")
+                pos = _skip_ws(text, pos + 1)
+                if text[pos] != "]":
+                    while True:
+                        item, pos = dec.raw_decode(text, pos)
+                        yield item
+                        pos = _skip_ws(text, pos)
+                        if text[pos] == ",":
+                            pos = _skip_ws(text, pos + 1)
+                            continue
+                        if text[pos] == "]":
+                            break
+                        raise ValueError(f"malformed '{key}' array")
+                pos += 1
+            else:
+                if text[pos] != "{":
+                    raise ValueError(f"'{key}' is not an object")
+                # Recurse; the nested walker leaves `pos` just past its closing brace.
+                sub = _walk_object(text, pos, path, depth + 1, dec, siblings, keep)
+                yield from sub
+                pos = siblings.pop("__end__")
+        else:
+            value, pos = dec.raw_decode(text, pos)
+            if key in keep:
+                siblings[f"{depth}.{key}"] = value
+        pos = _skip_ws(text, pos)
+        if text[pos] == ",":
+            pos = _skip_ws(text, pos + 1)
+            continue
+        if text[pos] == "}":
+            siblings["__end__"] = pos + 1
+            return
+        raise ValueError("malformed payload")
+
+
+# --------------------------------------------------------------------------------------
+# Stateless JSON-RPC tools/list
+# --------------------------------------------------------------------------------------
+
+
+def rpc_body_from_response(response: httpx.Response) -> str:
+    """Return the JSON-RPC reply text, whether the server answered JSON or a single SSE frame.
+
+    An SSE reply may carry notifications or pings ahead of the response: take the first
+    ``data:`` line that is a response (has ``result`` or ``error``) with our request id.
+    Kept textual so the caller can scan it without decoding the whole catalogue.
+    """
+    content_type = response.headers.get("content-type", "")
+    if "text/event-stream" not in content_type:
+        return response.text
+    for line in response.text.split("\n"):
+        if not line.startswith("data:"):
+            continue
+        data = line[5:].removeprefix(" ").rstrip("\r")
+        # Cheap pre-check before scanning: must look like our response envelope.
+        if f'"id": "{_LIST_ID}"' in data or f'"id":"{_LIST_ID}"' in data:
+            if '"result"' in data or '"error"' in data:
+                return data
+    raise ValueError("no JSON-RPC response frame in SSE reply")
+
+
+def mcp_tool_dict_to_catalogue_tool(server_name: str, item: Mapping[str, Any]) -> CatalogueTool | None:
+    """Flatten one ``tools/list`` result element (a plain dict) into bytes + card."""
+    name = item.get("name")
+    if not isinstance(name, str) or not name:
+        return None
+    description = item.get("description")
+    return make_catalogue_tool(
+        server_name=server_name,
+        name=name,
+        description=description if isinstance(description, str) else None,
+        input_schema=item.get("inputSchema") if isinstance(item.get("inputSchema"), Mapping) else None,
+        output_schema=item.get("outputSchema") if isinstance(item.get("outputSchema"), Mapping) else None,
+        annotations=item.get("annotations") if isinstance(item.get("annotations"), Mapping) else None,
+        meta=item.get("_meta") if isinstance(item.get("_meta"), Mapping) else None,
+    )
+
+
+def parse_tools_list_reply(server_name: str, body: str) -> tuple[list[CatalogueTool], str | None]:
+    """Scan one ``tools/list`` JSON-RPC reply into catalogue tools + the next cursor.
+
+    Raises :class:`StatelessListUnsupported` for a JSON-RPC error whose code says the
+    server will not serve this request, :class:`StatelessListError` for other RPC errors,
+    and ``ValueError`` for a malformed body.
+    """
+    siblings: dict[str, Any] = {}
+    tools: list[CatalogueTool] = []
+    for item in iter_array_at(body, ("result", "tools"), siblings, keep=frozenset({"error", "nextCursor", "id"})):
+        if isinstance(item, Mapping):
+            tool = mcp_tool_dict_to_catalogue_tool(server_name, item)
+            if tool is not None:
+                tools.append(tool)
+    error = siblings.get("0.error")
+    if isinstance(error, dict):
+        code = error.get("code")
+        message = error.get("message") or "unknown error"
+        if code in _UNSUPPORTED_RPC_CODES or "session" in str(message).lower():
+            raise StatelessListUnsupported(f"JSON-RPC {code}: {message}")
+        raise StatelessListError(f"JSON-RPC {code}: {message}")
+    cursor = siblings.get("1.nextCursor")
+    return tools, cursor if isinstance(cursor, str) and cursor else None
+
+
+async def fetch_catalogue_stateless(
+    client: httpx.AsyncClient,
+    *,
+    url: str,
+    headers: Mapping[str, str] | None,
+    server_slug: str,
+) -> ServerCatalogue:
+    """``tools/list`` (all pages) as stateless JSON-RPC POSTs, scanned into a catalogue.
+
+    ``url``/``headers`` are the server's MCP connection (the same URL and bearer token
+    the SDK path would use, e.g. ``…/mcp?includeOnlyServerSlugs=<slug>``).
+    """
+    request_headers = {
+        **(headers or {}),
+        "Content-Type": "application/json",
+        # Gateways may answer either; accept both so a streaming-only server still works.
+        "Accept": "application/json, text/event-stream",
+    }
+    tools: list[CatalogueTool] = []
+    cursor: str | None = None
+    for _ in range(_MAX_LIST_PAGES):
+        params: dict[str, Any] = {"cursor": cursor} if cursor else {}
+        try:
+            response = await client.post(
+                url,
+                headers=request_headers,
+                json={"jsonrpc": "2.0", "id": _LIST_ID, "method": "tools/list", "params": params},
+            )
+        except httpx.HTTPError as e:
+            raise StatelessListError(f"{url}: {e}") from e
+        if 300 <= response.status_code < 400:
+            # Redirect not followed (client built without follow_redirects): retrying via the
+            # SDK would follow it, so treat as a per-server failure, not a refusal.
+            raise StatelessListError(
+                f"{url} -> HTTP {response.status_code} redirect to {response.headers.get('location')!r} not followed"
+            )
+        if response.status_code in (400, 404, 405, 406):
+            # The endpoint rejects the request shape itself (no session, wrong method,
+            # unacceptable media) — a property of the endpoint, not of this fetch.
+            raise StatelessListUnsupported(f"{url} -> HTTP {response.status_code}: {response.text[:200]}")
+        if response.status_code >= 400:
+            raise StatelessListError(f"{url} -> HTTP {response.status_code}: {response.text[:300]}")
+        try:
+            body = rpc_body_from_response(response)
+            page, cursor = parse_tools_list_reply(server_slug, body)
+        except (ValueError, IndexError) as e:
+            # IndexError: the bounded scanner ran off the end of a truncated body.
+            raise StatelessListError(f"{url}: unparseable tools/list reply: {e}") from e
+        tools.extend(page)
+        if not cursor:
+            break
+    else:
+        raise StatelessListError(f"tools/list on '{server_slug}' exceeded {_MAX_LIST_PAGES} pages")
+    return build_server_catalogue(server_slug, tools, source="stateless")
+
+
+# --------------------------------------------------------------------------------------
+# SDK-session fallback
+# --------------------------------------------------------------------------------------
+
+
+def mcp_tool_to_catalogue_tool(server_name: str, tool: Any) -> CatalogueTool:
+    """Flatten an ``mcp.types.Tool`` into bytes + card (the pydantic object is dropped by the caller)."""
+    annotations = tool.annotations.model_dump(exclude_none=True) if getattr(tool, "annotations", None) else None
+    return make_catalogue_tool(
+        server_name=server_name,
+        name=tool.name,
+        description=tool.description,
+        input_schema=tool.inputSchema,
+        output_schema=getattr(tool, "outputSchema", None),
+        annotations=annotations,
+        meta=getattr(tool, "meta", None),
+    )
+
+
+async def fetch_catalogue_mcp(
+    session_factory: Callable[[], AbstractAsyncContextManager[Any]],
+    *,
+    server_slug: str,
+) -> ServerCatalogue:
+    """``tools/list`` (all pages) through an MCP SDK session, flattened page by page.
+
+    ``session_factory`` is typically ``lambda: client.session(server_slug)`` on a
+    ``MultiServerMCPClient`` — it owns transport, callbacks and ``initialize()``.
+    """
+    tools: list[CatalogueTool] = []
+    async with session_factory() as session:
+        cursor: str | None = None
+        for _ in range(_MAX_LIST_PAGES):
+            page = await session.list_tools(cursor=cursor)
+            for mcp_tool in page.tools or []:
+                tools.append(mcp_tool_to_catalogue_tool(server_slug, mcp_tool))
+            cursor = page.nextCursor
+            if not cursor:
+                break
+        else:
+            raise RuntimeError(f"tools/list on '{server_slug}' exceeded {_MAX_LIST_PAGES} pages")
+    return build_server_catalogue(server_slug, tools, source="mcp")
--- a/packages/agent-common/agent_common/core/tool_catalogue.py
+++ b/packages/agent-common/agent_common/core/tool_catalogue.py
@@ -202,9 +202,11 @@ class LazyMcpTool(BaseTool):
       and delegates to its coroutine, so interceptors / progress callbacks / result
       conversion are byte-for-byte the upstream behaviour.
 
-    The MCP connection (which carries the user's exchanged bearer token) is a private
-    attribute on the tool, never on the shared catalogue — catalogue bytes are shared
-    across users, connections are not.
+    The MCP connection and the per-call interceptors (which mint the user's bearer for
+    each call — see ``agent_common.core.token_provider``) are private attributes on the
+    tool, never on the shared catalogue: catalogue bytes are shared across users,
+    connections and credentials are not. A connection may still carry a static bearer
+    when no token provider is configured (standalone execution).
     """
 
     server_name: str
--- a/packages/agent-common/agent_common/core/tool_catalogue.py
+++ b/packages/agent-common/agent_common/core/tool_catalogue.py
@@ -1,0 +1,441 @@
+"""Raw-bytes tool catalogue: cards + lazily decoded schemas, no pydantic on the catalogue path.
+
+Why
+---
+Materialising a gateway's tool catalogue through the MCP SDK amplifies the wire size
+~7x into pydantic objects at the parse site (a 21.9 MB ``tools/list`` became ~663 MB of
+live blocks), while the parsed schemas are barely used afterwards: listings and the
+toolset selector need name/description/server, ``describe`` needs one schema, model
+binding needs a handful, and the MCP server validates arguments on invoke anyway.
+
+This module keeps the catalogue as **bytes**: every tool's input schema is stored as
+canonical JSON (``bytes``) next to a lightweight :class:`ToolCard` (name, description,
+parameter names, server). A :class:`LazyMcpTool` is a real ``BaseTool`` whose
+``args_schema`` is decoded from those bytes on first access — so every existing
+consumer (``convert_to_openai_tool``, ``tool.args``, ``describe_tool``, PTC signature
+rendering, the interface hash) keeps working unchanged and only pays for the tools it
+actually touches.
+
+Ingest is a strategy (see :mod:`catalogue_ingest`): a stateless JSON-RPC ``tools/list``
+over plain HTTP, or the same call through the MCP SDK session. Both produce the same
+:class:`ServerCatalogue`, and a process-wide :class:`CatalogueStore` interns catalogues
+by ``(server, interface_hash)`` so identical catalogues share one set of bytes across
+users.
+
+Dispatch delegates to ``langchain_mcp_adapters`` on first invocation: the one tool
+being called is rebuilt as an ``mcp.types.Tool`` (a single small pydantic object) and
+converted with ``convert_mcp_tool_to_langchain_tool`` so interceptors, progress
+callbacks and result conversion behave exactly as before.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import threading
+import time
+from collections.abc import Awaitable, Callable, Iterable, Mapping
+from dataclasses import dataclass, field
+from typing import Any, Literal, cast
+
+from langchain_core.tools import BaseTool, StructuredTool
+from pydantic import PrivateAttr
+
+logger = logging.getLogger(__name__)
+
+CatalogueSource = Literal["stateless", "mcp"]
+
+
+def canonical_bytes(value: Any) -> bytes:
+    """Serialise ``value`` to compact, key-sorted UTF-8 JSON.
+
+    Canonical so identical schemas hash identically regardless of ingest source.
+    """
+    return json.dumps(value, sort_keys=True, separators=(",", ":"), ensure_ascii=False).encode("utf-8")
+
+
+@dataclass(frozen=True, slots=True)
+class ToolCard:
+    """What listings need about a tool — never the schema body."""
+
+    name: str
+    description: str
+    param_names: tuple[str, ...]
+    server_name: str
+
+    @property
+    def first_line(self) -> str:
+        stripped = self.description.strip()
+        return stripped.splitlines()[0] if stripped else ""
+
+
+@dataclass(slots=True)
+class CatalogueTool:
+    """One tool: its card plus undecoded schema bytes and the small MCP extras."""
+
+    card: ToolCard
+    schema_bytes: bytes
+    output_schema_bytes: bytes | None = None
+    annotations: dict[str, Any] | None = None
+    meta: dict[str, Any] | None = None
+
+    @property
+    def name(self) -> str:
+        return self.card.name
+
+    def decode_schema(self) -> dict[str, Any]:
+        schema = json.loads(self.schema_bytes)
+        if not isinstance(schema, dict):
+            # An MCP inputSchema must be a JSON-schema object; tolerate garbage with an
+            # accept-anything schema rather than failing the whole tool.
+            return {"type": "object", "properties": {}}
+        return schema
+
+
+@dataclass(slots=True)
+class ServerCatalogue:
+    """All tools of one MCP server, as bytes + cards, with a stable interface hash."""
+
+    server_name: str
+    tools: dict[str, CatalogueTool]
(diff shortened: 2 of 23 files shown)


packages/agent-common/agent_common/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+        by_name[tool.name] = tool
+    return ServerCatalogue(
+        server_name=server_name,
+        tools=by_name,
+        interface_hash=compute_interface_hash(by_name.values()),
+        source=source,
+    )
+
+
+class LazyMcpTool(BaseTool):
```
alloy-cicd, line 192: **[correctness · confirmed] `LazyMcpTool` exposes no `coroutine`, so `_wrap_tool_with_agent_name` silently no-ops.**

`app/utils.py:43-45` does `getattr(tool, "coroutine", None)` and returns the tool unwrapped when it's missing. The previous adapter `StructuredTool`s had `.coroutine`; `LazyMcpTool` is a plain `BaseTool` (verified: `getattr(tool, 'coroutine', None) is None`). So all eight tools in `_SKILL_TOOLS_NEEDING_AGENT_NAME` (utils.py:269-281 — `console_create_skill`, `console_update_playbook`, …) are no longer wrapped and `agent_name` no longer defaults to `"orchestrator"` / maps `"self"`. The LLM omitting `agent_name` (as the prompt permits) now reaches console-backend with it missing. Nothing logs the skipped wrap and no test covers it.

Fix options: give `LazyMcpTool` a `coroutine` property returning `self._get_delegate().coroutine`-like callable (careful: that forces the delegate build), or make the wrapper fall back to `tool.ainvoke` when `.coroutine` is absent. Either way add a test in `test_discovery.py` asserting the wrap applies to a `LazyMcpTool`.

packages/orchestrator-agent/app/core/catalogue_ingest.py
```diff
@@ -0,0 +1,232 @@
...
+    pos = _skip_ws(text, 0)
+    if pos >= len(text) or text[pos] != "{":
+        raise ValueError("catalogue payload is not a JSON object")
+    pos = _skip_ws(text, pos + 1)
+    if text[pos] == "}":
+        return
+    while True:
+        key, pos = dec.raw_decode(text, pos)
+        pos = _skip_ws(text, pos)
+        if text[pos] != ":":
```
alloy-cicd: **[correctness · confirmed] Unguarded `text[pos]` → `IndexError` on truncated bodies, which escapes the fallback.**

Only line 80 bounds-checks; lines 83, 88, 92, 95, 101, 104, 110, 113 index after `_skip_ws` without checking `pos < len(text)`. Verified: `iter_tools_array('{')`, `('{"tools": [')`, `('{"tools":[{"toolName":"x"}')` all raise `IndexError`. `fetch_catalogue_rest` (l.185) only catches `ValueError`/`JSONDecodeError`, and `_ingest_server` only catches `RestCatalogue*`, so an HTTP 200 with a truncated body (proxy cut-off) makes the server land in `failed_servers` even though the MCP path would have worked.

Simplest fix: catch `(ValueError, IndexError)` at l.185, or add a bounds check in `_skip_ws`'s callers (raise `ValueError('truncated catalogue payload')`). Alternatively the whole scanner could be `json.loads(payload)['tools']` — pydantic is the 7x amplifier, not `json.loads`, and the body is already fully held as both `bytes` and `str` here.

packages/orchestrator-agent/app/core/catalogue_ingest.py
```diff
@@ -0,0 +1,232 @@
...
+    url = rest_tools_url(gateway_base_url, server_slug)
+    try:
+        response = await client.get(
+            url,
+            params={"schemas": "true"},
+            headers={"Authorization": f"Bearer {token}", "Accept": "application/json"},
+        )
+    except httpx.HTTPError as e:
+        raise RestCatalogueError(f"{url}: {e}") from e
+    if response.status_code in (404, 405):
```
alloy-cicd: **[correctness · confirmed] A per-slug 404 is treated as "gateway has no REST route" and poisons the memo for the process lifetime.**

`RestCatalogueUnsupported` → `discovery.py:330 set_rest_supported(base_url, False)`. But 404 is also the natural answer for a single stale/renamed/permission-scoped slug (listed by `/mcp-servers` but not served by `/tools`). One such server in one user's list disables REST for every server and every user until pod restart, with only a warning — exactly the slow path this PR exists to avoid.

Suggest: treat 405 (and 404 on the *collection* route, or a probe of a known slug) as route-absent; treat a per-slug 404 as `RestCatalogueError` (server-only fallback).

packages/agent-common/agent_common/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+        self._rest_cache: dict[str, tuple[ServerCatalogue, float]] = {}  # server -> (cat, expires_at)
+        self._rest_supported: dict[str, bool] = {}  # gateway base url -> supported
+
+    # -- interning -----------------------------------------------------------------
+    def intern(self, catalogue: ServerCatalogue) -> ServerCatalogue:
+        with self._lock:
+            held = self._interned.get(catalogue.server_name)
+            if held is not None and held.interface_hash == catalogue.interface_hash:
+                return held
+            self._interned[catalogue.server_name] = catalogue
```
alloy-cicd, line 380: **[correctness/efficiency · confirmed] Interning keeps only the latest catalogue per server, so differing per-user catalogues never share and evict each other.**

The module docstring (l.21) says interning is by `(server, interface_hash)`, but `_interned` is `server -> latest`. On the MCP path `tools/list` is per-user-token (l.324 says so): users A and B with different entitlements on server S alternate `intern()` calls, each replaces the other, and each gets a private copy — zero sharing, plus one extra strong ref pinning the most recent user's catalogue after they leave.

Suggest keying by `(server_name, interface_hash)` with a bounded size (small LRU per server, or `weakref.WeakValueDictionary` — `ServerCatalogue` would need `weakref_slot=True`).

packages/orchestrator-agent/app/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+    _connection: Any = PrivateAttr(default=None)
+    _callbacks: Any = PrivateAttr(default=None)
+    _interceptors: list[Any] | None = PrivateAttr(default=None)
+    _delegate: StructuredTool | None = PrivateAttr(default=None)
+
+    def __getattribute__(self, item: str) -> Any:
+        if item == "args_schema":
+            values = object.__getattribute__(self, "__dict__")
+            if values.get("args_schema") is None:
+                entry = object.__getattribute__(self, "__pydantic_private__")["_entry"]
```
alloy-cicd: **[correctness · confirmed, low] `KeyError('_entry')` when `args_schema` is read before `build_lazy_tools` binds `_entry`.**

`_entry = PrivateAttr()` has no default, so `__pydantic_private__` lacks the key until l.304 sets it. Verified: `LazyMcpTool(name='x', description='d', server_name='s').args_schema` → `KeyError: '_entry'`. Anything touching `args_schema` between construction and binding (a validator, `repr`, tests constructing the class directly) gets a confusing traceback. Use `.get("_entry")` and return `None` or raise a clear error.

packages/agent-common/agent_common/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+        self._lock = threading.Lock()
+        self._interned: dict[str, ServerCatalogue] = {}  # server -> latest catalogue
+        self._rest_cache: dict[str, tuple[ServerCatalogue, float]] = {}  # server -> (cat, expires_at)
+        self._rest_supported: dict[str, bool] = {}  # gateway base url -> supported
+
+    # -- interning -----------------------------------------------------------------
+    def intern(self, catalogue: ServerCatalogue) -> ServerCatalogue:
+        with self._lock:
+            held = self._interned.get(catalogue.server_name)
+            if held is not None and held.interface_hash == catalogue.interface_hash:
```
alloy-cicd, line 378: **[correctness · plausible] `interface_hash` covers only name/description/input schema, so `intern` can substitute a catalogue with different annotations/`_meta`.**

REST ingest never passes `meta` and its `annotations` are raw; MCP ingest passes `meta` and `model_dump(exclude_none=True)` annotations. If REST interns catalogue R and a later user falls back to MCP (503) producing M with the same hash, `intern` returns R and that user's tools lose `_meta`/hints (and vice versa). Low impact today (only `server_name`/`compression_enabled` are read from metadata), but the docstring promises the two paths are interchangeable. Either include annotations/meta bytes in the hash, or document that metadata is best-effort.

packages/agent-common/agent_common/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+
+    async def _arun(self, *args: Any, **kwargs: Any) -> Any:
+        delegate = self._get_delegate()
+        # The adapter coroutine takes ``runtime`` as an injected arg; the ToolNode only
+        # injects it when it can see it on ``tool.func``/``tool.coroutine``, which this
+        # class does not expose. Nothing on our call path reads it (the console
+        # attribution interceptor uses ContextVars; the progress callback only logs).
+        if delegate.coroutine is None:  # pragma: no cover — the adapter always sets a coroutine
+            raise RuntimeError(f"MCP tool '{self.name}' has no async implementation")
+        return await delegate.coroutine(runtime=None, **kwargs)
```
alloy-cicd, line 297: **[correctness · plausible, latent] `delegate.coroutine(runtime=None, **kwargs)` will `TypeError` if `runtime` ever arrives in `kwargs`.**

Because `args_schema` is a dict, `BaseTool._parse_input` passes `tool_input` through unfiltered, so a caller supplying `runtime` (a future interceptor/PTC bridge) yields `got multiple values for keyword argument 'runtime'`. Also `ToolCallRequest.runtime` is always `None` for these tools, unlike adapter-built tools inside `ToolNode`. Cheap hardening: `runtime = kwargs.pop("runtime", None)` and forward it.

packages/orchestrator-agent/app/core/discovery.py
```diff
@@ -284,26 +298,70 @@ def _discovery_semaphore(self) -> asyncio.Semaphore:
...
+            if store.rest_supported(base_url) is not False:
+                try:
+                    catalogue = await fetch_catalogue_rest(
+                        http_client, gateway_base_url=base_url, server_slug=server_name, token=rest_token
+                    )
+                except RestCatalogueUnsupported as e:
+                    store.set_rest_supported(base_url, False)
+                    logger.warning(f"Gateway has no REST catalogue route — using MCP tools/list for all servers ({e})")
+                except RestCatalogueError as e:
+                    logger.warning(f"REST catalogue fetch failed for '{server_name}', falling back to MCP: {e}")
```
alloy-cicd: **[resilience · plausible] Transient REST errors (5xx, timeout) are never retried; they fall straight to the heavy MCP path.**

`_get_catalogue_with_retry`'s backoff only sees exceptions that escape `_ingest_server`, and `RestCatalogueError` is swallowed here. During a gateway blip every cold-cache user's discovery skips REST and hits `tools/list` concurrently (the ~7x pydantic peak) — a load spike precisely when the gateway is degraded — and populates `_interned` from MCP instead of warming the shared REST cache. Consider retrying 502/503/504 on REST before falling back.

packages/agent-common/agent_common/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+    server_name: str
+    response_format: Literal["content", "content_and_artifact"] = "content_and_artifact"
+
+    _entry: CatalogueTool = PrivateAttr()
+    _connection: Any = PrivateAttr(default=None)
+    _callbacks: Any = PrivateAttr(default=None)
+    _interceptors: list[Any] | None = PrivateAttr(default=None)
+    _delegate: StructuredTool | None = PrivateAttr(default=None)
+
+    def __getattribute__(self, item: str) -> Any:
```
alloy-cicd, line 221: **[efficiency/simplification] `__getattribute__` intercepts every attribute read on every tool; `__getattr__` gives the same laziness for free.**

`BaseTool.arun` does dozens of `self.x` reads per call, the whitelist filter in discovery does `t.name`/`t.metadata` for ~700 tools per discovery, and pydantic's own machinery goes through this hook. Each intercept is a string compare + second `object.__getattribute__`. It also reaches into `__dict__`/`__pydantic_private__` by string key, which is fragile across pydantic versions.

Alternative: after `__init__`, `object.__getattribute__(self, '__dict__').pop('args_schema', None)` and implement `__getattr__` (fires only on a miss) that decodes, stores into `__dict__`, and returns; chain to `super().__getattr__` otherwise. `schema_decoded` becomes `'args_schema' in self.__dict__`. Or expose a `@property tool_call_schema`/`args` backed by a `_decoded` PrivateAttr and update `utils.py:55`.

packages/orchestrator-agent/app/core/catalogue_ingest.py
```diff
@@ -0,0 +1,232 @@
...
+    """``tools/list`` (all pages) through an MCP session, flattened page by page.
+
+    ``session_factory`` is typically ``lambda: client.session(server_slug)`` on a
+    ``MultiServerMCPClient`` — it owns transport, callbacks and ``initialize()``.
+    """
+    tools: list[CatalogueTool] = []
+    async with session_factory() as session:
+        cursor: str | None = None
+        for _ in range(_MAX_LIST_PAGES):
+            page = await session.list_tools(cursor=cursor)
```
alloy-cicd: **[reuse] Hand-rolled pagination duplicates `langchain_mcp_adapters.tools._list_all_tools`.**

The adapter already implements the cursor walk with `MAX_ITERATIONS=1000` and identical `nextCursor` handling. `tools = await _list_all_tools(session)` then `map(mcp_tool_to_catalogue_tool, tools)` removes `_MAX_LIST_PAGES` and the loop. (The "page by page" rationale doesn't buy memory: the SDK already materialises every page's pydantic Tools before returning either way.)

Related small reuse items: `CatalogueStore._rest_cache` re-implements `discovery_cache.TtlTokenCache` (same get/put/expiry, plus `max_entries`); `MCP_CATALOGUE_REST_INGEST` in `config.py:371` inlines the truthy-set literal — a `_bool_env` sibling to `_int_env` would give the same typo warning.

packages/agent-common/agent_common/core/tool_catalogue.py
```diff
@@ -0,0 +1,403 @@
...
+    return json.dumps(value, sort_keys=True, separators=(",", ":"), ensure_ascii=False).encode("utf-8")
+
+
+@dataclass(frozen=True, slots=True)
+class ToolCard:
+    """What listings need about a tool — never the schema body."""
+
+    name: str
+    description: str
+    param_names: tuple[str, ...]
```
alloy-cicd, line 64: **[simplification] Unused/derivable state carried per tool.**

`ToolCard.param_names` (computed at ingest, read only by tests — and a weaker second source of truth vs `decode_schema()`), `first_line`, `CatalogueTool.output_schema_bytes` (canonicalised for every tool, read only by `byte_size`), `ServerCatalogue.cards`/`byte_size`/`fetched_at` — none referenced in `app/`. `param_names` in particular is a tuple of strings per tool, the very per-tool footprint this PR shrinks. Drop what has no consumer; `ToolCard` can be `(name, description, server_name)`.

alloy-cicd (commented):

Reviewed at 2d11d3ca (7 files). Overall the bytes-first catalogue is a sound direction and the delegation to `convert_mcp_tool_to_langchain_tool` keeps dispatch semantics intact. Findings below, most severe first; each was verified against the checked-out branch (the first three by executing the code).

**Confirmed correctness**
1. `LazyMcpTool` has no `coroutine`, so `_wrap_tool_with_agent_name` (app/utils.py:43) silently no-ops for all eight `console_*` skill tools — `agent_name` defaulting to `orchestrator` is lost.
2. `iter_tools_array` raises `IndexError` on truncated bodies; `fetch_catalogue_rest` doesn't convert it, so the server fails instead of falling back to MCP.
3. A per-slug 404 permanently disables the REST path for the whole gateway/process.
4. `CatalogueStore.intern` keeps only the latest catalogue per server → per-user catalogues that differ never share and evict each other.
5. Accessing `args_schema` on a `LazyMcpTool` not built via `build_lazy_tools` raises `KeyError('_entry')`.

**Plausible / design**
6. Interning by `interface_hash` ignores annotations/`_meta`, so metadata from one ingest path can be served for another.
7. `_arun` hard-codes `runtime=None`; a `runtime` kwarg would `TypeError`.
8. Transient REST 5xx is never retried — under a gateway blip every cold discovery stampedes to the heavy MCP path.
9. `__getattribute__` hook runs on every attribute read of every tool; `__getattr__` after popping the field gives zero-cost laziness.
10. Reuse: adapter's `_list_all_tools`, `TtlTokenCache`, a `_bool_env` helper; several unused fields on the dataclasses.

Not flagged (checked and OK): `MCPTool(_meta=...)` alias, `model_copy`/`deepcopy` of `LazyMcpTool`, `content_and_artifact` handling, interceptor `server_name` threading, `is_retryable_mcp_error` still seeing raw MCP exceptions. The SDK's `_check_mcp_interface_changed` uses its own loader, so it does not force-decode these tools.

packages/agent-common/agent_common/core/catalogue_ingest.py
```diff
@@ -0,0 +1,327 @@
...
+        if response.status_code in (400, 404, 405, 406):
+            # The endpoint rejects the request shape itself (no session, wrong method,
+            # unacceptable media) — a property of the endpoint, not of this fetch.
+            raise StatelessListUnsupported(f"{url} -> HTTP {response.status_code}: {response.text[:200]}")
+        if response.status_code >= 400:
+            raise StatelessListError(f"{url} -> HTTP {response.status_code}: {response.text[:300]}")
+        try:
+            body = rpc_body_from_response(response)
+            page, cursor = parse_tools_list_reply(server_slug, body)
+        except (ValueError, IndexError) as e:
```
alloy-cicd: **[correctness · confirmed] `KeyError('__end__')` from the scanner escapes the fallback net.**

`_walk_object` returns early on an empty object (l.116-117) without setting `siblings["__end__"]`, so the parent's `pos = siblings.pop("__end__")` (l.147) raises `KeyError` when the path object itself is `{}`. Verified at this head:

```
parse_tools_list_reply('s', '{"jsonrpc":"2.0","id":"catalogue-tools-list","result":{}}')
→ KeyError: '__end__'
```

This `except (ValueError, IndexError)` doesn't catch it, `_ingest_server` only catches `StatelessList*`, so the exception reaches `_get_catalogue_with_retry`, is non-retryable, and the server lands in `failed_servers` — even though the SDK path would have returned an empty listing. (`{"result":{"tools":[]}}` is fine; only an empty `result` object trips it.)

Two-line fix: set `siblings["__end__"] = pos + 1` before the early `return` at l.117, and widen this clause to `except (ValueError, IndexError, KeyError)` — or simply `except Exception`, since any scanner defect here should mean "fall back for this server", never "fail the server".

packages/orchestrator-agent/app/utils.py
```diff
@@ -58,6 +58,24 @@ async def wrapped_coroutine(**kwargs: Any) -> Any:
...
+    token exchanges and a gateway-wide ``tools/list``. Hand them over instead: the
+    whitelist (``config.mcp_tools``) plus the console self-improvement tools every
+    sub-agent gets. Names not in the registry are left for the sub-agent to discover.
+    """
+    from langchain_core.tools import BaseTool
+
+    from agent_common.agents.dynamic_agent import DynamicLocalAgentRunnable
+
+    wanted = set(config.mcp_tools or []) | set(DynamicLocalAgentRunnable._CONSOLE_SELF_IMPROVEMENT_TOOLS)
+    return {name: tool_registry[name] for name in wanted if isinstance(tool_registry.get(name), BaseTool)}
```
alloy-cicd, line 80: **[correctness · plausible] Pre-resolved skill tools are handed over *after* the orchestrator-specific `agent_name` wrap.**

`_pre_resolved_tools_for` is called at l.529, after l.302-304 replaced every `_SKILL_TOOLS_NEEDING_AGENT_NAME` registry entry with `_wrap_tool_with_agent_name(tool, "orchestrator")` (defaults `agent_name` to `"orchestrator"` when `None`/`"self"`). Two consequences for a sub-agent:

- The self-improvement path is safe: `_wrap_with_agent_name` sets `kwargs["agent_name"] = self.name` explicitly, which the inner wrapper passes through.
- But if a sub-agent's `config.mcp_tools` whitelist itself names one of these tools (e.g. `console_update_playbook`), `_discover_mcp_tools` returns the pre-resolved copy *unwrapped by the sub-agent* (dynamic_agent.py:64-73), so a call that omits `agent_name` now silently targets the **orchestrator's** playbook/skills instead of failing validation as the SDK-discovered tool did before. (It also yields two same-named tools once the self-improvement set is added — pre-existing, but the pre-resolved one now carries a wrong default.)

Suggest excluding `_SKILL_TOOLS_NEEDING_AGENT_NAME` from the `config.mcp_tools` half of `wanted` (the self-improvement path already covers them with the right name), or snapshotting the registry for hand-over before l.302.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (2d11d3ca7f359167313f7f15366de26446190fe6..bfbf3f63fe1fe1eaa494dfc6fb5642fece9ebdb1)

The branch was rewritten (6 commits; catalogue moved to `agent-common`, Gatana REST route replaced by stateless JSON-RPC `tools/list`, plus sub-agent tool reuse and two PTC fixes). Ran the PR's test files at the new head: agent-common 64 passed, orchestrator 44 passed.

**Ledger — round 1**
| # | Finding | Disposition |
|---|---|---|
| 1 | `LazyMcpTool` lacks `coroutine` → `_wrap_tool_with_agent_name` no-op | **verified-fixed** — `coroutine` property (`tool_catalogue.py:235`) |
| 2 | `IndexError` on truncated body escapes fallback | **verified-fixed** — caught at `catalogue_ingest.py:275` (but see new finding A: `KeyError` still escapes) |
| 3 | Per-slug 404 poisons the REST memo process-wide | **verified-fixed by redesign** — REST route dropped; probe memo is now per MCP URL (per server) |
| 4 | Interning is latest-per-server, not `(server, hash)` | **still-open — deferred by author** (docstring at `tool_catalogue.py:356` still says `(server, interface_hash)`) |
| 5 | `KeyError('_entry')` on unbound tool | **verified-fixed** — clear `RuntimeError` |
| 6 | `interface_hash` ignores annotations/`_meta` | **still-open — deferred by author** |
| 7 | `runtime=None` hard-coded in `_arun` | **still-open — not addressed**, low/latent; no re-raise |
| 8 | Transient REST errors not retried | **rejected/moot** — REST path removed; stateless failures still fall straight to SDK (author deferred) |
| 9–11 | `__getattribute__`, reuse, unused fields | **rejected-by-author** — not re-raised |

**New in the delta (2 inline comments)**
- A. `catalogue_ingest.py:275` — `KeyError('__end__')` from `_walk_object` on `{"result":{}}` is not caught (executed to confirm), so a server fails instead of falling back to the SDK.
- B. `utils.py:76` — `_pre_resolved_tools_for` runs after the orchestrator-specific `agent_name="orchestrator"` wrap (l.302-304), so a sub-agent whose `mcp_tools` whitelist names a skill tool inherits the orchestrator's default.

**Follow-up material (not blocking, no inline comment):** `CatalogueStore.resolve()` serves entries interned from *other users'* per-token views to a sub-agent (whitelist is admin-controlled and the gateway still authorises the call, so exposure is limited to name/schema); HTTP 400 on the stateless probe permanently marks the URL unsupported (blast radius is one server); `_is_structured_success` also suppresses classification for any non-PTC tool output containing `<result`.

packages/agent-common/agent_common/core/catalogue_ingest.py
aartaria: Fixed in e0f926e6: the empty-object early return now records its end, and the fallback net is widened to LookupError so a scanner shape bug can never fail a server (it falls back to the SDK session). Test: parse_tools_list_reply("srv", '{"result": {}}') == ([], None).

packages/orchestrator-agent/app/utils.py
aartaria: Fixed in 26b90be1: the registry is snapshotted before the orchestrator wraps the skill tools, and sub-agents receive that unwrapped snapshot; the self-improvement path keeps applying the sub-agent's own agent_name. Test asserts snapshot < wrap < hand-over ordering.

packages/agent-common/agent_common/core/tool_catalogue.py
aartaria: Deferred, by design for now: the store's job is to dedupe identical catalogues and to let a sub-agent resolve names the orchestrator already fetched in this process. Keying by (server, hash) would keep more variants alive without changing behaviour; interning misses only cost memory, never correctness (every consumer holds its own reference).

packages/agent-common/agent_common/core/tool_catalogue.py
aartaria: Deferred: annotations/_meta are small and identical across users for a given server; if two catalogues differ only there, the substituted one is still semantically the same tool set. Will fold them into the hash if a real divergence shows up.

packages/agent-common/agent_common/core/tool_catalogue.py
aartaria: Acknowledged as latent: runtime is only injected when the ToolNode sees it on tool.func/coroutine; LazyMcpTool.coroutine has no runtime annotation, so kwargs never carries it today. Left as is to avoid speculative plumbing.

packages/agent-common/agent_common/core/tool_catalogue.py
aartaria: Fixed in 71c5f2b8 (rebuilt history): LazyMcpTool exposes a coroutine property; both agent_name wrappers work on it (covered by test_pre_resolved_tools_skip_exchange_and_discovery_entirely and the pass-down tests).

packages/agent-common/agent_common/core/tool_catalogue.py
aartaria: Not applicable: args_schema is a real pydantic field present in __dict__, so __getattr__ never fires for it (verified — a property approach recurses inside tool_call_schema). __getattribute__ is the only hook that works; the cost is one string compare per attribute read.

packages/agent-common/agent_common/core/tool_catalogue.py
aartaria: Kept: param_names feed the card-based listings the issue asks for (grep/describe without decoding), output_schema_bytes are needed to rebuild the mcp.types.Tool for dispatch with its outputSchema. Dropping them would force a decode on those paths.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (bfbf3f63fe1fe1eaa494dfc6fb5642fece9ebdb1..26b90be1305f1458f4086ec3e3b2c41a3e9edd32)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger — round 2**
| # | Finding | Disposition |
|---|---|---|
| A | `KeyError('__end__')` on `{"result":{}}` escapes the fallback net | **verified-fixed** (e0f926e6) — re-ran the repro at this head: `parse_tools_list_reply('s', '{"jsonrpc":"2.0","id":"catalogue-tools-list","result":{}}') == ([], None)`; truncated bodies now surface as `IndexError`, caught by the widened `LookupError` clause (`catalogue_ingest.py:276`) |
| B | Pre-resolved skill tools handed over after the `agent_name="orchestrator"` wrap | **verified-fixed** (26b90be1) — `unwrapped_registry = dict(tool_registry)` at `utils.py:310` precedes the wrap at l.313 and feeds the hand-over at l.518; checked that every other `tool_registry` mutation (l.177–213) happens before the snapshot, so nothing is lost |

**Ledger — round 1 (unchanged from round 2, all replied by author)**
#1, #2, #3, #5 verified-fixed · #4 (store keying), #6 (hash coverage), #7 (`runtime=None`) still-open, deferred by author with reasoning — not re-raised · #8 moot (REST path removed) · #9–11 rejected-by-author with reasoning — not re-raised.

Tests at this head: agent-common `test_tool_catalogue.py` 19 passed; orchestrator `test_utils_pre_resolved.py` + `test_discovery.py` + `test_catalogue_discovery.py` 24 passed.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (26b90be1305f1458f4086ec3e3b2c41a3e9edd32..67ac73716366e475a7e269bbaa52323ebf0cbbe6)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger**
- Round 3 raised no findings (clean round); nothing to re-classify.
- Round 2: A (`KeyError('__end__')`) and B (pre-wrapped skill tools handed to sub-agents) remain **verified-fixed**.
- Round 1: #1–3, #5 verified-fixed · #4, #6, #7 still-open, deferred by author with reasoning — not re-raised · #8 moot · #9–11 rejected-by-author — not re-raised.

**Delta reviewed:** one self-initiated commit (67ac7371) replacing the positional `gateway[0]` pick in `_resolve_catalogue_tools._connection_for` with an explicit lookup — the tool's own server slug when a connection is keyed by it, else the gateway-wide connection under `mcp_gateway_client_id`, else skip. Checked: store hits from the orchestrator's per-slug catalogues (e.g. `server_name="github"`) correctly fall through to the `mcp_gateway_client_id` connection since the sub-agent has no per-slug key; the `missing` computation still uses the same fallback; the `server_name != "console"` guard keeps console-named catalogues off the gateway path. Regression test asserts both a gateway and a console store hit bind to their respective tokens regardless of dict order. `tests/test_dynamic_agent.py`: 38 passed at this head.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (67ac73716366e475a7e269bbaa52323ebf0cbbe6..1ec132fbf04fc3ef3f09e2a52cbb82fb81279595)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger**
- Rounds 3 and 4 raised no findings; nothing to re-classify.
- Round 2: A (`KeyError('__end__')`) and B (pre-wrapped skill tools) remain **verified-fixed**.
- Round 1: #1–3, #5 verified-fixed · #4, #6, #7 still-open, deferred by author with reasoning — not re-raised · #8 moot · #9–11 rejected-by-author — not re-raised.

**Delta reviewed:** one new commit (1ec132fb) introducing `agent_common.core.token_provider` — a per-user `UserTokenProvider` (memoised per audience, re-minted within 90 s of `exp`, bounded by the subject token's `exp`, one exchange under concurrency) plus a `bearer_interceptor` that injects `Authorization` per call; discovery/sub-agent tools are now built on token-free connections. Checked, all OK:
- `langchain_mcp_adapters` merges `request.headers` into the connection for that call (`{**existing, **modified}`), so a token-free `call_connection` + interceptor is a working credential path; the interceptor order `[bearer, _console_attribution]` is safe because the attribution interceptor spreads `request.headers` before adding `x-nannos-context`.
- Sub-agent `_call_connection` strips only `Authorization` and only when a provider is present; the SDK *listing* connection keeps its bearer, which `fetch_catalogue_stateless`/`client.session` still need. Standalone (no provider) is unchanged.
- `_audience(server_name)` maps store hits from the orchestrator's per-slug catalogues (`github`, …) to the gateway audience and `console` to the console client id — consistent with `ToolDiscoveryService._audience_for_server`.
- Discovery-cache value is now a 3-tuple; `executor.py` is the only consumer that unpacks it. Cache-hit path hands the provider the current turn's subject token, so a rotated user token invalidates derived ones (`update_subject_token` clears the memo).
- Provider failures (Keycloak down / user token expired) propagate as tool errors and cache nothing — matches the stated intent.
- Tests at this head: agent-common `test_token_provider.py` + `test_dynamic_agent.py` 48 passed; orchestrator discovery/cache/executor suites 57 passed.

**Follow-up nits (no inline comment):** `discovery_cache.py:21` and `:168` still describe the cached value as `(tools, sub_agents)`; an opaque (non-JWT) exchanged token is deliberately re-minted on every call — fine for Keycloak, worth a one-line note if another IdP is ever used.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 6 (1ec132fbf04fc3ef3f09e2a52cbb82fb81279595..ce0c22f9646cb2b45a462be29fe7763be33b2f4e)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger**
- Round 5 raised no findings; its two summary-only nits are **verified-fixed** in e6b47816 (`discovery_cache.py` now describes the cached value as `(tools, sub_agents, token_provider)`; `token_provider.py` docstring states that an opaque token is re-exchanged on every ask).
- Rounds 3–4: no findings. Round 2: A and B remain **verified-fixed**.
- Round 1: #1–3, #5 verified-fixed · #4, #6, #7 still-open, deferred by author with reasoning — not re-raised · #8 moot · #9–11 rejected-by-author — not re-raised.

**Delta reviewed:** two commits. e6b47816 is docs only. ce0c22f9 adds `MCP_TOKEN_LEEWAY_SECONDS` (`_int_env`, default 90) and threads it into `ToolDiscoveryService.make_token_provider` → `UserTokenProvider(leeway_seconds=…)`; sub-agents inherit the same provider, so the knob applies to them too. `_int_env` tolerates bad values with a warning; an int is accepted where the provider types `float`. Test fixtures updated for the new setting. Orchestrator `test_discovery.py` + `test_catalogue_discovery.py` + `test_discovery_cache.py`: 41 passed at this head.

**Follow-up nit (no inline comment):** a negative `MCP_TOKEN_LEEWAY_SECONDS` is not clamped and would let a memoised bearer be reused past its `exp`; `max(0, …)` like `MCP_DISCOVERY_CONCURRENCY` uses would close that operator-error path.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 7 (ce0c22f9646cb2b45a462be29fe7763be33b2f4e..b4f414cd7869e42960cae8af158b304561cee75e)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger**
- Round 6 raised no findings; its summary-only nit (negative `MCP_TOKEN_LEEWAY_SECONDS` not clamped) is **verified-fixed** in b4f414cd — checked at import: `MCP_TOKEN_LEEWAY_SECONDS=-5` → `0`, `=120` → `120`.
- Round 5 nits verified-fixed (e6b47816). Rounds 3–4: no findings. Round 2: A and B remain **verified-fixed**.
- Round 1: #1–3, #5 verified-fixed · #4, #6, #7 still-open, deferred by author with reasoning — not re-raised · #8 moot · #9–11 rejected-by-author — not re-raised.

**Delta reviewed:** a single one-line commit wrapping the knob in `max(0, …)`, matching the `MCP_DISCOVERY_CONCURRENCY` pattern; `_int_env`'s bad-value fallback still applies first. Nothing else changed.

PR #173, opened 2026-08-27
feat(agent-runner): adopt the shared tool catalogue and call-time bearer tokens; remove CatalogueStore

### LINKED ISSUE #172: agent-runner: adopt the shared tool catalogue and call-time bearer tokens
## Summary

`agent-runner` still discovers MCP tools the pre-#170 way (`agent/core.py`, ~l.1105–1140): it exchanges the user's token for the gateway and console audiences, bakes the results into `StreamableHttpConnection.headers`, and runs `MultiServerMCPClient.get_tools()` over the **whole gateway** — the full MCP SDK parse into pydantic objects, one `StructuredTool` per tool, on every run.

Everything #152/#154 measured therefore still applies to it verbatim:

- **Memory** — the transient parse amplification (~7× wire size into live pydantic objects) on every `tools/list`; the #153/#156 size guard bounds the blast radius but not the cost.
- **Latency** — a gateway-wide SDK listing (~500 tools) plus 2 exchanges per run, before the agent can do anything.
- **Credentials** — bearers embedded in tool connections at discovery time; nothing re-mints if an exchanged token expires during a long run.

## Proposal

Reuse what `agent-common` now provides (landed in #170), so the runner has the same shape as an orchestrator sub-agent:

1. **Catalogue ingest** — `agent_common.core.catalogue_ingest.fetch_catalogue_stateless` (stateless JSON-RPC `tools/list`, per-user, no handshake, no pydantic on the catalogue path) with `fetch_catalogue_mcp` as fallback; intern in `CatalogueStore`; build `LazyMcpTool`s via `build_lazy_tools` / `make_lazy_tool`. Schemas decode lazily, so the runner only pays for the tools it binds/calls.
2. **Token provider** — `agent_common.core.token_provider.UserTokenProvider` + `bearer_interceptor`: one provider per run, tools on token-free connections, a fresh bearer per call (re-minted within the leeway of `exp`).
3. Keep the runner's own whitelist filtering and `_validate_tool_schema` step; the only things that change are *how* tools are obtained and *where* the token lives.
4. The PTC `eval` self-correction from #170 already reaches the runner through `build_sub_agent_graph`; nothing to do there.

## Acceptance

- A run lists tools via the stateless path (SDK fallback logged when an endpoint refuses), with no `mcp.types.Tool` objects retained after discovery.
- No `Authorization` header on any tool connection; `MCP_TOKEN_LEEWAY_SECONDS=100000` shows one exchange per tool call (the QA lever from #170), default shows one per audience.
- Cold-run tool discovery time and RSS after discovery recorded before/after in the PR.
- Existing agent-runner tests pass; new tests mirror `agent-common/tests/test_dynamic_agent.py::test_with_a_token_provider_tools_are_token_free_and_mint_per_call` for the runner.

## Related

- #154 / #170 — the catalogue, stateless listing and token provider this reuses.
- #156 — shares the inbound size guard with agent-runner/voice-agent (complementary: the guard protects the SDK fallback path; this issue removes most of the need for it).
- `voice-agent` uses the SDK's `LangGraphAgent`, not `agent-common`; if it needs the same treatment it is a separate, SDK-level change.

aartaria:
closes ringier-data/nannos#172

## What

`agent-runner` discovered MCP tools the pre-#170 way: two token exchanges baked into `StreamableHttpConnection.headers`, then `MultiServerMCPClient.get_tools()` over the **whole gateway** on every scheduled run. This gives a run the same shape as an orchestrator sub-agent, reusing what `agent-common` provides since #170:

- **Catalogue** — new `agent/mcp_tools.py::McpToolResolver`. Each server the whitelist needs is listed via `fetch_catalogue_stateless` (JSON-RPC `tools/list`, no handshake, no pydantic on the catalogue path) with `fetch_catalogue_mcp` as fallback. Endpoint refusal (400/404/405/406) is memoised per URL; a transient failure falls back for that listing only, and the fallback is logged. Every run lists with **its own token** — a Gatana profile can hide tools per user, so a listing is a per-user view and is never served to another run.
- **Tools** — every tool is a `LazyMcpTool`; schemas decode only for the tools the run binds/calls. No `mcp.types.Tool` survives discovery.
- **Credentials** — a per-run `UserTokenProvider` owns the user's token. Tool connections carry **no** `Authorization` header; `bearer_interceptor` mints one per call, re-exchanged within `MCP_TOKEN_LEEWAY_SECONDS` of `exp`, so an exchanged token expiring during a long run is re-minted instead of failing.
- The runner's whitelist filtering is unchanged (`console_*` → console MCP, everything else → gateway). Only servers the whitelist actually needs are listed/exchanged.

New env (all optional, documented in `.env.template`): `MCP_GATEWAY_CLIENT_ID` (default `gatana`), `MCP_CATALOGUE_STATELESS_LIST` (default `true`), `MCP_TOKEN_LEEWAY_SECONDS` (default 90 — set to `100000` for the one-exchange-per-call QA lever from #170).

Note on the issue text: the runner never had a `_validate_tool_schema` step of its own (that lives in `agent_common.agents.dynamic_agent`); `build_sub_agent_graph` is untouched, so the PTC `eval` self-correction still reaches the runner as before.

## Also in this PR: `CatalogueStore` removed (agent-common, orchestrator)

Reviewing the runner surfaced that `CatalogueStore` was unsound for the same reason: it kept **one catalogue per server** (whoever listed last) with no user dimension, and `dynamic_agent._resolve_catalogue_tools` resolved whitelisted names from it — so a sub-agent could be handed a tool its user's own listing does not offer. Its other job (sharing identical catalogues' bytes across users) is worth little now that the stateless path removed the parse amplification: ~200 KB of schema bytes per 500-tool view, held at most 60 s by the per-user discovery cache.

- `CatalogueStore` / `get_catalogue_store` / `reset_catalogue_store` deleted. The per-URL "serves stateless `tools/list`" memo — the only legitimately process-wide fact — moves to `catalogue_ingest` (`stateless_supported` / `set_stateless_supported` / `reset_stateless_memo`).
- `dynamic_agent`: `pre_resolved_tools` (the orchestrator's per-user discovery cache, which already carries the schemas) is the only source of names without a listing; anything else is listed on the sub-agent's own connection with the user's token — console names on `console`, gateway names on the gateway connection.
- Orchestrator `discovery._ingest_server` and the runner keep the catalogue they fetched instead of interning it.
- The "stateless `tools/list` → per-URL memo → SDK session" strategy both of them carried now lives once as `catalogue_ingest.fetch_catalogue()`; callers pass URL, bearer headers, httpx client and session factory and read `catalogue.source`.
- Tests flipped from "served from the store" to "listed per user" (agent-common 766 pass, orchestrator 684 pass).

## Before / after

Local fake FastMCP gateway (`stateless_http=True`, 500 tools with gateway-shaped schemas), whitelist of 5 tools, all imports hoisted out of the measured window, 3 runs each — loopback, so the network share of the SDK handshake and the 2 eager exchanges is understated:

| path | discovery | RSS after discovery | tracemalloc peak / retained | `mcp.types.Tool` alive after |
|---|---|---|---|---|
| **before** — `get_tools()` whole gateway | ~135 ms | +36 MB | 8.0 / 5.7 MB | **500** |
| **after** — stateless `tools/list` | **~100 ms** | **+8 MB** | 2.4 / 2.3 MB | 0 |
| after — SDK fallback path | ~175 ms | +39 MB | 8.1 / 5.1 MB | 0 |

Every run lists (per-user view); a name hidden from a user by their profile is never bound from another user's listing (`test_every_run_lists_with_its_own_token_and_never_binds_another_users_view`).

## Tests

`tests/test_mcp_tools.py` (12 new; 34 total pass, `ruff` clean, mypy error count 39 vs 41 on `main` — all pre-existing `import-untyped` for `agent_common`):
- runner mirror of `agent-common/tests/test_dynamic_agent.py::test_with_a_token_provider_tools_are_token_free_and_mint_per_call`: listing carries the bearer, tools carry none, one exchange per audience, interceptor memoised
- no `mcp.types.Tool` objects alive after discovery; schema decodes lazily on demand
- `MCP_TOKEN_LEEWAY_SECONDS` above token lifetime ⇒ one exchange per call
- only needed servers are listed; unresolved names are logged and skipped
- per-user views: a second user's run re-lists and does not get names their own listing lacks
- `scheduler_*` routed to the console MCP; a failing exchange fails discovery; listing honours `MCP_TIMEOUT_SECONDS`
- stateless refusal ⇒ endpoint memoised unsupported, SDK listing used the bearer, tool still token-free; transient error ⇒ fallback without memo; `MCP_CATALOGUE_STATELESS_LIST=false` ⇒ straight to SDK
- `agent.core` no longer references `MultiServerMCPClient`/`get_tools()`

Not run locally: a real run against the Gatana gateway (needs cluster credentials) — the stateless-vs-fallback decision against the real endpoint will show in the `Resolved N/M MCP tools … (gateway=stateless|mcp)` log line on first deploy.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no; defaults preserve behaviour, only the eager exchange goes away.

### DIFF
--- a/packages/agent-runner/agent/core.py
+++ b/packages/agent-runner/agent/core.py
@@ -49,6 +49,7 @@ from agent_common.core.model_factory import (
     require_default_model,
 )
 from agent_common.core.stream_watchdog import watch_stream_with_resume
+from agent_common.core.token_provider import DEFAULT_LEEWAY_S, UserTokenProvider
 from google.protobuf.json_format import MessageToDict, ParseDict
 from google.protobuf.struct_pb2 import Struct
 from object_storage import get_object_storage_service
@@ -56,8 +57,6 @@ from object_storage import get_object_storage_service
 if TYPE_CHECKING:
     from agent_common.core.sandbox_pool import SandboxPool
 from langchain_core.messages import HumanMessage
-from langchain_mcp_adapters.client import MultiServerMCPClient
-from langchain_mcp_adapters.sessions import StreamableHttpConnection
 from langgraph.checkpoint.memory import MemorySaver
 from langgraph.store.postgres.aio import AsyncPostgresStore
 from psycopg.rows import dict_row
@@ -67,11 +66,19 @@ from ringier_a2a_sdk.models import AgentStreamResponse, UserConfig
 from ringier_a2a_sdk.oauth import OidcOAuth2Client
 from ringier_a2a_sdk.utils.a2a_part_conversion import a2a_parts_to_content
 
+from agent.mcp_tools import McpToolResolver
+
 logger = logging.getLogger(__name__)
 
 _CONSOLE_BACKEND_URL = os.getenv("CONSOLE_BACKEND_URL", "http://localhost:5001")
 _CONSOLE_BACKEND_CLIENT_ID = os.getenv("CONSOLE_BACKEND_CLIENT_ID", "agent-console")
 _MCP_GATEWAY_URL = os.getenv("MCP_GATEWAY_URL", "https://nannos.gatana.nannos.ringier.ch/mcp")
+_MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
+# Stateless JSON-RPC tools/list (no SDK handshake/parse); off = always list through the SDK.
+_MCP_CATALOGUE_STATELESS_LIST = os.getenv("MCP_CATALOGUE_STATELESS_LIST", "true").lower() not in ("0", "false", "no")
+# How much validity a memoised exchanged bearer must keep to be reused for a tool call. Set it
+# above the exchanged tokens' lifetime to force one exchange per call (QA lever, see #170).
+_MCP_TOKEN_LEEWAY_SECONDS = max(0.0, float(os.getenv("MCP_TOKEN_LEEWAY_SECONDS", str(DEFAULT_LEEWAY_S))))
 _MCP_TIMEOUT_SECONDS = int(os.getenv("MCP_TIMEOUT_SECONDS", "300"))
 _DOCUMENT_STORE_S3_BUCKET = os.getenv("DOCUMENT_STORE_S3_BUCKET", "")
 _MAX_RECURSION_LIMIT = int(os.getenv("MAX_RECURSION_LIMIT", "50"))
@@ -1100,46 +1107,27 @@ class AgentRunner(BaseAgent):
 
         try:
             if mcp_tool_names:
-                # Keep the MCP session open for the entire graph execution so that
-                # tool closures can call back into the session when invoked.
-                allowed = set(mcp_tool_names)
-                has_console_tools = any(name.startswith("console_") for name in allowed)
-                has_gateway_tools = any(not name.startswith("console_") for name in allowed)
-
-                connections: dict[str, StreamableHttpConnection] = {}
-
-                # Add Gatana gateway connection for non-console tools
-                if has_gateway_tools:
-                    gatana_access_token = await (self._get_oauth2_client()).exchange_token(user_access_token, "gatana")
-                    connections["gateway"] = StreamableHttpConnection(
-                        transport="streamable_http",
-                        url=_MCP_GATEWAY_URL,
-                        headers={"Authorization": f"Bearer {gatana_access_token}"},
-                        timeout=mcp_timeout,
-                        sse_read_timeout=mcp_timeout,
-                    )
-
-                # Add console backend MCP connection for console_ tools
-                if has_console_tools:
-                    console_mcp_url = f"{_CONSOLE_BACKEND_URL}/mcp"
-                    console_token = await (self._get_oauth2_client()).exchange_token(
-                        user_access_token, _CONSOLE_BACKEND_CLIENT_ID
-                    )
-                    connections["console"] = StreamableHttpConnection(
-                        transport="streamable_http",
-                        url=console_mcp_url,
-                        headers={"Authorization": f"Bearer {console_token}"},
-                        timeout=mcp_timeout,
-                        sse_read_timeout=mcp_timeout,
-                    )
-
-                mcp_client = MultiServerMCPClient(connections)
-                all_tools = await mcp_client.get_tools()
-                tools = [t for t in all_tools if t.name in allowed]
+                # Tools come from the shared catalogue (stateless tools/list, SDK fallback) as
+                # LazyMcpTools on token-free connections; a per-run UserTokenProvider mints
+                # the bearer at call time, so a token expiring mid-run is re-exchanged.
+                resolver = McpToolResolver(
+                    token_provider=UserTokenProvider(
+                        user_access_token,
+                        self._get_oauth2_client().exchange_token,
+                        leeway_seconds=_MCP_TOKEN_LEEWAY_SECONDS,
+                    ),
+                    gateway_url=_MCP_GATEWAY_URL,
+                    gateway_client_id=_MCP_GATEWAY_CLIENT_ID,
+                    console_mcp_url=f"{_CONSOLE_BACKEND_URL}/mcp",
+                    console_client_id=_CONSOLE_BACKEND_CLIENT_ID,
+                    timeout=mcp_timeout,
+                    stateless_list=_MCP_CATALOGUE_STATELESS_LIST,
+                )
+                tools = await resolver.resolve(mcp_tool_names)
                 logger.info(
                     "Loaded %d/%d MCP tools for job %s: %s",
                     len(tools),
-                    len(all_tools),
+                    len(mcp_tool_names),
                     scheduled_job_id,
                     [t.name for t in tools],
                 )
--- a/packages/agent-runner/agent/core.py
+++ b/packages/agent-runner/agent/core.py
@@ -75,7 +75,11 @@ _CONSOLE_BACKEND_CLIENT_ID = os.getenv("CONSOLE_BACKEND_CLIENT_ID", "agent-conso
 _MCP_GATEWAY_URL = os.getenv("MCP_GATEWAY_URL", "https://nannos.gatana.nannos.ringier.ch/mcp")
 _MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
 # Stateless JSON-RPC tools/list (no SDK handshake/parse); off = always list through the SDK.
-_MCP_CATALOGUE_STATELESS_LIST = os.getenv("MCP_CATALOGUE_STATELESS_LIST", "true").lower() not in ("0", "false", "no")
+_MCP_CATALOGUE_STATELESS_LIST = os.getenv("MCP_CATALOGUE_STATELESS_LIST", "true").strip().lower() in {
+    "1",
+    "true",
+    "yes",
+}
 # How much validity a memoised exchanged bearer must keep to be reused for a tool call. Set it
 # above the exchanged tokens' lifetime to force one exchange per call (QA lever, see #170).
 _MCP_TOKEN_LEEWAY_SECONDS = max(0.0, float(os.getenv("MCP_TOKEN_LEEWAY_SECONDS", str(DEFAULT_LEEWAY_S))))
@@ -1123,14 +1127,7 @@ class AgentRunner(BaseAgent):
                     timeout=mcp_timeout,
                     stateless_list=_MCP_CATALOGUE_STATELESS_LIST,
                 )
-                tools = await resolver.resolve(mcp_tool_names)
-                logger.info(
-                    "Loaded %d/%d MCP tools for job %s: %s",
-                    len(tools),
-                    len(mcp_tool_names),
-                    scheduled_job_id,
-                    [t.name for t in tools],
-                )
+                tools = await resolver.resolve(mcp_tool_names)  # logs what was resolved and how
                 await _run_graph(tools + docstore_tools)
             else:
                 await _run_graph(docstore_tools)
--- a/packages/agent-runner/agent/mcp_tools.py
+++ b/packages/agent-runner/agent/mcp_tools.py
@@ -28,8 +28,8 @@ from datetime import timedelta
 from typing import Any
 
 import httpx
+from agent_common.agents.dynamic_agent import is_console_backend_tool
 from agent_common.core.catalogue_ingest import (
-    STATELESS_TIMEOUT_S,
     StatelessListError,
     StatelessListUnsupported,
     fetch_catalogue_mcp,
@@ -46,10 +46,15 @@ logger = logging.getLogger(__name__)
 GATEWAY_SERVER = "gateway"
 CONSOLE_SERVER = "console"
 
+# Console-backend tools (``console_*``, ``scheduler_*``) go to the console MCP; the rest to the gateway.
+is_console_tool = is_console_backend_tool
 
-def is_console_tool(name: str) -> bool:
-    """Console-backend MCP tools are ``console_*``; everything else lives on the gateway."""
-    return name.startswith("console_")
+# When each MCP URL was last listed by this process. Catalogues are per-user views and can
+# change on the gateway, so the shared store only serves a whitelist without a fresh
+# ``tools/list`` while the URL was listed within ``catalogue_ttl`` (default: the same 60 s the
+# orchestrator's discovery cache uses).
+_LISTED_AT: dict[str, float] = {}
+DEFAULT_CATALOGUE_TTL_S = 60.0
 
 
 class McpToolResolver:
@@ -69,11 +74,13 @@ class McpToolResolver:
         console_client_id: str,
         timeout: timedelta,
         stateless_list: bool = True,
+        catalogue_ttl_seconds: float = DEFAULT_CATALOGUE_TTL_S,
     ) -> None:
         self.token_provider = token_provider
         self.gateway_client_id = gateway_client_id
         self.console_client_id = console_client_id
         self.stateless_list = stateless_list
+        self.catalogue_ttl = catalogue_ttl_seconds
         self._urls = {GATEWAY_SERVER: gateway_url, CONSOLE_SERVER: console_mcp_url}
         self._timeout = timeout
         # Discovery statistics for the last resolve(); logged and surfaced for measurements.
@@ -121,11 +128,17 @@ class McpToolResolver:
             else:
                 store.set_stateless_supported(url, True)
                 self.stats["source"][server_name] = "stateless"
+                _LISTED_AT[url] = time.monotonic()
                 return store.intern(catalogue)
         client = MultiServerMCPClient({server_name: self._connection(server_name, bearer=bearer)})
         catalogue = await fetch_catalogue_mcp(lambda: client.session(server_name), server_slug=server_name)
         self.stats["source"][server_name] = "mcp"
+        _LISTED_AT[url] = time.monotonic()
         return store.intern(catalogue)
+
+    def _store_is_fresh(self, server_name: str) -> bool:
+        listed_at = _LISTED_AT.get(self._urls[server_name])
+        return listed_at is not None and time.monotonic() - listed_at < self.catalogue_ttl
 
     # -- resolution ----------------------------------------------------------------------
     async def resolve(self, wanted: Iterable[str]) -> list[BaseTool]:
@@ -152,9 +165,17 @@ class McpToolResolver:
         def _server_for(name: str) -> str:
             return CONSOLE_SERVER if is_console_tool(name) else GATEWAY_SERVER
 
+        # Exchange up front for every audience the run needs: discovery used to do this, and
+        # a user token that is expired or revoked must fail the run here, not surface as a
+        # run that "succeeded" without ever being able to call a tool.
+        for server in connections:
+            await self.token_provider.get(self.audience_for(server))
+
         tools: list[BaseTool] = []
         for name, (_held, entry) in store.resolve(names).items():
             server = _server_for(name)
+            if not self._store_is_fresh(server):
+                continue  # the store's copy may be stale for this process; list again below
             tools.append(
                 make_lazy_tool(
                     entry,
@@ -167,7 +188,8 @@ class McpToolResolver:
 
         missing = names - {t.name for t in tools}
         if missing:
-            async with httpx.AsyncClient(timeout=STATELESS_TIMEOUT_S, follow_redirects=True) as http_client:
+            # follow_redirects: console-backend's ``/mcp`` mount answers ``307 → /mcp/``.
+            async with httpx.AsyncClient(timeout=self._timeout.total_seconds(), follow_redirects=True) as http_client:
                 for server, connection in connections.items():
                     server_missing = {n for n in missing if _server_for(n) == server}
                     if not server_missing:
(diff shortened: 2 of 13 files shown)


packages/agent-runner/agent/mcp_tools.py
```diff
@@ -0,0 +1,203 @@
...
+
+logger = logging.getLogger(__name__)
+
+GATEWAY_SERVER = "gateway"
+CONSOLE_SERVER = "console"
+
+
+def is_console_tool(name: str) -> bool:
+    """Console-backend MCP tools are ``console_*``; everything else lives on the gateway."""
+    return name.startswith("console_")
```
alloy-cicd: 🤖 **Correctness:** `is_console_tool` only recognises `console_*`, but the console-backend MCP also serves `scheduler_*` tools (`scheduler_create_job`, `scheduler_list_jobs`, … — see `scheduler_router.py` operation_ids, and migration 075 seeds a local sub-agent whitelisting exactly those). The shared helper `agent_common.agents.dynamic_agent.is_console_backend_tool` already covers both prefixes (`CONSOLE_BACKEND_TOOL_PREFIXES = ("console_", "scheduler_")`) and is documented as the single source of truth.

Failure: a scheduled run of that sub-agent routes `scheduler_list_jobs` to `GATEWAY_SERVER` — it is listed against the gateway (not offered → silently dropped as unresolved), or, if a console catalogue was interned earlier in the process, `store.resolve` finds it and `make_lazy_tool` binds it to the gateway connection with the gateway bearer, so the call goes to the wrong server. Pre-existing in the old `startswith("console_")` check, but this PR re-implements the helper rather than reusing the correct one.

```suggestion
from agent_common.agents.dynamic_agent import is_console_backend_tool as is_console_tool
```
(and drop the local definition)

packages/agent-runner/agent/mcp_tools.py
```diff
@@ -0,0 +1,203 @@
...
+                (CONSOLE_SERVER, is_console_tool),
+            )
+            if any(predicate(n) for n in names)
+        }
+
+        def _server_for(name: str) -> str:
+            return CONSOLE_SERVER if is_console_tool(name) else GATEWAY_SERVER
+
+        tools: list[BaseTool] = []
+        for name, (_held, entry) in store.resolve(names).items():
```
alloy-cicd: 🤖 **Correctness (behaviour change):** `store.resolve(names)` hits are served unconditionally, and `CatalogueStore` has no TTL and no user dimension — its own docstring says catalogues "are per-user-token views, so they are never *served* from here in place of a fetch". The deleted code ran `tools/list` with *this* user's bearer on every run, so per-user entitlements and tool removals/schema changes were always fresh.

Failure scenarios in a long-lived runner pod: (a) user A's run interns the gateway catalogue; user B's run whitelists a tool B is not entitled to → it is bound to B's model anyway and only fails at call time with a gateway 403 (previously simply absent). (b) the gateway renames a required argument or disables a tool → every later run keeps the stale schema bytes until the pod restarts, because a name the store already holds never triggers a re-list. Consider honouring `ServerCatalogue.fetched_at` with a max age, or re-listing per run and using the store only for interning (as the orchestrator's discovery does).

packages/agent-runner/agent/mcp_tools.py
```diff
@@ -0,0 +1,203 @@
...
+
+        Names the store does not hold trigger one listing per server that can serve them;
+        names that no server lists are logged and skipped (same as the whitelist filter
+        the runner always applied).
+        """
+        started = time.monotonic()
+        names = set(wanted)
+        self.stats = {"source": {}, "from_store": 0, "listed": 0}
+        store = get_catalogue_store()
+        interceptors = [bearer_interceptor(self.token_provider, self.audience_for)]
```
alloy-cicd, line 125: 🤖 **Plausible behaviour change:** when every whitelisted name is already in the store, `resolve()` never calls `_list_server`, so no token exchange happens during discovery — the first exchange runs inside `bearer_interceptor` during a tool call. The old code exchanged both audiences up front, so an expired/revoked user token surfaced as an exception from `_run_langgraph_agent` and the run was recorded as `scheduler_status: failed`.

Failure: token revoked between scheduling and execution → the ToolNode turns the exchange error into a tool-error string, the model narrates "I couldn't access…", and the job is recorded as a *successful* run with a bogus summary. A cheap guard: eagerly `await token_provider.get(audience)` for each connected server at the end of `resolve()` (or in core.py before `_run_graph`).

packages/agent-runner/agent/mcp_tools.py
```diff
@@ -0,0 +1,203 @@
...
+                    server_name=server,
+                    connection=connections[server],
+                    tool_interceptors=interceptors,
+                )
+            )
+            self.stats["from_store"] += 1
+
+        missing = names - {t.name for t in tools}
+        if missing:
+            async with httpx.AsyncClient(timeout=STATELESS_TIMEOUT_S, follow_redirects=True) as http_client:
```
alloy-cicd: 🤖 **Plausible:** the stateless listing uses the hard-coded `STATELESS_TIMEOUT_S` (30 s) while `self._timeout` (`MCP_TIMEOUT_SECONDS`, 300 s) only applies to the SDK fallback. Unlike the orchestrator, which lists per server slug (`includeOnlyServerSlugs=<slug>`), the runner lists the *whole* gateway (~500 tools, tens of MB per the module docstring) in one stateless POST.

Failure: on a slow link the POST exceeds 30 s → `StatelessListError` (not `Unsupported`, so the endpoint is never memoised) → the SDK path re-lists everything; every cold run pays 30 s of pure waste plus the full parse, and raising `MCP_TIMEOUT_SECONDS` has no effect. Pass `timeout=self._timeout.total_seconds()` (or max of the two) to the httpx client.

packages/agent-runner/agent/mcp_tools.py
```diff
@@ -0,0 +1,203 @@
...
+            url=self._urls[server_name],
+            timeout=self._timeout,
+            sse_read_timeout=self._timeout,
+        )
+        if bearer is not None:
+            connection["headers"] = {"Authorization": f"Bearer {bearer}"}
+        return connection
+
+    # -- listing -------------------------------------------------------------------------
+    async def _list_server(self, server_name: str, http_client: httpx.AsyncClient) -> ServerCatalogue:
```
alloy-cicd, line 95: 🤖 **Reuse:** `_list_server` is a near line-for-line copy of `orchestrator-agent/app/core/discovery.py::_ingest_server` (stateless attempt → memoise supported/unsupported → SDK fallback → `store.intern`), differing only in how the bearer is obtained and in recording `stats`. Two copies of the fallback policy will drift (e.g. the orchestrator's follow-up fixes to the unsupported-code set or retry policy won't reach the runner). Suggest lifting it into `agent_common.core.catalogue_ingest` (e.g. `ingest_server(store, http_client, url, headers, server_slug, client_factory, stateless=...)`) and calling it from both.

packages/agent-runner/agent/core.py
```diff
@@ -1100,46 +1107,27 @@ async def _run_graph(tools: list) -> None:
...
+                    console_client_id=_CONSOLE_BACKEND_CLIENT_ID,
+                    timeout=mcp_timeout,
+                    stateless_list=_MCP_CATALOGUE_STATELESS_LIST,
+                )
+                tools = await resolver.resolve(mcp_tool_names)
                 logger.info(
                     "Loaded %d/%d MCP tools for job %s: %s",
                     len(tools),
-                    len(all_tools),
+                    len(mcp_tool_names),
```
alloy-cicd: 🤖 **Simplification:** this `Loaded %d/%d MCP tools` log duplicates the `Resolved %d/%d MCP tools … from the shared catalogue store, %d via tools/list` line `McpToolResolver.resolve()` emits a few microseconds earlier, with the same tool names. Either drop this one or pass `scheduled_job_id` into the resolver's log (the only extra information here).

packages/agent-runner/agent/core.py
```diff
@@ -67,11 +66,19 @@
...
+from agent.mcp_tools import McpToolResolver
+
 logger = logging.getLogger(__name__)
 
 _CONSOLE_BACKEND_URL = os.getenv("CONSOLE_BACKEND_URL", "http://localhost:5001")
 _CONSOLE_BACKEND_CLIENT_ID = os.getenv("CONSOLE_BACKEND_CLIENT_ID", "agent-console")
 _MCP_GATEWAY_URL = os.getenv("MCP_GATEWAY_URL", "https://nannos.gatana.nannos.ringier.ch/mcp")
+_MCP_GATEWAY_CLIENT_ID = os.getenv("MCP_GATEWAY_CLIENT_ID", "gatana")
+# Stateless JSON-RPC tools/list (no SDK handshake/parse); off = always list through the SDK.
+_MCP_CATALOGUE_STATELESS_LIST = os.getenv("MCP_CATALOGUE_STATELESS_LIST", "true").lower() not in ("0", "false", "no")
```
alloy-cicd: 🤖 **Convention drift (minor):** the orchestrator parses the same flag as `os.getenv(...).strip().lower() in {"1", "true", "yes"}` (`orchestrator-agent/app/models/config.py:375`), i.e. allowlist; here it is a denylist, so `MCP_CATALOGUE_STATELESS_LIST=""` (ConfigMap key present but blank) or `off` enables the fast path in the runner but disables it in the orchestrator. Same value, two semantics — worth aligning (a shared `env_bool` in agent_common would serve both).

aartaria:
Triage of the round-1 automatic review (commit 6b33c5ce):

1. **scheduler_\* routed to gateway** — fixed: uses the shared `is_console_backend_tool`.
2. **Store served with no TTL/user dimension** — fixed in part: a whitelist is served from the shared store only while this process listed the URL within 60s (same bound as the orchestrator's discovery cache), otherwise it re-lists. The user dimension is unchanged from how `dynamic_agent._resolve_catalogue_tools` already uses the store: every *call* still carries the user's own exchanged bearer, so the gateway enforces entitlements at call time; the whitelist is admin-configured.
3. **Expired user token no longer fails the run** — fixed: the provider exchanges up front for each audience the run needs (memoised, so no extra round trips), a failing exchange raises out of discovery as before.
4. **Fixed 30s listing timeout** — fixed: listing uses the run's `MCP_TIMEOUT_SECONDS`.
5. **Duplicates orchestrator `_ingest_server`** — discarded: sharing it means lifting the orchestrator's semaphore/config-coupled ingest into `agent-common`, a refactor beyond this PR. Noted as a follow-up candidate.
6. **Duplicate log lines** — fixed: only the resolver logs.
7. **Env parsing inconsistency** — fixed: allowlist parsing identical to the orchestrator's.

packages/agent-runner/agent/mcp_tools.py
```diff
@@ -0,0 +1,225 @@
...
+        # Exchange up front for every audience the run needs: discovery used to do this, and
+        # a user token that is expired or revoked must fail the run here, not surface as a
+        # run that "succeeded" without ever being able to call a tool.
+        for server in connections:
+            await self.token_provider.get(self.audience_for(server))
+
+        tools: list[BaseTool] = []
+        for name, (_held, entry) in store.resolve(names).items():
+            server = _server_for(name)
+            if not self._store_is_fresh(server):
```
alloy-cicd: 🤖 **Residual of round-1 #2 (minor):** the TTL is keyed by URL, so within the 60 s window a run for user B is still served the catalogue listed with user A's bearer. The commit message cites "the same bound as the orchestrator's cache", but that cache is per user (`discovery.py:286` — "discovery runs per user"), so the runner is looser than the orchestrator by the user dimension, not just by time. Consequence stays low (a tool B isn't entitled to is bound and 403s at call time rather than being absent), but two small things would settle it: either key `_LISTED_AT` by `(url, token identity)` when runs for different users share a pod, or state the accepted trade-off in the `_LISTED_AT` comment and update `CatalogueStore`'s docstring (`tool_catalogue.py:362`, "never *served* from here in place of a fetch"), which this path now contradicts.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (81fa61511f5a0a58f486b5eae7136701f98d3b8b..6b33c5ce1f522c557fba150fd96128af6d0196e2)

Ledger of round-1 findings:

| # | Finding | Disposition |
|---|---|---|
| 1 | `is_console_tool` missed `scheduler_*` | **verified-fixed** — aliased to shared `is_console_backend_tool`; `test_scheduler_tools_are_console_tools` covers it |
| 2 | Store hits served unconditionally (no TTL, no user dimension) | **partially fixed / still-open (minor)** — 60 s `_LISTED_AT` TTL bounds staleness; the user dimension is still absent (see inline) |
| 3 | Token exchange deferred to first tool call | **verified-fixed** — up-front `token_provider.get()` per needed audience; `test_a_failing_exchange_fails_discovery_even_on_a_store_hit` |
| 4 | Hard-coded 30 s stateless timeout | **verified-fixed** — uses `self._timeout` (`MCP_TIMEOUT_SECONDS`); `test_listing_honours_the_run_mcp_timeout` |
| 5 | `_list_server` duplicates orchestrator `_ingest_server` | **still-open** — not addressed and no reply; maintainability, fine as a follow-up |
| 6 | Duplicate `Loaded N/M MCP tools` log | **verified-fixed** |
| 7 | `MCP_CATALOGUE_STATELESS_LIST` denylist vs allowlist | **verified-fixed** — now `in {"1","true","yes"}` like the orchestrator |

Delta review: the fixes themselves introduce no new correctness or security regression. One inline note on the residual of #2. Nothing blocking from my side; the human reviewer decides.

packages/agent-runner/agent/mcp_tools.py
aartaria: Kept per-URL on purpose, documented in 136f39cc: keying by user would remove the cross-run sharing that is the point of the store (the whitelist is admin-configured, not user-chosen), and per-user entitlement is enforced at call time since every call carries that user's own bearer — a held name the user may not call still fails at the gateway. The residual exposure is a stale *schema* for up to 60 s, the same bound the orchestrator accepts. The `CatalogueStore` docstring contradiction predates this PR (`dynamic_agent` already resolves from the store) — noting it as an agent-common follow-up alongside the shared `_ingest_server` refactor rather than widening this PR.

aartaria: Revisiting after Andrea's input: Gatana profiles can hide tools of a server per user, so a listing is a genuine per-user view and per-URL sharing was wrong — a name in another user's listing could have been bound for a user whose profile hides it (it would fail at the gateway, but it should never be offered to the agent). Fixed by dropping the serve-from-store path and TTL altogether: every run does its own stateless tools/list with its own token; the store now only interns (dedups) identical catalogues, which is exactly its documented contract. Tests: `test_every_run_lists_with_its_own_token_and_never_binds_another_users_view`, `test_identical_listings_share_one_interned_catalogue`.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (6b33c5ce1f522c557fba150fd96128af6d0196e2..fa204fc1cc73fc415c034426fd86bae715227851)

Ledger of round-2 findings:

| # | Finding | Disposition |
|---|---|---|
| 2 (residual) | Store TTL keyed per URL, not per user | **verified-fixed** — author first documented the trade-off (136f39cc), then reversed it after confirming Gatana profiles hide tools per user: the serve-from-store path and `_LISTED_AT` TTL are gone; every run does its own `tools/list` with its own token and the store is intern-only, matching its docstring. Covered by `test_every_run_lists_with_its_own_token_and_never_binds_another_users_view` and `test_identical_listings_share_one_interned_catalogue`; PR description updated accordingly. |
| 5 | `_list_server` duplicates orchestrator `_ingest_server` | **still-open (follow-up)** — unchanged; maintainability only, not blocking |

Delta review: the simplification introduces nothing new — `resolve()` is now a straight list-per-needed-server loop, the up-front exchange, `MCP_TIMEOUT_SECONDS` timeout and `scheduler_*` routing from round 2 are preserved, `stats` and the log line were trimmed consistently, and `tests/test_mcp_tools.py` passes (13/13) at this head. Every previous finding is resolved and the delta introduces nothing new; no inline comments this round. The human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (fa204fc1cc73fc415c034426fd86bae715227851..fc9244b77d7c0126cee772d8d3f299b47c660db5)

Ledger of round-3 findings:

| # | Finding | Disposition |
|---|---|---|
| 5 | `_list_server` duplicates orchestrator `_ingest_server` | **still-open (follow-up)** — unchanged; with `CatalogueStore` gone both copies shrank to "stateless → memo → SDK fallback", so the lift into `agent_common` is now smaller, but it remains maintainability-only and not blocking |

Everything else from rounds 1–3 stayed fixed.

Delta review (`fc9244b7`, removal of `CatalogueStore` across agent-common / orchestrator / runner): no new correctness or security issue found. Checked specifically:
- no stray references to `CatalogueStore` / `get_catalogue_store` / `reset_catalogue_store` / `store.intern` / `store.resolve` remain anywhere under `packages/`;
- the stateless-capability memo moved to `catalogue_ingest` keeps the same per-URL semantics and both consumers (`discovery._ingest_server`, `McpToolResolver._list_server`) use it identically;
- `dynamic_agent._resolve_catalogue_tools` now lists each connection only when it has names of its class (console vs gateway) and binds by connection key, preserving the round-2 by-key-not-position guarantee (`test_listed_tools_bind_to_the_connection_by_key_not_position`); names with no reachable connection are still silently skipped, as before;
- tests at this head: agent-runner 34/34, orchestrator-agent 684/684, agent-common 749 pass — the 9 failures are all in `tests/test_thinking_blocks_round_trip.py`, untouched by this PR (pre-existing / environmental), plus one local-only collection error for `foundry_sdk`;
- `ruff` reports nothing in the changed files (existing errors are in untouched modules).

One stale-docs note, summary only: the PR description ("Before / after" and "Tests" sections) still cites `test_identical_listings_share_one_interned_catalogue` and "identical listings are interned once" — that test and behaviour were removed in this commit, so those two lines should be dropped or reworded.

Every previous finding is resolved or explicitly deferred as follow-up, and the delta introduces nothing new; no inline comments this round. The human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (fc9244b77d7c0126cee772d8d3f299b47c660db5..9c8c06c44ecf50832e94dc958bed444ef5456be6)

Ledger of round-4 findings:

| # | Finding | Disposition |
|---|---|---|
| 5 | `_list_server` duplicates orchestrator `_ingest_server` | **verified-fixed** — the "stateless → per-URL memo → SDK fallback" strategy now lives once as `catalogue_ingest.fetch_catalogue()`; both `McpToolResolver._list_server` and `ToolDiscoveryService._ingest_server` delegate to it with identical semantics (the `http_client is None` and `stateless=False` short-circuits are preserved, the runner reads `catalogue.source` instead of setting it), covered by the new `TestFetchCatalogueStrategy` (4 tests) |
| — | Stale PR-description lines citing the deleted interning test | **verified-fixed** — removed; the description now documents `fetch_catalogue()` instead |

Delta review (`9c8c06c4`): pure extraction, behaviour unchanged. Checked: the memo functions are no longer imported by either caller (single owner now), the log messages moved unchanged, callers' tests patch `agent_common.core.catalogue_ingest.fetch_catalogue_stateless/_mcp` rather than the (now non-existent) caller re-exports, and all suites pass at this head — agent-common 747 (pre-existing `test_thinking_blocks_round_trip.py` failures and local-only `foundry_sdk` collection error excluded, both untouched by this PR), agent-runner 34/34, orchestrator-agent 684/684. `ruff` reports nothing in the changed files.

Every previous finding is resolved and the delta introduces nothing new; no inline comments this round. The human reviewer decides.

PR #174, opened 2026-08-27
feat(agent-common): paginated, truncation-aware tool discovery search

aartaria:
closes #

## Problem
`tools.search` (PTC) and `search_tools` (PTC-off) returned a bare top-10 list. The model could not distinguish "exactly 10 tools matched" from "10 of 60 shown", so when the tool it needed ranked 11th it concluded the tool did not exist and stopped using search altogether. There was also no way to ask for more results, and ties in the token-overlap ranking fell back to registry order.

## Changes
- **New shared `agent_common/core/tool_search.py`** used by both discovery surfaces so they return the same shape:
  `{ matches: [{name, description}], total_matches, offset, shown, truncated, next_offset, hint }`
  `hint` is always present and states the situation in words (e.g. *"Showing 1-10 of 37 matches… call tools.search again with offset=10"*, *"All 4 matching tools shown"*, or which tokens matched nothing).
- Optional `limit` (default 10, clamped to 50) and `offset` args on both search tools.
- Deterministic ranking: whole-word hits beat substring hits (`repo` → `github_get_repo` before `gam_get_report`), then breadth of distinct tokens matched, then name. Full description is indexed; first line displayed.
- PTC discovery instruction (`graph_utils._PTC_DISCOVERY_INSTRUCTION`) and both tool descriptions now tell the model to check `truncated`/`next_offset` before concluding a tool does not exist.
- `call_tool` unknown-name suggestions reuse the same ranker.

## Tests
Updated existing tests to the new shape; added coverage for truncation + paging across 3 pages, limit clamping, whole-word-over-substring ranking, matching beyond the first description line, and an e2e check that the page metadata survives the QuickJS bridge. `agent-common`: 771 passed.

## Checklist
- [x] Does this PR introduce a breaking change? 🚨 — **Yes, for the model-facing contract only**: `search`/`search_tools` return an object instead of a list. No other Python code consumed the old list (only `dynamic_agent.py` imports from `tool_catalog`, using unchanged public names). Any prompt/skill text that assumed an array should be updated.


packages/agent-common/agent_common/core/tool_search.py
```diff
@@ -0,0 +1,168 @@
...
+    for tok in tokens:
+        if tok in entry["name_words"]:
+            weighted += 3
+            whole += 1
+        elif tok in entry["name_haystack"]:
+            weighted += 2
+        elif tok in entry["words"]:
+            weighted += 2
+            whole += 1
+        elif tok in entry["haystack"]:
```
alloy-cicd, line 119: **Stop-words inflate `total_matches` and make `truncated` nearly always true.**

`tokenize` keeps every `[a-z0-9]+` run, so a query like `"list campaigns for an advertiser"` yields tokens `for`, `an` (and `"get a report"` yields `a`). Those fall through to this substring tier (`tok in entry["haystack"]`, now the *full* description) and match almost every entry with weight 1. Result: `total_matches ≈ len(catalogue)`, `truncated=True`, and the hint tells the model to keep paging through hundreds of irrelevant tools — the exact confusion this PR is trying to remove, just inverted.

Pre-PR this was harmless (no truncation signal, first-line haystack only); now it directly corrupts the new contract.

Suggestion (any one of): skip the substring tiers for tokens shorter than 3 chars; require `matched >= 1` from the whole-word/name tiers before counting an entry; or only count an entry as a match if `weighted >= 2`.

Easy to verify: `rank(entries, "a")` over the real catalogue should not return every tool.

packages/agent-common/agent_common/core/tool_search.py
```diff
@@ -0,0 +1,168 @@
...
+            result["hint"] = (
+                f"No tool matched any of {tokens}. Try different or fewer words "
+                "(e.g. the object name alone) — the catalog may use other terminology."
+            )
+    elif start >= total:
+        result["hint"] = f"offset {start} is past the last match ({total} total); start from offset 0."
+    elif truncated:
+        result["hint"] = (
+            f"Showing {start + 1}-{end} of {total} matches. If the tool you need is not here, "
+            f"call {search_tool_name} again with offset={end}, or narrow the query."
```
alloy-cicd: **Hint uses kwarg syntax that is wrong for the PTC surface.**

For PTC-on the hint renders as `call tools.search again with offset=10`, but the tool takes a single object argument — the system prompt (`graph_utils.py:337-340`) and the tool description both say `tools.search({ query, offset: next_offset })`. A model following the hint literally writes `tools.search({query:'x'}, offset=10)` (SyntaxError) or ignores it and restarts at offset 0.

Suggest either dropping call syntax from the hint ("pass `next_offset` as `offset`") or letting each surface pass an example formatter alongside `search_tool_name`. Note `test_ptc_discovery.py:81` pins `"offset=10" in page1["hint"]` and would need updating.

packages/agent-common/agent_common/core/tool_search.py
```diff
@@ -0,0 +1,168 @@
...
+    for entry in entries:
+        key = _score(entry, tokens)
+        if key is not None:
+            scored.append((key, entry))
+    scored.sort(key=lambda pair: (tuple(-c for c in pair[0]), pair[1]["callable"]))
+    return [entry for _, entry in scored]
+
+
+def clamp_limit(limit: int | None) -> int:
+    if limit is None or limit < 1:
```
alloy-cicd, line 144: **`top_k` and `DEFAULT_LIMIT`/`MAX_LIMIT` are two sources of truth for the page size.**

- A caller-supplied `limit: 0` (or negative) lands here and becomes `DEFAULT_LIMIT` (10), not the configured `top_k` — e.g. `build_discovery_tools(catalog, top_k=5)` + `limit: 0` returns 10 while the description promises "default 5".
- `top_k=100` is silently capped to 50 (the removed `[:top_k]` slice honoured it), while the description still says "default 100".
- `_SearchArgs.limit` / `_SearchToolsArgs.limit` field descriptions hardcode `default {DEFAULT_LIMIT}` while the tool descriptions say `default {top_k}`.

Nothing outside tests passes `top_k` today, so simplest is to drop the `top_k` kwarg on both surfaces and let this module own the default (delete the two `_DEFAULT_TOP_K = DEFAULT_LIMIT` aliases). Otherwise thread it through: `clamp_limit(limit, default=top_k)` and interpolate `top_k` into the args-schema text too.

packages/agent-common/agent_common/core/tool_search.py
```diff
@@ -0,0 +1,168 @@
...
+        "truncated": truncated,
+        "next_offset": end if truncated else None,
+    }
+    if total == 0:
+        tokens = tokenize(query)
+        if not tokens:
+            result["hint"] = "Query contained no searchable words; use a few plain words describing the action."
+        else:
+            result["hint"] = (
+                f"No tool matched any of {tokens}. Try different or fewer words "
```
alloy-cicd: Nit: `{tokens}` interpolates the Python list repr (`No tool matched any of ['quantum', 'teleportation']`) into a message read by a JS-writing model. `', '.join(tokens)` or `' '.join(tokens)` reads cleaner.

packages/agent-common/agent_common/core/tool_search.py
```diff
@@ -0,0 +1,168 @@
...
+    truncated: bool
+    next_offset: int | None
+    hint: str
+
+
+def tokenize(text: str) -> list[str]:
+    return _TOKEN_RE.findall(text.lower())
+
+
+def first_line(description: str | None) -> str:
```
alloy-cicd, line 62: Nit (reuse): this is now the third copy of the first-line rule — `ToolCard.first_line` (`tool_catalogue.py:69`) and the inline expression in `ptc_signatures.py:195` do the same thing. Since this PR promotes the helper to a shared module, consider having those two call it so the "display description" rule cannot drift.

packages/agent-common/agent_common/core/ptc_discovery.py
```diff
@@ -77,31 +84,13 @@ def _build_entries(catalog: Sequence[BaseTool]) -> list[dict[str, Any]]:
...
-                "camel": camel,
-                "name": tool.name,
-                "desc": desc,
-                "name_haystack": f"{camel} {tool.name}".lower(),
-                "haystack": f"{camel} {tool.name} {desc}".lower(),
-                "tool": tool,
-            }
-        )
+        entry = make_entry(name=tool.name, callable_name=camel, description=tool.description)
+        entry["camel"] = camel
```
alloy-cicd: Nit: `entry["camel"]` always equals `entry["callable"]` (set by `make_entry`). `rank` sorts by `callable` while `_describe` indexes by `camel`; drop the extra key and use `by_camel = {e["callable"]: e for e in entries}`.

packages/agent-common/tests/test_tool_catalog.py
```diff
@@ -73,15 +69,30 @@ def test_contributes_three_meta_tools(self, middleware):
...
+        assert result["matches"] == []
+        assert result["total_matches"] == 0
+        assert "hint" in result
+
+    @pytest.mark.asyncio
+    async def test_search_pages_with_offset(self, middleware):
+        search = _meta_tool(middleware, CATALOG_SEARCH_TOOL_NAME)
+        page1 = await search.coroutine(query="campaign", limit=1)
+        assert page1["shown"] == 1
+        if page1["total_matches"] > 1:
```
alloy-cicd: The fixture is static (two `campaign` tools match), so this `if` either always runs or — after a fixture rename — silently turns the test into a no-op. Assert `page1["total_matches"] == 2` and `page1["truncated"] is True` unconditionally.

alloy-cicd (commented):

The shared `tool_search` module and the explicit `total_matches`/`truncated`/`next_offset` contract are a clear improvement over the bare `top_k` list. No blocking bugs found; the inline comments are about places where the new truncation signal can mislead the model, and a few contract/duplication nits.

**Summary of inline findings**
1. `tool_search.py:97` — stop-word tokens (`a`, `an`, `to`, `for`) hit the substring tier and match nearly every tool, so `total_matches ≈ catalogue size` and `truncated=True` for most natural-language queries — this undermines the very signal the PR introduces.
2. `tool_search.py:164` — the hint says `offset=10` (kwarg syntax) but the PTC surface takes an object arg `{ offset: next_offset }`; the two prose instructions disagree.
3. `tool_search.py:122` / `ptc_discovery.py:113` — `top_k` and `DEFAULT_LIMIT`/`MAX_LIMIT` are two sources of truth for the default page size (limit=0 → 10 not `top_k`; `top_k>50` silently capped; `_SearchArgs.limit` says "default 10" while the tool description says "default {top_k}").
4. `tool_search.py:156` — Python list repr leaks into the model-facing hint.
5. `tool_search.py:49` — third copy of the first-line helper (`ToolCard.first_line`, `ptc_signatures.py:195`).
6. `ptc_discovery.py:88` — `entry["camel"]` duplicates `entry["callable"]`.
7. `test_tool_catalog.py:92` — paging assertion is guarded by an `if` on a static fixture, so it can silently stop testing.

**Not in the diff, worth a follow-up:** `packages/agent-common/AGENTS.md:167` still documents `tools.search({query})` as returning bare matches; it should mention the paged `{ matches, total_matches, truncated, next_offset }` object.

aartaria:
Addressed all 7 round‑1 findings in 2275cd48, plus the AGENTS.md note:

1. Stop words / tokens < 3 chars now only match as a whole word of the *name* — `rank(entries, "a")` returns nothing; new test `test_stop_words_and_short_tokens_do_not_match_whole_catalog`.
2. Hint says "offset N (next_offset)" — no call syntax.
3. `top_k` kwarg dropped from both surfaces; `tool_search` owns `DEFAULT_LIMIT`/`MAX_LIMIT`.
4. Tokens joined in the hint.
5. `ToolCard.first_line` and `render_signature_block` reuse `tool_search.first_line`.
6. `entry["camel"]` removed.
7. Paging test asserts unconditionally.

agent-common: 772 passed.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (99615d7d60789ca5527d857036af3e5c7fefe324..2275cd4828bcad3b7cd1e1d9e5e91621763b7315)

| # | Finding | Disposition |
|---|---|---|
| 1 | Stop-words / short tokens inflate `total_matches` → `truncated=True` | ✅ verified-fixed — `_STOP_WORDS` + `_MIN_DESC_TOKEN_LEN` gate the substring/description tiers; short tokens still match via `name_words` (`ad` → `adGet`); new `test_stop_words_and_short_tokens_do_not_match_whole_catalog` |
| 2 | Hint used kwarg syntax `offset=10` (wrong for PTC object arg) | ✅ verified-fixed — hint now says "offset N (next_offset)", no call syntax; test updated |
| 3 | `top_k` vs `DEFAULT_LIMIT`/`MAX_LIMIT` two sources of truth | ✅ verified-fixed — `top_k` kwarg removed from `build_discovery_tools` and `ToolCatalogMiddleware`; descriptions interpolate `DEFAULT_LIMIT`; no remaining callers |
| 4 | Python list repr in no-match hint | ✅ verified-fixed — `' '.join(tokens)` |
| 5 | Third copy of `first_line` | ✅ verified-fixed — `ToolCard.first_line` and `render_signature_block` delegate to `tool_search.first_line` (no import cycle: `tool_search` imports only stdlib) |
| 6 | `entry["camel"]` duplicated `entry["callable"]` | ✅ verified-fixed |
| 7 | Conditional paging assertion in `test_search_pages_with_offset` | ✅ verified-fixed — asserts `total_matches == 2`, `truncated`, `next_offset == 1`, page-2 exhaustion unconditionally |
| — | `AGENTS.md:167` stale return shape (summary note) | ✅ verified-fixed |

**Delta review:** the fix commit introduces nothing new. `_score`'s `name_only` short-circuit correctly skips `matched += 1` so gated tokens neither match nor count toward breadth; removing `top_k` leaves `clamp_limit(None)` as the single default path. Ran `tests/test_ptc_discovery.py`, `tests/test_tool_catalog.py`, `tests/test_ptc_core_render_e2e.py` at `2275cd48`: 51 passed.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (2275cd4828bcad3b7cd1e1d9e5e91621763b7315..35d80d1eec152075454b3e00969bc1047f1e1c8f)

**Ledger:** round 2 closed with all seven round-1 findings verified-fixed and no new findings, so there was nothing carried over into this round (nothing still-open, nothing rejected-by-author).

**Delta review** (`35d80d1e`, author-initiated refinement of the round-1 stop-word fix): 1–2 letter tokens now match as a whole word in either the name *or* the description (previously name only), and never as a substring; stop words stay name-whole-word only; 3+ letter tokens keep the substring tiers. Checked:
- The reordering puts the whole-word-description tier (weight 2) ahead of the substring-name tier (weight 2). Weights are equal, so ranking by component 1 is unchanged; the only effect is that a token that is both a name substring and a description whole word now also counts toward the `whole` tie-breaker, which is the more sensible outcome. `test_search_whole_word_beats_substring` still passes.
- Gated tokens still `continue` before `matched += 1`, so they cannot inflate `total_matches`.
- New assertions cover `ga` → description whole-word hit only (not `gam`), and `gam` → substring tiers intact.
- `tests/test_ptc_discovery.py`, `tests/test_tool_catalog.py`, `tests/test_ptc_core_render_e2e.py` at `35d80d1e`: 30 passed.

PR #175, opened 2026-08-27
feat: Nannos Panel / embed-sdk (v2)

thecopy:
## Checklist


packages/orchestrator-agent/app/models/responses.py
```diff
@@ -37,6 +37,14 @@ class AgentStreamResponse(BaseAgentStreamResponse):
...
         default=None, description="Review policies specifying allowed decisions per action"
     )
+    client_action_request: Optional[dict] = Field(
+        default=None,
+        description=(
+            "Embedded Nannos: a client-action awaiting its RESULT from the embedding "
+            "client ({id, directive}). Emitted as input_required with the client-action "
+            "extension; the SDK executes the directive and auto-resumes with the result."
+        ),
+    )
```
aartaria: why can't we use the `action_requests`?

packages/agent-common/agent_common/core/tool_call_summarizer.py
```diff
@@ -0,0 +1,160 @@
...
+            lines.append(f"  arguments: {json.dumps(_compact_args(args), ensure_ascii=False, default=str)}")
+            lines.append("")
+        lines.append(f"Summarize these {len(calls)} tool call(s).")
+
+        result: ToolCallSummaries = await structured_model.ainvoke(
+            [
+                {"role": "system", "content": _SYSTEM_PROMPT},
+                {"role": "user", "content": "\n".join(lines)},
+            ]
+        )
```
aartaria: is the cost logger instrumented properly with sub-agent id and so on?

packages/agent-common/agent_common/core/client_action_tool.py
```diff
@@ -81,6 +143,18 @@ async def _client_action_handler(
...
+        # ROUND TRIP: pause the graph until the browser reports what happened.
+        # The directive rides the interrupt value (NOT the custom stream): the
+        # resume replays this handler from the top, and anything emitted before
+        # ``interrupt()`` would fire twice. ``tool_call_id`` is injected and
+        # stable across that replay — it is the id the client echoes back on
+        # its ``client_action_result`` decision.
+        logger.info(f"[CLIENT-ACTION] Awaiting result for directive: {directive}")
+        result = interrupt({"client_action_request": {"id": tool_call_id, "directive": directive}})
+        logger.info(f"[CLIENT-ACTION] Result received: {result}")
+        return _render_client_action_result(kind, result)
```
aartaria: would it still comply with the ordering batch interrupts?
```
            # 2. Static guards take priority
            if self._should_interrupt(tool_call):
                config = self.interrupt_on[tool_name]
                action_request, review_config = self._create_action_and_config(tool_call, config, state, runtime)
                # Stamp the stable per-call id on EVERY interrupted call (here a static
                # guard, no risk metadata) so the client can return one decision per
                # action_request and the resume path aligns them by id. Display-only:
                # ``args`` is never passed to the tool (approve replays the original call).
                action_request["args"] = {**action_request.get("args", {}), "_call_id": tool_call["id"]}
                action_requests.append(action_request)
                review_configs.append(review_config)
                interrupt_indices.append(idx)
                _risk_metadata.append({"source": "static_guard"})
                continue
```

aartaria:
🤖 **Automatic review** — Claude Code `/code-review` of `packages/orchestrator-agent` + `packages/agent-common` vs `origin/main` — disposition ledger

17 findings reported (7 finder angles, 15 candidates adversarially verified; 6 refuted and not listed).

### Fixed in this PR
| Finding | Commit |
|---|---|
| Destructive-verb risk floor skipped on risk-cache hits (`tool_risk_scorer.py`) | dd525a7b |
| `_resume_pending()` misfired on a 2nd interrupt in the same node run (`tool_call_summarizer.py`) | on branch (`hitl_resume.resume_will_return`) |
| `derive_mcp_server_slug` split at first token in tuple order (`alloy_post_get_report → alloy_post`) | dd525a7b |
| Gateway discovery failure swallowed → console-only toolset cached for the TTL (`discovery.py`) | dd525a7b |
| `streamed_chars` redundant with `len(streamed_text)`; `client_objects` written into config metadata twice | dd525a7b |
| `<current_page>`/`<client_objects>` on the last human message busted the prompt cache every turn — now a trailing flagged message; caching breakpoint skips it; both middlewares share `inject_embedded_context` | 5b13feda |

### Accepted as-is
- HITL bypass rules saved from sub-agent cards under `<tool>::_self` stop matching once those tools derive a real slug. Fails toward asking; users re-tick once. No migration.

### Still open — pick up later
**Plausible / low severity**
- [ ] `app/core/agent.py` `stream_subagent` (~L1482): attachments reach the embedded sub-agent as files under `/attachments/`, not as multimodal content blocks (the delegated `task` path builds `HumanMessage(content_blocks=...)`). Matters only if embedded agents should interpret images.
- [ ] `agent_common/middleware/tool_status.py` (~L114): the `/skills/` readFile pre-check returns "Loading skill X…" before `_extract_ptc_tool_calls`, so sibling tool calls in the same eval aren't named in the activity line. Cosmetic — HITL cards still name them.
- [ ] `app/utils.py` (~L586): `client_action` is registered on `user_config.client_objects` alone, without checking the client negotiated `urn:nannos:a2a:client-action:1.0`. Unreachable via console-backend (header always set, pinned by conformance test); hardening for third-party A2A clients.

**Reuse / simplification (no behavior change)**
- [ ] `app/core/executor.py` (~L693/710): protobuf-Struct unwrap copy-pasted for `clientObjects` and `pageContext`; `int(float(...))` re-implements `agent.py::_origin_int`. → one `_unwrap_struct(value, expected_type)`.
- [ ] `app/core/a2a_extensions.py` (~L261/287): `new_client_action_message` / `new_client_action_request_message` differ only in the DataPart key; five builders share the Part/ParseDict/Message boilerplate. → private `_data_part_message(...)`.
- [ ] `agent_common/core/tool_risk_scorer.py` (~L129, L312, L350): two hand-built fixed-score `ToolRiskEntry`s → `_fixed_entry(score)`; `_HARD_DESTRUCTIVE_KEYWORDS` is a subset of `_deterministic_fallback`'s `dangerous_keywords` → hoist and share.
- [ ] `agent_common/middleware/ptc_guard.py::_resolve_server_slug` vs `conditional_hitl.py::_get_server_slug`: same 4-step resolution in two files ("must match" per docstring; step 3 already sources the tool differently). → one shared resolver with an optional `tool` arg.

**Altitude**
- [ ] `notify_user` / `client_action` are special-cased by name in five frozensets across two packages (`score_tool_risk`, `_SELF_EVIDENT_TOOLS`, `_PTC_EXCLUDED_TOOL_NAMES`, `_SUPPRESSED_TOOLS`, `_ACTIVITY_LOG_EXCLUDED_TOOLS`). A per-tool `metadata` policy flag would replace them; worth doing before the next control tool is added.

**Efficiency nit**
- [ ] Full `clientObjects` / `pageContext` payloads (incl. form `values`) logged at INFO every turn — `executor.py` ~L701/718, `client_action_tool.py` ~L160/162. → counts/ids at INFO, payload at DEBUG.

PR #177, opened 2026-08-28
Feat/interrupt card

aartaria:
closes #

## Checklist


packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -190,6 +204,14 @@ async def awrap_tool_call(
         if tool_name in RESPONSE_TOOLS:
             return await handler(request)
 
+        # A refusal that arrived while this tool was waiting on authorization has to
+        # be honored BEFORE the tool runs. Below, the answer is only read where the
+        # call fails a second time — so once the user has completed the login in
+        # their browser the retry succeeds, and their "no" is read by no one.
+        veto = self._refusal_before_running(request, tool_name)
```
alloy-cicd, line 218: **Declined authorization vetoes every sibling tool call in the node.**

`_refusal_before_running` runs for *every* tool call and reads `pending_authorization_answer()`, which scans the task-wide `scratchpad.resume` list for any `{"authorization": {...}}` shape. All parallel tool calls in one ToolNode share that task, and nothing here compares the answer to the tool (or `tool_call_id`) it was asked for.

Scenario: the model calls `github_get_me` and `web_search` in parallel; GitHub needs auth → interrupt; user clicks Skip → the tools node re-runs both, and `web_search` returns `"The user DECLINED the authorization required by web_search"` without ever executing. Same for sub-agents (`utils.py` installs the same middleware).

Suggest keying the answer to the interrupt/tool it settled — `auth_requirement["tool"]` and the tool call id are both known when the interrupt is raised — and honoring it only for that call.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -298,15 +328,178 @@ async def awrap_tool_call(
...
+                ),
+            )
+            if intent == "approve":
+                verdict = "approved"
+            elif intent == "reject":
+                verdict = "declined"
+        logger.info(f"[AUTH MIDDLEWARE] Resumed auth interrupt for {tool_name}: {verdict}")
+
+        if verdict == "approved":
+            return await handler(request)
```
alloy-cicd: **Approved retry bypasses auth detection — contradicts the docstring above.**

Detection lives in `awrap_tool_call` *around* the handler, so `return await handler(request)` here returns whatever the retry produces unchecked. If the user clicks "Done, continue" before the OAuth flow actually completed, the retry returns the same `need-credentials` ToolMessage (or, on the exception path at L334, raises a new `ToolException` from inside the `except`) and the raw `authorizeUrl` reaches the model as prose — no second card, which is the exact failure this PR set out to remove. Secondary: the tool is executed twice per resume on this path.

Route the retry back through the detection (e.g. factor L223–268 into a helper and call it here), or re-raise into `awrap_tool_call`.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -188,13 +281,31 @@ def _build_interrupt_resume_map(cls, interrupts: Any, hitl_decisions: list, quer
         an interrupt (by ``call_id``). Non-HITL interrupts (auth, etc.) resume with the
         raw ``query``.
         """
-        decisions_by_id = {d["id"]: d for d in hitl_decisions if isinstance(d, dict) and "id" in d}
+        decisions_by_id = {d["id"]: d for d in (hitl_decisions or []) if isinstance(d, dict) and "id" in d}
```
alloy-cicd, line 295: (Anchored here; the defect is at **L324**, outside the diff.) **`hitl_decisions` can be `None` here → `TypeError: 'NoneType' object is not iterable`.**

`_extract_hitl_decisions` now returns `None` (L290 relies on that), and L284/L299 were updated with `or []`, but this generator was not. Reached when a client-action round trip is parked and the user types a message instead of the SDK sending its result DataPart — precisely the "user typed while parked" case the comment above says this branch handles. `authorization` is `None` too (no auth interrupt), so nothing short-circuits before this line.

`for d in (hitl_decisions or [])` fixes it.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -188,13 +281,31 @@ def _build_interrupt_resume_map(cls, interrupts: Any, hitl_decisions: list, quer
...
+                if authorization and not decisions_by_id:
+                    # The client answered an AUTHORIZATION prompt while this approval
+                    # was the pending question (the agent re-ran the blocked tool and
+                    # its guard asked again). Reject it explicitly, with the reason —
+                    # the synthetic blanket reject `_extract_hitl_decisions` falls back
+                    # to is safe but tells the model nothing it can act on.
+                    translated = structural_decisions({"authorization": authorization}, action_requests)
+                    if translated:
+                        per = translated
                 resume_map[intr.id] = {"decisions": per}
```
alloy-cicd, line 332: **Unrecognized authorization verdict resumes a HITL interrupt with zero decisions.**

When `dataParts` carries `{authorization: {decision: "<unknown word>"}}` and no `decisions` part: `hitl_decisions` is `None` → `per = []` (L299); `authorization` is truthy so L306 runs, but `authorization_verdict` returns `None` for the unknown word, `structural_decisions` returns `None`, and `per` stays `[]`. `{"decisions": []}` is then delivered against N `action_requests`, and `ConditionalHumanInTheLoopMiddleware` raises `Number of decisions (0) does not match (N)` — the turn dies with a system error where `main`'s blanket-reject fallback failed safe.

When `translated` is falsy and `per` is empty, fall back to an explicit reject (with the reply as reason) rather than an empty list.

alloy-cicd (commented):

Reviewed at `8461d4f` (46 files, +3852/−325). Python suites pass in the worktree (agent-common 98, orchestrator-agent 182); embed-sdk `tsc --noEmit` is clean (vitest could not run — no `node_modules` in the review worktree, so the new TS tests were read, not executed).

The in-task-auth wire contract itself checks out end to end: `urn:nannos:a2a:in-task-auth:1.0` is identical across the four registries, `AuthPayload.client_payload()` field names match `readAuthDataPart`, and the cockpit consumer only imports `useNannos`/`NannosHostAdapter`/`FieldBridge`/`pkce`, so the removed i18n keys are safe.

**Inline findings (most severe first):**

1. `auth_error_middleware.py:211` — the new pre-run veto refuses **every** tool call in the node once any declined authorization sits in the task's resume log; nothing scopes it to the tool that asked. *(high)*
2. `auth_error_middleware.py:447` — the `approved` retry calls `handler()` bare, bypassing auth detection; a still-unauthorized retry hands the raw `need-credentials` payload to the model instead of re-interrupting (the docstring claims the opposite). *(high)*
3. `executor.py:324` — `hitl_decisions` can now be `None` and the client-action fallback iterates it unguarded → `TypeError`. *(high)*
4. `executor.py:299` — an authorization answer with an unrecognized verdict yields `{"decisions": []}` against N pending calls; the removed blanket-reject fallback used to guard this. *(medium)*
5. `history-mapper.ts:142` / `send-payload.ts:110` — receipt `kind`/`outcome` are never persisted; authorization receipts reload as `Context:` chips. *(high)*
6. `keys.ts:221` — four new `thread.*` keys missing from `nannosStringKeys`. *(medium)*
7. `auth_error_middleware.py:384` — `_resume_decision` duplicates `authorization_verdict` with a drifted word list (`done`/`skip` missing). *(medium)*
8. `executor.py:1622` — `correlation_id` reads `metadata["tool_call_id"]`, which the auth interrupt value never sets; the field is always dropped. *(medium)*
9. `thread.tsx:812` — `countDecisions` counts every `output-available` dynamic-tool, including SDK-auto-answered client-action round trips, as "approved". *(low)*

Noted but not filed as inline: `stream_handler.py` `_current_turn_produced_nothing` shares `_has_text` (only `str` / `{type:'text'}` blocks) with `continue_on_truncation` — a provider that emits a different block shape would have a real answer rewritten to "did not manage to write the answer"; and `hitl_resume._next_interrupt_index` returning `0` (rather than raising) when the LangGraph scratchpad shape is unrecognized makes `resume_will_return()` report `True` for every interrupt after the first. Both are worth a look but depend on provider/LangGraph shapes I could not exercise here.

Submitted as a comment only — the human reviewer decides.

packages/embed-sdk/src/transport/history-mapper.ts
```diff
@@ -271,6 +271,7 @@ export function rowsToUIMessages(rows: RestMessageRow[]): NannosUIMessage[] {
           ...readAuthRequired(
```
alloy-cicd, line 289: (Anchored on this file's only hunk; the restore is at **L136–143**.) **Authorization receipts reload as `Context:` chips.**

`send-payload.ts:110` persists only `display.label` as `metadata.injectedDisplayText`; `kind: 'receipt'` and `outcome` never reach the wire, and this mapper hard-codes `kind: 'context'` on restore. Live thread: "Authorized GitHub · asked Nannos to retry" as a muted `ReceiptLine`. After reload / `loadOlderMessages`: the same row renders through `ContextChip` as "Context: Authorized GitHub" with a chevron that expands to the English agent-facing prompt. The auth-required *card* side of the mapper handles this asymmetry; the receipt side was missed.

Persist `display.kind`/`outcome` (or the `authorization` decision) in metadata and restore them here.

packages/embed-sdk/src/i18n/keys.ts
```diff
@@ -202,10 +236,9 @@ export const nannosStringKeys = [
   'conversations.deleteError',
```
alloy-cicd, line 240: (The gap is in the `thread.*` block just above, **L213–221**.) `nannosStringKeys` is missing the four new interface entries `thread.activityApproved`, `thread.activityRejected`, `thread.pendingOne`, `thread.pendingMany` (declared at L50–58; present in all four locale tables with matching placeholders). `satisfies readonly (keyof NannosStrings)[]` checks membership, not exhaustiveness, so nothing catches it.

Effect: a host building its override map from `nannosStringKeys` (the documented pattern at L9) can never translate the folded-step count or the pending pill, and `locales.test.ts`'s completeness check skips these keys.

packages/orchestrator-agent/app/middleware/auth_error_middleware.py
```diff
@@ -298,15 +328,178 @@ async def awrap_tool_call(
...
+        ``unclear`` rather than guessed at here.
+        """
+        if isinstance(resume, dict):
+            decision = resume.get("authorization")
+            if isinstance(decision, dict):
+                verdict = str(decision.get("decision") or "").lower()
+                message = str(decision.get("message") or "")
+                if verdict in ("approved", "approve", "completed"):
+                    return "approved", message
+                if verdict in ("declined", "decline", "reject", "rejected", "skipped"):
```
alloy-cicd: `_resume_decision` re-implements `agent_common.core.hitl_resume.authorization_verdict` (already imported alongside `pending_authorization_answer`/`classify_reply` in this file) with a drifted vocabulary: `hitl_resume` accepts `done` as approved and `skip` as declined (`_APPROVED_WORDS`/`_DECLINED_WORDS`, hitl_resume.py:50–51); this copy does not. The same `{"decision": "skip"}` is therefore read as declined by `structural_decisions` (executor path) but `unclear` here — and then pays for a fast-LLM classification of an empty message.

Harmless today because the SDK only sends `approved`/`declined`, but it is two readers of one wire field. Suggest `verdict, msg = authorization_verdict(resume); return (verdict or "unclear", ...)`.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -1459,11 +1596,41 @@ def _ext_active(uri: str) -> bool:
...
+                    # is a different fact, and only the producer can name it. It
+                    # stays EMPTY when unknown rather than falling back to the
+                    # tool: a `need-credentials` raised inside the sandbox is
+                    # reported against `eval`, and "Authorization needed for eval"
+                    # is worse than not naming anything at all.
+                    service=metadata.get("service") or "",
+                    resource=tool_name,
+                    auth_url=metadata.get("auth_url") or "",
+                    description=metadata.get("message") or "",
+                    correlation_id=metadata.get("tool_call_id") or "",
```
alloy-cicd, line 1657: `metadata.get("tool_call_id")` reads a key nothing sets: the `auth_requirement` dict built in `auth_error_middleware.py:239–252` has `tool`/`service`/`subagent`/`message`/`auth_url`/`error_code`/`timestamp` only, and `from_interrupt`/`auth_required` forward the interrupt's keys verbatim. So `correlation_id` is always `""` and dropped by `client_payload()` — the correlation the extension doc promises is dead on arrival. Either stamp `request.tool_call["id"]` into the interrupt value in the middleware, or drop the field until it can be populated.

packages/embed-sdk/src/panel/components/thread.tsx
```diff
@@ -755,6 +799,27 @@ function groupActivity(parts: MessagePart[], fold: boolean): Array<MessagePart |
...
+ * "1 approved, 1 rejected" for the collapsed label, or '' when the run holds no
+ * decisions. Counts settled HITL parts only: a pending one renders its card
+ * outside the group, and an activity line is not a decision.
+ */
+function countDecisions(parts: ActivityPart[], strings: ReturnType<typeof useStrings>): string {
+  let approved = 0;
+  let rejected = 0;
+  for (const part of parts) {
+    if (part.type !== 'dynamic-tool') continue;
+    if (part.state === 'output-available') approved += 1;
```
alloy-cicd, line 831: Minor: `countDecisions` treats every `dynamic-tool` in `output-available` as a user approval, but `a2a-transport.ts:200–216` settles client-action round trips (`_clientActionRequest`) and `allow-edits` auto-applies to `output-available` too — the panel answered those itself. A turn where the SDK auto-filled two forms and asked nothing reads "Worked through 5 steps · 2 approved". Filter with the same `isRoundTrip` guard used for the pending branch at L660 (which currently only applies there, so those parts also get an "Approved client_action" receipt at L671).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (8461d4feaa22acca825a5b3be64fcc91f368138c..8d6903fe2470cc26f33ecdb7e18d9091dbbee05c)

Delta: one commit, `8d6903fe fix(auth): settle the authorization for the call that asked for it` (11 files, +508/−166). No thread replies or description changes; every response is in the code.

| # | Finding | Disposition | Evidence at `8d6903fe` |
|---|---|---|---|
| 1 | Declined auth vetoed every sibling tool call | **verified-fixed** | `_auth_requirement` stamps `tool_call_id` (auth_error_middleware.py:625); executor echoes it onto the answer (executor.py:354–358); `_refusal_before_running` skips calls whose id differs (L653–658), honoring id-less legacy answers. Tests: `test_refusal_does_not_veto_a_sibling_tool_call`, `test_refusal_still_vetoes_the_call_it_was_asked_for`, `test_an_answer_without_a_call_id_is_honored_as_before` |
| 2 | Approved retry bypassed auth detection | **verified-fixed** | `_after_auth_interrupt` returns `_RETRY`; `awrap_tool_call` loops through `_call_and_detect` so the retry hits the same detection (L470–526). Test: `test_approved_retry_that_is_still_unauthorized_asks_again` (2 handler calls, 2 interrupts) |
| 3 | `hitl_decisions` `None` iterated → `TypeError` | **verified-fixed** | `(hitl_decisions or [])` at executor.py:341; test `test_a_parked_client_action_survives_a_typed_message` |
| 4 | Unreadable auth verdict → `{"decisions": []}` | **verified-fixed** | falls back to `reject_decisions(action_requests, _UNREADABLE_AUTHORIZATION_MESSAGE …)` (executor.py:317–328); test `test_an_unreadable_verdict_rejects_instead_of_sending_no_decisions` |
| 5 | Receipts reloaded as `Context:` chips | **verified-fixed** | `send-payload.ts:115–117` persists `injectedDisplayKind`/`injectedDisplayOutcome`; `history-mapper.ts:125–133` restores them. Console-backend stores the whole metadata dict as `jsonb` (`messages_service.py:442`), so the new keys survive the round trip. Tests in `send-payload.test.ts`, `history-mapper.test.ts` |
| 6 | Four `thread.*` keys missing from `nannosStringKeys` | **verified-fixed** | keys.ts:221–225 |
| 7 | `_resume_decision` duplicated `authorization_verdict` with drifted vocab | **verified-fixed** | now delegates to `authorization_verdict` (L676–679) |
| 8 | `correlation_id` read a key never set | **verified-fixed** | same `tool_call_id` stamp as #1; test `test_interrupt_value_carries_the_blocked_call_id` |
| 9 | `countDecisions` counted SDK-answered round trips | **verified-fixed** | shared `isClientActionRoundTrip` guard in count and receipt (thread.tsx:32–35, 59, 72); test `says nothing about a client-action round trip the SDK answered itself` |

rejected-by-author: none. still-open: none.

**Delta review:** no new issues. I looked specifically at the `while True` retry loop (each iteration calls `interrupt()`, which raises once the resume log is exhausted, so it is bounded by the number of stored answers), at the refactored `_call_and_detect` preserving the ToolMessage / `Command` / `ToolException` detection paths, and at the id-scoped veto's interaction with local sub-agents (the orchestrator's `task` call now correctly runs and `dynamic_tool_dispatch` delivers the held answer to the sub-agent's own middleware, whose interrupt value carries the matching inner call id).

Verification: orchestrator-agent `test_auth_error_middleware.py` + `test_agent_executor.py` → 89 passed; agent-common → 932 passed; the 9 failures are all `tests/test_thinking_blocks_round_trip.py` raising `ModuleNotFoundError: langchain_openai` in `model_factory.py:147`, and `test_foundry_cost_version.py` fails to collect on a missing `foundry_sdk` — both are packages absent from this review environment, and neither file is touched by the PR. embed-sdk `tsc`/`vitest` could not run in the review worktree (no `node_modules`); the new TS tests were read, and both `RestMessageRow.metadata` and the outgoing payload metadata are `Record<string, unknown>`, so the two new keys carry no typing risk.

Every previous finding is resolved and the delta introduces nothing new — clean incremental round, no inline comments. Submitted as a comment only; the human reviewer decides.

PR #179, opened 2026-08-31
feat: in-task auth on the Slack and Google Chat clients

### LINKED ISSUE #178: Slack and Google Chat clients: handle the in-task-auth extension instead of posting the gateway's prose
The web client (embed-sdk) negotiates `urn:nannos:a2a:in-task-auth:1.0` and renders its own authorization card when a tool needs the end-user's consent (the MCP gateway's `need-credentials`, surfaced as A2A's `auth-required` task state). The Slack and Google Chat clients do not: they never advertise the URN, so the status arrives as prose the gateway wrote **for the agent** ("You must tell the end-user to…") — and they post that text verbatim to the user.

Two consequences:

1. The user reads machine-to-machine instructions, with the authorize URL buried in them, and has to click a bare link in a paragraph.
2. There is no way to *answer* the interrupt. The parked task accepts a decision DataPart (`{"authorization": {"decision": "approved"|"declined"}}`), which retries the blocked tool or tells the agent to stop pushing the link. Without it the user has to type a sentence and the model has to guess whether it meant "done" or "no".

## Expected

Both clients:

- advertise `urn:nannos:a2a:in-task-auth:1.0` in `X-A2A-Extensions`;
- read the structured DataPart (falling back to status metadata, then to the URL in the prose, so a producer that never negotiated the URN still works);
- render their own card — Block Kit / cardV2 — naming the service and the tool, with the authorize link and both answers;
- resume with the `authorization` decision DataPart plus agent-facing prose, the same as the web client;
- never render the gateway's agent-facing text when a URL was found.

Sandbox/orchestration tool names (`eval`, `task`, `bash`, …) must not be shown as the thing being authorized — a `need-credentials` raised inside the sandbox is reported against `eval`.

aartaria:
closes ringier-data/nannos#178

Building on #177: the Slack and Google Chat clients now register for `urn:nannos:a2a:in-task-auth:1.0` and handle the in-task auth interrupt the way the web client does.

Before this, neither client advertised the URN, so an `auth-required` status arrived as prose the MCP gateway wrote **for the agent** ("You must tell the end-user to…") — and both clients posted it verbatim. The user read machine-to-machine instructions with the authorize URL buried inside, and had no way to *answer* the parked interrupt: typing a sentence left the model to guess whether it meant "done" or "no".

## What each client gained

1. **Negotiation** — `in-task-auth` added to `X-A2A-Extensions`. Emission is gated server-side on it, so without it the status stays prose.
2. **Reading the payload** — a new `src/utils/inTaskAuth.ts` per client, with the same precedence as the SDK's `readAuthRequired`: DataPart → status metadata → last-resort URL scrape out of the prose. A producer that never negotiated the URN therefore keeps working unchanged. It also carries over the SDK's `OPAQUE_TOOLS` filter, so a `need-credentials` reported against `eval` never becomes "Authorization needed for eval".
3. **Our own card** instead of the gateway's prose:
   - Slack: `buildAuthRequiredWidget` posted as its own message, with the timeline paused (`streamer.pause('Awaiting your authorization')`) and the in-flight record released the way a HITL pause does. The card's message becomes the decision receipt on resume.
   - Google Chat: `googleChatService.buildInTaskAuthCard`, posted privately; finalizing gained `suppressStatusText` so the prose is not posted beside the card that replaced it — anything streamed *before* the interrupt is still delivered from the artifacts.
   - The wire text survives only as the fallback when no URL was found anywhere.
4. **Answering the interrupt** — both buttons resume with `{"authorization": {"decision": "approved"|"declined"}}` plus the same English agent-facing prose the panel sends (the fallback for a server that never routed the DataPart). Slack: a new `listeners/actions/inTaskAuthButton.ts`, including an ack-only handler on the URL button so Slack does not warn "operation timeout". Google Chat: a new `in_task_auth_card` case in `buttonClickedHandler`.

One deliberate divergence from the panel: neither card walks through stages (Authorize → then "Done, continue"). A Google Chat `openLink` reports nothing back, and on Slack it would make the way forward depend on a `chat.update` succeeding — so all buttons are live from the start.

The orchestrator change is documentation only: the extension's contract still claimed there was no decision for a client to send, which #177 had already made untrue.

## Verification

- `npx tsc --noEmit` clean in both clients.
- `client-slack`: 77/77 jest tests pass (10 new); `client-google-chat`: 29/29 (7 new). The new tests pin the reading precedence (payload / metadata / prose), the plumbing-name filter, card contents and routing ids, and the answer shapes.
- `npx eslint .` cannot run in either client in this environment — the private `@alloy-ch/eslint-config-alloy` is not installed. Pre-existing, unrelated to this change.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no: the extension is additive and negotiated, and a producer that never sends the DataPart is handled by the metadata/prose fallbacks.

### DIFF
--- a/packages/client-google-chat/src/handlers/messageHandler.ts
+++ b/packages/client-google-chat/src/handlers/messageHandler.ts
@@ -825,6 +825,17 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
               statusEvent,
               config,
             );
+            if (authCardPosted) {
+              // Park the public spinner. Updating it in place is the ONLY thing
+              // that ever overwrites "🧠 Working…", and with the card carrying the
+              // prompt there is usually no final text to post — so without this
+              // the thread keeps a spinner that spins forever beside a card
+              // asking for a login. Written into the status line the block below
+              // already renders, rather than pushed here, or that same block
+              // would put the spinner straight back in this iteration.
+              statusMessage.thinking = '⏸️ Awaiting your authorization';
+              statusMessage.activity = '';
+            }
           }
 
           if (statusEvent.status.message?.extensions?.includes('urn:nannos:a2a:work-plan:1.0')) {
--- a/packages/client-google-chat/src/handlers/messageHandler.ts
+++ b/packages/client-google-chat/src/handlers/messageHandler.ts
@@ -17,6 +17,7 @@ import { HandlerDependencies } from './types.js';
 import { getSpinnerVerb } from '../utils/spinnerVerbs.js';
 import { FileStorageService } from '../services/fileStorageService.js';
 import { Config } from '../config/config.js';
+import { readAuthRequired } from '../utils/inTaskAuth.js';
 
 // ---------------------------------------------------------------------------
 // Types
@@ -394,6 +395,65 @@ async function processMessageAttachments(
     return processedFiles;
 }
 
+/**
+ * The in-task authorization interrupt: A2A's own `auth-required` state, whose
+ * payload schema is the in-task-auth extension.
+ *
+ * The status TEXT is the MCP gateway addressing the AGENT ("You must tell the
+ * end-user to…") — the interrupt fires in middleware, before the model, so no
+ * LLM ever rewrites it. Google Chat used to post it verbatim. The card is our
+ * own copy instead, built from the DataPart when the producer sent one and from
+ * metadata or the prose's URL when it did not.
+ *
+ * Returns whether the card was posted — the caller then keeps that prose out of
+ * the finalized message.
+ */
+async function processAuthRequiredEvent(
+  logger: Logger,
+  chatService: GoogleChatService,
+  inFlightTaskStore: IInFlightTaskStore,
+  projectId: string,
+  spaceId: string,
+  threadId: string,
+  userId: string,
+  accumulatedTask: Task,
+  statusEvent: TaskStatusUpdateEvent,
+  config: Config,
+): Promise<boolean> {
+  const statusMeta = (statusEvent.metadata ?? statusEvent.status.message?.metadata) as
+    | Record<string, unknown>
+    | undefined;
+  const prompt = readAuthRequired(statusEvent.status.message?.parts, statusMeta);
+
+  logger.info(
+    { taskId: accumulatedTask.id, tool: prompt.tool, service: prompt.service, hasUrl: !!prompt.authUrl },
+    `Received in-task authorization interrupt`,
+  );
+
+  try {
+    const authCard = chatService.buildInTaskAuthCard(config, prompt, { taskId: accumulatedTask.id });
+
+    await chatService.sendPrivateCardMessage(
+      projectId,
+      spaceId,
+      userId,
+      [authCard],
+      threadId,
+    );
+
+    await inFlightTaskStore.touch(accumulatedTask.id).catch((err) => {
+      logger.error(err, `Failed to update in-flight task for auth interrupt: ${err}`);
+    });
+
+    return true;
+  } catch (cardErr) {
+    // Falling back to the wire text is worse copy than the card, but it is the
+    // only thing left that carries the URL — a silent turn would strand the user.
+    logger.error(cardErr, `Failed to post in-task authorization card, falling back to text: ${cardErr}`);
+    return false;
+  }
+}
+
 async function processHumanInTheLoopEvent(
   logger: Logger,
   chatService: GoogleChatService,
@@ -691,6 +751,10 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
 
     let accumulatedTask: Task | null = null;
     let feedbackRequestData: { sub_agents?: string[] } | null = null;
+    // Set once the in-task authorization card is on screen: the `auth-required`
+    // status text is then the gateway's agent-facing prose, and finalizing must
+    // not post it beside the card that replaced it.
+    let authCardPosted = false;
     try {
       for await (const event of a2aClientService.sendMessageStream(a2aRequest, accessToken)) {
         logger.debug(`Stream event: ${_.get(event, 'kind')}`);
@@ -735,6 +799,21 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
             statusEvent.status.message?.extensions?.includes('urn:nannos:a2a:human-in-the-loop:1.0')
           ) {
             await processHumanInTheLoopEvent(
+              logger,
+              chatService,
+              inFlightTaskStore,
+              projectId,
+              spaceId,
+              threadId,
+              userId,
+              accumulatedTask,
+              statusEvent,
+              config,
+            );
+          }
+
+          if (statusEvent.status.state === 'auth-required' && !authCardPosted) {
+            authCardPosted = await processAuthRequiredEvent(
               logger,
               chatService,
               inFlightTaskStore,
@@ -851,6 +930,7 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
 
     const result = await handleTask({
       task: accumulatedTask,
+      suppressStatusText: authCardPosted,
       chatService,
       messageContext: {
         projectId,
--- a/packages/client-google-chat/src/utils/inTaskAuth.ts
+++ b/packages/client-google-chat/src/utils/inTaskAuth.ts
@@ -1,0 +1,136 @@
+/**
+ * The `urn:nannos:a2a:in-task-auth:1.0` extension, Google Chat side.
+ *
+ * A tool the agent wanted to use needs the END-USER's consent (the MCP
+ * gateway's `need-credentials`, surfaced as A2A's own `auth-required` task
+ * state). The state is protocol; the schema of what it carries is ours, and
+ * this reads it — the same precedence the Embed SDK uses (payload → metadata →
+ * prose), so both clients name the same service and link the same URL.
+ *
+ * The status text is NOT end-user copy: it is the gateway addressing the AGENT
+ * ("You must tell the end-user to…"), and the auth interrupt fires in
+ * middleware, before the model, so no LLM ever rewrites it. Google Chat used to
+ * post it verbatim. The card (googleChatService.buildInTaskAuthCard) writes our
+ * own copy from what this reads instead, and the wire text survives only as the
+ * fallback for a producer that gave us no URL to link.
+ */
+import { Part } from '@a2a-js/sdk';
+
+export const IN_TASK_AUTH_EXTENSION = 'urn:nannos:a2a:in-task-auth:1.0';
+
+/** First http(s) URL in a text blob — the fallback when nothing structured came. */
+const URL_IN_TEXT = /https?:\/\/[^\s<>"')]+/;
+
+/**
+ * Tool names that mean nothing to an end-user: sandbox and orchestration
+ * plumbing. The gateway call that needs authorization often runs INSIDE one of
+ * these (a `need-credentials` from an MCP call made in the sandbox is reported
+ * against `eval`), so naming them misinforms rather than informs. Anything else
+ * is shown verbatim — a mangled tool name is worse than a plain one.
+ * Kept in step with the Embed SDK's own list (auth-required-card.tsx).
+ */
+const OPAQUE_TOOLS = new Set(['eval', 'task', 'client_action', 'python', 'bash', 'shell', 'call_tool']);
+
+export interface AuthPrompt {
+  /** Where the user completes the flow. Without it there is no card, only text. */
+  authUrl?: string;
+  /** The tool call that needed the credential, when it is fit to name. */
+  tool?: string;
+  /** Whose credential this is (e.g. "github"), when the producer named one. */
+  service?: string;
+  /** The gateway's own words — agent-facing; shown only when there is no URL. */
+  message?: string;
+}
+
+/**
+ * The in-task-auth DataPart (`AuthPayload.client_payload()`), when one came.
+ *
+ * Only the first method carrying a URL is used: the model allows several ("try
+ * in order"), but a card offers one way forward, and a method without a URL is
+ * nothing a Slack button can act on.
+ */
+function readAuthDataPart(parts: Part[] | undefined): AuthPrompt | null {
+  for (const part of parts ?? []) {
+    if (part.kind !== 'data') continue;
+    const data = (part as { kind: 'data'; data: Record<string, unknown> }).data;
+    const requirement = data?.auth_requirement;
+    if (typeof requirement !== 'object' || requirement === null) continue;
+    const req = requirement as { service?: unknown; resource?: unknown; auth_methods?: unknown };
+    const methods = Array.isArray(req.auth_methods) ? req.auth_methods : [];
+    const withUrl = methods.find(
+      (m) => typeof (m as { auth_url?: unknown })?.auth_url === 'string' && (m as { auth_url: string }).auth_url
+    ) as { auth_url?: string } | undefined;
+    return {
+      ...(withUrl?.auth_url && { authUrl: withUrl.auth_url }),
+      // `resource` is the specific thing that needed the credential (the tool
+      // call); `service` is who it belongs to. The card names them differently.
+      ...(typeof req.resource === 'string' && req.resource && { tool: req.resource }),
+      ...(typeof req.service === 'string' && req.service && { service: req.service }),
+    };
+  }
+  return null;
+}
+
+/**
+ * The authorize target for an `auth-required` status, in descending order of how
+ * much the producer told us: the structured DataPart of the in-task-auth
+ * extension, then the status metadata (`auth_url` + the `tool` that asked), then
+ * — last resort — the first URL in the message text.
+ *
+ * That last resort scrapes a sentence written for the agent, and it stays only
+ * because a stranded user is worse than an ugly parse: producers that negotiated
+ * the extension never reach it.
+ */
+export function readAuthRequired(parts: Part[] | undefined, metadata?: Record<string, unknown>): AuthPrompt {
+  const text = (parts ?? [])
+    .filter((p) => p.kind === 'text')
+    .map((p) => (p as { kind: 'text'; text: string }).text)
+    .join('\n')
+    .trim();
+  const structured = readAuthDataPart(parts);
+  const fromMeta = metadata?.auth_url ?? metadata?.authUrl ?? metadata?.authorizeUrl;
+  const authUrl =
+    structured?.authUrl ??
+    (typeof fromMeta === 'string' && fromMeta ? fromMeta : (text.match(URL_IN_TEXT)?.[0] ?? undefined));
+  const rawTool = structured?.tool ?? (typeof metadata?.tool === 'string' ? metadata.tool : undefined);
+  const tool = rawTool && !OPAQUE_TOOLS.has(rawTool) ? rawTool : undefined;
+  // The same filter applies to the service: a producer filling it in from the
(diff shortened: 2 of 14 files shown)


packages/client-slack/src/listeners/events/messageHandler.ts
```diff
@@ -946,6 +947,55 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
...
+              ...prompt,
+              taskId: accumulatedTask.id,
+              contextId: accumulatedTask.contextId || '',
+              channelId,
+              threadTs,
+              planMessageTs: streamer.planTs,
+              streamMessageTs: streamer.ts,
+            });
+
+            await client.chat.postMessage({
```
alloy-cicd: 🔴 **Unguarded card post can strand the turn with no URL delivered.**

`interruptWidgetPosted = true` (line 968) and `streamer.pause(...)` (line 973) run *before* this `chat.postMessage`, and nothing catches a throw from it. If the post fails (`rate_limited`, `invalid_blocks`, transient network):

1. the exception lands in the stream-loop catch (lines 1083–1086, log-only),
2. the `if (interruptWidgetPosted) return` gate at lines 1105–1108 then skips `finalizeStreamedTask`, so the status text (which carries the auth URL) is never posted either,
3. `responsePosted = true` (line 996) was never reached, so the in-flight record is kept and the recovery loop re-drives the task ~5–7 min later via `handleTask` — which has no auth-card handling and posts the raw agent-facing gateway prose verbatim (`taskRecovery.ts:76-85` → `taskResponseHandler.ts:301-332`), exactly the text the comment above this branch says should never reach users.

Net: the user stares at a paused "Awaiting your authorization" timeline with no card and no link until the ugly-prose fallback arrives minutes later.

The Google Chat side of this same PR hardens exactly this: `processAuthRequiredEvent` wraps build+send in try/catch and returns false so the caller falls back to posting the wire text ("a silent turn would strand the user"). Suggest mirroring that here: wrap the build+post in try/catch, and only set `interruptWidgetPosted`/`responsePosted` after a successful post, so a failure falls through to `finalizeStreamedTask`'s text path. (The HITL branch above has the same unguarded shape — pre-existing, so fixing it here can be a follow-up, but this branch is new.)

packages/client-google-chat/src/handlers/messageHandler.ts
```diff
@@ -851,6 +930,7 @@ export async function handleIncomingMessage(msg: NormalizedMessage, deps: Handle
 
     const result = await handleTask({
       task: accumulatedTask,
+      suppressStatusText: authCardPosted,
```
alloy-cicd, line 944: 🔴 **Regression: the "🧠 Working..." spinner message is left permanently stale on every successful auth interrupt.**

The orchestrator's auth interrupt fires in middleware before final-answer streaming, so on the normal path `parts.textParts` is empty and intermediate-output artifacts are filtered out (lines 890–907). With `suppressStatusText: authCardPosted` now gating both text branches in `handleTask` (`taskResponseHandler.ts:152` and `:166`), `message` ends up `''`, the `if (message)` guard at `taskResponseHandler.ts:183` means `postOrUpdateMessage` never runs — and that update-in-place is the *only* code path that ever overwrites the spinner (`statusMessageId`, lines 184–191; there is no delete/cleanup anywhere in the package).

Pre-PR, the auth prose overwrote the spinner via the interrupted-state branch. Now the thread shows the private auth card next to a public status message frozen at its last "🧠 Working... [activity]" content, forever — and each auth interrupt leaves another one behind (the resume turn posts a fresh spinner). `handleTask` also still returns the stale spinner's id (`taskResponseHandler.ts:197`), which lines 946–949 record in `contextStore` as the turn's response message.

Suggestion: when `authCardPosted` is true and no message gets posted, finalize the spinner explicitly — e.g. update `statusMessageId` to a short neutral line ("⏸️ Awaiting your authorization"), which would also make the public thread reflect that the task is parked.

packages/client-google-chat/src/utils/taskResponseHandler.ts
```diff
@@ -153,7 +163,7 @@ export async function handleTask(params: HandleTaskResponseParams): Promise<{ me
   }
 
   // Final fallback: extract from status message (e.g. completed with no artifacts)
-  if (!message && task.status?.message?.parts) {
+  if (!message && !suppressStatusText && task.status?.message?.parts) {
```
alloy-cicd, line 170: 🟡 **`!suppressStatusText` here gates more than the auth prose.**

The flag's intent is "the auth card already rendered this status text", but this branch is the generic last-resort fallback for *any* terminal task whose only content is `status.message` (e.g. `failed`/`canceled` with the explanation only there and no artifacts). If a stream ever carries an auth card and then ends in such a state, the explanation is silently dropped and the user is left with a card for a task that already died.

With the current orchestrator this is latent — `auth-required` is only ever emitted as the stream's terminal status (`executor.py:1626-1674`), so the two can't co-occur today. But the gate is broader than its stated purpose; scoping the suppression to the interrupted-state branch at line 152 (the one that would repeat the card's text) would keep this fallback intact for free.

packages/client-google-chat/src/utils/inTaskAuth.ts
```diff
@@ -0,0 +1,136 @@
+/**
```
alloy-cicd, line 1: 🟡 **Third hand-synced copy of the in-task-auth protocol reader.**

The first ~133 lines of this file (`IN_TASK_AUTH_EXTENSION`, `URL_IN_TEXT`, `OPAQUE_TOOLS`, `AuthPrompt`, `readAuthDataPart`, `readAuthRequired`, `authSubject`, `authResumeText`, `authorizationDataPart`) are byte-identical to `packages/client-slack/src/utils/inTaskAuth.ts` — only the header comment differs — and the same reader already lives in `packages/embed-sdk/src/transport/demux.ts:126-185` / `auth-required-card.tsx` (`OPAQUE_TOOLS`, the resume prose, the `{authorization:{decision}}` shape). The doc comments codify the manual sync ("Kept in step with the Embed SDK's own list"), and the ~120-line test suites are duplicated too.

This is pure protocol (payload→metadata→prose precedence, the resume contract the orchestrator middleware routes on), not surface rendering — the next schema tweak (a new `auth_method` shape, a new `OPAQUE_TOOLS` entry, tuned resume prose) needs three synchronized edits with nothing enforcing agreement, and a missed one degrades silently because the fallback paths mask it. I realize per-client util copies are the established idiom here (`fileUtils`, `oauthCallback`, `taskRecovery`, …) and no shared TS package exists between the clients yet — but this file is the strongest candidate yet for starting one: extract the protocol half into a shared workspace module (leaving only the widget/card builders per client) with one test suite. At minimum, the two chat clients could share one copy so the sync burden drops from 3 to 2.

packages/client-slack/src/utils/inTaskAuth.ts
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * The status text is NOT end-user copy: it is the gateway addressing the AGENT
+ * ("You must tell the end-user to…"), and the auth interrupt fires in
+ * middleware, before the model, so no LLM ever rewrites it. Slack used to post
+ * it verbatim. The card below is our own copy instead, and the wire text
+ * survives only as the fallback for a producer that gave us no URL to link.
+ */
+import { Part } from '@a2a-js/sdk';
+
+export const IN_TASK_AUTH_EXTENSION = 'urn:nannos:a2a:in-task-auth:1.0';
```
alloy-cicd: 🔵 **Dead export that looks load-bearing.** `IN_TASK_AUTH_EXTENSION` is exported here (and in the Google Chat twin) but nothing imports it — the URN is hardcoded inside the `X-A2A-Extensions` header string in both `a2aClientService.ts` files (slack `:67`, google-chat `:55`). A future version bump edits the constant, sees green, and never touches the header where the value actually lives. Either build the header from the constant or drop the export.

packages/client-slack/src/utils/inTaskAuth.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/** First http(s) URL in a text blob — the fallback when nothing structured came. */
+const URL_IN_TEXT = /https?:\/\/[^\s<>"')]+/;
```
alloy-cicd: 🔵 **`URL_IN_TEXT` captures trailing sentence punctuation.** The class `[^\s<>"')]+` excludes whitespace and a few closers but not `.` `,` `;`, so prose like "visit https://gateway.example/oauth/begin." yields an Authorize button linking `…/begin.` → 404. Low severity: the in-repo producer's template (`responses.py:66-71`) puts the URL on its own line, and its traffic never reaches this scrape branch anyway (metadata always carries `auth_url`) — but this fallback exists precisely for producers that negotiated nothing, and the tests only cover newline-delimited URLs. Trimming `[.,;:!?]+$` from the match (and one test case) would close it. Same regex in the Google Chat twin (`inTaskAuth.ts:22`).

alloy-cicd (commented):

Reviewed the in-task-auth flow end-to-end on both clients, traced against the producer side (`executor.py`, `responses.py`, `a2a_extensions.py`) and the existing HITL/recovery paths. The wire contracts all line up: the `auth_requirement` payload shape matches `AuthPayload.client_payload()`, the `{"authorization": {"decision"}}` resume DataPart matches `_extract_authorization_decision`, button parameter encoding matches the `app.ts` dispatcher, and the `a2a_extensions.py` docstring change accurately describes the executor's routing.

**Findings (inline comments):**

1. 🔴 **Slack**: the auth-card `chat.postMessage` is unguarded, and `interruptWidgetPosted` is latched before the post — a single failed post strands the turn on a paused "Awaiting your authorization" timeline with no card and no URL, until the recovery loop re-posts the raw gateway prose ~5–7 min later (`messageHandler.ts:985`)
2. 🔴 **Google Chat**: regression — on a successful auth-card turn the public "🧠 Working..." spinner message is never overwritten or cleared; pre-PR the status text replaced it (`messageHandler.ts:933`)
3. 🟡 **Google Chat**: `suppressStatusText` also gates the final no-artifacts fallback, broader than its stated intent (`taskResponseHandler.ts:166`)
4. 🟡 ~133 lines of protocol parsing are byte-identical between the two clients, and a third hand-synced copy lives in `embed-sdk` (`inTaskAuth.ts:1`)
5. 🔵 `IN_TASK_AUTH_EXTENSION` is exported but never used; the URN is hardcoded in both `X-A2A-Extensions` headers (`inTaskAuth.ts:18`)
6. 🔵 `URL_IN_TEXT` captures trailing sentence punctuation in the prose-scrape fallback (`inTaskAuth.ts:21`)

**Out-of-diff observation:** `client-email` was not updated — `packages/client-email/src/services/a2aClientService.ts:106-109` still advertises no `urn:nannos:a2a:in-task-auth:1.0`, while its webhook handler (`a2aWebhookHandler.ts:60`, `:132`) delivers `auth-required` by emailing `payload.message` verbatim — exactly the agent-facing gateway prose this PR's comments describe as unfit to show users. If email is intentionally out of scope, fine, but worth a follow-up ticket.

**Checked and cleared:** the `statusEvent.metadata ?? status.message?.metadata` precedence (the executor always carries `auth_url`/`tool` in the event-level dict on every emission path, and the `??` chain deliberately mirrors `demux.ts`); the one-shot `!authCardPosted` gate (the orchestrator emits `auth-required` only as the stream's terminal status); the Google Chat unknown-action-→-declined coercion (only the two decision buttons carry `cardId: 'in_task_auth_card'`; the Authorize button is pure `openLink`); Google Chat's card-post failure fallback to wire text (correct, and it also restores the spinner overwrite); and the resume plumbing on both clients, which faithfully mirrors the HITL precedent.

packages/client-slack/src/listeners/events/messageHandler.ts
aartaria: Confirmed and fixed in 872a3e22. The flags were the whole bug: setting `interruptWidgetPosted` before the post meant a throw took the status text down with the card, and the kept in-flight record turned that into the gateway prose arriving minutes later — the exact text this branch exists to suppress.

Now the build + pause + post sit in a try/catch and both flags are set only after `postMessage` returns, so a failure falls through to `finalizeStreamedTask` deliberately: its text is ugly, but it carries the URL. The comment says that out loud so the ordering is not "tidied" back later.

Leaving the pre-existing HITL branch alone as you suggested — same shape, but not this PR.

packages/client-google-chat/src/handlers/messageHandler.ts
aartaria: Confirmed and fixed in 872a3e22. Real regression, and worse than cosmetic: the parked card and a spinner that never stops are contradictory instructions to the user.

Fixed one layer up from your suggestion, in the caller rather than in `handleTask`: on a posted card the status line becomes `⏸️ Awaiting your authorization` via `statusMessage.thinking`, which the existing render block at the end of the status-update branch then pushes. Setting it directly in `processAuthRequiredEvent` would not have survived — that same block re-renders the spinner in the same iteration and would have overwritten it.

packages/client-google-chat/src/utils/taskResponseHandler.ts
aartaria: Working as intended — but the code did not say so, and now it does (872a3e22).

Both branches read the same `status.message`; gating only the interrupted-state one would let the generic fallback post the identical prose right after, so the narrower version is not actually narrower in effect. Agreed it is latent either way, since the orchestrator only emits `auth-required` terminally.

I considered adding a `task.status.state === "auth-required"` guard to pin the scope, and dropped it: the caller only ever sets the flag on that state, so it would be defensive code for a state that cannot occur. The docstring now records the reasoning instead.

packages/client-google-chat/src/utils/inTaskAuth.ts
aartaria: Agreed on the substance, deliberately not in this PR — filed as #180.

The blocker is packaging, not taste: `client-slack` and `client-google-chat` are not npm workspaces (the root `package.json` lists only `embed-sdk` and `console-frontend`), so each installs independently and builds its own image. Sharing means either workspace membership plus Dockerfile changes for both clients, or a published internal package — a decision bigger than the feature that surfaced it, and the wrong thing to make silently inside a PR about auth cards.

#180 also notes the embed-sdk copy, which differs only because it reads wire-shaped parts rather than `@a2a-js/sdk` types.

packages/client-slack/src/utils/inTaskAuth.ts
aartaria: Fixed in b2871a8b (pushed before this round landed, so it shows up in the same delta). Removed from both clients rather than wired up: the negotiation header spells every URN out inline, so a single-URN constant would not have been the source of truth it looked like.

packages/client-slack/src/utils/inTaskAuth.ts
aartaria: Confirmed and fixed in 872a3e22. The class now has to end on a character a URL can actually end on: `[^\s<>"')]*[^\s<>"').,;:!?\]}]`. Regression test added in both clients covering a trailing full stop and a trailing comma — the two shapes the gateway prose actually produces.

Worth noting this path only runs for a producer that never negotiated the URN; with the extension active the URL arrives as data. It was still a real broken link on that path.

aartaria:
**Round 1 triage** — all six findings answered: four fixed, one working-as-intended (documented), one deferred with an issue.

| # | Finding | Disposition |
|---|---|---|
| 1 🔴 | Slack: unguarded card post strands the turn | **fixed** `872a3e22` — flags set only after a successful post, failure falls through to finalize |
| 2 🔴 | Google Chat: spinner left stale on every auth interrupt | **fixed** `872a3e22` — status line parked as `⏸️ Awaiting your authorization` |
| 3 🟡 | `suppressStatusText` gates more than the auth prose | **as intended**, now documented — both branches read the same `status.message`; gating one would let the other re-post it |
| 4 🟡 | Third hand-synced copy of the protocol reader | **deferred** → #180. The clients are not npm workspaces, so sharing is a packaging decision bigger than this PR |
| 5 🔵 | `IN_TASK_AUTH_EXTENSION` exported but unused | **fixed** `b2871a8b` — removed from both clients |
| 6 🔵 | `URL_IN_TEXT` captures trailing punctuation | **fixed** `872a3e22` + regression tests in both clients |

The out-of-diff gap you flagged is filed too: **#181** — `client-email` still emails the gateway's agent-facing prose on `auth-required`.

Also in this delta, found while answering a question about the contract rather than by the review: `b2871a8b` corrects the extension docstring, which claimed a reply without a DataPart is "handed to the model node". It is not — a fast-LLM classifier grades it first (`AuthErrorDetectionMiddleware._after_auth_interrupt` → `classify_reply`), and a failed classification never counts as consent. Both clients now note that their fallback prose is *graded*, so it has to stay blunt about the decision.

Verification after the fixes: `tsc --noEmit` clean in both clients; `client-slack` 78/78, `client-google-chat` 30/30.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (a78817fe0a34d017309a610964876542f5ac1612..872a3e225c1726258b120d27ab224285ef89d769)

Every finding from round 1 is resolved, and the delta introduces nothing new. No inline comments this round.

1. 🔴 Slack unguarded auth-card post — **verified-fixed** (872a3e22). Build + pause + post now sit in a try/catch, and `interruptWidgetPosted`/`responsePosted` are set only after `chat.postMessage` returns, so a failed post falls through to `finalizeStreamedTask` and the status text still delivers the URL. The comment records the ordering's intent so it doesn't get "tidied" back.
2. 🔴 Google Chat stale spinner — **verified-fixed** (872a3e22). On a posted card, `statusMessage.thinking` is parked at "⏸️ Awaiting your authorization" (activity cleared) *before* the render block at the end of the status-update branch, which is the right layer: I confirmed the auth-required event matches none of the extension branches that rewrite `statusMessage`, so the parked line survives to the unconditional `updateMessage` and, `auth-required` being terminal, nothing overwrites it afterwards. On a card-post failure the wire-text fallback still overwrites the spinner, so both paths now clean it up.
3. 🟡 `suppressStatusText` gating the generic final fallback — **rejected-by-author, accepted**. The reasoning holds: both branches read the same `status.message`, so gating only the interrupted-state branch would let the fallback post the identical prose right after; the docstring now records this, including why a `state === 'auth-required'` guard was considered and dropped. Settled.
4. 🟡 Three-way duplication of the protocol reader — **still-open by design, deferred to #180**. The packaging blocker (the two clients are not npm workspaces; sharing means workspace membership + Dockerfile changes or a published internal package) is a fair reason not to fold it into this PR. Tracked, so settled for this review.
5. 🔵 Dead `IN_TASK_AUTH_EXTENSION` export — **verified-fixed** (b2871a8b). Removed from both clients; a workspace grep confirms no remaining TS references (the Python and embed-sdk constants are separate, as before).
6. 🔵 `URL_IN_TEXT` trailing punctuation — **verified-fixed** (872a3e22). I exercised the new regex (`[^\s<>"')]*[^\s<>"').,;:!?\]}]`) against trailing `.`, `,`, `;`, `!`, multiple trailing dots, and a bare single-char-path URL — all correct, and backtracking strips runs of punctuation, not just one. Regression tests added in both clients.

**Delta checks beyond the fixes:** the new Slack ordering keeps `streamer.pause` before the post (deliberate — the card must render below the paused timeline), and on failure the paused "Awaiting your authorization" line stays accurate while finalize posts the URL, so no residual issue there. The expanded classifier documentation in `a2a_extensions.py` and both `authResumeText` doc comments matches the middleware (`classify_reply` at `auth_error_middleware.py:429`; a `None` verdict hands the words to the model node, never counting as consent). The new doc claims are accurate.

A clean incremental round — nothing blocking from this reviewer; over to the human reviewer.

PR #183, opened 2026-09-01
fix: two tasks for one sub-agent no longer overwrite each other

aartaria:
## The bug

A sub-agent's memory is its LangGraph checkpoint, and that checkpoint is addressed by conversation and agent name alone:

```python
_effective_thread_id = f"{orchestrator_conversation_id}::{subagent_type}"
# mirroring agents/dynamic_agent.py::get_thread_id -> f"{context_id}::dynamic-{name}"
```

Two `task` calls to the **same** `subagent_type` in one assistant message therefore ran on **one thread**. Their writes interleaved, the last writer won, and the loser's conversation was gone.

Observed: a parallel "who am I on GitHub" + "list campaigns" delegation. The GitHub sub-agent resumed on the campaign-listing sub-agent's state, never called `github_get_me` again, and the turn ended telling the user *"the agent did not return GitHub identity info — it ended up calling the Naonous Riad campaigns tool instead"*. The authorization it had parked on was never answered. Collateral in the same traces: `ValueError('already closed')` and `<error type="ConcurrentEval">` from the two sub-agents sharing one QuickJS context, which is keyed off the same `thread_id`.

## Why not just isolate the threads?

Separate threads make two parked tasks *distinguishable* but still not *addressable*, and two layers downstream need to address them:

| Layer | With two parked tasks for one agent |
|-------|-------------------------------------|
| Sub-agent checkpoint / interrupt id | collide — fixable by isolation |
| Client → orchestrator auth answer | `authorizationDataPart` sends a verdict with **no interrupt id**, so one "Done, continue" answers **both** prompts — including one whose card the user never saw |
| Model → orchestrator continuation | nothing can name which parked task a follow-up continues: `TaskToolSchema` (description, subagent_type) belongs to deepagents |

The last two each need a contract change — in every client, and in a library we do not own. One live task per agent removes the need for either: there is never a second candidate to address.

## The change

**Refuse the second concurrent call.** `surplus_same_agent_call(state, tool_call)` reads *only* the assistant message that issued the call — a snapshot committed before the tools superstep, identical for every parallel branch — and picks the owner **by position, not arrival**. Nothing mutable is observed, so there is no check-then-act race (the "is this agent already running?" phrasing would be racy twice over: nothing is durably written until the superstep commits, and LangGraph deliberately hides one task's writes from its siblings in the same superstep). The resume replay reaches the same verdict.

Checked in **both** `wrap_tool_call` and `awrap_tool_call`, ahead of the registry lookup — the built-in agents the fall-through path reaches share the thread the same way, in fact worse (deepagents hands its sub-agent the parent's own `configurable`).

The refusal is deliberately **not** `status="error"`: it is a policy decision, and an error status would earn it an `[ERROR_TYPE: system_error]` prefix inviting a bug report for something working as designed.

**Tell the model up front.** `_ONE_TASK_PER_AGENT_GUIDANCE` is appended to the task tool description via the existing `_enhance_task_tool_schema`, countering deepagents' own usage note 1 (*"Launch multiple agents concurrently whenever possible"*) for the one case it does not survive. Guidance, not schema: the description is a string we may rewrite freely, while `TaskToolSchema` is deepagents' and `atask`'s signature would reject an extra argument. The refusal is the backstop.

## What still works

- **Different agents in parallel** — untouched, including two that both need authorization.
- **Sequential re-delegation** — a later `task` call with a new `tool_call_id` continues the agent's thread, which is how a sub-agent remembers earlier delegations and how a parked `input-required`/`auth-required` task is resumed.

## QA (traces on dev)

Same parallel prompt that produced the original failure, conversation `627e620f`:

1. **Turn 1** — step 10: `__pregel_push 1` (GitHub) dispatched → `github_get_me` → `need-credentials` → one auth prompt. `__pregel_push 2` (campaigns) refused. Only **one** `task:general-purpose` run in the trace.
2. **Turn 2** — resumed with the approved authorization, correctly matched to the single pending prompt. The GitHub task resumed on **its own** state, `github_get_me` succeeded. The refusal was not re-executed (already answered in turn 1, so the replay left it settled). At step 18 the model **re-delegated the campaigns work sequentially on its own**, exactly as the refusal instructs → 103 campaigns.
3. Final answer carried both halves, with the real identity: *"You are Andrea Artaria — username `aartaria` (GitHub ID: 10273710)"* + campaigns.

## Notes

- 788 tests pass; 18 new in `tests/test_concurrent_same_agent_dispatch.py` (verdict, determinism across a resume replay, the refusal message, short-circuit before dispatch on both paths) plus 2 for the guidance. `test_returns_original_when_registry_empty` asserted `result is original`, which the guidance append intentionally changes — replaced with a test of the new contract.
- `ruff check` clean. `ruff format` deliberately not run: the repo's `lint` recipe is `ruff check .` only, and formatting `dynamic_tool_dispatch.py` would rewrite ~100 unrelated import lines to 88 columns.
- Known cosmetic wart: `notify_user` may still narrate *"Launching two agents in parallel"* before the guard declines one. Not user-visible as an error; the up-front guidance is the real fix for the narration.
- Out of scope: two **concurrent orchestrator turns** on one conversation (two user messages in quick succession, or a scheduled run landing mid-turn) contend for the same sub-agent thread on a different axis, above the tool node. Not addressed here, and I have not verified how the executor serializes them.


packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -661,6 +740,28 @@ def __init__(
...
+            "[DISPATCH] Refusing concurrent task for '%s' (call %s); owned by call %s",
+            subagent_type,
+            str(tool_call.get("id"))[:8],
+            owner_call_id[:8],
+        )
+        return ToolMessage(
+            content=CONCURRENT_SAME_AGENT_MESSAGE.format(agent=subagent_type),
+            name="task",
+            tool_call_id=tool_call["id"],
+        )
```
alloy-cicd, line 824: **The refusal is indistinguishable from a real sub-agent result, and it can be shown to the end user as the answer.**

This `ToolMessage(name="task", ...)` enters the turn looking exactly like a delegation result, and two consumers key off that:

1. `StreamHandler.parse_agent_response` (`app/handlers/stream_handler.py:425-458`): with `include_subagent_output=true`, it walks `reversed(current_turn_messages)` and adopts the **most recent** `task` ToolMessage with non-empty extracted content. `ToolNode` writes outputs in `tool_calls` order, and the refused call is by construction a *later* sibling than the owner — so the refusal always lands after the owner's real result and wins the reverse scan (`json.loads` on it raises, so the full refusal string is adopted). Scenario: model emits `task(github-agent, "who am I")` + `task(github-agent, "list repos")`, the first succeeds, the model finalizes with `include_subagent_output=true, message=""` (which the system prompt tells it to do) — the user's entire visible reply becomes *"This call was NOT executed: sub-agent 'github-agent' is already running another task… Do not retry…"* and the real answer is dropped.
2. `_extract_recently_called_subagents` (`stream_handler.py:60-93`) counts the refusal as a sub-agent that ran, so `is_phantom_subagent_completion()` is satisfied by a turn whose only surviving `task` output is a refusal, and the executor's `_DELEGATION_NUDGE` re-entry is skipped.

Suggestion: tag the refusal (e.g. `additional_kwargs={"concurrent_task_refusal": True}` or a distinct `name`) and skip tagged messages in both consumers. None of the new tests cover the stream-handler side.

packages/orchestrator-agent/docs/subagent-flow.md
```diff
@@ -259,6 +260,60 @@ The `general-purpose` subagent is **NOT** in `subagent_registry`. It's handled s
...
+own. One live task per agent removes the need for either: there is never a second
+candidate to address.
+
+### What still works
+
+- **Different agents in parallel** — untouched, including two that both need authorization.
+- **Sequential re-delegation** — a later `task` call with a new `tool_call_id` continues
+  the agent's thread, which is how a sub-agent remembers earlier delegations and how a
+  parked `input-required`/`auth-required` task is resumed (`a2a_tracking` keeps
+  `context_id` and clears `task_id` only once the task completes).
```
alloy-cicd, line 347: **This claim doesn't hold in the very scenario the guard creates.** `A2ATrackingMiddleware.before_model` reads only `messages[-1]` (`app/middleware/a2a_tracking.py:107-114`). When a same-agent pair is refused, the refusal ToolMessage is always the *later* sibling and therefore lands last — so the owner's ToolMessage, the one carrying `additional_kwargs["a2a_metadata"]` (`task_id`, `context_id`, `is_complete`), is never read that turn. If the owner returned `input-required`/`auth-required` (parked), its ids are dropped and the next delegation to that agent starts without the parked `task_id` — the parked task is never resumed, which is exactly what this bullet promises still works.

Two cheap hardenings:
- make `before_model` scan all trailing ToolMessages of the step (or skip tagged refusals) rather than only `messages[-1]`;
- pin a test asserting `CONCURRENT_SAME_AGENT_MESSAGE` never matches the stale-task heuristic at `a2a_tracking.py:138` (`"task" in content and "does not exist" in content`) — the current wording escapes it only by luck, and a future edit that says "does not exist" would delete the *owner's* live `task_id` and set `is_complete=True`.

packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -484,6 +484,85 @@ def _build_subagent_resume_command(runnable: Any, interrupt_obj: Any, user_decis
...
+            c
+            for c in calls
+            if isinstance(c, dict)
+            and c.get("name") == "task"
+            and (c.get("args") or {}).get("subagent_type") == subagent_type
+        ]
+        owner = siblings[0].get("id") if siblings else None
+        if len(siblings) < 2 or owner == call_id:
+            return None
+        return owner
```
alloy-cicd: **Fail-open when the owning sibling has no `id`.** `owner = siblings[0].get("id")` can be `None` (LangChain's `ToolCall["id"]` is `Optional[str]` — streaming reassembly, replayed/synthetic history such as reconstructed delegation turns). Then `len(siblings) >= 2` and `owner != call_id`, so `return owner` returns `None` — and both call sites read `if owner:` as "may run". Net effect: an id-less *first* sibling disables the guard for the whole group, silently (the log line only fires on refusal), and every sibling dispatches onto one thread — the exact corruption this PR exists to prevent. The new tests cover a missing id only on the argument (`ids=["no-id"]`), never on the sibling in state.

Safer: decide by position (refuse anything that is not `siblings[0]` by identity/index) instead of comparing possibly-`None` ids, or fail closed with a warning when the owner id is missing.

Two adjacent nits: the `if siblings else None` arm is unreachable (the call that matched `call_id` is itself in `siblings`, since both call sites only run this for `name == "task"` and the argument's own `subagent_type`), and the callers' `if owner:` (vs `is not None`) is what turns a falsy owner into "allowed" — worth making that decision in one place.

packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -2396,6 +2533,11 @@ def wrap_tool_call(
         # Special handling for "task" tool (subagent dispatch)
         # Try dynamic registry first, fall back to handler (SubAgentMiddleware) for general-purpose
         if tool_name == "task":
+            # Guard BEFORE the registry lookup: the built-in agents this falls
+            # through to share the thread the same way, so both paths need it.
+            owner = surplus_same_agent_call(request.runtime.state, tool_call)
+            if owner:
+                return self._refuse_concurrent_same_agent(tool_call, owner)
```
alloy-cicd: **A duplicated call with an unknown/typo'd `subagent_type` now gets "already running" instead of "does not exist".** The guard runs before anything checks existence. Previously both calls fell through to deepagents, whose handler answers *"We cannot invoke subagent X because it does not exist, the only allowed types are […]"* — teaching the model the valid names. Now call 1 gets that, while call 2 gets `CONCURRENT_SAME_AGENT_MESSAGE`, which asserts the opposite: *"'X' is NOT missing, broken or unavailable — never tell the user it is"* and "the other call is running and will return its own result" (it isn't and won't). The model holds contradictory evidence about the same name in one turn and is forbidden from reporting unavailability. Cheap fix: only refuse when `subagent_type` resolves (registry or built-in names).

Also, the comment's justification looks inaccurate: the deepagents fallback runs the sub-agent inline against the parent's state — there is no `{conv}::{agent}` checkpoint thread on that path (`_effective_thread_id` exists only on the registry path, and `a2a_tracking.py:19,77` states the built-in general-purpose agent doesn't use A2A tracking). So on the fall-through path the refusal isn't preventing memory corruption; it's serializing stateless one-shots. Worth correcting the comment (and the same claim in `subagent-flow.md`) even if the behavior stays.

packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -484,6 +484,85 @@ def _build_subagent_resume_command(runnable: Any, interrupt_obj: Any, user_decis
...
+    LangGraph replays byte-identical on a resume, so the sibling that won the
+    first attempt wins the replay too — a refusal cannot flip into a second
+    execution half way through a turn, nor an execution into a refusal.
+    """
+    call_id = tool_call.get("id")
+    subagent_type = (tool_call.get("args") or {}).get("subagent_type")
+    if not call_id or not subagent_type:
+        return None
+
+    messages = (state.get("messages") if hasattr(state, "get") else None) or []
```
alloy-cicd: **Silent fail-open, and a duck-type probe the rest of the file doesn't use.** Every other state access in this middleware treats state as a mapping directly (`subagent_state["messages"]`, `state.get(...)`), and the same `request.runtime.state` is already passed into `_adispatch_task_tool` untouched — so `hasattr(state, "get")` advertises a state shape that can't occur today. The risk is tomorrow: if the state schema ever becomes a Pydantic model (or a `ToolNode` list-input path reaches this), this line turns the guard into a no-op with no log line, and every test in `test_concurrent_same_agent_dispatch.py` still passes because they all hand it a dict. Suggest `state.get("messages") or []` plus a `logger.warning` on the unreadable-state branch, so a regression is visible.

packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -834,6 +953,20 @@ def _enhance_system_prompt_agents(
 
         return SystemMessage(content_blocks=new_blocks)
 
+    def _append_one_task_per_agent_guidance(self, task_tool_dict: dict[str, Any]) -> dict[str, Any]:
+        """Append the one-task-per-agent rule to the task tool description, once."""
+        function_dict = task_tool_dict.get("function", {})
+        description = function_dict.get("description", "")
+        if not description or self._ONE_TASK_PER_AGENT_GUIDANCE in description:
```
alloy-cicd: **The one real behavior here is a fail-open: an empty description silently drops the rule** while `surplus_same_agent_call` still refuses at runtime — the model walks into a refusal it was never warned about and is told not to retry. Same for a task tool supplied as a flat dict (no `"function"` key): `function_dict` is `{}`, `description` is `""`, no-op, no warning (the analogous marker-missing fallback at the bottom of `_enhance_task_tool_schema` does warn). Since the guidance is self-contained, appending it even to an empty description costs nothing.

Minor: the substring idempotence check is belt-and-braces today (both call sites rebuild the dict fresh from `original_tools` each bind), but it silently stops matching the *old* text if the guidance wording is ever edited — `test_one_task_per_agent_guidance_is_appended_once` re-enhances with the same constant so it wouldn't notice.

packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -2494,6 +2636,12 @@ async def awrap_tool_call(
             # orchestrator's astream loop via tool call detection, not here.
             # Removed duplicate emission to prevent status history duplicates.
 
+            # Guard BEFORE the registry lookup: the built-in agents this falls
+            # through to share the thread the same way, so both paths need it.
+            owner = surplus_same_agent_call(request.runtime.state, tool_call)
+            if owner:
+                return self._refuse_concurrent_same_agent(tool_call, owner)
```
alloy-cicd: Minor: this verdict-then-refuse pair (plus its comment) is duplicated verbatim in `wrap_tool_call`, and the two pieces are only ever used together. A single helper (`refusal = self._concurrent_same_agent_refusal(request); if refusal: return refusal`) would keep the pairing, the `if owner:` truthiness decision, and any future change (status, metric, tagging per the other comments) in one place — a third `task` entry point that calls `surplus_same_agent_call` but forgets the refusal would dispatch anyway with no test failing.

packages/orchestrator-agent/tests/test_concurrent_same_agent_dispatch.py
```diff
@@ -0,0 +1,285 @@
...
+# ---------------------------------------------------------------------------
+
+
+class TestGuardShortCircuitsDispatch:
+    @pytest.fixture
+    def middleware(self):
+        return DynamicToolDispatchMiddleware()
+
+    @pytest.mark.asyncio
+    async def test_async_refuses_without_dispatching(self, middleware, monkeypatch):
```
alloy-cicd, line 341: Convention note: `packages/orchestrator-agent/AGENTS.md` (§Testing) says *"Use real graph execution for middleware integration tests"*, and the repo follows it elsewhere (`test_hitl_reject_integration.py`, `test_adoption_postgres_fork.py`). These wiring tests build a `MagicMock` request with a hand-built `runtime.state` dict and monkeypatch the dispatch away, so they can't catch the guard sitting at the wrong point in the real middleware chain or `request.runtime.state` drifting from the hand-built shape — which matters more than usual here given the guard's several silent fail-open paths. (The pure-verdict tests above are unit tests and rightly outside this rule.)

alloy-cicd (commented):

The guard itself is well built: the verdict is a pure function of the issuing assistant message (deterministic across resume replays), it short-circuits before `get_stream_writer()`/dispatch on both paths, and the refusal deliberately avoids `status="error"` to keep the bug-report machinery quiet. Tests cover the verdict well. The findings below are mostly about what happens to the refusal ToolMessage *after* it enters the turn — several downstream consumers treat any `task` ToolMessage as a real sub-agent result.

**Inline findings (ranked):**
1. The refusal can be shown to the end user as the sub-agent's answer (`include_subagent_output` adopts the most recent non-empty `task` ToolMessage, and the refusal always lands after the owner's result).
2. The refusal landing last shadows the owner's `a2a_metadata` in `A2ATrackingMiddleware.before_model` (which reads only `messages[-1]`), so a parked `input-required`/`auth-required` owner task is not resumable — contradicting the doc's "What still works" claim.
3. `surplus_same_agent_call` fails open when the owning sibling has no `id`.
4. A duplicated call with an unknown/typo'd `subagent_type` is told the agent is "already running… NOT missing, broken or unavailable" — the guard runs before the registry lookup ever checks existence.
5. Smaller: silent fail-open branches, guidance dropped on empty description, duplicated guard ritual, mocked-graph wiring tests.

**Non-inline notes:**
- **Conventions:** `packages/orchestrator-agent/AGENTS.md` has a `## Critical Design Decisions` section for exactly this class of non-obvious contract, and it is injected into the orchestrator's own system prompt via `PlaybookInjectionMiddleware` — the new one-live-task-per-agent rule is documented in `docs/subagent-flow.md` but not there. Comparable feature PRs (#175, #161, #113) all updated it.
- **Scope observations** (fine to leave out of this PR, but worth a note in the doc): the invariant is enforced only for siblings within one assistant message. Cross-message routes to the same collision remain open: (a) a stall-timeout abort cancels the consumer but a remote A2A sub-agent keeps executing on `{conv}::{agent}` while the model is told the task failed and retries; (b) two orchestrator replicas on one `context_id` each see a lone sibling; (c) the refusal's own advice — "wait for the running task's result and delegate the remainder afterwards" — steers the model into the pre-call pending-interrupt path (`Command(resume=...)` replaces the freshly built `HumanMessage`), so if the owner parked on auth, the follow-up delegation silently resumes the *old* task and drops the new description. A thread-claim at dispatch (the `StreamCoordinator.try_register/release` pattern already in `ringier-a2a-sdk/server/executor.py`) would own the invariant at the resource itself and subsume all of these.
- Because `DynamicToolDispatchMiddleware` is outermost and returns the refusal directly, `RepeatedToolCallMiddleware` never observes refused calls — a model that ignores "Do not retry" and re-issues the same pair each turn is not counted toward `max_repeats`.
- `_iter_subagent_stream_with_stall_timeout`'s docstring still explains `dispatch_id` as existing to "disambiguate concurrent sub-agent dispatches that share the same subagent_thread_id" — now declared impossible by construction; worth reconciling.

aartaria:
## Round 1 addressed — commit 22d16443

Thanks, this was a genuinely load-bearing review. Findings 1–2 were the important ones and I confirmed both by running the pre-fix path: with the refusal untagged, `parse_agent_response` returns *"This call was NOT executed: sub-agent 'github-agent' is already running…"* as the **entire** user-visible reply, and `is_phantom_subagent_completion` returns `False` for a turn whose only `task` output is a refusal. Both are now regression-tested.

**1. Refusal shown as the sub-agent's answer — fixed.** Tagged via `additional_kwargs["concurrent_task_refusal"]`, in a new neutral module `app/middleware/task_refusal.py` (`a2a_tracking` importing `dynamic_tool_dispatch` would have risked a cycle through `app.middleware.__init__`). `parse_agent_response` and `_extract_recently_called_subagents` both skip tagged messages. Test in `test_stream_handler.py` beside its `TestIncludeSubagentOutput` siblings.

**2. `a2a_tracking` shadowing — fixed, and the doc claim with it.** `before_model` now walks back over the trailing ToolMessages to the last non-refusal result instead of taking `messages[-1]`. `test_a2a_tracking_refusal.py` covers the parked owner keeping its `task_id`, the completed owner still clearing it, refusal-only recording nothing, and your suggested pin that the wording never matches the stale-task heuristic.

**3. Fail-open on an id-less owner — fixed.** Ownership is positional now (`next(i for i, c in enumerate(siblings) …)`), never an id comparison, so an id-less owner still owns and later siblings are refused (`test_owner_without_an_id_still_owns`). Your two adjacent nits went with it: the unreachable `if siblings else None` arm is gone, and the truthiness decision now lives once in `_concurrent_same_agent_refusal`.

**4. Unknown/typo'd `subagent_type` — fixed, and you were right about the comment.** The guard is limited to `subagent_registry` members. The fall-through path has no `{conv}::{agent}` thread of its own, so refusing there was preventing nothing; the comment and `subagent-flow.md` both said otherwise and are corrected. A typo'd name now reaches deepagents' "does not exist, the only allowed types are […]" for **both** calls. Costs nothing in practice — the orchestrator registers its own `general-purpose` and the `subagent_type` enum is built from the registry.

**5. Duck-type state probe — fixed.** Direct `state["messages"]` with a `logger.warning` on the unreadable branch, plus `test_unreadable_state_warns` so a future state-schema change surfaces instead of silently disarming the guard.

**6. Guidance dropped on empty description — fixed.** Appended regardless of description content; a tool dict with no `function` entry logs a warning instead of no-opping.

**7. Duplicated guard ritual — fixed.** Both paths call `_concurrent_same_agent_refusal(request)`.

**8. Real graph execution — added.** `tests/test_concurrent_same_agent_graph.py` runs `create_agent` with the production middleware chain and a scripted model emitting the same-agent pair. It also pins the assumption the whole tag design rests on — that the refusal lands *after* the owner's result — against real LangGraph rather than a hand-built dict. Sub-agent internals are still stubbed at `_adispatch_task_tool` (a real dispatch needs a checkpointer, the runnable protocol and an LLM round trip for the human-message build); everything up to and including the guard is production code.

**Conventions:** the rule is now in `AGENTS.md` §Critical Design Decisions, which as you note is injected into the orchestrator's own prompt — so it reinforces `_ONE_TASK_PER_AGENT_GUIDANCE` rather than duplicating it.

**Scope observations — accepted, recorded, not fixed here.** `subagent-flow.md` gains a "What this does NOT cover" section naming all three of your routes (second turn on one conversation, stall-abort leaving a remote executing, and re-delegation into a parked task silently dropping the new description via `Command(resume=…)`), and points at `StreamCoordinator.try_register/release` as the thread-claim shape that would subsume them. I'd rather land the enforced invariant plus an honest map than widen this PR into a resource-claim refactor — happy to be argued out of that.

**`RepeatedToolCallMiddleware`:** true that refused calls never reach it, but I think the loop is still caught: `task` is exempt from `max_tool_repeats` only, not from identical-args `max_repeats`, and a model re-issuing the same pair each turn re-issues the **owner** call too, which does reach the detector. Tell me if you read that differently.

**`dispatch_id` docstring:** reconciled — it no longer claims same-thread concurrency is what it disambiguates, and points at the routes that genuinely still can.

802 tests pass; `ruff check` clean on the touched files.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (320e88a76f475e3b0ca6f2e8ee76650147869e37..22d164438f36bf76432f571d11d324a0452da243)

**Every round-1 finding is resolved and the delta introduces nothing new. No inline comments this round.** I re-verified each fix against the code (not just the commit message) and ran the PR's test files at the new head: 95 passed, including the real-graph integration test.

**Ledger:**
1. Refusal shown to the user as the sub-agent's answer — **verified-fixed.** Tagged via `task_refusal.py` (`concurrent_task_refusal` in `additional_kwargs`); both `stream_handler` consumers skip it (`parse_agent_response` adoption loop and `_extract_recently_called_subagents`, so a refusal-only turn is again a phantom that triggers the delegation nudge). The ordering the tag design rests on (refusal lands after the owner) is pinned against real LangGraph execution in `test_concurrent_same_agent_graph.py`.
2. Owner's `a2a_metadata` shadowed in `A2ATaskTrackingMiddleware.before_model` — **verified-fixed.** Walks back over trailing ToolMessages to the last real result; `test_a2a_tracking_refusal.py` covers the parked-owner, completed-owner, and refusal-only cases, and pins that the refusal wording never trips the stale-task "does not exist" heuristic.
3. Fail-open on an id-less owning sibling — **verified-fixed.** Ownership is positional; an id-less owner still owns (`test_owner_without_an_id_still_owns`), and the dead `if siblings else None` branch is gone.
4. Unknown/typo'd `subagent_type` told "already running" — **verified-fixed.** Guard scoped to `subagent_registry` (`known_agents`); a name outside it falls through to deepagents' "does not exist" answer. The incorrect "built-ins share the thread the same way" justification is corrected in code, `subagent-flow.md`, and `AGENTS.md`.
5. Silent fail-open on unreadable state — **verified-fixed.** Direct `state["messages"]` read with a `logger.warning` on the unreadable branch, covered by a `caplog` test.
6. Guidance dropped on empty description — **verified-fixed.** Appended regardless; a tool dict with no `function` entry warns that the rule is enforced but not announced.
7. Duplicated verdict-then-refuse ritual — **verified-fixed.** Single `_concurrent_same_agent_refusal` helper at both call sites.
8. Mocked-graph wiring tests — **verified-fixed.** `test_concurrent_same_agent_graph.py` runs the production middleware chain via `create_agent` with a scripted model, per AGENTS.md §Testing.

The round-1 summary notes are also addressed: the rule is in `AGENTS.md` §Critical Design Decisions, the out-of-scope collision routes are recorded in the new "What this does NOT cover" doc section, and the `dispatch_id` docstring no longer claims same-message concurrency.

**Follow-up material only (not blocking, pre-existing):**
- `before_model` still records tracking for only *one* ToolMessage per step — with two *different* agents returning results in parallel, the earlier agent's `a2a_metadata` is dropped exactly as `messages[-1]` always dropped it. The new walk-back loop is the natural place to process all trailing real results if that ever bites.
- `RepeatedToolCallMiddleware` still never observes refused calls (round-1 note, unchanged) — a model re-issuing the same refused pair each turn isn't counted toward `max_repeats`.

aartaria:
## Round 2 — agreed, with one correction on the loop-detection note

Thanks. Both follow-ups noted; one of them I think we can close outright.

**`RepeatedToolCallMiddleware` — not actually a gap.** The detector never observes ToolMessages: `aafter_model` reads `last_ai_message.tool_calls` (`loop_detection_middleware.py:259,272`) and hashes tool+args from the model's *emitted* calls, before dispatch runs at all. So a refused sibling is tracked exactly like any other call, and a model re-issuing the same pair each turn accumulates `max_repeats` on **both** calls — my round-1 reply was right for the wrong reason (I argued via the owner call; in fact the refused call is counted too). Nothing to change.

**`before_model` recording only one ToolMessage per step — agreed, real, pre-existing.** Two *different* agents returning in parallel still lose the earlier one's `a2a_metadata`, exactly as `messages[-1]` always did; my walk-back loop just made that shape visible. It is a live bug for parallel delegation to two A2A agents (the earlier agent's `task_id` is dropped, so a parked task there is unresumable) and the walk-back is indeed the natural place to fix it — collect every trailing real result and merge per `subagent_type`.

I'd rather not fold that into this PR: it changes tracking behaviour for a case this PR does not touch, and it wants its own tests for two-different-agents-in-parallel (including one parked). Happy to open it as a follow-up issue, or to do it here if you'd prefer it not to sit as a known bug next to a doc section that now describes the neighbourhood accurately.

aartaria:
## Round-2 follow-up taken in this PR — commit 7dcc84e3

Fixed rather than deferred: it's a live bug and, as you say, the walk-back was the natural place.

`before_model` now collects **every** trailing ToolMessage, skips refusals, and folds each real result into one update keyed by `subagent_type`, so parallel results merge instead of overwriting. Two different agents delegated in parallel both get recorded — previously the earlier one's `task_id`/`context_id` were dropped, leaving a parked task there unresumable.

The per-message logic moved out to `_track_tool_message` / `_subagent_type_for` with behaviour unchanged, plus two corrections that only become visible once more than one message is processed:

- **the tracking map is now copied per agent.** The previous `dict(state.get("a2a_tracking"))` was shallow, so nested records stayed shared with state and were mutated in place — harmless-looking with one writer, wrong as soon as results merge. Pinned by `test_records_in_state_are_not_mutated`.
- **the stale-task branch no longer creates an empty record** for an agent with no `task_id` to clear.

`tests/test_a2a_tracking_refusal.py` → `tests/test_a2a_tracking_step_results.py`, which is what it now covers: parallel different agents (one parked, one completed), a refusal landing after both, no in-place mutation, and the earlier refusal cases.

805 tests pass; `ruff check` clean on the touched files.

That closes both round-2 follow-ups — the loop-detection one by the `aafter_model` reading noted above, this one by fixing it.

packages/orchestrator-agent/app/middleware/a2a_tracking.py
```diff
@@ -108,71 +110,108 @@ def before_model(self, state: A2ATrackingState, runtime: Runtime[ContextT]) -> D
...
+            agent_record = tracking.get(subagent_type) or {}
+            if "task_id" in agent_record:
+                old_task_id = agent_record.pop("task_id")
                 # Also mark as incomplete to prevent re-injection
-                current_tracking[subagent_type]["is_complete"] = True
+                agent_record["is_complete"] = True
                 logger.info(f"[A2A MIDDLEWARE before_model] Cleared stale task_id {old_task_id} for {subagent_type}")
-                return {"a2a_tracking": current_tracking}
+                return True
+            return False
```
alloy-cicd: **Behavior change in this delta: the no-stale-id case now returns instead of falling through, dropping the result's `a2a_metadata`.**

Before this commit, the stale-task branch returned an update *only* when it actually cleared a `task_id`; otherwise control fell through to the `a2a_metadata` processing below. Now `return False` on line 194 ends the message's processing, so the metadata is never read.

That matters because line 180 is a **content heuristic**, not an error channel — a legitimate sub-agent answer can contain the phrase (a GitHub/Jira agent replying *"that task does not exist — did you mean X?"* is a very natural `input-required` park). On the first delegation to that agent there is no `task_id` to clear, so:

```
result text contains "task … does not exist"
+ no prior record (or a record without task_id)
+ a2a_metadata = {task_id: t-1, context_id: ctx-1, state: input-required}
→ nothing recorded
```

I confirmed the divergence by invoking `before_model` directly on that message shape at both commits:

| | no prior record | prior `context_id` only |
|---|---|---|
| `22d16443` | records `task_id` + `context_id` | records `task_id` + `context_id` |
| `7dcc84e3` | `None` | `None` |

The consequence is exactly the failure this PR set out to fix: a parked task whose `task_id` was never recorded cannot be resumed, and the next delegation starts blank. It also loses `context_id`, so the sub-agent conversation restarts.

Suggestion: keep the early return only for the case that actually cleared something, and let everything else continue to the metadata block:

```python
agent_record = tracking.get(subagent_type) or {}
if "task_id" in agent_record:
    old_task_id = agent_record.pop("task_id")
    agent_record["is_complete"] = True
    logger.info(...)
    return True
# else: nothing stale to clear — fall through, the result may still
# carry ids worth recording (the phrase is a heuristic, not an error code)
```

Worth a test either way: no current test exercises "phrase present, nothing to clear, metadata present" (all a2a/tracking suites pass at this head).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (22d164438f36bf76432f571d11d324a0452da243..7dcc84e31a1f931efabee753e7a317b0612a2fa5)

**Ledger of previous findings:** all 8 round-1 findings remain **verified-fixed** (re-confirmed at this head; nothing in the delta touches those fixes). The first of the two round-2 non-blocking follow-ups — `before_model` recording only one ToolMessage per step, losing the earlier agent's `a2a_metadata` when two *different* agents return in parallel — is now **verified-fixed**, and it was a real bug, not just a shape concern: `before_model` collects every trailing ToolMessage, skips refusals, and folds each result into one map keyed by `subagent_type`. The per-agent record copy (`{agent: dict(record) ...}`) is a good catch of its own — the previous shallow `dict()` shared nested records with state and mutated them in place. The second follow-up (`RepeatedToolCallMiddleware` never observing refused calls) is **still-open** and remains non-blocking. Nothing was rejected-by-author.

**One new finding in the delta**, inline below: the stale-task branch changed from *fall through* to *return early*, so a real result whose text happens to contain "task … does not exist" now has its `a2a_metadata` dropped when there is no stale `task_id` to clear. I verified this by running `before_model` at both commits — at `22d16443` it recorded `task_id`/`context_id`, at `7dcc84e3` it returns `None`. Low severity (needs the phrase in a genuine result), but it is the same failure mode this PR exists to fix, reached through a narrow door, and no test covers it.

I ran the PR's test files plus every a2a/tracking-related suite at this head: 82 and 70 passed respectively, so the regression is genuinely untested rather than caught-and-accepted.

aartaria:
## Round 3 — regression confirmed and fixed, commit 5d5a7642

Good catch, and correctly diagnosed: the extraction turned that branch from *fall through* into *return early*, and I didn't notice because the phrase-present-nothing-to-clear path had no test on either side of the change.

Fixed as suggested — the branch returns only when it actually cleared a `task_id` (still breaking the retry loop it exists for), and otherwise falls through to the metadata block exactly as the single-message version did.

Two tests, in a class named for the underlying reason (the phrase is a heuristic, not an error code):
- `test_ids_are_recorded_when_there_is_nothing_stale_to_clear` — your exact shape: phrase in the text, no prior record, `a2a_metadata` carrying an `input-required` park. Fails against 7dcc84e3, passes now.
- `test_a_stale_id_is_still_cleared_and_the_metadata_left_alone` — the loop-breaker still wins when there *is* a stale id, and the cleared `task_id` is not re-set from the same message's metadata. That semantic was previously unpinned too.

807 pass; `ruff check` clean.

On the remaining follow-up (`RepeatedToolCallMiddleware` never observing refused calls): I think it can be closed rather than carried. The detector doesn't read ToolMessages at all — `aafter_model` hashes tool+args off `last_ai_message.tool_calls` (`loop_detection_middleware.py:259,272`), before dispatch runs — so a refused sibling is tracked like any other call and a model re-issuing the same pair accumulates `max_repeats` on both. If you read that differently, say so and I'll dig further; otherwise I'd call the ledger clear.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (7dcc84e31a1f931efabee753e7a317b0612a2fa5..5d5a7642d14a7915bfc13229d87b45d43df22f1b)

**Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.**

**Ledger:**
- **Round 3 finding** (stale-task branch returning early, dropping `a2a_metadata` when a genuine result merely mentions a missing task) — **verified-fixed.** The `return False` is gone; the branch now returns only when it actually cleared a stale `task_id`, and otherwise falls through to the metadata block. I re-ran the exact probe that found the regression: at this head a parked result containing the phrase records `task_id=task-parked-1` and `context_id=ctx-1` both with no prior record and with a prior `context_id` only — byte-identical to the pre-`7dcc84e3` behavior. The loop-breaker still fires when there *is* a stale id (clears `task_id`, sets `is_complete`, keeps `context_id`), and the record held in state is still not mutated in place. Both new tests pin the two halves.
- **Rounds 1 and 2 findings** (all 8, plus the parallel-different-agents tracking follow-up) — **verified-fixed**, unchanged by this delta.
- **Round 2 follow-up:** `RepeatedToolCallMiddleware` never observing refused calls — **still-open**, still non-blocking.
- Nothing was rejected-by-author.

The fix is minimal and I checked the one thing that could have gone wrong in it: the fall-through path reassigns `agent_record` from `tracking.setdefault(...)` at line 220, so the throwaway `{}` built for the heuristic check at line 187 is discarded rather than aliased — no orphan record, no lost write. Full suite at this head: **807 passed**, matching the commit message.

**Follow-up material only (pre-existing, not introduced here):** the `logger.warning` at line 181 announces "Clearing stale task_id from state to prevent retry loop" before it is known whether anything will be cleared, so it now also fires on genuine answers that merely contain the phrase. That was the behavior before this PR too, and the fix deliberately restores it, but the wording will read as a false alarm in logs; moving the warning inside the `if "task_id" in agent_record:` block would make it truthful. Cosmetic, no functional impact.

aartaria:
## Round 4 clean — cosmetic point taken, 6f738855

Took the log-line follow-up rather than carrying it: the warning now fires inside the branch that actually clears, and says what was true ("…for X while holding a task_id. Clearing it…"). You were right that it read as a false alarm on every genuine answer containing the phrase — and since the round-3 fix makes those results *common* on that code path rather than rare, leaving it would have made the log actively misleading about the change we just made. No functional change; 807 still pass.

Not re-dispatching a round 5 for a log line unless you'd like one.

## Ledger as I read it

- Rounds 1–3: all findings **verified-fixed** by you at each head.
- Round 4: nothing new; cosmetic point now addressed.
- **`RepeatedToolCallMiddleware`** — I'd close this rather than carry it, and I'll state the evidence once more rather than re-argue: `aafter_model` never looks at ToolMessages. It hashes tool name + args off `last_ai_message.tool_calls` (`loop_detection_middleware.py:259,272`), which runs *before* dispatch, so a refused sibling is tracked exactly like an executed one and a model re-issuing the same pair accumulates `max_repeats` on both calls. `task` is exempt from `max_tool_repeats` only. If that reading is wrong I'd genuinely like to know where, but absent a specific rebuttal I don't think there's a gap here to track.

Everything else from the original scope stands as reviewed: guard, tag, consumers, docs, and the parallel-results tracking fix that came out of your round-2 note.

PR #184, opened 2026-09-01
fix(agent-common): don't orphan a tool result when force-stopping a loop

### LINKED ISSUE #182: Summarization can orphan tool results → Bedrock 400 'toolResult blocks exceed toolUse blocks', bricking the conversation
## Symptom

A long-running agent conversation dies with a hard 400 from the model gateway (Bedrock via LiteLLM):

```
BedrockException - {"message":"The number of toolResult blocks at messages.64.content
exceeds the number of toolUse blocks of previous turn."}
```

Raised from `langgraph` task `model` during a normal `astream` turn; there is no fallback model group, so the run fails outright. Once the conversation history is in this state the thread is bricked — every subsequent turn replays the same invalid history and fails the same way.

## Analysis

The Anthropic/Bedrock message format requires every `toolResult` block in a user turn to match a `toolUse` block in the immediately preceding assistant turn. The error means the effective message list handed to the model contained more `ToolMessage`s than the preceding `AIMessage` had `tool_calls` — i.e. orphaned tool results.

The prime suspect is the deepagents summarization middleware (`deepagents/middleware/summarization.py`, pinned `>=0.7.11,<0.8.0`), which rewrites the request messages via `request.override(messages=truncated_messages)`:

- `_apply_event_to_messages` reconstructs the effective history as `[summary_message] + messages[cutoff_index:]` using a **stored** `cutoff_index` from a previous summarization event, with no validation that the slice starts at a safe boundary. If the boundary splits a (parallel) tool-call turn — the `AIMessage` with the `tool_use` blocks falls before the cutoff while its `ToolMessage`s fall after — the model sees tool results with no matching tool use.
- The stored index is also fragile against any later mutation of the state message list (messages removed/merged by other middleware, `RemoveMessage`, resumed checkpoints): the index silently points at a different boundary than the one it was computed for.
- The same class of risk exists at `_determine_cutoff_index` time if the retention window boundary is not snapped to a turn boundary that keeps AIMessage+ToolMessage groups intact.

None of our own middleware in the stack rewrites history (our truncation-recovery middleware only appends a nudge `HumanMessage`), so the corruption almost certainly enters at the summarization layer.

## Suggested fix

Two layers, independently useful:

1. **Boundary snapping**: when applying/choosing a cutoff, snap it so that a `ToolMessage` is never separated from its parent `AIMessage` (move the cutoff back to before the `AIMessage`, or forward past the last of its `ToolMessage`s). If upstream deepagents already fixed this in a newer release, bump the pin instead.
2. **Last-line guard in `agent-common`**: a cheap validation pass in a late `wrap_model_call` middleware that walks the final message list and drops/stubs orphaned `ToolMessage`s (and, symmetrically, synthesizes error `ToolMessage`s for dangling `tool_calls`) before the request leaves the process. This makes the failure self-healing instead of bricking the thread, whatever middleware introduced the mismatch.

A regression test should cover: parallel tool calls straddling the summarization cutoff; re-applying a stored summarization event after the message list has shifted.

aartaria:
closes #182

## What was actually wrong

A conversation died with a hard 400 from Bedrock — `The number of toolResult blocks at
messages.N.content exceeds the number of toolUse blocks of previous turn` — and stayed
dead: the illegal messages sit in the persisted checkpoint, so every later turn replays
them and fails identically.

Reconstructing the failing thread's message list refutes the issue's summarization
hypothesis: that thread had never been summarized (full unsummarized history, below the
trigger, no summary message, no summarization event). The corruption is written by
`RepeatedToolCallMiddleware`, via two independent defects.

### 1. The response tool was treated as a loop candidate

`FinalResponseSchema` is how a turn *delivers its answer* — once per turn — and the
sub-agent pass-through variant carries byte-identical arguments every time:

```json
{"include_subagent_output": true, "message": "", "task_state": "completed"}
```

`tool_call_history` is cumulative for the whole thread, never reset at a turn boundary, so
with `max_repeats=5` / `max_tool_repeats=10` / `force_stop_after=3` a block is arithmetic,
not bad luck. The failing thread walked exactly that ladder:

| call | verdict |
|---|---|
| 6th identical-args | `same_args` block — *"called 5 times with identical arguments"* |
| 7th identical | block — *"called 6 times"* |
| 12th call to the tool | `same_tool` block — *"called 12 times (with 5 different argument sets)"* |
| 13th call | 13−10 = 3 ≥ `force_stop_after` → **force-stop** |

`dispatch_tools` already exempts `task`/`eval` from `max_tool_repeats`, but nothing
exempted the terminal response tools from anything. They are now excluded from detection
altogether (`RESPONSE_TOOLS`, opt-out via `response_tools=set()`): there is no loop to
break there, only an answer to deliver.

### 2. Force-stop orphaned the tool result

Force-stop rebuilt the `AIMessage` without its `tool_calls` (same `id`, upserted by the
message reducer) so routing would fall through to `END`. But the model node's
structured-output branch returns `{"messages": [output, ToolMessage(...)]}` — the
`AIMessage` **and** its result in one update — so the result is always already there. The
rebuilt message removed the call it answered, leaving:

| # | type | `tool_calls` | `tool_call_id` |
|---|------|--------------|----------------|
| n | `ai` | `[]` (`finish_reason == "tool_calls"`) | — |
| n+1 | `tool` | — | `tooluse_…` (matches nothing) |

Force-stop now ends the run by jumping to `END` instead of rewriting the `AIMessage`, so
stopping a tool cannot orphan its output. This also covers the HITL path, which answers a
rejected call while deliberately keeping the call on the `AIMessage`.

Ending the run is not enough on its own, though: a call left unanswered in the checkpoint
is the mirror image of the orphan — the next turn sends a `tool_use` with no `tool_result`,
which providers reject just as hard (the repo already treats that shape as invalid, see
`_seal_dangling_tool_calls`, which only ever runs on the adoption path). So force-stop
first *seals* every call on the turn that has no result yet — blocked ones with their
`BLOCKED` message, any other unanswered call (an exempt response tool, or one outside a
`tool_name`-filtered instance, neither of which counts toward `all_tracked_blocked`) with a
"not executed" result — and then jumps to `END`. Only ever appending results means no
existing message is rewritten, so neither failure shape can be reintroduced.

Additionally: a blocked call that already has a result no longer receives a second,
`BLOCKED` result — two `toolResult` blocks for one `toolUse` is the same illegal shape.

## Scope

Deliberately write-side only. An earlier revision of this PR also added a
request-boundary middleware that repaired already-corrupted message lists; it was dropped
as not worth the complexity. **Consequence: threads already carrying an orphan in their
checkpoint stay broken** and need to be started fresh — this PR stops new ones being
created.

## Verification

- `agent-common` 961, `orchestrator-agent` 769, `agent-runner` 35, `ringier-a2a-sdk` 390 — all passing
- New regression tests: the exemption (both response tools, plus the opt-out), force-stop
  against an already-answered call, sealing of every unanswered call on a mixed
  tracked/untracked turn, no duplicate `BLOCKED` result, and a check that the hook actually
  declares `can_jump_to=["end"]` (the graph ignores `jump_to` otherwise).

### Graph-level tests

`tests/test_loop_detection_graph_e2e.py` runs a real agent graph with a real checkpointer
and asserts the provider's own contract against both the persisted history and the message
list handed to the model on the *resumed* turn — the turn that actually died in production.

`assert_tool_pairing_valid` is the offline proxy for that contract, enforced per turn as the
provider does: every result answers a call on the immediately preceding `AIMessage`, every
call has exactly one result, any other message closes the group, and ids are scoped per turn
rather than globally.

Seven tests:

- a plain tool loop force-stops, ends the run, and leaves a valid checkpoint;
- the conversation resumes cleanly afterwards;
- the production shape — repeated `ToolStrategy` turns (where the model node emits the
  `AIMessage` and its result in one update) walking the same block-then-force-stop ladder as
  the failing thread, in three turns instead of thirteen;
- the terminal response tool is never blocked or force-stopped;
- **the repeat counter survives the stop** — asserted behaviourally: on the turn after a
  stop the identical call is refused and the tool does not execute again, so a lost
  `tool_call_history` (which would silently re-run the loop one turn later) fails the test;
- **the "do NOT retry" guidance is replayed to the model** on the next turn, attached to the
  refused call — it previously got no result at all;
- the invariant helper itself rejects an orphan, a dangling call, and a result separated
  from its call.

**Verified non-vacuous against every revision this PR passed through**, by swapping in the
older middleware and re-running: 4 of 7 fail against pre-fix (`abd4b7bc`), and 4 of 7
against the intermediate `jump_to: "end"`-without-sealing revision (`bb63fcb6`). The two
sets differ, so the suite pins both failure directions — the orphaned result and the
dangling call — not just the original bug.

## Checklist

### DIFF
--- a/packages/agent-common/agent_common/middleware/loop_detection_middleware.py
+++ b/packages/agent-common/agent_common/middleware/loop_detection_middleware.py
@@ -36,13 +36,23 @@ import json
 import logging
 from typing import Annotated, Any
 
-from langchain.agents.middleware.types import AgentMiddleware, AgentState, PrivateStateAttr
+from langchain.agents.middleware.types import AgentMiddleware, AgentState, PrivateStateAttr, hook_config
 from langchain_core.messages import AIMessage, ToolCall, ToolMessage
 from langgraph.runtime import Runtime
 from langgraph.typing import ContextT
 from typing_extensions import NotRequired
 
 logger = logging.getLogger(__name__)
+
+#: Terminal response tools. These are how a turn *delivers its answer*, not tools the
+#: model chooses between — every turn ends in exactly one of them, and the sub-agent
+#: pass-through variant carries byte-identical arguments every time
+#: (``{"task_state": "completed", "message": "", "include_subagent_output": true}``).
+#: ``tool_call_history`` is cumulative for the whole thread, so tracking them makes a
+#: block inevitable on any sufficiently long conversation: ~6 pass-through turns trip
+#: ``max_repeats``, ~11 turns trip ``max_tool_repeats``. Blocking the response tool is
+#: never right — there is no loop to break, only an answer to deliver.
+RESPONSE_TOOLS: frozenset[str] = frozenset({"FinalResponseSchema", "SubAgentResponseSchema"})
 
 
 class LoopDetectionState(AgentState):
@@ -84,6 +94,8 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
     - dispatch_tools: Tools exempt from max_tool_repeats (e.g. dispatch/meta-tools
       like 'task' that delegate to sub-agents). These are still subject to
       max_repeats (same args) detection.
+    - response_tools: Terminal response tools, exempt from detection altogether
+      (see ``RESPONSE_TOOLS``).
     """
 
     state_schema = LoopDetectionState  # type: ignore[assignment]
@@ -97,6 +109,7 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
         window_size: int = 10,
         force_stop_after: int = 3,
         dispatch_tools: set[str] | None = {"task"},
+        response_tools: frozenset[str] | set[str] | None = None,
     ):
         """Initialize loop detection middleware.
 
@@ -106,13 +119,15 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
             max_tool_repeats: Same tool threshold (default: 5)
             window_size: Sliding window size (default: 10)
             force_stop_after: After blocking a tool this many consecutive times,
-                strip tool_calls from the AIMessage to force the graph to END.
+                end the run (jump to END) instead of letting the model retry.
                 This prevents infinite block-retry loops when the model ignores
                 error messages. (default: 3)
             dispatch_tools: Set of tool names exempt from max_tool_repeats.
                 Dispatch/meta-tools (e.g. 'task') delegate to sub-agents and are
                 expected to be called many times with different arguments. They are
                 still subject to max_repeats (same args) detection.
+            response_tools: Terminal response tools, excluded from tracking entirely
+                (see ``RESPONSE_TOOLS``). Pass an empty set to disable the exemption.
         """
         super().__init__()
         self.tool_name = tool_name
@@ -121,11 +136,12 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
         self.window_size = window_size
         self.force_stop_after = force_stop_after
         self.dispatch_tools = dispatch_tools or set()
+        self.response_tools = frozenset(RESPONSE_TOOLS if response_tools is None else response_tools)
         logger.info(
             f"RepeatedToolCallMiddleware initialized: tool_name={tool_name}, "
             f"max_repeats={max_repeats}, max_tool_repeats={max_tool_repeats}, "
             f"window_size={window_size}, force_stop_after={force_stop_after}, "
-            f"dispatch_tools={self.dispatch_tools}"
+            f"dispatch_tools={self.dispatch_tools}, response_tools={sorted(self.response_tools)}"
         )
 
     @property
@@ -161,6 +177,9 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
         Returns:
             True if this middleware should track this tool call.
         """
+        # Terminal response tools are never loop candidates — see ``RESPONSE_TOOLS``.
+        if tool_call["name"] in self.response_tools:
+            return False
         return self.tool_name is None or tool_call["name"] == self.tool_name
 
     def _build_error_message(self, info: dict[str, Any]) -> str:
@@ -226,6 +245,7 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
 
         return False, 0, ""
 
+    @hook_config(can_jump_to=["end"])
     async def aafter_model(
         self,
         state: LoopDetectionState,
@@ -237,6 +257,8 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
         - Only blocks looping tool calls with error ToolMessages
         - Lets non-looping calls execute normally
         - Updates history state for tracking
+        - Never rewrites an existing message: a tool call that already has a result is
+          left alone, so a stop can never orphan a tool result
 
         Args:
             state: Current agent state
@@ -251,13 +273,26 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
             return None
 
         last_ai_message = None
-        for message in reversed(messages):
-            if isinstance(message, AIMessage):
-                last_ai_message = message
+        last_ai_index = -1
+        for index in range(len(messages) - 1, -1, -1):
+            if isinstance(messages[index], AIMessage):
+                last_ai_message = messages[index]
+                last_ai_index = index
                 break
 
         if not last_ai_message or not last_ai_message.tool_calls:
             return None
+
+        # Tool calls that already have a result at this point. ``after_model`` hooks run
+        # innermost-first, so a result can already exist before this one runs — the HITL
+        # middleware answers a rejected call while keeping the call itself, and the model
+        # node emits an AIMessage together with its ToolMessage for structured-output
+        # responses. Answering such a call a second time would leave the turn with two
+        # toolResult blocks for one toolUse, which Anthropic models reject with a hard
+        # 400 on *every* subsequent turn (the conversation is bricked, not just the turn).
+        already_answered = {
+            message.tool_call_id for message in messages[last_ai_index + 1 :] if isinstance(message, ToolMessage)
+        }
 
         # Get current history (per-tool tracking).
         # Deep-copy: shallow .copy() shares inner lists, and .append() below
@@ -330,9 +365,9 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
             logger.debug(f"[LOOP DETECTION] Tracked {len(last_ai_message.tool_calls)} tool call(s)")
             return {"tool_call_history": history}
 
-        # Check if we should force-stop by stripping tool_calls from the AIMessage.
-        # This prevents the infinite block-retry loop where the model ignores error
-        # messages and keeps retrying the same tool call.
+        # Check if we should force-stop and end the run. This prevents the infinite
+        # block-retry loop where the model ignores error messages and keeps retrying
+        # the same tool call.
         should_force_stop = any(
             info["repeat_count"] - self.max_repeats >= self.force_stop_after
             if info["loop_type"] == "same_args"
@@ -345,28 +380,31 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
         all_tracked_blocked = len(blocked_calls) == len(tracked_calls)
 
         if should_force_stop and all_tracked_blocked:
-            # Strip tool_calls from AIMessage by returning a modified copy with same ID.
-            # LangGraph's message reducer upserts by ID, so this replaces the original.
-            # With no tool_calls, the graph routes to END instead of tools node.
+            # End the run by jumping to END, leaving the message history untouched.
+            #
+            # This used to rebuild the AIMessage without its tool_calls (same id, upserted
+            # by the message reducer) so that routing would fall through to END. That
+            # rewrite is what orphans a tool result: any result already emitted for those
+            # calls — by the HITL middleware, or by the model node's own structured-output
+            # branch, which returns the AIMessage and its ToolMessage in one update — is
+            # left answering a tool call that no longer exists. The gateway then rejects
+            # the whole conversation on every later turn. Stopping a tool must never
+            # orphan its output, so express the stop as what it actually is (end the run)
+            # instead of as message surgery.
             blocked_names = [info["tool_name"] for info in blocked_calls]
             logger.warning(
                 f"[LOOP DETECTION] Force-stopping: {blocked_names} blocked "
-                f"{self.force_stop_after}+ consecutive times. Stripping tool_calls to end loop."
-            )
-
-            # Preserve any text content the model generated alongside the tool calls
-            modified_ai = AIMessage(
-                content=last_ai_message.content or "",
-                id=last_ai_message.id,
-                response_metadata=last_ai_message.response_metadata,
+                f"{self.force_stop_after}+ consecutive times. Ending the run."
             )
             return {
                 "tool_call_history": history,
-                "messages": [modified_ai],
+                "jump_to": "end",
             }
 
         # Build error ToolMessages for blocked calls (only blocked ones!)
-        # Frame as permanent failure (not rate-limiting) so models don't stubbornly retry
+        # Frame as permanent failure (not rate-limiting) so models don't stubbornly retry.
+        # Calls that already carry a result are skipped: a second result for the same
+        # ``tool_call_id`` is one toolResult block too many for the preceding turn.
         error_messages = [
             ToolMessage(
                 content=self._build_error_message(info),
@@ -375,7 +413,15 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
                 status="error",
             )
             for info in blocked_calls
+            if info["tool_call"]["id"] not in already_answered
         ]
+
+        if not error_messages:
+            logger.info(
+                "[LOOP DETECTION] All blocked call(s) already have results; "
+                "recording history only to avoid duplicate tool results"
+            )
+            return {"tool_call_history": history}
 
         logger.info(
             f"[LOOP DETECTION] Blocked {len(blocked_calls)} looping tool call(s): "
--- a/packages/agent-common/agent_common/middleware/loop_detection_middleware.py
+++ b/packages/agent-common/agent_common/middleware/loop_detection_middleware.py
@@ -54,6 +54,12 @@ logger = logging.getLogger(__name__)
 #: never right — there is no loop to break, only an answer to deliver.
 RESPONSE_TOOLS: frozenset[str] = frozenset({"FinalResponseSchema", "SubAgentResponseSchema"})
 
+#: Result given to a call that never ran because the run was force-stopped over a
+#: *different* looping call on the same turn.
+_STOPPED_BEFORE_EXECUTION = (
+    "BLOCKED: the run was stopped because another tool call on this turn was looping. This call was not executed."
+)
+
 
 class LoopDetectionState(AgentState):
     """Extended agent state with tool call tracking for loop detection.
@@ -396,10 +402,32 @@ class RepeatedToolCallMiddleware(AgentMiddleware[LoopDetectionState, ContextT]):
                 f"[LOOP DETECTION] Force-stopping: {blocked_names} blocked "
                 f"{self.force_stop_after}+ consecutive times. Ending the run."
             )
-            return {
-                "tool_call_history": history,
-                "jump_to": "end",
-            }
+
+            # Answer every call that has no result yet before ending. A tool call left
+            # unanswered in the checkpoint is the mirror image of the orphaned result —
+            # the next turn sends a tool_use with no tool_result, which providers reject
+            # just as hard (cf. ``_seal_dangling_tool_calls`` on the adoption path).
+            # Appending results can never remove a call, so this direction is always safe.
+            blocked_by_id = {info["tool_call"]["id"]: info for info in blocked_calls}
+            stop_messages = [
+                ToolMessage(
+                    content=(
+                        self._build_error_message(blocked_by_id[tool_call["id"]])
+                        if tool_call["id"] in blocked_by_id
+                        else _STOPPED_BEFORE_EXECUTION
+                    ),
+                    tool_call_id=tool_call["id"],
+                    name=tool_call["name"],
+                    status="error",
+                )
+                for tool_call in last_ai_message.tool_calls
+                if tool_call.get("id") and tool_call["id"] not in already_answered
+            ]
+
+            update: dict[str, Any] = {"tool_call_history": history, "jump_to": "end"}
+            if stop_messages:
+                update["messages"] = stop_messages
+            return update
 
         # Build error ToolMessages for blocked calls (only blocked ones!)
         # Frame as permanent failure (not rate-limiting) so models don't stubbornly retry.
--- a/packages/agent-common/tests/test_loop_detection_graph_e2e.py
+++ b/packages/agent-common/tests/test_loop_detection_graph_e2e.py
@@ -1,0 +1,329 @@
+"""Graph-level tests for loop force-stop (issue #182).
+
+The unit tests call ``aafter_model`` directly and take on faith that the state update it
+returns does the right thing once LangGraph applies it. These run a *real* agent graph with
+a real checkpointer instead, because the bug that bricked conversations was never visible
+in a single hook call: it was written into the checkpoint, and only the *next* turn died.
+
+So the assertions here are about what ends up persisted, and about what the model is handed
+when the conversation is resumed — which is exactly what the provider validates:
+
+* every ``ToolMessage`` answers a ``tool_call`` on the immediately preceding ``AIMessage``
+  (no orphaned result → ``toolResult blocks … exceeds … toolUse blocks``), and
+* every ``tool_call`` has exactly one result (no dangling call → ``tool_use`` with no
+  ``tool_result``).
+"""
+
+from __future__ import annotations
+
+from typing import Any, Optional
+
+import pytest
+from langchain.agents.factory import create_agent
+from langchain.agents.structured_output import ToolStrategy
+from langchain_core.callbacks import CallbackManagerForLLMRun
+from langchain_core.language_models.chat_models import BaseChatModel
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, ToolMessage
+from langchain_core.outputs import ChatGeneration, ChatResult
+from langchain_core.tools import StructuredTool
+from langgraph.checkpoint.memory import InMemorySaver
+from pydantic import BaseModel, Field
+
+from agent_common.middleware.loop_detection_middleware import RepeatedToolCallMiddleware
+
+
+def assert_tool_pairing_valid(messages: list[BaseMessage]) -> None:
+    """Assert the message list satisfies the provider's tool-call/result contract.
+
+    This is the offline proxy for what Bedrock enforces; a list that fails here is one the
+    gateway rejects with a hard 400, permanently, because it lives in the checkpoint.
+    """
+    answered: dict[str, int] = {}
+    open_calls: dict[str, str] = {}
+
+    for position, message in enumerate(messages):
+        if isinstance(message, ToolMessage):
+            assert message.tool_call_id in open_calls, (
+                f"orphaned tool result at index {position}: tool_call_id "
+                f"{message.tool_call_id!r} matches no tool call on the preceding AIMessage"
+            )
+            answered[message.tool_call_id] = answered.get(message.tool_call_id, 0) + 1
+            assert answered[message.tool_call_id] == 1, (
+                f"duplicate tool result at index {position} for {message.tool_call_id!r}"
+            )
+        elif isinstance(message, AIMessage):
+            unanswered = [cid for cid in open_calls if cid not in answered]
+            assert not unanswered, (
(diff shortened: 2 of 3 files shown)


packages/agent-common/agent_common/middleware/loop_detection_middleware.py
```diff
@@ -344,7 +357,12 @@ async def aafter_model(
         tracked_calls = [tc for tc in last_ai_message.tool_calls if self._matches_tool_filter(tc)]
         all_tracked_blocked = len(blocked_calls) == len(tracked_calls)
 
-        if should_force_stop and all_tracked_blocked:
+        # Never strip a tool call whose result is already in the message list (see
+        # ``already_answered``): the result would outlive its call and orphan itself.
+        # There is nothing to force-stop in that case anyway — the call has run.
+        force_stop_is_safe = not any(tc.get("id") in already_answered for tc in last_ai_message.tool_calls)
```
alloy-cicd: **Force-stop is now disarmed all-or-nothing, not just for the answered call.**

`force_stop_is_safe` scans *every* tool call on the last `AIMessage`, so one already-answered call suppresses the force-stop for all of them. `after_model` hooks run innermost-first, i.e. `ConditionalHumanInTheLoopMiddleware` (and the auth/`client_action` paths) run *before* this one and inject a `ToolMessage` while keeping the tool call on the `AIMessage` (`conditional_hitl.py:268-270`, langchain `_process_decision` returns `(tool_call, tool_message)` for `reject`/`respond`).

Concrete scenario: the model repeatedly calls a guarded tool and the user rejects it each round. Every round HITL injects a rejection result for that call, so `already_answered` is non-empty on every superstep, `force_stop_is_safe` is always `False`, and the escape hatch never fires — the block/retry loop this middleware exists to terminate now runs forever (routing falls through to step 6 of `model_to_tools`, back to the model). Secondary effect: blocked entries deliberately skip the sliding-window trim so `tool_call_history[tool]` grows without bound in the checkpoint.

A narrower guard would keep both properties: either strip only the *unanswered* tool calls, or force-stop via `{"jump_to": "end"}` instead of rewriting the `AIMessage` — that ends the loop without ever orphaning a result.

packages/agent-common/agent_common/middleware/tool_result_integrity.py
```diff
@@ -0,0 +1,133 @@
...
+
+import logging
+from collections.abc import Awaitable, Callable
+
+from langchain.agents.middleware import AgentMiddleware, ModelRequest, ModelResponse
+from langchain_core.messages import AIMessage, BaseMessage, ToolMessage
+
+logger = logging.getLogger(__name__)
+
+_MISSING_RESULT = (
```
alloy-cicd: **The synthesized result asserts something the guard cannot know.**

`"Nothing was executed. Assume no effect and continue."` is a factual claim about a tool that, in the most likely dangling-call scenario, *did* run. The docstring names "a summarization cutoff" as a source of dangling calls — a cutoff that keeps the `AIMessage` but drops its `ToolMessage` means the tool executed and its result was discarded. Same for a run cancelled/killed between tool execution and the checkpoint write.

Failure scenario: the model calls `gmail_send` (or any side-effecting tool), it succeeds, summarization drops the result. On the next turn this middleware tells the model the send never happened and to assume no effect — the model resends the mail. Duplicate side effects are worse than the model being told the outcome is unknown.

Suggested wording: state that the result is missing and the outcome is *unknown*, and tell the model to verify before retrying rather than to assume no effect.

packages/agent-common/agent_common/middleware/tool_result_integrity.py
```diff
@@ -0,0 +1,133 @@
...
+                            tool_call_id=call_id,
+                            name=tool_name,
+                            status="error",
+                        )
+                    )
+            continue
+
+        if isinstance(message, ToolMessage):
+            # A result with no tool-calling assistant turn in front of it at all.
+            dropped += 1
```
alloy-cicd: **The repair drops the orphan but leaves the empty assistant turn that came with it.**

In the exact production shape this PR describes, the checkpoint holds `AIMessage(content=last_ai_message.content or "", tool_calls=[])` (`loop_detection_middleware.py:376-379`) followed by the orphaned structured-output `ToolMessage`. With `ToolStrategy` + `FinalResponseTextStripMiddleware` that content is almost always `""`.

This branch drops the orphan, which clears the `toolResult` 400 — but the request now carries an assistant turn with empty content in the middle of the history, which Bedrock Converse rejects in its own right (`The text field in the ContentBlock object at messages.N.content.0 is blank`). If that fires, the already-bricked conversations this middleware is meant to rescue stay bricked, just with a different error.

Worth either dropping content-less, tool-call-less `AIMessage`s in the same pass, or confirming the gateway filters them before relying on the self-healing claim.

packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -729,6 +730,12 @@ async def _forward_to_active_subagents(context_id: str, messages: list) -> None:
             # the checkpointer see the cleaned message (prevents the model from
             # imitating its own text+tool-call pattern on later turns).
             FinalResponseTextStripMiddleware(),
+            # Truly last: repair an illegal tool-call/tool-result pairing (orphaned
```
alloy-cicd: **"Truly last" / "deliberately innermost" is not what the assembled stack does.**

This list is passed to `create_deep_agent`, and `deepagents._apply_custom_middleware` inserts user middleware *after the last core entry, ahead of the tail* — the tail being `extra_middleware`, `_ToolExclusionMiddleware`, `AnthropicPromptCachingMiddleware`, `BedrockPromptCachingMiddleware`, `MemoryMiddleware`, `HumanInTheLoopMiddleware`. Those all wrap the model call *inside* `ToolResultIntegrityMiddleware`.

Nothing in today's tail rewrites tool-call/tool-result pairing, so there is no live failure — but the invariant the comment states (and that `tool_result_integrity.py`'s "Keep it late in the stack" relies on) is not actually enforced, and the same comment is duplicated in `graph_utils.build_common_middleware_stack`. Worth correcting the comment, or asserting the position in a test so a future deepagents tail addition doesn't silently slip underneath the guard.

alloy-cicd:

Reviewed `bugfix/orphaned-tool-results` (4 source files + 3 test files). The core approach is sound: I verified against the installed `langchain` / `langgraph` / `deepagents` sources that `model_to_tools` excludes already-answered calls (so the loop-detection routing change is correct), that `ModelRequest.override` exists, and that `repair_tool_result_pairing` is logically correct on every message shape I could construct.

Four findings, posted as inline comments:

1. **`loop_detection_middleware.py:363`** — `force_stop_is_safe` is all-or-nothing across every tool call on the last `AIMessage`. Since `after_model` hooks run innermost-first, `ConditionalHumanInTheLoopMiddleware` injects a `ToolMessage` while keeping the call, so a user repeatedly rejecting a looping guarded tool keeps `already_answered` non-empty every superstep and force-stop never fires — an unbounded block/retry loop, with `tool_call_history` growing without bound because blocked entries skip the window trim.
2. **`tool_result_integrity.py:38`** — `_MISSING_RESULT` claims "Nothing was executed. Assume no effect", but the docstring's own summarization-cutoff example means the tool *did* run. A side-effecting tool can be re-executed; the text should say the outcome is unknown.
3. **`tool_result_integrity.py:91`** — the repair drops the orphan but leaves the empty-content `AIMessage` that force-stop wrote. Bedrock Converse rejects a blank assistant content block, so the bricked-thread shape this PR targets may still 400, just differently.
4. **`graph_factory.py:733`** — the "Truly last / deliberately innermost" comment (duplicated in `graph_utils.py`) is inaccurate; `deepagents._apply_custom_middleware` places user middleware ahead of the tail. No live failure, but the stated invariant is unenforced.

Submitted as **COMMENT** — the human reviewer decides.

aartaria:
### Triage of the automated review (round 1)

The PR has been reworked since: the request-boundary repair middleware is gone (not worth
the complexity — fix the root cause instead), and force-stop no longer rewrites messages.

1. **`force_stop_is_safe` is all-or-nothing / HITL rejection starves force-stop** — valid,
   and it pointed at the real design flaw. Confirmed in the upstream HITL middleware:
   `_process_decision` returns `(tool_call, tool_message)` on *reject*, i.e. it keeps the
   call and answers it, so `already_answered` would indeed stay non-empty and force-stop
   would never fire. **Fixed at the root**: force-stop now returns `{"jump_to": "end"}`
   (hook declared with `@hook_config(can_jump_to=["end"])`) and leaves the message history
   untouched. No message rewrite means no orphan is possible and no guard is needed, in
   the HITL path or anywhere else.
2. **`_MISSING_RESULT` overclaims "nothing was executed"** — moot, the middleware is gone.
   Fair point regardless.
3. **Repair leaves an empty-content `AIMessage` that Bedrock may still reject** — moot,
   the middleware is gone. Also no longer reachable: force-stop no longer produces that
   message.
4. **"Truly last / deliberately innermost" comment doesn't match `_apply_custom_middleware`** —
   moot, both comments were removed with the middleware.

Separately, and not from the review: the actual trigger was that `FinalResponseSchema` was
being loop-tracked at all. That is now fixed too — see the PR description.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
```diff
@@ -345,28 +380,31 @@ async def aafter_model(
...
-                content=last_ai_message.content or "",
-                id=last_ai_message.id,
-                response_metadata=last_ai_message.response_metadata,
+                f"{self.force_stop_after}+ consecutive times. Ending the run."
             )
             return {
                 "tool_call_history": history,
-                "messages": [modified_ai],
+                "jump_to": "end",
             }
```
alloy-cicd: **Force-stop now leaves the blocked tool call unanswered — the mirror-image 400.**

The old force-stop stripped `tool_calls`, so the run ended with a well-formed history. `jump_to: "end"` returns *before* the `error_messages` block below, so nothing answers the blocked calls, and the last `AIMessage` keeps them. In the ordinary force-stop path the call has no result yet (a blocked call never reaches the tools node — the earlier rounds answer it with a `BLOCKED` `ToolMessage`, but this branch returns first), so the checkpoint is persisted as:

| # | type | content |
|---|------|---------|
| n | `ai` | `tool_calls=[read_file id=tc-9]` |

The next user turn appends a `HumanMessage` directly after that assistant turn, so the request carries a `toolUse` with no matching `toolResult`. Bedrock Converse rejects it (`Did not find 1 tool_result block(s) at the beginning of this turn`), and — same as the orphan this PR fixes — the offending messages live in the checkpoint, so every later turn replays them. Bricked thread, opposite direction.

This is not speculative: `_seal_dangling_tool_calls` in `orchestrator-agent/app/middleware/dynamic_tool_dispatch.py:487-517` exists precisely because "sending `tool_use` with no `tool_result` … providers reject outright", and `test_adoption_postgres_fork.py:284` records that sealing happens **only in the forked copy** — "the run's own thread still ends with the dangling call, unsealed". Nothing on the live resume path repairs it, and `DynamicToolDispatchMiddleware`'s empty-`AIMessage` filter deliberately preserves messages that carry tool calls (`test_empty_ai_message_filter.py:121`). The request-boundary repair that would have caught this was removed in the same commit.

`test_force_stop_strips_tool_calls_after_threshold` (renamed in spirit but not in name) pins the broken shape: `tc-123` has no `ToolMessage`, and the test asserts `"messages" not in result`.

Two aggravating cases:

* `all_tracked_blocked` only counts tracked calls, so an `AIMessage` carrying a blocked call **plus** a now-exempt `RESPONSE_TOOLS` call force-stops with *both* left unanswered.
* The run ends on an `AIMessage` whose content is `""` (structured output), so the user also gets an empty final answer where the old code preserved `last_ai_message.content`.

Suggested fix — answer every unanswered call *and* end the run, which keeps both invariants:

```python
if should_force_stop and all_tracked_blocked:
    ...
    stop_results = [
        ToolMessage(
            content=self._build_error_message(info),
            tool_call_id=info["tool_call"]["id"],
            name=info["tool_name"],
            status="error",
        )
        for info in blocked_calls
        if info["tool_call"]["id"] not in already_answered
    ]
    # plus a result for any untracked call on the same AIMessage
    update: dict[str, Any] = {"tool_call_history": history, "jump_to": "end"}
    if stop_results:
        update["messages"] = stop_results
    return update
```

Appending results is not the rewrite that caused #182 — it never removes a call, so it cannot orphan anything, and `already_answered` still guards the duplicate case.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (1dcaec7ee2f02cf8da13eae7930fe41c09022328..bb63fcb638308a78e6a6a3821f36dc554b4ef0e2)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | `force_stop_is_safe` disarmed force-stop all-or-nothing (HITL path) | **verified-fixed** — `force_stop_is_safe` is gone; the stop is expressed as `jump_to: "end"`, so an already-answered call no longer suppresses it. `test_force_stop_never_rewrites_messages` covers the HITL shape directly. |
| 2 | `_MISSING_RESULT` asserted "Nothing was executed" | **rejected-by-author** (dissolved) — `tool_result_integrity.py` deleted; PR scope is now write-side only. |
| 3 | Repair left an empty-content `AIMessage` | **rejected-by-author** (dissolved) — same deletion, and force-stop no longer creates that message at all. |
| 4 | "Truly last / deliberately innermost" comment was inaccurate | **rejected-by-author** (dissolved) — comment and registration removed from both `graph_factory.py` and `graph_utils.py`. |

Dropping the request-boundary middleware is a defensible scope call, and the PR description states the consequence (already-corrupted threads stay broken) plainly. No argument from me.

**New in this delta: 1 finding**

`jump_to: "end"` fixes the orphaned *result* but introduces the dangling *call* — force-stop returns before the `BLOCKED` `ToolMessage` block, so the blocked call is persisted unanswered and the next turn sends a `toolUse` with no `toolResult`. The repo's own `_seal_dangling_tool_calls` documents that providers reject this outright, and it runs only on the fork path, not the live thread. Details and a suggested fix inline.

Verified as correct in the delta: `@hook_config(can_jump_to=["end"])` is picked up from the async-only override (`_get_can_jump_to` falls through to `aafter_model`), `jump_to` is `EphemeralValue` so it cannot leak into the next turn, the mid-chain `after_model` node does get a conditional jump edge, `response_tools=set()` correctly opts out (the guard is `is None`, not falsy), and both `FinalResponseSchema` / `SubAgentResponseSchema` match the real class names.

Two follow-up-only notes, not blocking: `test_force_stop_strips_tool_calls_after_threshold` no longer describes what it tests, and `RESPONSE_TOOLS` duplicates class names as strings across package boundaries (`orchestrator-agent`'s `FinalResponseSchema` isn't importable from `agent-common`, so this is reasonable — but a rename would silently un-exempt the tool).

Submitted as **COMMENT** — the human reviewer decides.

aartaria:
### Triage of the automated review (round 2)

**Finding: force-stop trades the orphaned result for a dangling call** — valid, confirmed,
and fixed in 29c3e03.

I checked the claim rather than taking it: `_seal_dangling_tool_calls`
(`orchestrator-agent/app/middleware/dynamic_tool_dispatch.py:487`) exists precisely because
a `tool_use` with no `tool_result` is rejected outright, and it runs only when adopting a
forked run — never on a live thread. So ending the run without answering the blocked call
did leave the live checkpoint in the mirror-image bad shape. The aggravating case is real
too: `all_tracked_blocked` counts only *tracked* calls, so force-stop can fire on a turn
that also carries an exempt response tool or a call outside a `tool_name`-filtered
instance.

Fix, as suggested: force-stop now answers **every** call on the turn that has no result yet
— blocked ones with their `BLOCKED` message, any other unanswered call with a "not
executed" result — and then jumps to END. Appending a result can never remove a call, so
that direction cannot orphan anything.

New test `test_force_stop_seals_every_unanswered_call_on_the_turn` covers the mixed
tracked/untracked turn; the two existing force-stop tests now assert the seal as well.

Suites after the change: agent-common 954, orchestrator-agent 769, agent-runner 35,
ringier-a2a-sdk 390.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (bb63fcb638308a78e6a6a3821f36dc554b4ef0e2..29c3e0351d4dcc6d683717893c6040b1bac60631)

**Ledger**

| Round | Finding | Disposition |
|---|---|---|
| 2 | Force-stop left the blocked call unanswered → dangling `tool_use`, mirror-image 400 | **verified-fixed** |
| 1 | `force_stop_is_safe` disarmed force-stop all-or-nothing | verified-fixed (round 2) |
| 1 | `_MISSING_RESULT` asserted "Nothing was executed" | rejected-by-author (`tool_result_integrity.py` deleted) |
| 1 | Repair left an empty-content `AIMessage` | rejected-by-author (same deletion) |
| 1 | "Truly last / deliberately innermost" comment inaccurate | rejected-by-author (code removed) |

**Every previous finding is resolved and the delta introduces nothing new.**

The fix goes further than what I suggested, correctly: sealing iterates `last_ai_message.tool_calls` rather than just `blocked_calls`, so a call that `all_tracked_blocked` ignores — an exempt `RESPONSE_TOOLS` call, or one outside this instance's `tool_name` filter — also gets a result instead of being left dangling. That gap was real (`_matches_tool_filter` excludes both from `tracked_calls`) and I had only mentioned it in passing.

Checked in this delta:

- Every `tool_call` on the turn is either in `already_answered` or gets exactly one new `ToolMessage`, so the turn's `toolUse`/`toolResult` counts match in both directions — no orphan, no dangling call.
- Appending never removes a call, so the sealing cannot reintroduce #182; the `already_answered` guard still prevents a second result for the HITL/structured-output case (`test_force_stop_never_rewrites_messages` still asserts `"messages" not in result` there).
- `blocked_by_id` routes each blocked call to its own `_build_error_message(info)`; only genuinely-unexecuted calls get `_STOPPED_BEFORE_EXECUTION`, so the model is not told a looping call "was not executed".
- `messages` and `jump_to` in one update is fine — LangGraph applies the node's writes before the conditional edge reads `state.get("jump_to")`, and `jump_to` is `EphemeralValue` so it cannot leak into the next turn.
- The resulting shape (assistant `tool_use` → sealed `tool_result` → next human turn) is the same one `_seal_dangling_tool_calls` deliberately produces on the adoption path, so it is already treated as valid here.
- `agent-common` suite: 40/40 in `tests/test_loop_detection_middleware.py`, run against this head. Nothing else in the repo depends on the old force-stop shape (the only other `force-stop` mentions are prose comments).

One follow-up-only note, not blocking: the PR description still says force-stop "ends the run by jumping to `END` and never touches the message history" and does not mention sealing at all — worth a refresh so the description matches the third fix.

Submitted as **COMMENT** — the human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (29c3e0351d4dcc6d683717893c6040b1bac60631..e4bec425f9471561f6b665330c7b8445d471d678)

**Ledger**

| Round | Finding | Disposition |
|---|---|---|
| 3 | *(no findings — clean round)* | — |
| 3 | Follow-up note: PR description didn't mention sealing | **verified-fixed** — description now documents the seal-then-jump sequence and why appending is safe |
| 2 | Force-stop left the blocked call unanswered → dangling `tool_use` | verified-fixed (round 3) |
| 1 | `force_stop_is_safe` disarmed force-stop all-or-nothing | verified-fixed (round 2) |
| 1 | `_MISSING_RESULT` / empty `AIMessage` / "truly last" comment | rejected-by-author (`tool_result_integrity.py` deleted) |

**Every previous finding is resolved and the delta introduces nothing new.**

The delta is test-only: 329 lines of graph-level cover with a real `create_agent` graph, a real `InMemorySaver`, and assertions against both the persisted history and the message list handed to the model on the resumed turn. That is the right level for this bug — the unit tests inspect a state update, but the failure was always in the checkpoint and only killed the *next* turn.

I verified the commit message's non-vacuousness claim rather than taking it on trust, by swapping in older versions of `loop_detection_middleware.py` and re-running the new file:

- against the pre-fix middleware (`abd4b7bc`): **3 of 5 fail**, exactly as claimed — including `assert 2 == 1`, the duplicate result for `answer-2` on the structured-output ladder.
- against the round-2 middleware (`bb63fcb6`, `jump_to: "end"` with no sealing): **2 of 5 fail**, catching the dangling-call regression I raised in round 2.

That second run is worth calling out: `test_conversation_resumes_cleanly_after_a_force_stop` modestly disclaims itself as non-discriminating, which is true of the *original* stripping bug but not of the intermediate one — it is precisely the test that fails on the dangling call. The suite now pins both failure directions, not just the one in the issue title.

Also checked: `assert_tool_pairing_valid` is itself guarded by `test_pairing_helper_catches_both_failure_shapes` and genuinely rejects both shapes; `AnswerSchema` is deliberately named outside `RESPONSE_TOOLS` so the structured-output test exercises force-stop rather than the exemption; the `assert self.calls < 25` guards mean a non-terminating run fails loudly instead of hanging the suite; and the pydantic mutable-default fields are safe (deep-copied per instance, and both tests pass an explicit list). 45/45 pass on this head.

Two follow-up-only notes, neither blocking: `assert_tool_pairing_valid` does not check *adjacency* — a `ToolMessage` separated from its `AIMessage` by a `HumanMessage` would pass the helper but not the provider — and its `answered` dict is never reset per turn, so the helper quietly assumes globally unique `tool_call_id`s. Separately, langgraph emits a deprecation warning when deserializing the test-local `FinalResponseSchema` from the checkpoint ("will be blocked in a future version"), which will eventually need `allowed_msgpack_modules` or a module-level schema.

Submitted as **COMMENT** — the human reviewer decides.

aartaria:
### Triage of the automated review (round 4) + two more test cases

**Both follow-up notes taken** (744210d4). `assert_tool_pairing_valid` was looser than the
provider rule on exactly the two axes flagged:

- *Adjacency*: the rule is per turn — results must follow their `AIMessage` immediately. A
  `ToolMessage` separated from its call by a `HumanMessage` passed the helper but would not
  pass the gateway. Any non-result message now closes the group.
- *Global ids*: a `tool_call_id` legitimately reused in a later turn would have been read as
  a duplicate result. Ids are now scoped per group.

Both shapes added to the helper's own self-test. Still discriminating after the change:
4 of 7 fail against pre-fix, 4 of 7 against the intermediate `jump_to: "end"`-without-sealing
revision.

*The langgraph deserialization warning I could not reproduce* — no warning on this head,
under the full suite or with `-W always`. Possibly environment-specific; leaving it.

**Two further cases added** (92ac2696), covering the turn *after* a stop:

- **The repeat counter survives the stop.** `tool_call_history` is private state in the
  checkpoint; if a stop lost it, the next turn would start from zero and just run the loop
  again, one turn later. Asserted behaviourally — after a stop the identical call is refused
  and the tool does not execute once more. Verified discriminating: fails when the history
  is made not to carry across turns, and against pre-fix.
- **The "do NOT retry" guidance is still readable.** The `BLOCKED` result is what tells the
  model to stop retrying and answer with what it has, so the next turn must replay it
  verbatim — and the force-stopped call in particular used to get no result at all, leaving
  a bare refusal with no reason.

`agent-common` now 961 passing.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (e4bec425f9471561f6b665330c7b8445d471d678..744210d4802cfddcfd46ff10aaf04e50af9a91b7)

**Ledger**

| Round | Finding / note | Disposition |
|---|---|---|
| 4 | *(no findings — clean round)* | — |
| 4 | Note: `assert_tool_pairing_valid` didn't enforce adjacency | **verified-fixed** — any non-result message now closes the group |
| 4 | Note: helper assumed globally unique `tool_call_id`s | **verified-fixed** — ids scoped per group |
| 4 | Note: langgraph msgpack deprecation warning on the test-local schema | **still-open** (non-blocking, test-only) |
| 3 | Note: PR description didn't mention sealing | verified-fixed (round 4) |
| 2 | Dangling `tool_use` after `jump_to: "end"` | verified-fixed (round 3) |
| 1 | `force_stop_is_safe` all-or-nothing | verified-fixed (round 2) |
| 1 | `_MISSING_RESULT` / empty `AIMessage` / "truly last" comment | rejected-by-author (`tool_result_integrity.py` deleted) |

**Every previous finding is resolved and the delta introduces nothing new.**

Test-only delta, and it does more than close my two notes. `92ac2696` pins two properties nothing had covered: that `tool_call_history` survives the force-stop (asserted behaviourally — the tool must not execute again on the next turn, which is the failure a state-shaped assertion would have missed), and that the `BLOCKED` guidance is replayed verbatim to the model on the following turn. The second is the one that closes the loop on the whole PR: the force-stopped call used to get no result at all, so the model was refused with no reason and would simply retry.

I re-verified the commit messages' discrimination claims rather than trusting them, by swapping in the older middleware and re-running the suite:

- vs. pre-fix (`abd4b7bc`): **4 of 7 fail** — as claimed.
- vs. the intermediate jump-to-end-without-sealing revision (`bb63fcb6`): **4 of 7 fail** — as claimed.

Note the two sets differ (`test_structured_output_turns_never_corrupt_the_checkpoint` and `test_response_tool_is_never_force_stopped` catch only the original bug; `test_conversation_resumes_cleanly_after_a_force_stop` and `test_do_not_retry_guidance_reaches_the_model_on_the_next_turn` catch only the intermediate one; `test_blocking_still_fires_on_the_turn_after_a_force_stop` catches both). The suite discriminates against every revision this PR has been through, not just the starting point. 47/47 pass on this head.

I also read the rewritten helper closely, since a tightened invariant is itself a place to introduce a false positive. It holds up: `_close` reads the rebound `open_calls`/`answered`/`opened_at` through the closure so it always validates the group that is actually ending; it is called before the reset, so the group is checked at the right moment; an `AIMessage` without `tool_calls` correctly closes without opening; a leading `ToolMessage` still trips the orphan branch; and the final `_close(len(messages))` catches a trailing dangling call. The new self-tests cover both the separated-result shape (which the old helper accepted) and legitimate id reuse across turns (which a global `answered` set would have misread as a duplicate).

One follow-up-only note, not blocking: the PR description's graph-test section is now a revision behind — it still says "three of the five fail" where it is four of seven, and does not mention the counter-survival or guidance-replay tests or the per-turn tightening.

Submitted as **COMMENT** — the human reviewer decides.

Issue #192, opened 2026-09-02
console-backend: a scheduled job accepts a delivery channel its owner cannot receive on


aartaria:
## Summary

Setting a `delivery_channel_id` on a scheduled job verifies only that the channel exists. Nothing checks that the job's owner can actually *receive* on that channel — i.e. that they have completed the authorize flow with the installation the channel belongs to. A job can therefore be created, saved and scheduled in a state where its notifications are guaranteed to be undeliverable, and the user learns nothing until the first run has already silently failed.

## Mechanism

`SchedulerService._validate_delivery_channel` (`packages/console-backend/console_backend/services/scheduler_service.py`) exists precisely so that a bad channel id fails at creation rather than at the database, but it stops at existence:

```python
channel = await self._delivery_channel_repo.get_channel_by_id(db, channel_id)
# ... raises if the channel does not exist; no ownership/reachability check
```

At dispatch time the engine loads the channel and hands its secret to the push sender (`_build_dispatch`, `scheduler_engine.py`). The receiving client then authenticates as that channel's installation and tries to resolve the job owner *within that installation* — which fails if the owner never authorized it. See #191 for the receiving half.

Because a user is typically authorized with one installation and merely a member of the other workspaces, choosing the "wrong but valid" channel is easy to do deliberately and impossible to detect from the UI.

## Why it matters

The information needed to catch this is available at creation time, when the fix is one action away and the user is present to take it ("authorize <installation> first, or these notifications will not reach you"). Deferring it to dispatch turns a solvable setup problem into an invisible one: the job runs on schedule, forever, delivering nothing.

## What a fix needs
To address the issue of scheduled jobs being created with delivery channels that the owner cannot actually receive on, the solution discussed involves improving both the user experience and backend validation:
* UI/UX Improvement: Instead of simply showing all possible delivery channels, the interface will distinguish between active and inactive channels. Channels that the user has not yet interacted with (e.g., authorized) will appear grayed out in the dropdown, accompanied by a hint instructing the user to interact with that specific channel (e.g., by messaging Nannos on Google Chat or Slack) to activate it for receiving notifications
* Tooling Adaptation: MCP tools will be adapted to provide actionable feedback when a job creation fails due to an unauthorized channel. Instead of a silent failure, the tool will explicitly guide the user on how to resolve the issue, such as providing the specific commands or steps needed to authenticate the channel
[...]

Issue #191, opened 2026-09-02
client-slack: an undeliverable scheduled-run notification is dropped silently


aartaria:
## Summary

When a scheduled run's completion notification cannot be delivered, the notification is discarded and nothing tells the user. The run itself is recorded as successful, so the only trace is a `WARNING` line in the client's logs.

Observed in practice: a scheduled job whose delivery channel pointed at one Slack installation completed normally and reported success, but no DM ever arrived. The user found out only by reading service logs.

## Mechanism

`handleA2ANotification` resolves the recipient with `findByOidcSubAndTeam(payload.user_sub, botInstallation.teamId)` and, on a miss, logs and returns:

```
// packages/client-slack/src/handlers/a2aNotificationHandler.ts — handleA2ANotification
const userAuth = await userAuthStorage.findByOidcSubAndTeam(schedulerPayload.user_sub, botInstallation.teamId);
if (!userAuth) {
  logger.warn(`[A2ACallback] No Slack user found for oidcSub=... in team=...`);
  return;
}
```

The lookup misses whenever the job owner has no `user_auth` row for the installation that owns the delivery channel. That is a normal, reachable state: `user_auth` rows are keyed `(user_id, team_id)` and written only by that installation's OIDC authorize callback, so a user who is a member of a workspace — and who chats happily with a *different* installation — has no row until they have completed the authorize flow with *that* one.

The failure is invisible from the user's side because the interactive path resolves identity by `getToken(userId, teamId)` and never reads `oidc_sub`. Only the notification path queries by subject. An account can therefore work perfectly in conversation while being unreachable for scheduled delivery.

The same silent return covers the other delivery preconditions — a missing `botToken`, and a `conversations.open` that fails — each of which ends in `logger.warn` and no user-visible signal.

## Why it matters

Scheduled runs exist to tell you something happened while you were not watching. A notification path that can fail without saying so converts the whole feature into a silent no-op, and the user's only debugging tool is server logs they normally cannot see. Worse, the run is marked delivered-adjacent: `delivered=(job.delivery_channel_id is not None)` is recorded from the *presence* of a channel, not from an actual delivery.

## What a fix needs

- An undeliverable notification must surface to the user through some other route — an in-app notification against the job, at minimum — rather than terminating in a warning.
- Distinguish "no recipient could be resolved" from "delivery was attempted and failed": the first is a configuration problem the user can fix (authorize with that installation), the second is transient.
[...]

PR #194, opened 2026-09-02
console-backend: notify on new bug reports — console_list_bug_reports MCP tool, created_after filter, in-app notification on filing

### LINKED ISSUE #193: console-backend: notify on new bug reports — expose console_list_bug_reports (+ created_after) and write an in-app notification on filing
## Problem

There is no way to be notified when a bug report is filed. A user who wants "tell me when someone files a bug report" has no path at all today:

- `GET /api/v1/bug-reports` (`list_bug_reports`) is REST-only — it is not exposed as an MCP tool, so a **watch job cannot use it as a `check_tool`**. `bug_report_mcp_tools.py` exposes create / update-status / set-external-link, but no list.
- Filing a bug report writes nothing to the in-app notification inbox, unlike the other lifecycle events (agent activation, secret sharing, group membership, approvals) that all go through `NotificationService.create_notification`.

## Scope

Three changes, all in `packages/console-backend`:

**1. `console_list_bug_reports` MCP tool.** Add an MCP-shaped route to `bug_report_mcp_tools.py` (`tags=["MCP"]`, `operation_id="console_list_bug_reports"`), mirroring the three tools already in that file. The existing REST route cannot simply be tagged — the MCP tools in this file are separate routes by convention.

Access control follows the REST endpoint exactly: administrators see all reports, everyone else sees only their own. State this in the tool description so a caller does not promise more coverage than the tool delivers. No new capability gate — unlike `console_set_bug_report_external_link`, listing your own reports is not triage-only.

**2. `created_after` filter on the list query.** Add `created_after: datetime | None` to `BugReportService.list_bug_reports` / the repository query (`WHERE created_at > :ts`; ordering is already `created_at DESC`), and expose it on both the REST and MCP routes.

This is a correctness fix, not just an optimisation. Without it, a change-detecting watch has to diff the whole first page against `prev` (`last_check_result`), which means (a) the full result set is stored on the job on every poll, and (b) with `limit=50`, more than 50 reports arriving between two polls silently loses the tail. With it, the payload is bounded by the window rather than by the page.

A watch can then be configured as:

```
check_tool:        console_list_bug_reports
check_args:        {limit: 50, status_filter: open}
check_args_exprs:  {created_after: strftime(now - duration("600s"), "%Y-%m-%dT%H:%M:%S%z")}
cel_expr:          prev == null ? [] : result.data.filter(r, !(r.id in prev.data.map(p, p.id)))
destroy_after_trigger: false
```

Notes on that expression, for whoever writes the docs:
- the `prev == null` guard is load-bearing — without it the first poll reports every pre-existing report as new;
- returning a *list* rather than a bool gates on non-empty **and** puts the new reports in `condition_evaluation.extracted`, so the notification names them;
- window at ~2x the poll interval so a skipped tick does not open a gap, and keep the id-diff for dedup across the overlap;
- `destroy_after_trigger` defaults to `true`, which disables the job after one trigger — a standing watch must set it false.

**3. `BUG_REPORT_FILED` notification type.** Add the enum value (DB enum + `NotificationType`) and write a notification from `BugReportService.create_bug_report` for users with the `bug_reports` triage capability. Same shape as the existing lifecycle notifications; must not fail the bug-report creation if the notification write fails.

## Out of scope

Routing in-app notifications to a registered delivery channel (the event-driven path that would make the watch job unnecessary for this and every other event Nannos itself writes). That is a separate design with its own ADR — it makes console-backend an outbound push sender for the first time, and it has to decide what "delivered" means when a channel accepts a notification for a user it cannot resolve (cf. #191, #192). Item 3 here is the seam it would plug into.

## Acceptance

- [ ] `console_list_bug_reports` appears in the MCP tool list and returns the caller's own reports (all reports for an administrator)
- [ ] `created_after` filters both the REST and MCP list responses
- [ ] a watch job using the configuration above fires once per newly filed report, and not on its first poll
- [ ] filing a bug report creates an in-app notification for triagers, and a failing notification write does not fail the report

### LINKED ISSUE #198: users: introduce is_service_account so notification audiences can exclude machine identities structurally
## Problem

There is no way to tell a machine identity from a person in the `users` table, so anything that selects an audience by standing has to guess. Two machine rows exist in every deployment:

- **`system`** — seeded by migration 041 with `role = 'admin'` to own auto-provisioned agents.
- **`service-account-<client>`** — auto-onboarded by `dependencies.py` from a client-credentials token the first time a service (e.g. the orchestrator) calls console-backend.

Neither has a person behind it, and neither has an inbox anyone reads.

This bit twice while adding bug-report notifications (#193 / #194):

1. The audience keyed on `role`, which picked up `system` (role 'admin').
2. Re-keyed on `is_administrator` — which migration 041 leaves FALSE, so the exclusion looked redundant and was removed. In a live deployment an operator had promoted that account, so it collected notifications until the exclusion came back.
3. With the exclusion back in by id, `service-account-orchestrator` was **still** in the audience in that same deployment: it is flagged `is_administrator` there too. A hardcoded single id fixes one instance of the class.

The seed's column values are a starting state, not an invariant, and an id list is not a category. What is missing is the category.

## Proposal

Add `is_service_account BOOLEAN NOT NULL DEFAULT FALSE` to `users`, set it where those rows are created, and let audience queries filter on it.

**Migration**

- add the column
- backfill `TRUE` for `id = 'system'` and for rows whose email or username carries Keycloak's `service-account-` convention
- the down migration drops the column

**Where it gets set**

`dependencies.py` auto-onboards from token claims, which is where the signal is: a Keycloak client-credentials token carries `preferred_username = service-account-<clientId>` and no personal identity claims. That call site has the payload, so the flag can be set at creation rather than inferred later. Worth confirming against a real token before trusting a single claim — the lesson from #194 is to reason from the rows a live database holds rather than from what a convention implies.

**What consumes it**

`BugReportService._notify_administrators` replaces its hardcoded `id != 'system'` exclusion with `AND is_service_account IS FALSE`, which also covers the orchestrator's account. Any future notification audience gets the same filter for free — see the note in `packages/console-backend/AGENTS.md`.

## Out of scope

Whether these accounts should be `is_administrator` at all. They are in at least one deployment, nothing in the code puts them there (`FIRST_USER_IS_ADMIN` fires for the first user only), and this issue is about not depending on that either way.

## Acceptance

- [ ] `users.is_service_account` exists, defaults FALSE, and is TRUE for `system` and for auto-onboarded service accounts
- [ ] a service account newly onboarded from a client-credentials token is flagged at creation
- [ ] the bug-report notification audience excludes service accounts by flag, not by id, with a test that promotes both machine rows to `is_administrator` first
- [ ] `AGENTS.md` guidance updated to point at the flag instead of recording that no marker exists

aartaria:
closes #193
closes #198

## What

Being notified when a bug report is filed had no path at all. `GET /api/v1/bug-reports` was REST-only, so a scheduled watch could not name it as a `check_tool`, and filing a report wrote nothing to the notification inbox that every other lifecycle event uses.

**1. `console_list_bug_reports` MCP tool.** An MCP-shaped route alongside the three already in `bug_report_mcp_tools.py`. Visibility follows the REST endpoint exactly — administrators see every report, everyone else only their own — and the tool description says so, because an empty result otherwise reads as "none at all" rather than "none of yours". Triage capability deliberately does not widen it: it governs acting on a report, not seeing every user's.

The path is `/mcp/list`, not `/mcp-list`: `bug_report_router` is registered first and owns `GET /{report_id}`, which would match a one-segment path and 404 on a report id of "mcp-list". The `PATCH` tools have no such problem — their REST siblings are two segments too. There's a test pinning this.

**2. `created_after` on the list**, both routes. A correctness fix rather than an optimisation: without it a change-detecting watch must diff the whole first page against `prev` (`last_check_result`), which stores the full result set on the job on every poll and, at `limit=50`, silently loses the tail when more than 50 reports arrive between two polls.

A standing watch can then be configured as:

```
check_tool:        console_list_bug_reports
check_args:        {limit: 50, status_filter: open}
check_args_exprs:  {created_after: strftime(now - duration("600s"), "%Y-%m-%dT%H:%M:%S%z")}
cel_expr:          prev == null ? [] : result.data.filter(r, !(r.id in prev.data.map(p, p.id)))
destroy_after_trigger: false
```

The `prev == null` guard is load-bearing (without it the first poll reports every pre-existing report as new); returning a list rather than a bool gates on non-empty *and* puts the new reports in `condition_evaluation.extracted`, so the notification names them; the window is ~2x the poll interval so a skipped tick opens no gap, with the id-diff deduplicating the overlap.

**3. `BUG_REPORT_FILED` notifications.** Filing writes to every triager's inbox, after the report is committed and best-effort — the same ordering the other notification producers use, so a failing inbox insert cannot lose a bug report.

**4. `users.is_service_account`** (#198, folded in here because it replaces this PR's own machine-account exclusion). Machine identities had no marker, so audiences selected by standing had to guess — and guessing went wrong three times in review: keyed on `role`, which caught the seeded `system` account; re-keyed on `is_administrator`, which a live deployment had promoted that account into; then excluded `system` by id, which left the orchestrator's own service account, flagged administrator in that same deployment, still collecting notifications nobody reads.

Migration 088 adds the column and backfills `system` plus rows carrying Keycloak's `service-account-` email convention. New rows are flagged at creation by `dependencies.token_is_service_account`, from `preferred_username` — the one claim where the issuer *states* it. Absence of identity claims looks like a second signal and is not: `require_auth_or_bearer_token` onboards a person from a token carrying nothing but a `sub`, and `test_dependencies.py` says so. Set on insert only, like `is_administrator`, so an operator's correction survives the next call.

`_notify_administrators` filters on the flag, covering both machine rows at once.

**5. One naming convention for the MCP paths.** `/mcp/list` was the odd one out beside `/mcp-create`, `/{report_id}/mcp-status` and `/{report_id}/mcp-external-link`; it used two segments to dodge a collision with the REST router's `GET /{report_id}`. The collision is now fixed at its source — the MCP router is registered first, so its literal segments win the match — and the path is `/mcp-list`. Safe because every path in that router is a literal, and the test asserts both directions: `mcp-list` reaches the tool, a real report id still reaches `GET /{report_id}`.

## Worth a reviewer's attention

**The seeded `system` user is an admin.** Migration 041 creates it with `role = 'admin'` to own auto-provisioned agents. The triage audience is the first notification audience selected by *role* rather than group membership, so it is the first one that reaches it — an inbox nobody opens, one row per bug report forever. Excluded, with a test that asserts the seed still looks that way and a note in `AGENTS.md`, because the next role-selected audience will hit the same thing.

The audience is derived from `SYSTEM_ROLE_CAPABILITIES` rather than hardcoded, so a role that gains or loses `triage` is not silently left in or out of it.

**The regenerated SDK carries unrelated drift.** `npm run gen-sdk` also picked up `deleteConversation` / `renameConversation` — endpoints already on `main` whose client was never regenerated. Regeneration is all-or-nothing and the convention is to commit it with API changes, so it's included rather than hand-trimmed. `tsc -b` is clean.

## Out of scope

Routing in-app notifications to a registered delivery channel — the event-driven path that would make a watch job unnecessary for this and every other event Nannos itself writes. It makes console-backend an outbound push sender for the first time and has to settle what "delivered" means when a channel accepts a notification for a user it cannot resolve (cf. #191, #192), so it gets its own ADR. Change 3 here is the seam it plugs into.

## Verification

- `uv run pytest tests` in `packages/console-backend`: **1674 passed**, including 20 in `test_bug_report_router.py` (8 new: `created_after` boundary, offset-less `created_after` under a pinned UTC+14 process timezone, out-of-range paging on both surfaces, MCP list visibility/filters/no-shadowing, administrator fan-out, non-admin triager exclusion, reporter exclusion, `system` exclusion, survival of a failing notification)
- the timezone normalisation was verified by removing it and watching its test fail, then restoring it — on a UTC runner that test is vacuous without the pinned `TZ`, which is what round 2 of the review caught
- `console_list_bug_reports` confirmed present in the app's MCP tool list, with `page` / `limit` / `status_filter` / `created_after` and no required params
- `npx tsc -b` in `packages/console-frontend`: clean
- `ruff check` adds no findings on the touched files (the package has no ruff config or lint gate; the repo's existing 22 findings are untouched)

## Review

Three rounds of automatic review, now clean. Round 1 raised six findings: five taken as reported, one (the scheduled-watch pattern) fixed in part with the remainder declined on the PR. Round 2 verified those and raised one more — the timezone test was vacuous on a UTC runner, and a reply of mine had claimed the opposite; fixed in `17fb7795` and verified empirically this time. Round 3 found nothing.

Every finding has a reply on its thread saying what happened and why, including the three sub-points declined (`clock_timestamp()` for `created_at`, keyset pagination, dropping the `COUNT(*)` on the polling path).

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no. Both new query params are optional, and the migration only adds an enum value.


packages/console-backend/console_backend/services/bug_report_service.py
```diff
@@ -45,8 +71,93 @@ async def create_bug_report(
...
+        bug report forever.
+        """
+        if self.notification_service is None:
+            return
+
+        try:
+            recipients = await db.execute(
+                text("""
+                    SELECT id FROM users
+                    WHERE (role = ANY(:triage_roles) OR is_administrator IS TRUE)
```
alloy-cicd: 🤖 **Notification audience is wider than read visibility: approvers are notified about reports they cannot open.**

`_TRIAGE_ROLES` includes `approver` (`bug_reports: {"read", "triage"}`), so non-admin approvers get the notification — including a 200-char snippet of the description. But every read path is admin-or-owner only: `GET /{report_id}` 404s for them (`bug_report_router.py:94`), and both list endpoints filter to `user.id` for non-administrators (the MCP list even says so deliberately). The frontend bug-reports page is `<AdminRoute>`-gated too. So an approver receives content they are denied on every read surface and cannot view the report the notification points at (they can only act blind via the triage-gated PATCH routes). Either the audience should match who can read, or the read paths should honor the triage capability the way `PATCH /{id}/status` already does.

packages/console-backend/console_backend/repositories/bug_report_repository.py
```diff
@@ -99,6 +101,12 @@ async def list_bug_reports(
             conditions.append("status = :status")
             params["status"] = status.value
 
+        if created_after is not None:
+            # Strictly greater than, so a window whose lower bound is a report's own
+            # timestamp does not return that report again.
+            conditions.append("created_at > :created_after")
```
alloy-cicd, line 116: 🤖 **A timezone-naive `created_after` silently shifts the window by the app container's UTC offset.**

`created_at` is `TIMESTAMPTZ` (migration 046), and this value is bound straight into raw `text()` SQL. FastAPI accepts an offset-less ISO string (`created_after=2026-09-02T08:00:00`) as a naive datetime — no 422 — and asyncpg's timestamptz encoder does `obj.astimezone(utc)`, which for a naive datetime assumes the **backend process's local timezone**. On any non-UTC deployment the window shifts by the offset with no error, so the advertised rolling-watch pattern re-delivers or skips reports. The tests never hit this because they feed back the API's own offset-aware `created_at`. Fix: normalize (`if created_after.tzinfo is None: created_after = created_after.replace(tzinfo=timezone.utc)`) or reject naive input — `scheduler_service.py:106` and `cel_condition.py:511` already do this elsewhere.

packages/console-backend/console_backend/routers/bug_report_mcp_tools.py
```diff
@@ -134,6 +139,59 @@ async def update_bug_report_status_mcp(
...
+    response_model=BugReportListResponse,
+    tags=["MCP"],
+    operation_id="console_list_bug_reports",
+    summary="List bug reports, most recently filed first.",
+    description=(
+        "List bug reports, newest first. Administrators see every report; everyone else sees only "
+        "the reports they filed themselves — so an empty result means 'none of yours', not 'none at all'. "
+        "Filter with `status_filter` (open, acknowledged, investigating, resolved) and `created_after` "
+        "(an ISO-8601 timestamp; only reports filed strictly after it are returned). `created_after` is "
+        "what a scheduled watch should use to look at a rolling window rather than the whole first page."
```
alloy-cicd: 🤖 **The recommended scheduled-watch pattern can silently lose reports.** Two mechanisms:

1. `created_at` defaults to `NOW()` = transaction-start time, but visibility follows commit time, and the filter is strict `>`. The filing transaction stamps `NOW()` at the auth-lookup query and only commits after the audit insert + re-select, so a slow filing A can become visible *after* a faster, later-stamped B was already seen. A watch that advanced its cursor to B's `created_at` never sees A — permanently.
2. Results are `ORDER BY created_at DESC` with `limit ≤ 100`; a burst larger than the limit means the newest-seen cursor skips the oldest reports in the window. `meta.total` reveals the truncation, but the description never tells the caller to page on it.

Cheap hardening: tell watchers to overlap their window / page until `total` is exhausted, add a stable ORDER BY tiebreaker, and consider `clock_timestamp()` for `created_at` (or keyset pagination on id).

packages/console-backend/console_backend/routers/bug_report_router.py
```diff
@@ -53,6 +54,7 @@ async def list_bug_reports(
     page: int = 1,
     limit: int = 50,
     status_filter: BugReportStatus | None = None,
+    created_after: datetime | None = None,
```
alloy-cicd: 🤖 **This route (unlike its new MCP twin) still accepts `page=0` / negative values, producing a 500.**

The MCP list declares `Query(1, ge=1)` / `Query(50, ge=1, le=100)`; here `page`/`limit` are bare `int`s and only the upper bound of `limit` is clamped (`if limit > 100`). `page=0` → `offset = (0-1)*50 = -50` in `bug_report_repository.py:122` → Postgres `OFFSET must not be negative` → unhandled 500 for any authenticated user (`limit=-1` likewise). Pre-existing, but the PR touches this function and its sibling now validates — worth aligning:

```suggestion
    page: int = Query(1, ge=1),
    limit: int = Query(50, ge=1, le=100),
    status_filter: BugReportStatus | None = None,
    created_after: datetime | None = None,
```
(requires `Query` in the imports; the `if limit > 100` clamp then becomes dead and can go — note this changes REST behavior from silent clamp to 422, matching MCP.)

packages/console-backend/console_backend/routers/bug_report_mcp_tools.py
```diff
@@ -134,6 +139,59 @@ async def update_bug_report_status_mcp(
...
+        ),
+    ),
+    user: User = Depends(require_auth_or_bearer_token),
+) -> BugReportListResponse:
+    service = _get_bug_report_service(request)
+
+    # Same visibility rule as the REST list endpoint: administrators see everything,
+    # everyone else only their own reports. Triage capability deliberately does not
+    # widen this — it governs acting on a report, not seeing every user's.
+    user_id_filter = None if user.is_administrator else user.id
```
alloy-cicd: 🤖 **This handler duplicates the REST `list_bug_reports` body line for line, with divergent validation.**

Same visibility rule, same service call, same `BugReportListResponse`/`PaginationMeta` assembly as `bug_report_router.py:50-78` — only auth dependency, path, and OpenAPI metadata differ (and those must stay). The visibility rule now lives in two places kept in sync only by the comment above; the next change (e.g. letting triagers see all reports) will update one surface and not the other. The validation also diverges: REST silently clamps `limit>100` while MCP 422s, and REST lacks `ge=1` entirely. Suggest a single service-level helper, e.g. `BugReportService.list_for_user(db, user, status, created_after, page, limit) -> BugReportListResponse` (the current `list_bug_reports` service method is a pure pass-through, so it can absorb this), with both routes as one-line wrappers.

packages/console-backend/console_backend/repositories/bug_report_repository.py
```diff
@@ -99,6 +101,12 @@ async def list_bug_reports(
...
+            params["created_after"] = created_after
+
         where_clause = f"WHERE {' AND '.join(conditions)}" if conditions else ""
 
         count_query = text(f"SELECT COUNT(*) FROM bug_reports {where_clause}")
```
alloy-cicd, line 121: 🤖 **No index supports the admin-side `created_after` poll this PR advertises.**

The only indexes on `bug_reports` are `(conversation_id)` and `(user_id, created_at DESC)` (migration 046). Administrators list with `user_id` unset, so the advertised scheduled-watch query — `WHERE created_at > $1 ORDER BY created_at DESC` — plus the unconditional `COUNT(*)` with the same WHERE both scan the table on every poll, with cost growing with total reports rather than window size. Cheap fix in migration 086: `CREATE INDEX idx_bug_reports_created_at ON bug_reports (created_at DESC);` (minor: the COUNT could also be skipped or replaced by `LIMIT limit+1` for the polling path, since a watcher never reads `meta.total`).

alloy-cicd (commented):

Reviewed PR #194 (feat(console-backend): notify on new bug reports). 6 verified findings posted as inline comments:

1. **Audience vs. visibility mismatch** (`bug_report_service.py:109`) — non-admin approvers are notified (with a 200-char description snippet) about reports that every read path (GET by id, both lists, the AdminRoute-gated frontend page) denies them.
2. **Timezone-naive `created_after`** (`bug_report_repository.py:107`) — a naive datetime bound to the TIMESTAMPTZ column is interpreted in the app container's local timezone by asyncpg, silently shifting the advertised rolling-watch window; other paths in the repo normalize, this one does not.
3. **Scheduled-watch pattern can lose reports** (`bug_report_mcp_tools.py:157`) — `created_at` is transaction-start `NOW()` while visibility is commit-time, so a strict `>` cursor can skip slow-committing reports permanently; a burst larger than `limit` is truncated newest-first and the tool description never says to page on `meta.total`.
4. **REST list accepts `page=0` / negative `limit` → 500** (`bug_report_router.py:54`) — pre-existing, but the PR fixed only the MCP twin (`Query(ge=1)`), leaving the two siblings validating inconsistently.
5. **MCP list duplicates the REST list handler line for line** (`bug_report_mcp_tools.py:179`) — the visibility rule now lives in two places with divergent limit semantics (silent clamp vs. 422); extract a shared service helper.
6. **No index for the admin-side `created_after` poll** (`bug_report_repository.py:112`) — the only indexes are `(conversation_id)` and `(user_id, created_at DESC)`; the advertised watch query plus its `COUNT(*)` scan the table on every poll. Migration 086 could add `(created_at DESC)`.

Checked and clean: route registration order and the two-segment `/mcp/list` workaround, `role = ANY(list)` binding, migration format and `ALTER TYPE` usage, SDK regeneration, test conventions, and the mid-service commit/rollback pattern (matches the codebase's other notification producers and is stricter than the incumbents). A suspected migration-before-code deploy gap was refuted — `entrypoint.sh` runs migrations before uvicorn starts.

packages/console-backend/console_backend/services/bug_report_service.py
aartaria: Accepted — and the resolution was a call for the repo owner rather than mine, so I asked. The audience is now `is_administrator IS TRUE` only, matching read visibility: nothing is sent to someone who would be 404d on the report it points at.

Two things fell out of fixing it:

- `role = ANY(triage_roles) OR is_administrator` was hiding a second bug. The seeded `system` user (migration 041) has `role = admin` with `is_administrator` left FALSE, so the role arm was picking up a service account with no person behind it. On the flag alone it drops out by construction, so the explicit `id != 'system'` exclusion I had added is gone as dead code — the test now pins the DB fact that keeps it out (`is_administrator IS FALSE` on that row) rather than asserting a condition that can no longer fire.
- The underlying inconsistency you identified — `triage` permits writes on every accessible report but no read path honours it — is filed as #197, with both options written out (widen the read paths, or drop `triage` from `approver` as unusable). `_notify_administrators` documents the current choice and points there.

packages/console-backend/console_backend/repositories/bug_report_repository.py
aartaria: Accepted, fixed as suggested — `created_after.replace(tzinfo=timezone.utc)` when naive, in the repository so both routes are covered by one normalisation.

You were right that the tests could not have caught it: they fed the API its own offset-aware `created_at` back. There is now one that constructs the boundary deliberately without an offset (`.astimezone(utc).replace(tzinfo=None)`) and asserts the window still lands where UTC says it should, so a regression shows up on a UTC dev box rather than only on a non-UTC deployment.

packages/console-backend/console_backend/routers/bug_report_mcp_tools.py
aartaria: Partly accepted. Splitting the two mechanisms, because they do not land the same way:

**Burst beyond `limit` — accepted, real.** The tool description now tells the caller to keep paging while `meta.total` exceeds what came back, and says what happens if they do not (the oldest reports in the window are the ones they never see). I also added `id DESC` as an `ORDER BY` tiebreaker, so two reports filed inside the same clock tick cannot be ordered arbitrarily between two pages.

**Commit-time visibility — real in general, but not against the pattern as documented.** Your scenario advances a cursor to the newest-seen `created_at`; the documented pattern deliberately does not do that, because a watch has no state to hold a watermark in (`prev` is the previous *result*, and once a window comes back empty there is nowhere left to carry "newest id seen"). It uses an overlapping window — `now - 600s` on a 5-minute poll — plus an id-diff against `prev`. A report stamped at T that commits at T+90s is still inside the next poll's window, so it is picked up, and the id-diff stops it being reported twice. That is exactly why the window is 2x the interval rather than 1x, and the description now says so in those terms rather than leaving the reader to infer it.

**Declined for this PR:** `clock_timestamp()` for `created_at` and keyset pagination on id. Both are the right direction, and both change table-wide semantics for every reader of `bug_reports` — well past a PR that adds a filter and a notification. Worth their own issue if the watch pattern gets real use.

packages/console-backend/console_backend/routers/bug_report_router.py
aartaria: Accepted, applied as suggested — `Query(1, ge=1)` / `Query(50, ge=1, le=100)`, dead clamp removed, `Query` imported.

Confirming the behaviour change you flagged, since it is client-visible: `limit>100` now 422s instead of silently clamping. The only caller of this endpoint is the console's bug-reports page, which sends `page>=1, limit=20` and is admin-gated, so nothing in-tree notices. Taking the 422 rather than keeping the clamp was deliberate: it is the same contract as the MCP twin, which is half of what the duplication finding below is about.

There is a test covering `page=0`, `page=-1`, `limit=0`, `limit=-1` and `limit=101` against both surfaces.

packages/console-backend/console_backend/routers/bug_report_mcp_tools.py
aartaria: Accepted, and done the way you suggested: `BugReportService.list_for_user(db, user, status, created_after, page, limit) -> BugReportListResponse`. The visibility rule and the response assembly now live there once, and both routes are the wrapper you described — the pass-through `list_bug_reports` service method had no other callers, so it was replaced rather than left alongside.

The validation divergence is gone too (see the reply above): both surfaces declare `ge=1` / `le=100`. What stays per-surface is only what has to — the auth dependency, the path, and the OpenAPI metadata the model reads.

packages/console-backend/console_backend/repositories/bug_report_repository.py
aartaria: Accepted — `CREATE INDEX IF NOT EXISTS idx_bug_reports_created_at ON bug_reports (created_at DESC)` added to migration 086, with the down migration dropping it (the enum value above it stays irreversible, as Postgres gives no way to remove one).

Confirmed your reading of the index situation: migration 046 leaves only `(conversation_id)` and `(user_id, created_at DESC)`, and the latter cannot serve a range scan on `created_at` with `user_id` unset — which is exactly the administrator poll this PR advertises.

**Declined:** skipping the `COUNT(*)` or replacing it with `LIMIT limit+1` on the polling path. `meta.total` is now load-bearing for a watcher — the description tells it to page until the total is exhausted — so the count has a consumer rather than being dead weight, and it is the same query shape the console's paginated table needs.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (ffbb44f2c9e0944087a5bc4f83bf54ec47da0987..da84a1dcbbff8a1afcf85f5a1165b4499d9d9e73)

1. **Audience vs. read visibility** — **verified-fixed.** The recipient query is now `is_administrator IS TRUE` only, matching every read path; the underlying `triage`-capability inconsistency is tracked in #197. The fix also removed the now-dead `system`-user exclusion correctly: the seed has `is_administrator` FALSE, and the test now pins that database fact.
2. **Timezone-naive `created_after`** — **verified-fixed** in the repository (`replace(tzinfo=timezone.utc)` when naive), covering both routes with one normalisation. One caveat on its new test — see the inline comment.
3. **Scheduled-watch report loss** — **partly fixed, remainder rejected-by-author with sound reasoning.** The burst arm is addressed (tool description now instructs paging on `meta.total`; `id DESC` tiebreaker added). The commit-time-visibility arm is declined against the documented pattern, and the reasoning holds: the documented watch uses an overlapping window (2× the poll interval) plus an id-diff against `prev`, not a max-`created_at` cursor, so a slow-committing report is still inside the next window. `clock_timestamp()` / keyset pagination declined as table-wide semantics beyond this PR — reasonable; not re-raised.
4. **REST `page=0` → 500** — **verified-fixed.** Both surfaces declare `Query(1, ge=1)` / `Query(50, ge=1, le=100)`, the dead clamp is gone, and `test_list_rejects_out_of_range_paging` covers both. The acknowledged behaviour change (`limit>100` now 422s instead of clamping) is documented and its only in-tree caller sends `limit=20`.
5. **Duplicated list handler** — **verified-fixed.** Visibility and response assembly live once in `BugReportService.list_for_user`; both routes are thin wrappers, and the pass-through service method was replaced rather than left alongside.
6. **Missing `created_at` index** — **verified-fixed.** `idx_bug_reports_created_at (created_at DESC)` added to migration 086 with a correct down migration. Keeping the `COUNT(*)` is a reasoned decline (`meta.total` is now load-bearing for the paging instruction) — accepted.

**Delta review:** one new minor finding, posted inline — the new timezone regression test is vacuous on a UTC runner and cannot guard the fix it was written for. Checked and clean: mixing `ALTER TYPE ... ADD VALUE` with `CREATE INDEX` in migration 086 is safe (the index does not use the new enum value in the same transaction, and migrations 016/026/033 set the precedent); the `id DESC` tiebreaker is stable (though not covered by the new single-column index — an `ORDER BY` on ties still sorts in memory, negligible at page size); the regenerated SDK carries only the updated tool description.

packages/console-backend/tests/test_bug_report_router.py
```diff
@@ -260,3 +260,299 @@ async def test_update_status_audit_log(
...
+
+    # A window starting after the newest report returns nothing — the shape a watch
+    # sees on a quiet poll.
+    response = await client_with_db.get("/api/v1/bug-reports", params={"created_after": newer["created_at"]})
+    assert response.status_code == 200
+    assert response.json()["meta"]["total"] == 0
+
+
+@pytest.mark.asyncio
+async def test_created_after_without_an_offset_is_read_as_utc(
```
alloy-cicd, line 300: 🤖 **This test cannot fail on a UTC machine, so it does not guard the normalisation it was written for.**

Without the fix, asyncpg encodes a naive datetime via `astimezone(utc)`, which assumes the *process's local timezone*. On a runner with `TZ=UTC` (the GitHub Actions default, and most dev containers), the unnormalised path already encodes the naive boundary as UTC — so deleting the `replace(tzinfo=timezone.utc)` in `bug_report_repository.py` leaves this test green. The reply on the round-1 thread says a regression would "show up on a UTC dev box"; it would not — it still only shows on non-UTC deployments.

To make it bite, pin a non-UTC process timezone for the test's duration, e.g.:

```python
monkeypatch.setenv("TZ", "Pacific/Kiritimati")  # UTC+14, no DST
time.tzset()
```

(restore + `tzset()` afterwards — `monkeypatch` handles the env var, a `finally` the `tzset`). With that, the naive-boundary assertion fails by 14 hours if the normalisation is ever removed. Alternatively, a narrow unit test asserting the repository binds a tz-aware value would do the same job without touching process state.

aartaria: You are right, and my reply on the round-1 thread was wrong — I claimed the test would catch a regression on a UTC dev box, which is exactly the case where it cannot. With `TZ=UTC`, "read the naive value as local" and "read it as UTC" are the same operation, so the assertion holds with or without the normalisation.

Fixed as you suggested: the test now pins `TZ=Pacific/Kiritimati` (UTC+14, no DST) via `monkeypatch.setenv` + `time.tzset()`, restoring both in a `finally`. At +14 the two readings are 14 hours apart, which is far more than the gap between the two reports, so the unfixed code pulls in the report the boundary was meant to exclude.

And this time I checked the property instead of asserting it: with the two normalisation lines deleted, the test fails —

```
>       assert [r["id"] for r in response.json()["data"]] == [newer["id"]]
E       AssertionError: assert [older_id] == [newer_id]
```

then passes again with them restored. Full suite re-run to confirm the `TZ` pinning does not leak into anything downstream.

packages/console-backend/console_backend/repositories/bug_report_repository.py
aartaria: Correction to my reply above: the claim that the new test "shows up on a UTC dev box" was wrong, and the round-2 review caught it. On `TZ=UTC` the naive-as-local and naive-as-UTC readings are identical, so the test passed with or without the fix — it verified nothing. It now pins `TZ=Pacific/Kiritimati` for its duration, and I confirmed by deleting the normalisation and watching it fail. Details on the round-2 thread.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (da84a1dcbbff8a1afcf85f5a1165b4499d9d9e73..17fb77952bd8522febdc3c4097225392295c7527)

1. **Timezone test vacuous on a UTC runner** — **verified-fixed.** The test now pins `TZ=Pacific/Kiritimati` (UTC+14, no DST) via `monkeypatch.setenv` + `time.tzset()`, restoring both in a `finally`, exactly as suggested. The author verified the property empirically: with the normalisation deleted the test fails (`[older_id] == [newer_id]`), and passes with it restored. The mechanics check out on inspection too — the TZ pin happens after the reports are filed and the boundary computed (report timestamps come from Postgres, not the process clock), so only the naive-bind path under test is affected, and the dropped secondary assertion (the one-hour-earlier window) was the non-discriminating one.

All six round-1 findings remain in their round-2 dispositions (5 verified-fixed; the scheduled-watch finding partly fixed with the remainder reasonably declined).

**Delta review:** every previous finding is resolved and the delta (one test file) introduces nothing new. No inline comments this round.

packages/console-backend/console_backend/services/bug_report_service.py
aartaria: Correction, from testing against a real deployment rather than the migration.

I removed the explicit `system` exclusion in `da84a1dc` as dead code, reasoning that migration 041 leaves `is_administrator` FALSE on that row so an audience filtering on the flag could never reach it. Round 2 verified that removal. **We were both wrong.** A live console database has `system` promoted to `is_administrator = TRUE`, and a bug report filed through the client UI produced this:

```
user_id                                | type
system                                 | bug_report_filed   ←
fc9670a5-…                             | bug_report_filed
3281e3c9-… (service-account-orchestrator) | bug_report_filed
704019ec-… (test@local.dev)            | bug_report_filed
```

The seed default is not what deployments contain — nothing in the code promotes it (the only such path is `FIRST_USER_IS_ADMIN`, first user only), so an operator did, and any deployment may have done the same. The exclusion is restored, and its test now promotes `system` before filing, so it asserts the exclusion rather than the seed value that happened to make it redundant.

The general lesson for this file: filter on the flag *and* exclude the account. One is reasoning from the migration, the other from the data.

packages/console-backend/AGENTS.md
```diff
@@ -272,6 +272,20 @@ from console_backend.service_instances import sub_agent_service
...
 
+### Notification Audiences Beyond Group Membership
+Two things to get right when an audience is chosen by standing rather than by group
+membership (`BugReportService._notify_administrators` is the worked example):
+
+- **`role = 'admin'` is not the same as `is_administrator`.** Migration 041 seeds a
+  `system` user with `role = 'admin'` — it owns auto-provisioned agents — while leaving
+  `is_administrator` FALSE. Selecting on the flag leaves it out; selecting on the role
+  picks up a service account with no person behind it, and fills an inbox nobody opens.
+  An audience that must select by role has to exclude it explicitly.
```
alloy-cicd: 🤖 **This bullet still teaches the reasoning the new commit disproves.**

It says "Selecting on the flag leaves it out" and only an audience "that must select by role has to exclude it explicitly" — but `3548f15a`'s own rationale is that the flag is not reliable (deployments promote `system`, and a promoted one was collecting notifications), so the exclusion is required *regardless of how the audience is selected*. The service docstring was updated to say exactly that; this doc, which exists to teach the next audience author, now contradicts both the code and the incident that motivated it. Suggest rewording the bullet to the commit's own lesson: filter on the flag *and* exclude the account — one is reasoning from the migration, the other from the data.

packages/console-backend/console_backend/services/bug_report_service.py
```diff
@@ -45,30 +61,143 @@ async def create_bug_report(
...
+
+        try:
+            recipients = await db.execute(
+                text("""
+                    SELECT id FROM users
+                    WHERE is_administrator IS TRUE
+                      AND status = 'active'
+                      AND deleted_at IS NULL
+                      AND id != :actor_id
+                      AND id != :system_user_id
```
alloy-cicd: 🤖 **The evidence that motivated this fix shows a second service account the single-id exclusion misses — worth confirming it is an intended recipient.**

The notification dump in the thread lists `service-account-orchestrator` (3281e3c9-…) among the `bug_report_filed` recipients. If that account is what its name says — a machine identity with `is_administrator = TRUE` and no person reading its inbox — it has exactly the problem `system` had, and excluding by the one hardcoded id fixes one instance of the class rather than the class. Non-blocking either way: if that account's inbox is actually consumed (e.g. surfaced through the orchestrator), a word in the docstring saying so would stop the next reader re-flagging it; if it is not, the durable fix is a structural marker (an `is_service_account` column or a naming convention the query can exclude) — follow-up material, not this PR.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (17fb77952bd8522febdc3c4097225392295c7527..3548f15a65d507c4c22a0addd957453e432f5c58)

**Ledger:** all findings from rounds 1–3 keep their dispositions (verified-fixed, or partly-fixed-with-reasoned-decline for the scheduled-watch remainder). No previous finding is re-opened by this delta — with one correction to the record: round 2 verified the removal of the `system` exclusion as dead code, and the author's live-deployment evidence now shows that verification was wrong in practice. Migration 041's seed (`is_administrator` FALSE) is not what deployments contain — a promoted `system` account was collecting notifications. The restored explicit exclusion (`id != 'system'`) is the right call, and the reworked test (promote first, then assert exclusion) now pins the exclusion itself rather than the seed value. **Verified-fixed, correctly this time.**

**Rebase check:** the branch was rebased onto main, which took migration number 086 for #196's delivery-channel column; the PR's migration is correctly renumbered to 087 with content unchanged, and no other PR file was altered by the rebase.

**Delta review:** two minor inline comments, neither blocking — the AGENTS.md bullet this PR added still states the flag-based reasoning the new commit disproves, and the author's own evidence table shows a second service-account-shaped recipient (`service-account-orchestrator`) the single-id exclusion does not cover, worth confirming as intended.

packages/console-backend/AGENTS.md
aartaria: Accepted — the doc that exists to teach the lesson was still teaching the disproved version. Rewritten so the guidance is the one the evidence supports: exclude machine accounts explicitly, and do not infer that a filter already excludes them. It now names why the inference failed — the seed's column values are a starting state, not an invariant, and nothing in the code promotes that account beyond `FIRST_USER_IS_ADMIN` for the very first user — and records that there is no structural marker yet, so exclusion is by id.

packages/console-backend/console_backend/services/bug_report_service.py
aartaria: Confirmed, and deferred rather than patched — with reasoning, since I have now been wrong twice in this area by reasoning ahead of the data.

`service-account-orchestrator` is a real row: console-backend creates it on first contact from the orchestrator's own token, so it exists in every deployment. In the database I looked at it is flagged `is_administrator` and did receive a notification, alongside a second account with no email address. So the class is real, exactly as you describe.

What I am not doing is extending the hardcoded exclusion to it, or matching on a `service-account-` prefix. That trades a known-correct exclusion for a naming convention nobody has committed to, and it would not cover the no-email account either. The class fix is a structural marker — an `is_service_account` column, or an account-type enum — set at the point these rows are created, which is a migration plus a change to user upsert and belongs nowhere near a PR about bug-report notifications.

Two things make deferring safe rather than convenient: those accounts are only in the audience because an operator promoted them (nothing in the code does), and the blast radius is an unread inbox row rather than a leak — the notification carries a description snippet, and these accounts have no one reading them. Filing the follow-up; the `AGENTS.md` note now records that the marker does not exist yet, so the next audience author is not left to rediscover this.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (3548f15a65d507c4c22a0addd957453e432f5c58..f544ad3a95a2913b55120ba69fac87c7d68b8cb4)

**Ledger of round-4 findings:**

1. **Stale AGENTS.md bullet** — **verified-fixed** (`2d0971ab`, then superseded in `f544ad3a` by the flag-based guidance). The doc now teaches the lesson the evidence supports and points at `is_service_account` instead of an id list.
2. **Second service-account recipient** — **verified-fixed structurally** (`f544ad3a`): `users.is_service_account` (migration 088) marks the category, `token_is_service_account` flags new rows at creation from Keycloak's `preferred_username` convention (with a well-argued refusal to treat sparse claims as a machine signal), `_notify_administrators` filters on the flag, and the reworked test promotes both machine rows before filing and asserts only the human administrator is notified. One caveat on the backfill's coverage of *existing* rows — see the inline comment.

All findings from rounds 1–3 keep their dispositions.

**Delta review:** one inline comment (backfill coverage, minor). Checked and clean: the route-order swap in `app.py` is safe — every MCP route's distinguishing tail segment is a literal (`mcp-list`, `mcp-create`, `mcp-status`, `mcp-external-link`), so registering the MCP router first cannot capture any REST-bound request, and the new test asserts both directions; the path rename `/mcp/list` → `/mcp-list` breaks no external consumer since the route only exists on this branch; the `upsert` column/value lists align and the flag is correctly insert-only; migration 088's down migration is complete; the regenerated SDK churn is mechanical. One wording nit, not worth a thread: the `app.py` comment says "every MCP path here is a literal segment" — the item-level tools start with `{report_id}`; the safety argument actually rests on the literal tails, as the module docstring correctly states.

packages/console-backend/sqlmigrations/ddl/088_users_is_service_account.sql
```diff
@@ -0,0 +1,21 @@
...
+UPDATE users SET is_service_account = TRUE WHERE id = 'system';
+
+-- Keycloak names a client's service account 'service-account-<clientId>', and console-backend
+-- auto-onboards one the first time a service calls it with a client-credentials token. Matching
+-- the convention backfills the rows that already exist; new ones are flagged at creation from
+-- the token's own claims, so this pattern is not load-bearing going forward.
+UPDATE users
+   SET is_service_account = TRUE
+ WHERE is_service_account IS FALSE
+   AND (email LIKE 'service-account-%' OR email LIKE '%@service-account.%');
```
alloy-cicd, line 18: 🤖 **The backfill may miss the very row that motivated this fix — worth verifying against the live database before this ships.**

Existing machine rows are caught only by *email* patterns, but the auto-onboarding path fills `email` from `payload.get("email", "")`, and a Keycloak client-credentials token typically carries no email claim — so an already-onboarded service account can sit here with an empty email, match neither pattern, and stay unflagged. Since the flag is insert-only (deliberately, and the upsert never re-touches it), such a row is never corrected by future calls either: the notification bug persists on exactly the deployment that surfaced it, unless its orchestrator row happens to have a convention-shaped email.

Two cheap options, either sufficient: (a) confirm on the live database that the orchestrator row's email actually matches one of these patterns, and flag it manually if not (the docstring already blesses operator correction); or (b) let the token flip the flag one-directionally on upsert — set TRUE when `token_is_service_account(payload)` says so, never set FALSE — which self-heals any pre-existing row on its next call while still protecting an operator's corrections in the direction that matters (a wrongly-flagged person losing notifications).

aartaria: Accepted, and fixed with your second suggestion: the flag is now monotonic on upsert rather than insert-only —

```sql
is_service_account = users.is_service_account OR EXCLUDED.is_service_account,
```

so a machine account the backfill missed corrects itself on its next call, and a token lacking the claim can never re-classify a person as a machine.

Your reasoning holds even though it happens to be moot on the deployment that surfaced this: I checked the live rows, and that orchestrator account carries `service-account-orchestrator@…` as its email, so the pattern did catch it (`is_service_account = t`, and its notifications stopped). But the email is not guaranteed — a client-credentials token need not carry the claim at all — and "the row exists but nothing will ever flag it" is precisely the shape of the bug this column was added to end. Insert-only made the fix depend on a convention holding in a table the migration cannot see.

Two tests now: one that a later call claiming `False` does not clear it, one that a row created unflagged is raised by a later call that recognises it. The first was passing vacuously at first — my edit had left `@pytest.mark.asyncio` on the helper rather than the test, so it never ran; pytest-asyncio strict mode caught it.

The `app.py` nit is fixed too. You are right that "every MCP path is a literal segment" is wrong for the item-level tools; the accurate claim is that each MCP path is distinguished by a literal the REST paths lack — `mcp-create`/`mcp-list` at the collection level, and an `mcp-`-prefixed tail after `{report_id}`, so `/{report_id}/mcp-status` cannot capture `/{report_id}/status`.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 6 (f544ad3a95a2913b55120ba69fac87c7d68b8cb4..0aa5c1134ecd3666415355b85fa3aaab786fd99a)

**Ledger of round-5 findings:**

1. **Backfill coverage of pre-existing machine rows** — **verified-fixed.** The flag is now monotonic on upsert (`is_service_account = users.is_service_account OR EXCLUDED.is_service_account`), exactly the suggested one-directional semantics: a machine account the email-pattern backfill missed corrects itself on its next call, and a token lacking the claim can never re-classify a person. The author also verified the live database directly — the orchestrator row carried a convention-shaped email, so the backfill had caught it and its notifications stopped — making this a durability fix rather than a live gap. Two tests cover both directions (never cleared; raised on an existing row), and the author caught and fixed a vacuously-passing variant of the first (a misplaced `@pytest.mark.asyncio`) before it landed.
2. **`app.py` comment overstating "every MCP path is a literal segment"** — **verified-fixed.** The comment now states the accurate invariant: each MCP path is distinguished by a literal segment the REST paths lack, so `/{report_id}/mcp-status` cannot capture `/{report_id}/status`.

All findings from rounds 1–5 keep their dispositions.

**Delta review:** every previous finding is resolved and the delta (three files: the SQL upsert, its docstring, the tests, and the comment) introduces nothing new. The `OR` is null-safe (the column is `NOT NULL DEFAULT FALSE` on both sides), the upsert conflict path exercises it correctly, and the documented trade-off (an operator clearing the flag on a row whose token still claims service-account sees it return) is stated where the next reader will look. No inline comments this round.

PR #196, opened 2026-09-02
fix(scheduler): format scheduled notifications for the delivering channel

### LINKED ISSUE #195: Scheduled notifications are not formatted for the channel that delivers them
## Problem

A scheduled job that notifies a chat client delivers text the client cannot render. Slack shows the raw source — `### Heading`, `**bold**`, `| tables |`, `[label](url)` — instead of formatted output, and Google Chat has the same problem with its own markup.

The formatting machinery already exists, but only for interactive turns: a client sends `messageFormatting` in its A2A request metadata, and `UserPreferencesMiddleware` turns that into a `<message_formatting format="slack">` block in the system prompt. A scheduled run has no client on the other end to send it, so it defaults to `markdown` — and nothing between the agent and the delivery channel rewrites the output.

The current workaround is to write "format the message for Slack" into the job prompt or the sub-agent's system prompt by hand. That is wrong twice: it has to be remembered per job, and a sub-agent's system prompt is written once and reused, while the same agent may notify Slack for one job and the web console for the next.

## Expected

Correct formatting per delivery channel is built in. The channel declares how it renders text; every writer in the scheduled path is told those rules without a job author asking.

## Notes

- The renderer is a property of the **delivery channel**, not of the job or the agent, so the format belongs on `delivery_channels` and clients can declare it when they self-register.
- The rules are currently spelled out inline in `UserPreferencesMiddleware`; a second copy in the scheduled path would drift. They belong in one shared place.
- Three writers are involved: a local sub-agent (system prompt), a remote sub-agent (owns its own system prompt, so the rules have to ride the message), and the scheduler's own one-shot notification writer for a triggered watch.

aartaria:
closes #195

Scheduled notifications arrived on Slack as raw Markdown. The formatting machinery already existed — a client sends `messageFormatting` in its request metadata and `UserPreferencesMiddleware` turns it into a `<message_formatting format="slack">` block — but only an interactive turn sends it. A cron-started run has no client on the other end, so it fell back to Markdown, and nothing between the agent and the channel rewrites output.

The renderer is a property of the **delivery channel**, so that is where the format now lives, and every writer in the scheduled path is told the rules without a job author asking.

### What changed

- **`agent_common.core.message_formatting`** (new): one table of per-channel rules plus `normalize_message_formatting` / `formatting_prompt_block`. `UserPreferencesMiddleware` reads from it instead of spelling the rules out inline — a second copy in the scheduled path would have drifted from the interactive one. The shared table also closes gaps the old Slack text missed: `<url|label>` links, bullets, tables.
- **Migration 086**: `delivery_channels.message_formatting`, `NOT NULL DEFAULT 'markdown'`, CHECK-constrained to the four known formats, with the rows the two chat clients already registered backfilled. The Slack and Google Chat clients declare their format when they self-register, so a new installation is correct without operator action.
- **`scheduler_engine._build_message_args`**: fetches the channel once and puts its format in the dispatch metadata under the same key an interactive client uses. (The same fetch still yields `push_config`, so this adds no query.)
- **agent-runner**: reads it off the message and appends the rules to a local sub-agent's system prompt — a sub-agent's prompt is written once and reused, while the same agent may notify Slack for one job and the web console for the next, so the channel cannot be baked into it. A remote sub-agent owns its own system prompt and `SubAgentInput` carries no per-message metadata, so there the rules ride the message as a trailing instruction. Foundry sub-agents are untouched: their query API decides its own output.
- **`_write_notification`** (the scheduler's own one-shot writer, used when a triggered watch's author left the message empty) now asks for markup-free prose. It is one or two sentences that go out verbatim on whichever channel the job notifies, and plain text renders correctly everywhere.

### Verification

- New tests: 7 in agent-common (normalisation incl. protobuf-Struct-shaped junk, per-format blocks, Markdown adding nothing), 6 in agent-runner (metadata hand-off, remote message carrier, the local injection point), 2 in console-backend (a Slack-bound job dispatches with the channel's rules and still gets its push target from the same single fetch).
- Suites green locally: console-backend 1665 passed, agent-runner 43, orchestrator `test_user_preferences_middleware` 29 + `test_runtime`/thinking 25, agent-common formatting 7. Both TS clients typecheck clean.
- `ruff check` passes on every touched file. Two touched files fail `ruff format --check`, both identically on `main` — pre-existing, not introduced here.

### Follow-ups (not in this PR)

- `console-frontend`'s generated API types are stale for the new response field (`npm run gen-sdk` needs a running backend), and `DeliveryChannelsPage` has no Format column — worth adding if operators should be able to override what a bot declares.

## Checklist

### DIFF
--- a/packages/agent-common/agent_common/core/message_formatting.py
+++ b/packages/agent-common/agent_common/core/message_formatting.py
@@ -1,0 +1,78 @@
+"""Channel-dependent message formatting instructions.
+
+The text an agent writes is rendered by whatever client delivers it, and those
+renderers disagree: Slack reads mrkdwn (``*bold*``, no headers), Google Chat its
+own markup, a web console real Markdown. So the format is a property of the
+DELIVERY CHANNEL, not of the agent — and the agent has to be told which one it
+is writing for, because nothing downstream rewrites its output.
+
+Interactive turns get this from the client's ``messageFormatting`` request
+metadata. Scheduled runs have no client on the other end: the format comes from
+the delivery channel the job notifies, is carried in the dispatch metadata under
+the same key, and lands here. One table so a channel's rules are stated once and
+every writer (orchestrator turn, scheduled sub-agent, scheduler-written
+notification) obeys the same ones.
+"""
+
+from __future__ import annotations
+
+from typing import Literal
+
+MessageFormatting = Literal["markdown", "slack", "google-chat", "plain"]
+
+DEFAULT_MESSAGE_FORMATTING: MessageFormatting = "markdown"
+
+#: Rendering rules per channel, as prose for the model. ``markdown`` is absent on
+#: purpose: it is standard behaviour and an instruction saying so only spends tokens.
+FORMATTING_RULES: dict[str, str] = {
+    "slack": (
+        "Format responses using Slack mrkdwn syntax: *bold* for emphasis, _italic_ for secondary "
+        "emphasis, `code` for inline code, ```code blocks``` for multi-line code. "
+        "Use <https://url|label> for links and a leading '• ' for bullets. "
+        "Avoid markdown syntax that Slack doesn't support (e.g. # headers, **bold**, "
+        "[links](url), | tables |)."
+    ),
+    "google-chat": (
+        "Format responses using Google Chat markup syntax: *bold* for emphasis, _italic_ for secondary "
+        "emphasis, ~strikethrough~ for strikethrough, `code` for inline code, ```code blocks``` for "
+        "multi-line code. Use plain URLs for links (they are auto-linked). "
+        "Avoid markdown syntax that Google Chat doesn't support (e.g. # headers, **bold**, "
+        "[links](url), | tables |)."
+    ),
+    "plain": (
+        "Use plain text only. Do not use any formatting syntax "
+        "(no markdown, no bold, no code blocks). Keep responses simple and readable."
+    ),
+}
+
+
+def normalize_message_formatting(value: object) -> MessageFormatting:
+    """Coerce a metadata value to a known format, defaulting to ``markdown``.
+
+    Metadata reaches us from A2A clients and, over gRPC, through a protobuf Struct, so
+    an unexpected type or a channel format written by an older client must degrade to
+    the default rather than raise inside a scheduled run.
+    """
+    if isinstance(value, str):
+        candidate = value.strip().lower()
+        if candidate in FORMATTING_RULES or candidate == DEFAULT_MESSAGE_FORMATTING:
+            return candidate  # type: ignore[return-value]
+    return DEFAULT_MESSAGE_FORMATTING
+
+
+def formatting_rules(value: object) -> str:
+    """The bare rules sentence for a format, or ``""`` for plain Markdown."""
+    return FORMATTING_RULES.get(normalize_message_formatting(value), "")
+
+
+def formatting_prompt_block(value: object) -> str:
+    """The rules as a ``<message_formatting>`` block, or ``""`` when none are needed.
+
+    Returned without surrounding blank lines; callers place it (system-prompt
+    addendum, user-preferences block, one-shot prompt) as they need.
+    """
+    rules = formatting_rules(value)
+    if not rules:
+        return ""
+    fmt = normalize_message_formatting(value)
+    return f'<message_formatting format="{fmt}">\n{rules}\n</message_formatting>'
--- a/packages/agent-common/agent_common/core/message_formatting.py
+++ b/packages/agent-common/agent_common/core/message_formatting.py
@@ -16,11 +16,13 @@ notification) obeys the same ones.
 
 from __future__ import annotations
 
-from typing import Literal
+from typing import Literal, get_args
 
 MessageFormatting = Literal["markdown", "slack", "google-chat", "plain"]
 
 DEFAULT_MESSAGE_FORMATTING: MessageFormatting = "markdown"
+
+KNOWN_FORMATS = frozenset(get_args(MessageFormatting))
 
 #: Rendering rules per channel, as prose for the model. ``markdown`` is absent on
 #: purpose: it is standard behaviour and an instruction saying so only spends tokens.
@@ -55,7 +57,7 @@ def normalize_message_formatting(value: object) -> MessageFormatting:
     """
     if isinstance(value, str):
         candidate = value.strip().lower()
-        if candidate in FORMATTING_RULES or candidate == DEFAULT_MESSAGE_FORMATTING:
+        if candidate in KNOWN_FORMATS:
             return candidate  # type: ignore[return-value]
     return DEFAULT_MESSAGE_FORMATTING
 
@@ -71,8 +73,8 @@ def formatting_prompt_block(value: object) -> str:
     Returned without surrounding blank lines; callers place it (system-prompt
     addendum, user-preferences block, one-shot prompt) as they need.
     """
-    rules = formatting_rules(value)
+    fmt = normalize_message_formatting(value)
+    rules = FORMATTING_RULES.get(fmt)
     if not rules:
         return ""
-    fmt = normalize_message_formatting(value)
     return f'<message_formatting format="{fmt}">\n{rules}\n</message_formatting>'
--- a/packages/agent-runner/agent/core.py
+++ b/packages/agent-runner/agent/core.py
@@ -42,7 +42,11 @@ from agent_common.a2a.structured_response import A2A_PROTOCOL_ADDENDUM, SubAgent
 from agent_common.agents.foundry_agent import create_foundry_local_subagent
 from agent_common.core.document_store_tools import create_document_store_tools
 from agent_common.core.graph_utils import build_sub_agent_graph
-from agent_common.core.message_formatting import formatting_prompt_block, normalize_message_formatting
+from agent_common.core.message_formatting import (
+    formatting_prompt_block,
+    formatting_rules,
+    normalize_message_formatting,
+)
 from agent_common.core.model_factory import (
     create_model,
     get_default_model,
@@ -188,6 +192,21 @@ def _current_time_context(timezone_name: str | None) -> str:
         except Exception:
             pass
     return line
+
+
+def _build_sub_agent_system_prompt(system_prompt: str, message_formatting: str) -> str:
+    """The sub-agent's own prompt, plus the response protocol and the channel's rules.
+
+    The rendering rules cannot live in the stored system prompt: it is written once and
+    reused across every job, while the same agent may notify Slack for one job and the
+    web console for the next. Assembling them per run is what makes correct formatting
+    built in rather than something each job's author has to remember to ask for.
+    """
+    parts = [system_prompt, A2A_PROTOCOL_ADDENDUM]
+    formatting_block = formatting_prompt_block(message_formatting)
+    if formatting_block:
+        parts.append(formatting_block)
+    return "\n\n".join(parts)
 
 
 def _extract_message_metadata(task: Task) -> dict[str, Any]:
@@ -896,6 +915,7 @@ class AgentRunner(BaseAgent):
             return await self._run_foundry_agent(
                 sub_agent_cfg=sub_agent_cfg,
                 prompt=prompt,
+                message_formatting=message_formatting,
                 user_config=user_config,
                 scheduled_job_id=scheduled_job_id,
                 scheduled_job_run_id=scheduled_job_run_id,
@@ -978,17 +998,7 @@ class AgentRunner(BaseAgent):
 
         llm = create_model(model_name, thinking_level=thinking_level)
 
-        # Append the A2A response protocol addendum so the LLM knows to use SubAgentResponseSchema
-        full_system_prompt = system_prompt + "\n\n" + A2A_PROTOCOL_ADDENDUM
-
-        # ...and the delivery channel's rendering rules. A sub-agent's system prompt is
-        # written once and reused across every job, so the channel cannot be baked into
-        # it: the same agent may notify Slack for one job and the web console for the
-        # next. Injecting it here is what makes correct formatting built in rather than
-        # something each job's author has to remember to ask for.
-        formatting_block = formatting_prompt_block(message_formatting)
-        if formatting_block:
-            full_system_prompt += "\n\n" + formatting_block
+        full_system_prompt = _build_sub_agent_system_prompt(system_prompt, message_formatting)
 
         mcp_timeout = timedelta(seconds=_MCP_TIMEOUT_SECONDS)
 
@@ -1200,6 +1210,7 @@ class AgentRunner(BaseAgent):
         user_config: UserConfig,
         scheduled_job_id: int,
         scheduled_job_run_id: int,
+        message_formatting: str = "markdown",
     ) -> tuple[str | None, str | None]:
         """Run a Foundry query-API agent using agent-common's foundry module.
 
@@ -1209,6 +1220,10 @@ class AgentRunner(BaseAgent):
             user_config: Authenticated user context.
             scheduled_job_id: For logging.
             scheduled_job_run_id: For tracking the conversation.
+            message_formatting: Rendering rules of the delivery channel. The query API
+                takes a single `userInput` string and no system prompt, so they can only
+                go into the prompt; whether the Foundry-side agent honours them is its
+                own business, but a run that is never told cannot get it right.
         Returns:
             (result_summary, task_state) — see _collect_stream_text.
         """
@@ -1239,9 +1254,12 @@ class AgentRunner(BaseAgent):
             sub_agent_config_version_id=sub_agent_cfg.get("sub_agent_config_version_id"),
         )
 
+        formatting_block = formatting_prompt_block(message_formatting)
+        foundry_prompt = f"{prompt}\n\n{formatting_block}" if formatting_block else prompt
+
         # Stream the foundry runnable via the A2A SubAgentInput interface
         input_data = SubAgentInput(
-            messages=[{"role": "user", "content": prompt}],
+            messages=[{"role": "user", "content": foundry_prompt}],
         )
         result_summary, task_state = await _collect_stream_text(compiled_subagent["runnable"], input_data)
 
@@ -1304,9 +1322,8 @@ class AgentRunner(BaseAgent):
                 (scheduled_job_runs.conversation_id) — the prerequisite for a
                 later orchestrator delegation to resume that conversation via
                 the conversation-origin extension.
-            message_formatting: Rendering rules of the delivery channel. A remote
-                agent owns its own system prompt, so the rules ride the message as a
-                trailing instruction — the only place this side can put them.
+            message_formatting: Rendering rules of the delivery channel, forwarded as
+                A2A message metadata so the remote agent applies them itself.
 
         Returns:
             (result_summary, task_state) — see _collect_stream_text.
@@ -1351,22 +1368,24 @@ class AgentRunner(BaseAgent):
             # Fallback to plain text prompt
             messages_input = [{"role": "user", "content": prompt}]
 
-        # The channel's rendering rules as a final instruction. There is no system prompt
-        # to append to on this path (the remote agent has its own) and SubAgentInput
-        # carries no per-message metadata, so the message itself is the only carrier.
-        formatting_block = formatting_prompt_block(message_formatting)
-        if formatting_block:
-            messages_input = [*messages_input, HumanMessage(content=formatting_block)]
-
         # orchestrator_conversation_id feeds A2AClientRunnable's contextId
         # waterfall (_extract_tracking_ids), putting the run task's contextId on
         # the wire. The remote keys its checkpoints by the contextId it
         # receives, so the run's stored conversation_id then names a real,
         # resumable conversation on the executing side.
+        # The channel's rendering rules travel as message metadata, not as an extra
+        # message: a remote agent owns its system prompt, and A2AClientRunnable puts
+        # `messageFormatting` on the wire under the key an interactive client uses, so the
+        # remote applies them through its own request-metadata path. Appending an
+        # instruction message instead would land in the remote's checkpointed
+        # conversation, where a later turn can read it as part of the task.
         input_data = SubAgentInput(
             messages=messages_input,
             scheduled_job_id=scheduled_job_id,
             orchestrator_conversation_id=context_id,
+            # Only when there is something to say: plain Markdown is the remote's default
+            # too, so sending it would put a no-op instruction on the wire.
+            message_formatting=message_formatting if formatting_rules(message_formatting) else None,
         )
         result_summary, task_state = await _collect_stream_text(runnable, input_data)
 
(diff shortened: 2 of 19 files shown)


packages/agent-runner/agent/core.py
```diff
@@ -876,6 +890,7 @@ async def _execute_sub_agent(
                 scheduled_job_id=scheduled_job_id,
                 scheduled_job_run_id=scheduled_job_run_id,
                 context_id=context_id,
+                message_formatting=message_formatting,
             )
         elif agent_type == "foundry":
             return await self._run_foundry_agent(
```
alloy-cicd, line 922: **Foundry sub-agents never receive the channel's formatting.** `_execute_sub_agent` forwards `message_formatting` to `_run_langgraph_agent` (line 893) and `_run_remote_agent` (line 912), but this `foundry` branch omits it and `_run_foundry_agent` has no such parameter — its `SubAgentInput` is built from the bare prompt with nothing appended.

Concretely: a scheduled job whose sub-agent is type `foundry` with a Slack delivery channel still delivers `### heading` / `**bold**` — the exact bug this PR fixes — silently, and no test covers this branch. Verification for a fix: `_run_foundry_agent` accepts the kwarg and the block reaches its prompt/messages.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -268,6 +273,7 @@ async def upsert_channel_by_installation(
             description=data.description,
             webhook_url=data.webhook_url,
             secret=data.secret,
+            message_formatting=data.message_formatting,
```
alloy-cicd, line 277: **A client that omits the field resets the channel to `markdown` on re-registration.** `DeliveryChannelCreate.message_formatting` defaults to `"markdown"` (not `None`), so this line always carries a concrete value into `DeliveryChannelUpdate`, where only `None` means "leave unchanged".

Sequence: migration 086 backfills a Slack channel to `'slack'` → a not-yet-upgraded slack-bot replica (or the email client, or any third-party A2A client that hasn't been taught the field) reboots and re-registers → the channel silently reverts to `markdown` and Slack notifications regress to raw Markdown until the new client image lands. An operator-corrected value is likewise overwritten on every bot boot.

Fix options: make `DeliveryChannelCreate.message_formatting: MessageFormatting | None = None`, or only include it in the update when `"message_formatting" in data.model_fields_set`.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -312,6 +312,18 @@ async def _build_message_args(
...
+        # the only place that knows both the job and its delivery channel, and sent under
+        # the key an interactive client uses (`messageFormatting`), so agent-runner and
+        # the orchestrator apply the same rules whether a human or a cron started the run.
+        # Without it a Slack notification arrives as raw Markdown — '### heading',
+        # '**bold**' — because nothing downstream rewrites the agent's output.
+        channel: dict[str, Any] | None = None
+        if job.delivery_channel_id is not None:
+            channel = await self._delivery_channel_repo.get_channel_for_dispatch(db, job.delivery_channel_id)
+        message_formatting = (channel or {}).get("message_formatting") or "markdown"
+        metadata["messageFormatting"] = message_formatting
```
alloy-cicd: **Voice-call jobs inherit the text channel's formatting rules.** `metadata["messageFormatting"]` is set unconditionally here, before the `if job.voice_call:` branch below rewrites `metadata["sub_agent_id"]` to the voice agent. A watch with `voice_call=True` and a Slack delivery channel therefore dispatches the voice agent with `messageFormatting="slack"`, and agent-runner injects the mrkdwn block (`*bold*`, `` `code` ``, `<url|label>`, leading `• `) into the phone agent's system prompt — markup that TTS reads aloud or mangles.

Suggest forcing `"plain"` (or omitting the key) inside the `job.voice_call` branch.

packages/console-backend/sqlmigrations/ddl/086_delivery_channel_message_formatting.sql
```diff
@@ -0,0 +1,32 @@
...
+ALTER TABLE delivery_channels
+    ADD COLUMN message_formatting TEXT NOT NULL DEFAULT 'markdown';
+
+ALTER TABLE delivery_channels
+    ADD CONSTRAINT delivery_channels_message_formatting_chk
+    CHECK (message_formatting IN ('markdown', 'slack', 'google-chat', 'plain'));
+
+-- Backfill the channels already registered by the two chat clients, so existing
+-- installations are correct before their bots next re-register on boot.
+UPDATE delivery_channels SET message_formatting = 'slack' WHERE client_id LIKE '%slack%';
```
alloy-cicd: **The backfill guesses the platform from a `client_id` substring, but `client_id` is operator-configured.** It is the Keycloak `azp` (or the user's `sub` for session-authenticated registrations), and nothing guarantees it names the platform — the repo's own `example-k8s-deployment/base/client-slack.yaml` sets `OIDC_CLIENT_ID: "email-client"`, so a deployment following the shipped example gets **no** Slack backfill. The asymmetric case is worse: any unrelated client whose id happens to contain `slack` (or contains `_`, which is a `LIKE` wildcard here) gets its channels mislabeled `slack`, which the CHECK constraint won't catch and which never self-heals unless that client re-registers.

Since both TS clients now send `message_formatting` on their boot-time upsert, the miss case self-corrects on the next bot restart anyway — so the two `UPDATE`s buy little and carry the mislabeling risk. Consider dropping them, or restricting to exact known ids. (Note this interacts with the upsert-clobbering comment on `delivery_channel_repository.py:276`: as written, the backfill's effect doesn't survive an old client's reboot either.)

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -26,6 +26,7 @@ def _row_to_response(row: Any) -> DeliveryChannelResponse:
         name=row["name"],
         description=row["description"],
         webhook_url=row["webhook_url"],
+        message_formatting=row["message_formatting"],
```
alloy-cicd, line 29: Minor (deploy-order coupling): `row["message_formatting"]` indexes the new column unconditionally, and every read path funnels through `_row_to_response` — so if app pods roll before migration 086 is applied (or 086 is rolled back under live code), every delivery-channel GET/list **and** `get_channel_for_dispatch` raise, failing both the console page and every scheduled dispatch with a `delivery_channel_id`. `DeliveryChannelResponse` already declares a `"markdown"` default, so `row.get("message_formatting") or "markdown"` (and tolerance in the dispatch SELECT) would make the deploy order-independent. Skip if migrations are strictly ordered before rollout in all environments.

packages/agent-runner/agent/core.py
```diff
@@ -1321,6 +1351,13 @@ async def _run_remote_agent(
             # Fallback to plain text prompt
             messages_input = [{"role": "user", "content": prompt}]
 
+        # The channel's rendering rules as a final instruction. There is no system prompt
+        # to append to on this path (the remote agent has its own) and SubAgentInput
+        # carries no per-message metadata, so the message itself is the only carrier.
+        formatting_block = formatting_prompt_block(message_formatting)
+        if formatting_block:
+            messages_input = [*messages_input, HumanMessage(content=formatting_block)]
```
alloy-cicd: Design concern: appending the formatting block as a **separate trailing `HumanMessage`** has two side effects. (1) The remote keys its checkpoints by the propagated `context_id` (per the comment just below), so this bare `<message_formatting …>` turn is persisted in the remote's conversation history and replayed on a later resume with no accompanying task — a single-turn remote that treats the last user message as the request may answer the formatting note instead of the task. (2) The comment's claim that "SubAgentInput carries no per-message metadata" is worth revisiting: `SubAgentInput` already carries `scheduled_job_id` end-to-end into the outbound A2A `Message.metadata`, and the orchestrator already reads `messageFormatting` from merged request metadata (`executor.py:861`) — so a `message_formatting` field on `SubAgentInput` would reuse the existing interactive-path mechanism, keep it out of the conversation history, and fix the foundry gap for free.

Cheaper local fix if the metadata route is out of scope: merge the block into the final message's content instead of adding a new turn.

packages/agent-runner/tests/test_agent_runner_security.py
```diff
@@ -370,3 +373,161 @@ async def test_execute_sub_agent_forwards_context_id_to_remote(self, agent_runne
...
+        """The injection point, guarded by source rather than by running a graph.
+
+        A sub-agent's system prompt is written once and reused across jobs, so the
+        channel cannot be baked into it — the same agent may notify Slack for one job
+        and the web console for the next. Building the full prompt without this append
+        is exactly the regression that puts raw Markdown into Slack.
+        """
+        source = inspect.getsource(core.AgentRunner._run_langgraph_agent)
+        assert "formatting_prompt_block(message_formatting)" in source
+        assert "full_system_prompt += " in source
```
alloy-cicd: Test quality: this asserts on `inspect.getsource(...)` text, so it passes on broken behaviour (block appended in the wrong order relative to `A2A_PROTOCOL_ADDENDUM`, or computed but appended to a prompt that's never used) and fails on harmless refactors (renaming the local, extracting a `_build_system_prompt` helper, a formatter splitting the line). The two remote-agent tests in this same class already show the harness for the behavioural version: patch the graph factory the way they patch `make_a2a_async_runnable`, capture the system prompt, and assert `'format="slack"' in captured_system_prompt` after the A2A addendum.

packages/agent-common/agent_common/core/message_formatting.py
```diff
@@ -0,0 +1,78 @@
...
+def formatting_prompt_block(value: object) -> str:
+    """The rules as a ``<message_formatting>`` block, or ``""`` when none are needed.
+
+    Returned without surrounding blank lines; callers place it (system-prompt
+    addendum, user-preferences block, one-shot prompt) as they need.
+    """
+    rules = formatting_rules(value)
+    if not rules:
+        return ""
+    fmt = normalize_message_formatting(value)
```
alloy-cicd: Nit: `formatting_prompt_block` normalizes twice — once inside `formatting_rules(value)` and again here. A single-pass version is also easier to read and drops the `# type: ignore` pressure upstream:

```python
fmt = normalize_message_formatting(value)
rules = FORMATTING_RULES.get(fmt, "")
if not rules:
    return ""
return f'<message_formatting format="{fmt}">\n{rules}\n</message_formatting>'
```

Relatedly, line 58's validity test (`candidate in FORMATTING_RULES or candidate == DEFAULT_MESSAGE_FORMATTING`) encodes "valid" as an accident of markdown having no rules row; `frozenset(get_args(MessageFormatting))` derives the canonical set from the Literal and removes the two-clause condition.

alloy-cicd (commented):

The mechanism is sound (formatting as a property of the delivery channel, resolved once at dispatch, rules centralised in `agent_common.core.message_formatting`) and the migration/tests follow repo conventions. Eight inline comments below: three behavioural gaps (foundry sub-agents, re-registration clobbering the backfill, voice calls inheriting text rules), one migration-backfill risk, and four smaller items.

Two notes on files outside this diff:

1. **`orchestrator-agent/app/core/executor.py:861-863`** has the same dual-key lookup (`messageFormatting` / `message_formatting`) that `agent-runner/agent/core.py:663-665` now duplicates — but the executor does **not** pass the value through `normalize_message_formatting`, so a client sending e.g. `"Slack"` is normalized on the scheduled path and hits the `Literal`-typed `UserConfig.message_formatting` raw on the interactive path. A shared `message_formatting_from_metadata(meta)` helper next to `normalize_message_formatting` would close both the duplication and the drift.
2. **`console-frontend/src/api/generated/types.gen.ts`** was not regenerated — `DeliveryChannelCreate/Update/Response` there still lack `message_formatting`, so the next `gen-sdk` run produces a dirty diff and, until then, an operator cannot inspect or correct a channel's formatting from the console (relevant given the backfill and clobbering comments below).

packages/agent-runner/agent/core.py
aartaria: Fixed in dcceb219. `_execute_sub_agent` now forwards `message_formatting` to `_run_foundry_agent` too. The query API takes a single `userInput` string and no system prompt, so the rules can only be appended to the prompt there — whether the Foundry-side agent honours them is its own business, but a run that is never told cannot get it right. Covered by `test_execute_sub_agent_forwards_the_format_to_foundry`.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Fixed in dcceb219. `DeliveryChannelCreate.message_formatting` now defaults to `None`, meaning "unchanged": `create_channel` omits the key so the column default applies, and the upsert's update path already skips `None` fields. New DB test `test_reregistration_without_a_format_keeps_the_stored_one` pins it — a Slack channel survives a re-registration from a client that does not declare a format.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Fixed in dcceb219. `messageFormatting` is now set only when `not job.voice_call`. Covered by `test_a_voice_call_is_not_told_how_to_render_text`.

packages/console-backend/sqlmigrations/ddl/086_delivery_channel_message_formatting.sql
aartaria: Fixed in dcceb219 — the UPDATEs are gone. Confirmed your example: `example-k8s-deployment/base/client-slack.yaml` sets `OIDC_CLIENT_ID: "email-client"`. Clients declare their format on every boot's self-registration, so existing rows are corrected there anyway; the migration comment now says why there is no backfill.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Discarding this one as a non-issue for this deployment model: `packages/console-backend/entrypoint.sh` runs `./migrate.sh` under `set -e` before exec'ing uvicorn, so new code cannot serve a request before its own migration has applied. The reverse order during a rolling update (old pod, new column) is safe — old code reads `SELECT *` and ignores the extra column.

packages/agent-runner/agent/core.py
aartaria: Agreed, and fixed in dcceb219 along the lines you suggested. `SubAgentInput` now carries `message_formatting`, and `A2AClientRunnable._from_human_messages_to_a2a` puts it on the outgoing A2A message metadata as `messageFormatting` — the same key an interactive client sends, and the same channel `scheduled_job_id` already uses. So the remote applies its own rules through its normal request-metadata path, nothing lands in its checkpoint, and the trailing `HumanMessage` is gone. It is only sent when it has something to say (plain Markdown is the remote's default too). Covered in agent-common (`test_message_formatting_rides_the_metadata`) and agent-runner (`test_a_remote_agent_is_told_in_the_metadata`).

packages/agent-runner/tests/test_agent_runner_security.py
aartaria: Fixed in dcceb219. The prompt assembly moved into `_build_sub_agent_system_prompt(system_prompt, message_formatting)`, and `TestSubAgentSystemPrompt` now asserts on the returned string instead of the source text. The `inspect` import is gone.

packages/agent-common/agent_common/core/message_formatting.py
aartaria: Fixed in dcceb219. `formatting_prompt_block` normalizes once and reads `FORMATTING_RULES` directly, and validity is now `frozenset(get_args(MessageFormatting))` so the tuple cannot drift from the Literal.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -312,6 +312,20 @@ async def _build_message_args(
...
+        # the key an interactive client uses (`messageFormatting`), so agent-runner and
+        # the orchestrator apply the same rules whether a human or a cron started the run.
+        # Without it a Slack notification arrives as raw Markdown — '### heading',
+        # '**bold**' — because nothing downstream rewrites the agent's output.
+        # Not for a voice call: nothing is rendered there, and mrkdwn rules in a phone
+        # agent's prompt are instructions about a medium it is not using.
+        channel: dict[str, Any] | None = None
+        if job.delivery_channel_id is not None:
+            channel = await self._delivery_channel_repo.get_channel_for_dispatch(db, job.delivery_channel_id)
+        if not job.voice_call:
```
alloy-cicd: Minor, non-blocking: the guard keys on the *intent* (`job.voice_call`) rather than the *outcome* (the dispatch actually going to the voice agent). In the degraded path below — `voice_call=True` but `_resolve_voice_agent_id` returns `None` (line 362's warning) — the dispatch proceeds to the regular sub-agent as a text run, and that run now goes unformatted where before this commit it would have carried the channel's rules. Downstream degrades safely (`normalize_message_formatting(None)` → `markdown`), and the path is already a flagged misconfiguration, so this is only worth changing if you touch this code anyway — e.g. by setting the key inside the `voice_call` branch when the voice agent turns out to be unavailable.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (a0cbd7430d98ded37cb3c349927ed4c6617b53d9..dcceb219eafffcbca52b56cc13dfa5e4ec335e69)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Foundry sub-agents drop `message_formatting` | **verified-fixed** — forwarded to `_run_foundry_agent`, block appended to the `userInput` prompt (the query API's only carrier); covered by `test_execute_sub_agent_forwards_the_format_to_foundry` |
| 2 | Re-registration resets channel to `markdown` | **verified-fixed** — `DeliveryChannelCreate.message_formatting` is now `None`-defaulted meaning "unchanged"; `create_channel` omits the key so the column default applies; pinned by `test_reregistration_without_a_format_keeps_the_stored_one` |
| 3 | Voice calls inherit text formatting rules | **verified-fixed** — `messageFormatting` set only when `not job.voice_call`; covered by `test_a_voice_call_is_not_told_how_to_render_text` (one minor edge inline below) |
| 4 | Migration backfill guesses from `client_id` substring | **verified-fixed** — both `UPDATE`s removed, rationale documented in the migration comment |
| 5 | `_row_to_response` deploy-order coupling | **rejected-by-author, rejection verified** — `packages/console-backend/entrypoint.sh` runs `./migrate.sh` under `set -e` before exec'ing uvicorn, so new code cannot serve before its migration; old-code/new-column direction is safe. Not re-raised. |
| 6 | Remote formatting as trailing `HumanMessage` | **verified-fixed** — now rides `SubAgentInput.message_formatting` → A2A message metadata under `messageFormatting` (same key as interactive clients), sent only when the format has rules; covered in agent-common and agent-runner tests |
| 7 | `inspect.getsource` test assertion | **verified-fixed** — prompt assembly extracted to `_build_sub_agent_system_prompt`, `TestSubAgentSystemPrompt` asserts on the returned string, `inspect` import gone |
| 8 | Double-normalize / two-clause validity nit | **verified-fixed** — single normalize in `formatting_prompt_block`, `KNOWN_FORMATS = frozenset(get_args(MessageFormatting))` |

The summary-level notes from round 1 were also addressed: the orchestrator executor now normalizes through the shared helper, closing the interactive/scheduled drift. (The console-frontend generated SDK types remain un-regenerated — follow-up material, noted once here, not blocking.)

**Delta review:** the fixes are clean and well-tested; one minor non-blocking edge inline. Micro-nit not worth a thread: `test_the_channels_rules_are_appended`'s trailing comment says "The response protocol still comes first" but the assertion checks the system prompt's position relative to the formatting block, not the protocol addendum's.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Fixed in 405e9d93 — I was there anyway. The decision now keys on `is_voice_dispatch`, set only when `_resolve_voice_agent_id` actually returned an id, and moved below the voice branch so it can see that outcome. Fetching the channel stays where it was (one query, still feeding `push_config`). Covered by `test_a_voice_job_that_finds_no_voice_agent_still_formats_its_text`.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (dcceb219eafffcbca52b56cc13dfa5e4ec335e69..405e9d9390a2e580c2e7ad076345d13a96a4df5b)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Voice-call formatting guard keyed on intent (`job.voice_call`) rather than outcome | **verified-fixed** — the decision now keys on `is_voice_dispatch`, set only when `_resolve_voice_agent_id` actually returned an id, and moved below the voice branch so it sees that outcome; the channel is still fetched exactly once. Covered by `test_a_voice_job_that_finds_no_voice_agent_still_formats_its_text`. |

(All eight round-1 findings were settled in round 2: seven verified-fixed, one rejected with verified reasoning.)

**Delta review:** every previous finding is resolved and the code delta introduces nothing new — no inline comments this round.

One scope note, not a code finding: the delta also adds `docs/adr/0006-release-notes-generated-locally-from-commits.md`, an ADR about release-note generation in the justfile that has no connection to this PR's scheduler-formatting change. If it rode along by accident, consider moving it to its own PR so the history of both decisions stays legible; if it's intentional, no objection to its content from this review's angle.

aartaria:
Round 3's scope note was right, and it was my error: `docs/adr/0006-release-notes-generated-locally-from-commits.md` was an unrelated untracked file sitting in my checkout that a `git add -A` swept into 405e9d93. Removed by amending the tip commit — the PR head is now 777e091d and its diff is scheduler-formatting only. The ADR is untouched on disk as an untracked file, so it can go out in its own PR by whoever wrote it.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (405e9d9390a2e580c2e7ad076345d13a96a4df5b..130e8bea7b9e691a3a6c1c8b054fc5fa356779f1)

**Ledger of round-3 findings:**

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Scope note: unrelated ADR (`docs/adr/0006-release-notes-...md`) rode along in the branch | **verified-fixed** — removed via rebase (`405e9d93` → `777e091d`, same scheduler fix without the ADR) |

(Round 3 had no code findings; the round-2 voice-dispatch fix and all eight round-1 findings remain settled.)

**Delta review:** every previous finding is resolved and the delta introduces nothing new — no inline comments this round. The only new commit, `130e8bea`, changes no behaviour: it documents on `SubAgentInput.message_formatting` and `_build_sub_agent_system_prompt` that the field is set only on the stand-alone agent-runner dispatch (where the sub-agent's text is delivered verbatim) and left `None` when the orchestrator routes (it composes the delivered message and applies the channel's rules itself), and pins the delegation path with `test_delegation_does_not_impose_a_channel_format_on_the_sub_agent`. The documented invariant matches the code as reviewed in rounds 1–3.

Nothing outstanding from the automated review across all four rounds.

PR #201, opened 2026-09-03
fix(risk-gate): never classify a tool that could not be fetched

### LINKED ISSUE #199: risk gate: an unresolvable tool name is classified from the name alone, persisted, and answered with a dead-end error
## Problem

When the model names a tool that cannot be resolved, the risk gate classifies it anyway — from the name alone — and persists the result as if it were a real profile. The agent then gets a dead-end error it cannot act on.

Observed end to end. Under PTC the orchestrator's registry is exposed inside `eval`, so its system prompt advertises camelCase identifiers:

```ts
/** Create a bug report for an unrecoverable error. */
async function consoleCreateBugReport(input: { description: string; task_id?: string | null }): Promise<unknown>
```

The model called that name as a **native** tool call instead of inside `eval`. What followed:

1. `ConditionalHITLMiddleware` scores before dispatch. `_get_tool_instance("consoleCreateBugReport", ctx)` is an exact-match lookup against `tool_registry`, which holds `console_create_bug_report` → `None`. `_get_server_slug` misses for the same reason → falls back to `_self`.
2. `score_tool_risk(name, args, tool=None, …)` misses the cache and reaches the LLM branch with `description=""` and `input_schema={}` — the classification prompt reads *"No description available / No schema available"*.
3. The result is cached **and persisted** with `schema_hash=""`.
4. Dispatch then fails with `Error: Tool 'consoleCreateBugReport' is not available`, which the model relayed to the user as "unavailable in this session" — against the standing instruction never to report a tool as missing (`orchestrator-agent/app/models/config.py`).

`risk_factors` is derived from the schema (parameter name → glob patterns → score) and matched against arguments locally by `ToolRiskEntry.match_args`. With no schema the classification cannot name any control parameter, so it returns `{}` — a profile asserting "this tool has no risk-bearing parameters", produced for a tool nobody looked at.

## It is not a one-off

Every camelCase name the model has guessed natively has left a row behind, all filed under `_self` because the slug lookup missed too:

```
consoleCreateBugReport  | _self | 0.4 | schema_hash '' | 2026-09-02
consoleSearchSkills     | _self | 0.1 | ''             | 2026-08-28
consoleListMcpServers   | _self | 0.1 | ''             | 2026-08-17
consoleGrepMcpTools     | _self | 0.5 | ''             | 2026-08-17
githubSearchIssues      | _self | 0.1 | ''             | 2026-08-17
readFile                | _self | 0.3 | ''             | 2026-06-10
FinalResponseSchema     | _self | 0.0 | ''             | 2026-06-08
SubAgentResponseSchema  | _self | 0.1 | ''             | 2026-05-27
```

Each cost a classification call, and each is indistinguishable in the catalogue from a profile derived from a real schema. `schema_hash = ''` is the only trace, and nothing reads it as "never inspected".

Note `readFile`, `FinalResponseSchema` and `SubAgentResponseSchema`: those names *are* callable — the middleware's `_platform_tools` and the dispatcher's `static_tools` are different sets, so a tool the gate cannot fetch may still execute. Any fix has to keep gating those rather than waving them through.

## Wanted behaviour

**If the tool cannot be fetched, do not score it. Tell the agent what went wrong and how to fix the call.**

1. **Resolve the way the dispatcher does** before concluding anything is unresolvable — registry plus the static/platform sets — so a legitimately callable tool is neither mis-scored nor blocked.
2. **Still unresolvable → no classification.** No LLM call, no `cache.put`, no `persist_entry`. There is no schema to reason about, so there is nothing to cache.
3. **Answer the call with a corrective `ToolMessage`** instead of letting it fall through to `"Tool ... is not available"`. The middleware already has the mechanism: an artificial `ToolMessage` plus dropping the call from the revised list, as denials and client actions do. The message should name the defect and the fix — when the name matches a PTC-exposed tool (compare on `to_camel_case`), say it is reachable inside `eval` as `tools.<camelName>`; otherwise point at discovery. `agent_common/core/graph_utils._not_a_function_hint` already does exactly this for the inside-`eval` direction and is the model to follow.
4. **Delete the eight rows above**, and consider whether the risk API should hide entries whose `schema_hash` is empty.

## Acceptance

- [ ] a native tool call naming a PTC-exposed tool is answered with a hint naming `eval` and the camelCase identifier, and no risk classification runs
- [ ] no row is written to `tool_risk_scores` for a tool that could not be fetched
- [ ] a callable tool the gate cannot fetch (a static/platform tool) is still resolved and still gated — regression test for `FinalResponseSchema`
- [ ] the eight existing phantom rows are removed

aartaria:
closes ringier-data/nannos#199

## What was wrong

Under PTC the orchestrator's registry is exposed inside `eval`, so the system prompt advertises camelCase identifiers. The model emitted one of them (`consoleCreateBugReport`) as a **native** tool call, and every layer failed open:

1. the gate's exact-match lookup missed (the registry is keyed `console_create_bug_report`), and so did the slug lookup → `_self`;
2. `score_tool_risk` reached the LLM branch with `description=""` and `input_schema={}` — the classification prompt read *"No description available / No schema available"*, so `risk_factors` came back `{}`: a profile asserting "this tool has no risk-bearing parameters", produced for a tool nobody looked at;
3. that result was cached **and persisted** with `schema_hash = ''`, indistinguishable in the catalogue from a real profile;
4. dispatch then answered `Error: Tool '...' is not available`, which the model relayed to the user as "unavailable in this session" — against the standing instruction never to report a tool as missing.

Eight such rows were already on record, one per camelCase name a model had guessed.

## What changed

- **`score_tool_risk` never classifies a call it cannot fetch** — no LLM call, no `cache.put`, no `persist_entry`. The cache *lookup* still runs ahead of the guard: the static guards seeded in migration 057 carry `schema_hash = ''` on purpose and must be honoured whether or not the instance is fetchable (the existing destructive-floor-on-cache-hit test catches exactly this). The call is still gated, on the same name-based fallback the LLM branch itself falls back to on failure.
- **`ConditionalHITLMiddleware` answers the call instead of letting it fall through.** A native call whose name is a camelCase alias of a known tool gets a corrective `ToolMessage` — naming `eval` and `tools.<camelName>` under PTC, the snake_case name otherwise — and is dropped from the revised call list, the mechanism denials and client actions already use. Never scored, never dispatched.
- **Static/platform tools are handed to the gate**: the orchestrator's own (time, presigned-url, copy_file, notify_user, `FinalResponseSchema`) and each sub-agent's bound set. A callable tool the registry does not hold is now resolved and gated on its real schema rather than dropping to a name-only score.
- **The dispatcher's dead end is replaced** with the same corrective advice, for the paths that reach it anyway.
- **Migration 089** deletes the eight phantom rows.

## Two judgement calls worth reviewing

- **The refusal fires only on a positive signal** (the name *is* a camelCase alias of a known tool), never on "the gate could not find it". The gate's view is `tool_registry ∪ platform_tools ∪ server maps`, but the dispatcher also resolves tools registered directly with `ToolNode` — `write_todos`, the filesystem tools, the response schemas. Refusing on absence alone would break callable tools, which is what the issue warns about for `readFile` / `FinalResponseSchema` / `SubAgentResponseSchema`. Those keep flowing through the gate; they are simply no longer classified or persisted, and in the orchestrator they are now fetchable and properly gated.
- **The risk API does *not* hide entries with an empty `schema_hash`** (the issue's "consider"). The 057 static guards — `read_personal_file`, `console_create_bug_report`, `docstore_search`, the skill-mutation guards — all carry `schema_hash = ''` deliberately, so that predicate would hide hand-written policy. Migration 089 deletes by exact name instead, with the empty hash only as a safety catch so a row since re-scored against a real schema is left alone.

## Verification

- `agent-common`: 958 passed (`test_foundry_cost_version.py` needs the absent `foundry_sdk`; the 9 `test_thinking_blocks_round_trip.py` failures are pre-existing on `origin/main` — verified by stashing).
- `orchestrator-agent`: 815 passed. `agent-runner`: 45 passed.
- Migration 089 applied against a throwaway PostgreSQL 18 seeded with 057 + the eight phantom rows: `DELETE 8`, all 13 seeded static guards intact.
- `just test-db` could not run here — local port 4000 is occupied by an unrelated service, so `psql` got an HTTP response during SSL negotiation. Environment limitation, not a migration failure.

## Checklist


packages/agent-common/agent_common/middleware/conditional_hitl.py
```diff
@@ -420,9 +499,17 @@ async def aafter_model(self, state: AgentState[Any], runtime: Runtime[ContextT])
...
         if not action_requests:
-            return None
+            if not corrective_messages:
+                return None
+            last_ai_msg.tool_calls = [
+                tool_call
+                for idx, tool_call in enumerate(last_ai_msg.tool_calls)
+                if idx not in corrective_messages
+            ]
+            return {"messages": [last_ai_msg, *corrective_messages.values()]}
```
alloy-cicd, line 603: 🔴 **Dropping the aliased tool call orphans its ToolMessage and can brick the thread.**

This early return (and the mirror at lines 543-546 / the `revised_tool_calls` write-back at 590-592) rebuilds `last_ai_msg.tool_calls` *without* the aliased call while returning a `ToolMessage` that carries that call's `tool_call_id`. Since `add_messages` upserts by id, the checkpointed AIMessage is replaced by the stripped one, so the history permanently contains a `tool_result` with no matching `tool_use`. This is exactly the failure documented in `loop_detection_middleware.py:388-399` and fixed in #184 (commit `5daad04b`): Bedrock/Anthropic hard-400s on every later turn ("toolResult blocks … exceeds the number of toolUse blocks") and the thread is unrecoverable. Nothing downstream sanitizes orphans (`langchain_anthropic` only normalizes ids; `_seal_dangling_tool_calls` fixes the opposite direction).

A second effect when **all** calls in the turn are aliases: `tool_calls == []`, and langchain's `model_to_tools` router returns `end_destination` on `len(tool_calls) == 0` (factory.py:1866-1868) *before* its artificial-tool-message branch — so the corrective hint is never fed back to the model and the user gets the (typically empty) AIMessage content as the final answer. The new test `test_unresolvable_call_alongside_a_real_one` asserts `ai.tool_calls == []`, encoding the bug as intent.

Upstream `HumanInTheLoopMiddleware._process_decision` **keeps** the tool call on reject for precisely this reason (`human_in_the_loop.py:321-333`: `return tool_call, tool_message`). Keeping the aliased call in `tool_calls` and letting the corrective ToolMessage answer it gives the same "not dispatched" outcome, keeps the history provider-legal, and makes the router loop back to the model so the hint is actually read.

packages/agent-common/agent_common/core/tool_risk_scorer.py
```diff
@@ -179,13 +181,34 @@ async def score_tool_risk(
...
+        # "No description available / No schema available". `risk_factors` then comes
+        # back `{}`: a profile asserting "this tool has no risk-bearing parameters",
+        # derived from a tool nobody looked at, and it used to be cached AND persisted
+        # with an empty `schema_hash` — indistinguishable in the catalogue from a
+        # profile derived from a real schema. Every camelCase name a model guessed
+        # natively left such a row behind (see migration 089).
+        #
+        # The call is still gated: this is the same name-based fallback the LLM branch
+        # itself falls back to on failure. It is only never classified and never stored.
+        score = _deterministic_fallback(tool_name)
```
alloy-cicd: 🟠 **The new early return bypasses `_destructive_floor` — regression vs. the pre-PR path.**

`_deterministic_fallback` checks the safe prefixes (`get_`/`list_`/`search_`/`read_`/`fetch_`/`describe_` → 0.3, line 380) *before* the dangerous-keyword check, so an unfetchable tool named e.g. `read_and_remove_file` or `get_or_delete_x` scores **0.3** — under the 0.8 threshold — and auto-executes with no approval card. Pre-PR, the `tool is None` case fell through to the LLM branch where `_destructive_floor(tool_name)` (0.9 for `delete`/`remove`/`drop`/`destroy`, applied at line 221) guaranteed an interrupt for such names. The floor is applied on the cache-hit path (line 172) and the LLM path but not here.

Cheap fix: `score = max(_deterministic_fallback(tool_name), _destructive_floor(tool_name))`. (The `cache is None` branch at line 158 has the same hole, but that one is pre-existing, not from this PR.)

packages/agent-common/agent_common/middleware/conditional_hitl.py
```diff
@@ -118,6 +119,61 @@ def _client_action_tool_message(decision: dict[str, Any], tool_call: ToolCall) -
...
+
+    snake = next((n for n in known_names if to_camel_case(n) == tool_name), None)
+    if snake is None or snake == tool_name:
+        return None
+
+    # Imported lazily: ``graph_utils`` reaches back into this module (it builds the
+    # middleware), so a module-level import would close the cycle.
+    from agent_common.core.graph_utils import code_interpreter_ptc_enabled
+
+    if code_interpreter_ptc_enabled():
```
alloy-cicd: 🟠 **The PTC branch tells the model to call `tools.<camel>` in `eval` without checking the tool is actually exposed there.**

`snake` is resolved against `_known_tool_names`, which is a strict superset of the PTC-exposed set: it includes `notify_user` (in `_PTC_EXCLUDED_TOOL_NAMES`, `graph_utils.py:320-329` — it *always fails* inside eval — yet it lands in `_platform_tools` via `get_static_tools(..., with_notify_user=True)` in `graph_factory.py:663-665`), the raw listers stripped by `_without_raw_listers` (`graph_utils.py:771-785`), and any registry tool outside `whitelisted_tool_names`. For those, the hint is factually wrong: the model goes into `eval`, `tools.<camel>` doesn't exist, and `_not_a_function_hint` bounces it back out — at least one wasted round-trip, potentially a ping-pong.

Suggestion: take this branch only when `snake` is in the current PTC exposure set (or at minimum skip names in `_PTC_EXCLUDED_TOOL_NAMES` and the raw listers); otherwise fall through to the non-PTC wording, which is correct for natively bound tools.

packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -651,7 +660,9 @@ async def _forward_to_active_subagents(context_id: str, messages: list) -> None:
         # ConditionalHumanInTheLoopMiddleware: uses interrupt() to pause and ask for user
         # confirmation before executing guarded tools (self-improvement, privacy, bug reports).
         # Supports argument-based conditions (e.g., docstore_search only when include_personal=True).
-        hitl_middleware = _create_hitl_middleware()
+        hitl_middleware = _create_hitl_middleware(
+            self.get_static_tools(with_response_tool=True, with_notify_user=True)
```
alloy-cicd: 🟡 **The orchestrator's deep-agent filesystem/todo tools are still on the `tool is None` path.**

`create_deep_agent`'s `ls`/`read_file`/`write_file`/`edit_file`/`write_todos` are ToolNode-bound and appear in neither `tool_registry` nor these `platform_tools`, so `_get_tool_instance` returns `None` for them and the new early return in `score_tool_risk` gives `write_file` a flat 0.6 (< 0.8, never gated), with no LLM classification, no cache/DB write, and no `risk_factors` arg-pattern matching — ever, unless a sub-agent (whose stack now passes `platform_tools=all_tools` + `fs_middleware.tools`) happens to classify the same `name::_self` key first. Sub-agents got the fix; the orchestrator didn't.

To be fair, the pre-PR "classification" of these was the empty-schema junk this PR is deleting, so gating outcomes barely change today — but the PR's stated goal (score statically bound tools on their real schema) is left unmet for the orchestrator's own fs tools. Passing them here (they're available from the deep-agent middlewares) closes it cheaply.

alloy-cicd (commented):

The core direction is right (fetchable static tools, no more phantom classification), but the camel-alias corrective path re-introduces the orphaned-tool-result bug class this repo fixed in #184, and the new `tool is None` early return drops the destructive-name floor. Details inline.

**Findings (ranked):**
1. 🔴 Dropping the aliased tool call while emitting its ToolMessage orphans the tool result (provider 400 on the next turn) and, when *all* calls are aliases, routes to END so the hint is never read — `conditional_hitl.py:507-512`, `:543-546`
2. 🟠 The `tool is None` early return bypasses `_destructive_floor` — an unfetchable `read_and_remove_*`-style name now scores 0.3 and auto-executes — `tool_risk_scorer.py:198`
3. 🟠 The PTC hint sends the model into `eval` without checking the tool is actually exposed there (`notify_user`, raw listers, non-whitelisted registry tools) — `conditional_hitl.py:154`
4. 🟡 Migration 089's hardcoded 8-name list leaves other empty-`schema_hash` phantom rows in place, and an empty stored hash permanently shadows the real schema-derived profile — `089_delete_phantom_tool_risk_scores.sql:22`
5. 🟡 The orchestrator's gate still can't fetch the deep-agent filesystem/todo tools, so they land on the new fallback path (never classified, no `risk_factors`) — `graph_factory.py:664`
6. 🔵 The dispatcher fallback's alias detection is narrower than the gate's (registry-only) — hardening — `dynamic_tool_dispatch.py:731`
7. 🔵 `known_names` is rebuilt per model turn and the camel reverse-lookup is an O(registry) regex scan — `conditional_hitl.py:361`

Minor, no inline comment: `_unresolved_tool_content` and `_camel_alias_tool_message` are two hand-maintained copies of the same alias-detection + hint logic (already diverging in wording and known-name set — a shared helper in `agent-common` would serve both); and the new module-level `from langchain_quickjs._prompt import to_camel_case` in `conditional_hitl.py` adds ~0.5 s of cold-import to every agent, including PTC-disabled ones (the file already imports `graph_utils` lazily for exactly this reason).

packages/console-backend/sqlmigrations/ddl/089_delete_phantom_tool_risk_scores.sql
```diff
@@ -0,0 +1,37 @@
...
+-- appear. These are the ones already on record.
+--
+-- Deleted by exact name, never by `schema_hash = ''`: the static guards seeded in
+-- 057 carry an empty hash on purpose (they are hand-written policy, not derived
+-- profiles). The hash predicate here is only a safety catch, so that a row which
+-- has since been re-scored against a real schema is left alone.
+DELETE FROM tool_risk_scores
+WHERE server_slug = '_self'
+  AND schema_hash = ''
+  AND tool_name IN (
```
alloy-cicd, line 52: 🟡 **The hardcoded name list under-deletes, and leftover empty-hash rows shadow real profiles forever.**

The phantom generator was generic — pre-PR, *any* `tool is None` call persisted a row with `schema_hash = ''` (the presence of `FinalResponseSchema`/`SubAgentResponseSchema` in this list shows it fired on real static-tool names, not just camel guesses). Rows this list misses — e.g. `get_current_time`, `copy_file`, presigned-url phantoms called out in the new `graph_factory.py:83-95` comment, or any other name a model guessed in prod — are worse than dead weight after this PR: `ToolRiskCache.get` treats an empty stored hash as "can't disprove, trust it" (`tool_risk_cache.py:203`, `_schema_confirms_unchanged`), and the stale-hit path *extends* `updated_at`, so a surviving phantom row permanently wins over the schema-derived profile the PR just made possible.

A precise predicate exists: `persist_entry` only ever wrote `schema_hash = ''` for unfetchable tools, and migration 057 seeds exactly 13 empty-hash rows (11 on `server_slug='console'`, plus `read_personal_file`/`docstore_search` on `'_self'`). So `DELETE ... WHERE schema_hash = '' AND NOT (server_slug = 'console' OR tool_name IN ('read_personal_file','docstore_search'))` is complete and can't touch the seeds.

Also worth a deploy note: `ToolRiskCache._do_refresh` is an additive merge ("never remove", `tool_risk_cache.py:366`), so the DELETE doesn't reach warm pods — fine if this ships with a rollout restart, surprising otherwise.

packages/orchestrator-agent/app/middleware/dynamic_tool_dispatch.py
```diff
@@ -714,6 +716,38 @@ async def _maybe_adopt_run_thread(
...
+    ``Error: Tool 'X' is not available`` is a dead end: the model has nothing to
+    act on, and it has been observed relaying it to the user as "unavailable in
+    this session" — against the standing instruction never to report a tool as
+    missing (``models/config.py``). Say what the defect is and what to do next.
+
+    The camelCase case (the model lifted a ``tools.<camelName>`` identifier out of
+    the PTC prompt and emitted it as a native call) is normally answered earlier,
+    by the risk gate; this covers the paths that reach dispatch anyway.
+    """
+    known = set(user_context.tool_registry or {})
```
alloy-cicd: 🔵 **Hardening: this fallback's alias detection is narrower than the gate's.**

`known` is registry-only, while the HITL twin `_known_tool_names` also unions platform tools and both server maps, and `_lookup_tool` resolves `static_tools` too. On any path that reaches dispatch without the async gate (the sync `after_model` has no alias handling; a future instantiation with non-empty `static_tools`), a camel alias of a static tool gets the generic "look it up with `console_grep_mcp_tools`" hint — discovery that only indexes MCP catalog tools and can never surface them. Not a live bug today (the async gate answers these first, and production `static_tools=[]`), but the two copies of this logic already disagree on both wording and known-name set; a shared helper in `agent-common` used by both sites would keep them from drifting further.

packages/agent-common/agent_common/middleware/conditional_hitl.py
```diff
@@ -295,6 +354,11 @@ async def aafter_model(self, state: AgentState[Any], runtime: Runtime[ContextT])
         interrupt_indices: list[int] = []
         # Store risk metadata per interrupt for inclusion in the payload
         _risk_metadata: list[_RiskMetadata] = []
+        # Calls answered here instead of being scored or dispatched, by index:
+        # a camelCase PTC identifier emitted as a native call. See
+        # _camel_alias_tool_message.
+        corrective_messages: dict[int, ToolMessage] = {}
+        known_names: set[str] = self._known_tool_names(getattr(runtime, "context", None))
```
alloy-cicd: 🔵 **Perf: this set is rebuilt on every model turn, and the miss path is an O(registry) regex scan.**

`_known_tool_names` unions the full `context.tool_registry` (hundreds of names for catalog-mode users) on every `aafter_model`, even when there are no tool calls or all of them resolve; and each unresolvable name then runs `to_camel_case` (a regex substitution) over the entire set (line 146). Building the set lazily on the first unknown name — or better, a memoized `camel → snake` dict so the lookup is O(1) — removes both costs from the hot path. (Also: on line 147, the `snake == tool_name` disjunct is unreachable — if the camel form equals the name, the name was in `known_names` and line 145 already returned.)

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (a3a731773f631bd33d5f7efb465bdee3bcf07b49..6796fd9db8c0280271546091987196a9893d6159)

1. 🔴 Orphaned tool result / END routing (`conditional_hitl.py`) — **verified-fixed.** The early-return path now returns the AIMessage unchanged and the post-interrupt rebuild appends the answered call to `revised_tool_calls`; the two tests that asserted `tool_calls == []` now assert the pairing invariant, plus a new all-alias test (`test_every_answered_call_keeps_a_matching_tool_call`). Verified the kept-call semantics against the router: an answered call is excluded from `pending_tool_calls`, so it is never dispatched, and the non-empty list keeps the turn off the END destination.
2. 🟠 Destructive floor bypass (`tool_risk_scorer.py:204`) — **verified-fixed.** `max(_deterministic_fallback, _destructive_floor)` with a regression test (`test_unfetchable_destructive_tool_still_clears_the_floor`, asserts the 0.3 trap first).
3. 🟠 PTC hint without exposure check (`conditional_hitl.py`) — **verified-fixed.** `_reachable_inside_eval` gates the `eval` wording on the checkpointed `ptc_exposed_tool_names` set, falling back to excluding `_PTC_EXCLUDED_TOOL_NAMES` + the raw listers. Verified the state key holds snake_case names (`graph_utils.py:903`, `exposed_names = {tool.name ...}`), matching the membership test; both new tests cover `notifyUser` and a non-exposed registry tool.
4. 🟡 Migration 089 name list (`089_...sql`) — **rejected-by-author, reasoning verified.** `ToolRiskScoreUpsertRequest.schema_hash` does default to `''` (`tool_risk_router.py:53`), so a blanket empty-hash predicate would delete admin-created overrides — the round-1 suggestion was wrong on that point. The known gap (unlisted phantoms keep shadowing via `_schema_confirms_unchanged`) and the warm-cache rollout-restart requirement are now documented in the migration header; the gap is tracked separately. Accepted.
5. 🟡 Orchestrator deep-agent fs/todo tools on the `tool is None` path (`graph_factory.py:664`) — **still-open.** Not addressed by the delta and not answered on the thread; the original comment stands. Low severity (gating outcomes barely change vs. the pre-PR empty-schema classification), follow-up material rather than blocking.
6. 🔵 Dispatcher alias detection narrower than the gate's (`dynamic_tool_dispatch.py`) — **verified-fixed.** `_unresolved_tool_content` now unions `self.static_tools` (a name-keyed dict, so the union is over names), and the second commit tailors the fallthrough advice per PTC mode, naming the raw listers only when the registry holds them. Tests cover all three branches.
7. 🔵 Perf: per-turn name-set build + O(registry) camel scan (`conditional_hitl.py`) — **verified-fixed** (adequately). The set build and camel scan now run only when the two-dict instance lookup misses, and the unreachable `snake == tool_name` disjunct is gone.

**Delta review:** no new issues. `packages/agent-common` `test_conditional_hitl.py` + `test_tool_risk_scorer.py`: 37 passed; `orchestrator-agent` `test_unresolved_tool_hint.py`: 8 passed. One non-blocking observation for the record: `_ptc_exposed_tool_names` reads a key written by the PTC middleware's own `aafter_model` in the same super-step, so depending on hook ordering the gate may see the previous turn's exposure set (or `None` on the first turn) — the `None` fallback and the turn-to-turn stability of the exposure set make this at worst a one-turn-stale hint, which the code's fallback semantics already accept.

No inline comments this round. Submitted as COMMENT — the human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (6796fd9db8c0280271546091987196a9893d6159..94cd41a24eb9dddbd34f658976daf92544e66a2d)

**Ledger of round-2 dispositions:**

1. 🟡 Orchestrator deep-agent fs/todo tools on the `tool is None` path (`graph_factory.py`) — the one finding still open after round 2 — **verified-fixed** by `94cd41a`, on two fronts:
   - **Policy over classification:** migration 090 seeds `tool_risk_scores` rows for the eight ToolNode builtins, making the table the authority exactly as 057 does for the static guards (the cache lookup runs ahead of the unfetchable-tool guard and an empty `schema_hash` never mismatches, so the rows are honoured with no code change). Verified: the `ON CONFLICT (tool_name, server_slug)` target matches the table's primary key (057:13); the eight names don't collide with any 057 seed; the overwrite-on-conflict trade-off (a hand-tuned admin row for these names is lost) is documented; the `/memories/*` factor at 0.75 works as written — `risky_values` keys are compiled as glob patterns (`tool_risk_cache.py:97`, `_glob_to_regex`); and none of the eight names trips `_destructive_floor`, as the header claims.
   - **Exhaustive gate:** `deep_agent_builtin_tools` re-derives the builtin instances (no hardcoded list to rot — and the new pinning test fails the suite if deepagents renames one), the orchestrator registers them alongside its static tools, and `platform_tools_are_exhaustive=True` licenses answering an invented name immediately instead of scoring it. I probed the exhaustive verdict for false refusals and found none: `task` and `eval` are skipped before the check (`conditional_hitl.py:455`), `client_action` lives in the per-user `tool_registry` (`utils.py:589`) so it resolves, the response schema and `notify_user` come via `get_static_tools(with_response_tool=True, with_notify_user=True)`, and `deep_agent_builtin_tools(self.backend_factory)` receives the same backend value `create_deep_agent` itself gets. The flag defaults to `False`, so sub-agents keep the absence-proves-nothing behaviour, with a test pinning each of the three modes. Even a hypothetical miss fails soft — a corrective ToolMessage the model can recover from, not a dropped call.

All other round-1/round-2 findings were already resolved or settled (the migration-089 predicate remains rejected-by-author with verified reasoning).

**Delta review:** no new issues. `test_conditional_hitl.py`: 30 passed; `test_unresolved_tool_hint.py`: 10 passed.

Every previous finding is now resolved and the delta introduces nothing new — a clean round. No inline comments. Submitted as COMMENT — the human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (94cd41a24eb9dddbd34f658976daf92544e66a2d..8fa3f6facb6a0c2bc163bbaf4759af52bf6ae1d2)

**Ledger:** all findings from rounds 1–3 remain resolved (six verified-fixed, one — the migration-089 broad predicate — rejected-by-author with verified reasoning). Nothing was reopened.

The delta is a self-directed refinement, not a response to an open finding: it reverts the seeding half of `94cd41a` (migration 090) on the correct observation that its rationale — "the gate cannot fetch these tools" — was voided by the same commit that introduced it (`deep_agent_builtin_tools` made them fetchable). The replacement shape is verified sound:

- **089 extended to the builtin junk rows.** The eight deep-agent names are appended to the exact-name DELETE so the now-fetchable builtins re-classify from their real schemas instead of being pinned forever by an empty-hash guess (`_schema_confirms_unchanged` never mismatches an empty stored hash). Verified the 057 policy seeds are untouchable by this list: `read_personal_file` and `docstore_search` do not appear in it, and the `schema_hash = ''` conjunct still protects any row since re-scored against a real schema. The self-healing distinction the header draws (derived rows re-derive on next use; 057 rows are hand-written policy with nothing to re-derive from) is accurate.
- **`persist_entry` refuses an empty `schema_hash`** — the structural half. Verified there is exactly one production caller (`tool_risk_scorer.py:241`, the LLM path), where the hash is a truncated SHA-256 hex digest of a fetched schema and can never be empty — so the guard changes behaviour only for the regression it exists to prevent. The 057 seeds and admin overrides are written through the console API, not through the cache, so hand-written policy is unaffected, and with this the empty hash becomes unambiguous ("policy") as the commit message argues. Covered by a new test.
- **Test adjustments consistent:** the 090 pinning test is gone with the migration it pinned; the remaining fetchability test now also asserts each builtin exposes a non-empty input schema, which is the property the classify-from-schema path actually depends on. Deleting the never-shipped 090 inside the same PR leaves no deployment-ordering hole, and the migration sequence stays contiguous at 089.

**Delta review:** no new issues. `agent-common` `test_tool_risk_scorer.py` + `test_conditional_hitl.py`: 41 passed; `orchestrator-agent` `test_unresolved_tool_hint.py`: 9 passed.

Every previous finding is resolved and the delta introduces nothing new — a clean round. No inline comments. Submitted as COMMENT — the human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (8fa3f6facb6a0c2bc163bbaf4759af52bf6ae1d2..bb420b27b92542973a907e03bf449263b127cf1a)

**Ledger:** all findings from rounds 1–4 remain resolved (six verified-fixed, one rejected-by-author with verified reasoning). Nothing was reopened.

The delta is one new migration, `090_restore_weakened_policy_guards.sql`, prompted by the author's own QA against a real database: the phantom-classification path had not only added junk rows but *overwritten* seeded 057 policy via the replace-all-columns upsert — silently disabling two privacy guards (`read_personal_file` 1.00→0.30, `docstore_search` losing its `include_personal` factor) plus `console_create_bug_report`. 089 cannot repair those (they are the policy rows themselves, wearing guessed values), so 090 re-asserts them. Verified:

- **Fidelity:** the 13 rows are byte-for-byte identical to 057's seeds — names, slugs, base scores, `risk_factors`, and `allowed_actions` all match. The migration restores policy; it does not quietly change it.
- **Weakenings-only semantics:** the `ON CONFLICT … DO UPDATE … WHERE` fires only when the stored `base_score` is strictly below the seed, or when seeded non-empty factors were emptied. For the eleven 1.0 guards the second condition can never fire (their seeded factors are `{}`), so a guard whose score was raised by an admin is untouched, and re-running is a no-op — matching the header's safety argument, and the tie toward re-asking the user on ambiguity is the right default for privacy guards.
- **Structural durability:** resetting `schema_hash = ''` marks the row as policy — the cache lookup then always hits it (`_schema_confirms_unchanged`), the LLM branch never runs for it, and the round-4 `persist_entry` guard means no derived profile can be written over it again. The overwrite class this repairs cannot recur under the PR's code.
- **Hygiene:** numbering is contiguous and unique (090 was freed when the round-3 seed migration was deleted within this PR, never having shipped); the deliberate no-op `rambler down` matches repo precedent and is correctly argued (rolling back would mean re-weakening a privacy guard); the warm-cache deploy note carries over from 089.

One marginal corner, for the record rather than as a finding: a policy row whose `allowed_actions` alone was corrupted (score and factors intact) is not repaired — no QA evidence suggests that combination exists, and the phantom writes observed always lowered the score, which the predicate catches.

**Delta review:** no new issues — every previous finding is resolved and the delta introduces nothing new. A clean round; no inline comments. Submitted as COMMENT — the human reviewer decides.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 6 (bb420b27b92542973a907e03bf449263b127cf1a..4f578f7be48c401b0951e579b91d69150a61f703)

**Ledger:** all findings from rounds 1–5 remain resolved (six verified-fixed, one rejected-by-author with verified reasoning). The round-5 summary's for-the-record corner — a policy row whose `allowed_actions` alone were corrupted would not be repaired — is now **verified-fixed** as well: the author checked the real database, found the corner was live on four of the thirteen guards (`edit` added to `console_import_skill`, `console_activate_skill`, `read_personal_file`, `docstore_search`, all seeded approve-or-reject only), and extended migration 090 accordingly.

The delta is that one predicate clause plus a corrected damage account in the header. Verified:

- **Containment semantics are right.** `NOT (tool_risk_scores.allowed_actions <@ EXCLUDED.allowed_actions)` fires exactly on a *widened* set: JSONB array containment is order- and duplicate-insensitive, so a stored `["approve","edit","reject"]` against a seeded `["approve","reject"]` triggers the repair, while an admin who *narrowed* to `["approve"]` stays contained and is untouched, and a stored set equal to the seed makes a re-run a no-op. All three clauses remain one-directional, preserving the weakenings-only safety argument; the author's PG18 verification covered the widened-with-intact-score repair, the narrowed-and-raised guard left alone, and idempotence.
- **The header now matches the data:** seven of thirteen guards score-weakened (six below the 0.80 gate — the self-improvement guards included, not only the privacy pair), four actions-widened. The account and the predicate agree.
- Residual corner-of-corner, for the record only: a row that fires any clause has *all* columns reset to seed, so an admin who had deliberately raised a score on a row that separately had its actions widened would see the score reset too — but the phantom upsert replaced whole rows, so that mixed state cannot arise from the defect being repaired, and the documented tie-break (err toward asking) covers it.

**Delta review:** no new issues — every previous finding is resolved and the delta introduces nothing new. A clean round; no inline comments. Submitted as COMMENT — the human reviewer decides.

PR #202, opened 2026-09-03
test(agent-common): pin the risk scorer to tool-calling structured output

aartaria:
Follow-up to #188 — tests only, no production code touched.

#188 switched `_score_tool_via_llm` from the langchain-openai >= 0.3 default of `method="json_schema"` to `method="function_calling"`. That change is correct and now merged, but it landed without a test, and it is invisible at the call site: delete the keyword and all 987 agent-common tests still pass, while risk classification silently degrades in production.

## What the two tests pin

**1. The classification path must ask for `method="function_calling"`.**
Verified live against a gateway fronting several providers, with a risk-bearing tool (`fs_delete_path`, params `path` / `recursive` / `confirm`):

| model | `json_schema` (the default) | `function_calling` (#188) |
|---|---|---|
| `gpt-4o-2024-08-06` (Azure) | **400, 3/3** — `Invalid schema for response_format 'ToolRiskOutput': 'required' is required to be supplied and to be an array including every key in properties. Extra required key 'risky_values' supplied.` | OK, `base_score=0.9`, factors `path`, `recursive`, `confirm` |
| `claude-haiku-4-5` (Bedrock) | OK, but **`risk_factors={}`, 3/3** | OK, `base_score=0.95`, factors `path`, `recursive` |
| `gemini-3.5-flash`, `gemini-3.1-pro-preview`, `claude-sonnet-4.6` | OK | OK |

One defect, two symptoms. The rejected key is `risky_values`, the open `dict` inside `RiskyValuesOutput`: strict mode cannot express an open map, so Azure refuses the request while Bedrock/Vertex accept it and drop the map.

The silent symptom is the worse one. An empty `risk_factors` gets persisted with a **real** schema hash, so — unlike a hard failure, which falls through to the deterministic fallback and retries next time — it never self-corrects, and per-argument risk matching stays off for that tool permanently. See #199 / #201 for why an un-self-correcting risk row matters.

`langchain_openai` also emits a client-side `UserWarning` for this schema ("Specify `method=\"function_calling\"` instead"), so it is not gateway-specific.

**2. `ToolRiskOutput` still has the open maps strict mode cannot express.**
If that test ever fails, the schema has been flattened to fixed keys and the method override may no longer be needed — the point is that the two get revisited together rather than either being deleted in isolation.

## Verification

- `test_risk_classification_asks_for_tool_calling_not_strict_json_schema` fails when `method="function_calling"` is removed from the call (mutation-checked: `assert None == 'function_calling'`), and passes with it.
- Full agent-common suite: **987 passed**.
- `ruff check`: 4 pre-existing findings in this file on `origin/main`, same 4 after — this change adds none.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no, tests only.

### DIFF
--- a/packages/agent-common/tests/test_tool_risk_scorer.py
+++ b/packages/agent-common/tests/test_tool_risk_scorer.py
@@ -14,6 +14,9 @@ from agent_common.core.tool_risk_scorer import (
     _DESTRUCTIVE_FLOOR_SCORE,
     _destructive_floor,
     _deterministic_fallback,
+    _score_tool_via_llm,
+    RiskyValuesOutput,
+    ToolRiskOutput,
     score_tool_risk,
 )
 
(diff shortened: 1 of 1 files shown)


aartaria:
closed in favour of #233

PR #205, opened 2026-09-07
fix(scheduler): rank draft-generator tools server-side, fail loudly on an empty draft

### LINKED ISSUE #200: scheduler: generate-job-draft returns a 200 of nulls — the whole tool catalogue goes into the prompt, and an empty generation is reported as success
## Problem

`POST /api/v1/scheduler/generate-job-draft` answers `200 OK` with every field `null` for a request as ordinary as *"notify me every time a new bug report is filed"*. The form silently stays empty and the user has no idea why.

Two separate defects, stacked.

### 1. An empty generation is rendered as a successful draft

`gateway_chat_json` returns `{}` when it cannot find a JSON object in the model's reply. That is deliberate, and its docstring says why:

> Returns `{}` when there is no object to be found, which every caller already treats as "no usable output" and answers with its own fallback.

`generate_job_draft` has no such fallback. `{}` flows into `_build_draft`, which skips every `None` value, and the endpoint returns `ScheduledJobDraft()` — all fields empty — with a `200`. A total generation failure is therefore indistinguishable from "the request implied nothing", both to the UI and in the logs, where nothing is recorded at all.

The two other callers of that helper (condition generation, the watch judge) do handle the empty case.

### 2. The prompt carries the entire tool catalogue

`SchedulerPage.tsx` fetches the full MCP tool list and posts it verbatim:

```ts
const { data: mcpToolsData } = useQuery(consoleListMcpToolsOptions());
const mcpTools = mcpToolsData?.tools ?? [];
…
const result = await generateJobDraft(mcpTools as unknown as Record<string, unknown>[], aiQuery);
```

and the endpoint serialises all of it into the prompt — name, description **and full `input_schema`** per tool, at `indent=2`. In one environment the discovered registry holds ~560 tools across 33 servers, which is on the order of a hundred thousand tokens of schema, sent to the `chat:low` tier for an answer that is twenty lines of JSON (`max_tokens=1024`). The signal the model needs — one tool name — is buried in the noise, and the reply comes back unparseable, which lands in defect 1.

### And the tool list should not come from the client at all

The endpoint reads sub-agents and delivery channels server-side on purpose, per its own docstring:

> The sub-agents and delivery channels the model may choose from are read here rather than accepted from the caller, so a generated job can only ever reference something the user can already reach.

with a comment recording that a wider offer (`is_admin=True`) produced jobs the user could not save. Tools break that rule: they arrive in the request body. So the payload is both enormous and untrusted, and a caller can offer the generator tools the user cannot reach.

## Proposal

**Rank server-side; send a handful.** The scorer already exists — `console_grep_mcp_tools` (`mcp_router.py`) scores tools against a query across name, description and schema with documented weights. Reuse it against `data.query`, put roughly the top 15 into the prompt with their schemas, and read the catalogue server-side via the same path that endpoint uses. Keep accepting the request's `tools` field but ignore it, so a deploy does not break the UI mid-flight, then drop it.

**Fail loudly.** When the model produces nothing usable after the repair pass, answer with an error (503, matching the existing "AI generation service unavailable" path) instead of a `200` of nulls, and log the query plus the raw reply length so the next occurrence is diagnosable. A draft where the model genuinely inferred only some fields is a different outcome and should stay a `200`.

## Notes

- Prediction worth checking first: the same request against a small tool list should succeed today. That would confirm prompt size as the cause rather than something in the prompt's instructions.
- `generate-condition` takes the same shape of input but is narrower (it sees one tool's real response), so it is not affected by the catalogue size — worth confirming rather than assuming.

## Acceptance

- [ ] the draft endpoint selects candidate tools server-side from the user's own catalogue, and no longer depends on the request's `tools` field
- [ ] the prompt carries a bounded number of tools rather than the whole registry
- [ ] an unusable generation returns an error, not a `200` with an empty draft, and is logged
- [ ] "notify me every time a new bug report is filed" produces a watch on `console_list_bug_reports`

aartaria:
closes ringier-data/nannos#200

## What

`POST /api/v1/scheduler/generate-job-draft` answered `200 OK` with every field `null` for a request as ordinary as *"notify me every time a new bug report is filed"*. Two defects, stacked (see the issue):

1. **The prompt carried the whole tool catalogue.** The UI posted every MCP tool with its full `input_schema`; the endpoint serialised all of it into a `chat:low` prompt. The one tool name the model needed was buried in ~100k tokens of schema and the reply came back unparseable. Tools were also the only offer that came from the request body, unlike sub-agents and channels which are read server-side so a draft can only reference what the user can reach.
2. **An empty generation was rendered as a successful draft.** `gateway_chat_json` returns `{}` when it finds no JSON object, by contract; the draft path had no fallback, so `{}` became `ScheduledJobDraft()` and a total failure was reported as success with nothing in the logs.

## How

**Backend** (`console-backend`)
- The scorer behind `/mcp/tools/search` is extracted into a pure `rank_mcp_tools(tools, query, top_k)`; the search endpoint calls it unchanged.
- `generate_job_draft` reads the user's own catalogue via `_list_mcp_tools`, ranks it against the query and offers the top `_DRAFT_TOOL_CANDIDATES = 15` with schemas. A `check_tool` outside the offered set is discarded, like an unreachable sub-agent id. An unreadable catalogue is a 503, not a tool-less prompt.
- A draft with nothing in it is a **422** (`"The model produced no usable draft — rephrase the request."`) with the query, model, candidate count and raw reply keys logged; `gateway_chat_json` now also logs (and returns `{}`) when it finds no object or an unparseable one. A partly-filled draft stays a 200.
  - ⚠️ **Deviation from the issue:** #200 proposed 503. Review round 1 pointed out that `generate-condition` answers the identical case with 422 and that a retry-on-503 layer would re-run a deterministic content miss, so this PR uses 422. Easy to flip back if 503 is preferred.
- An invented `check_tool` is discarded together with its `check_args`/`check_args_exprs`/`cel_expr`/`llm_condition`; a non-string `check_tool` is dropped rather than 500ing.
- `GenerateJobDraftRequest.tools` is removed. Backend and frontend ship as one console package, so there is no older UI to keep compatible.

**Frontend** (`console-frontend`)
- `generateJobDraft(query)` no longer posts the catalogue; the Generate button no longer waits for the catalogue to load. The create dialog now shows the backend error detail instead of a generic message. The full tool list is still fetched for the form's tool picker, as before.
- SDK regenerated from the exported OpenAPI spec (only the request type and endpoint description changed).

## Verification

- New `tests/test_generate_job_draft_endpoint.py` (15 tests): a bounded, ranked handful in the prompt; the request body carries only `query`; the bug-report query offers and accepts `console_list_bug_reports`; invented tool discarded; unreadable catalogue → 503; no-JSON, two-objects-in-prose and unknown-fields replies → 422 with the query logged (patched at the raw-text level so the salvage runs); partial draft → 200; `rank_mcp_tools` ordering/cap/purity.
- console-backend full suite: 1709 passed. `ruff check` clean on touched files.
- console-frontend: `tsc -b` and `eslint` clean.
- Not verified against a live model: the acceptance criterion "the bug-report request produces a watch on `console_list_bug_reports`" is covered at the ranking level (that tool ranks first for the query against a 31-tool catalogue) and by the endpoint accepting it; a manual run against a real deployment is worth doing before closing.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — The `tools` request field is removed from `generate-job-draft`. Frontend and backend deploy together as the console, so no consumer is left behind.

### DIFF
--- a/packages/console-backend/console_backend/routers/scheduler_router.py
+++ b/packages/console-backend/console_backend/routers/scheduler_router.py
@@ -254,6 +254,22 @@ def _coerce_id(value: object, allowed: set[int]) -> int | None:
     return candidate
 
 
+def _coerce_name(value: object, allowed: set[str]) -> str | None:
+    """A generated name, or None when it is not a string or not among the offered ones.
+
+    The string guard matters: a model asked for "the single best-matching tool" answers
+    with a list when several match, and a bare `in` on a set would raise on it.
+    """
+    if not isinstance(value, str):
+        if value is not None:
+            logger.info("Discarding generated name %r: not a string", value)
+        return None
+    if value not in allowed:
+        logger.info("Discarding generated name %r outside the offered set", value)
+        return None
+    return value
+
+
 def _agent_choices(sub_agents: list[SubAgent]) -> list[dict[str, Any]]:
     """The sub-agents a generated job may pick from, as the prompt sees them.
 
@@ -309,19 +325,38 @@ async def _candidate_tools(request: Request, user: User, query: str) -> list[MCP
     except HTTPException:
         raise
     except Exception as exc:
-        logger.warning("Could not read the tool catalogue for draft generation: %s", exc)
+        # When the gateway starts failing this is the only line an operator gets, so
+        # it carries the upstream status where there is one, and the traceback.
+        upstream = getattr(getattr(exc, "response", None), "status_code", None)
+        logger.warning(
+            "Could not read the tool catalogue for draft generation (%s, upstream status %s)",
+            type(exc).__name__,
+            upstream,
+            exc_info=True,
+        )
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
             detail="Could not read the tool catalogue — try again shortly.",
         ) from exc
     candidates = rank_mcp_tools(catalogue.tools, query, _DRAFT_TOOL_CANDIDATES)
-    logger.info(
-        "Draft generation offers %d of %d tools for query %r: %s",
-        len(candidates),
-        len(catalogue.tools),
-        query,
-        [t.name for t in candidates],
-    )
+    # An empty offer is legitimate — a task job needs no tool, and the request may
+    # simply share no vocabulary with the catalogue — but it is worth its own line, since
+    # the model can then only name a tool from memory, which is discarded afterwards.
+    if not candidates:
+        logger.info(
+            "Draft generation offers no tools for query %r (%d in the catalogue%s)",
+            query,
+            len(catalogue.tools),
+            "" if catalogue.tools else " — empty, e.g. under impersonation",
+        )
+    else:
+        logger.info(
+            "Draft generation offers %d of %d tools for query %r: %s",
+            len(candidates),
+            len(catalogue.tools),
+            query,
+            [t.name for t in candidates],
+        )
     return candidates
 
 
@@ -351,9 +386,11 @@ async def generate_job_draft(
     request and cut to a handful — see `_DRAFT_TOOL_CANDIDATES`.
     """
     candidate_tools = await _candidate_tools(request, current_user, data.query)
+    # Compact on purpose: pretty-printed schema is a third more tokens for whitespace the
+    # model does not need, and the whole prompt is re-sent on every CEL repair round.
     tools_summary = json.dumps(
         [{"name": t.name, "description": t.description, "input_schema": t.input_schema} for t in candidate_tools],
-        indent=2,
+        separators=(",", ":"),
     )
     allowed_tool_names = {t.name for t in candidate_tools}
     # Offer only what this user can reach: the model picks from these, and anything
@@ -506,11 +543,14 @@ async def generate_job_draft(
     generated["sub_agent_id"] = _coerce_id(result.get("sub_agent_id"), allowed_agent_ids)
     generated["delivery_channel_id"] = _coerce_id(result.get("delivery_channel_id"), allowed_channel_ids)
     # A tool name outside the offered set is an invention — the model saw only the
-    # candidates — and would produce a job whose check fails on its first run.
-    check_tool = result.get("check_tool")
-    if check_tool is not None and check_tool not in allowed_tool_names:
-        logger.info("Discarding generated check_tool %r: not among the offered tools", check_tool)
-        generated["check_tool"] = None
+    # candidates — and would produce a job whose check fails on its first run. The
+    # arguments and condition were written against that invented tool, so they go with
+    # it: the form applies each of them independently, and pre-filled arguments under a
+    # tool the user then picks by hand are exactly the first-run failure being avoided.
+    generated["check_tool"] = _coerce_name(result.get("check_tool"), allowed_tool_names)
+    if generated["check_tool"] is None and result.get("check_tool") is not None:
+        for key in ("check_args", "check_args_exprs", "cel_expr", "llm_condition"):
+            generated.pop(key, None)
     if isinstance(generated.get("check_args"), str):
         try:
             generated["check_args"] = json.loads(generated["check_args"])
@@ -522,8 +562,9 @@ async def generate_job_draft(
     # generation that produced no usable output (no JSON in the reply, or none of the
     # fields asked for). Rendering that as a 200 left the form silently empty and the
     # logs silent with it. A draft the model filled only partly is a different outcome
-    # and stays a success.
-    if not draft.model_dump(exclude_none=True):
+    # and stays a success. 422 rather than 503, as /generate-condition answers the same
+    # case: the service is up, and a retry-on-503 layer must not re-run a content miss.
+    if not draft.model_fields_set:
         logger.warning(
             "Draft generation produced nothing usable for query %r (model %s, %d candidate tools, raw keys %s)",
             data.query,
@@ -532,8 +573,8 @@ async def generate_job_draft(
             sorted(result),
         )
         raise HTTPException(
-            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-            detail="The model produced no usable draft — try again or rephrase the request.",
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="The model produced no usable draft — rephrase the request.",
         )
     return draft
 
--- a/packages/console-backend/console_backend/routers/scheduler_router.py
+++ b/packages/console-backend/console_backend/routers/scheduler_router.py
@@ -49,10 +49,19 @@ from ..services.llm_gateway import gateway_chat_json
 from ..services.scheduler_engine import SchedulerEngine
 from ..services.scheduler_service import _UNSET, SchedulerService
 from ..utils.timezones import resolve_timezone
+from .mcp_router import MCPTool, _list_mcp_tools, rank_mcp_tools
 
 logger = logging.getLogger(__name__)
 
 router = APIRouter(prefix="/api/v1/scheduler")
+
+#: How many tools the draft generator shows the model. The registry can hold hundreds
+#: of tools across dozens of servers — with every schema included that is on the order
+#: of a hundred thousand tokens for an answer of twenty lines, and the one name the
+#: model needs is buried in it (the reply came back unparseable, as a 200 of nulls).
+#: Fifteen leaves room for a query that matches a family of tools (list/get/search
+#: variants across servers) while keeping the prompt within a few thousand tokens.
+_DRAFT_TOOL_CANDIDATES = 15
 
 
 #: Why check_args must never carry a date: they are stored once and sent unchanged on
@@ -287,14 +296,45 @@ def _get_scheduler_service(request: Request) -> SchedulerService:
     return request.app.state.scheduler_service  # type: ignore[no-any-return]
 
 
+async def _candidate_tools(request: Request, user: User, query: str) -> list[MCPTool]:
+    """The tools worth offering the draft generator for `query`: the user's own catalogue,
+    ranked by the search endpoint's scorer, cut to `_DRAFT_TOOL_CANDIDATES`.
+
+    Read here rather than accepted from the caller for the same reason the sub-agents
+    and channels are: a generated job may only reference what this user can reach, and
+    the request body is not where that is decided.
+    """
+    try:
+        catalogue = await _list_mcp_tools(request, user)
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.warning("Could not read the tool catalogue for draft generation: %s", exc)
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="Could not read the tool catalogue — try again shortly.",
+        ) from exc
+    candidates = rank_mcp_tools(catalogue.tools, query, _DRAFT_TOOL_CANDIDATES)
+    logger.info(
+        "Draft generation offers %d of %d tools for query %r: %s",
+        len(candidates),
+        len(catalogue.tools),
+        query,
+        [t.name for t in candidates],
+    )
+    return candidates
+
+
 @router.post(
     "/generate-job-draft",
     response_model=ScheduledJobDraft,
     summary="Draft a whole scheduled job from a one-line description.",
     description=(
-        "Given the available MCP tools and a natural-language request, returns a partial "
-        "ScheduledJobCreate: job type, schedule, check tool and arguments, condition, "
-        "outcome and delivery. Fields it cannot infer are omitted for the caller to fill in."
+        "Given a natural-language request, returns a partial ScheduledJobCreate: job type, "
+        "schedule, check tool and arguments, condition, outcome and delivery. The tools, "
+        "sub-agents and channels the draft may reference are the caller's own, read "
+        "server-side. Fields it cannot infer are omitted for the caller to fill in; a "
+        "generation that infers nothing at all is an error, not an empty draft."
     ),
 )
 async def generate_job_draft(
@@ -305,17 +345,17 @@ async def generate_job_draft(
 ) -> ScheduledJobDraft:
     """Generate a whole job from one sentence: type, schedule, tool, condition, outcome.
 
-    The sub-agents and delivery channels the model may choose from are read here rather
-    than accepted from the caller, so a generated job can only ever reference something
-    the user can already reach.
+    The tools, sub-agents and delivery channels the model may choose from are read here
+    rather than accepted from the caller, so a generated job can only ever reference
+    something the user can already reach. Tools are additionally ranked against the
+    request and cut to a handful — see `_DRAFT_TOOL_CANDIDATES`.
     """
+    candidate_tools = await _candidate_tools(request, current_user, data.query)
     tools_summary = json.dumps(
-        [
-            {"name": t.get("name"), "description": t.get("description"), "input_schema": t.get("input_schema")}
-            for t in data.tools
-        ],
+        [{"name": t.name, "description": t.description, "input_schema": t.input_schema} for t in candidate_tools],
         indent=2,
     )
+    allowed_tool_names = {t.name for t in candidate_tools}
     # Offer only what this user can reach: the model picks from these, and anything
     # outside the offered ids is discarded after the call.
     # The same set create_job validates against — offering anything wider produces a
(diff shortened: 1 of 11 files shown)


packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -465,13 +505,37 @@ async def _generate(instruction: str) -> dict[str, Any]:
     generated["schedule_kind"] = _coerce_enum(ScheduleKind, result.get("schedule_kind"))
     generated["sub_agent_id"] = _coerce_id(result.get("sub_agent_id"), allowed_agent_ids)
     generated["delivery_channel_id"] = _coerce_id(result.get("delivery_channel_id"), allowed_channel_ids)
+    # A tool name outside the offered set is an invention — the model saw only the
+    # candidates — and would produce a job whose check fails on its first run.
+    check_tool = result.get("check_tool")
+    if check_tool is not None and check_tool not in allowed_tool_names:
```
alloy-cicd: **Crash path:** `check_tool not in allowed_tool_names` does a set-membership test on an unvalidated model value. If the model answers `"check_tool": ["a", "b"]` or `{"name": …}` (a real failure mode for "the single best-matching tool" when several match), `in` on a set raises `TypeError: unhashable type` — and this line is outside the `try` around `_generate`, so the endpoint 500s. Before this PR a non-string `check_tool` was quietly dropped by `_build_draft`'s per-field `ValidationError` handling.

Every neighbour is type-guarded (`_coerce_enum` checks `isinstance(value, str)`, `_coerce_id` catches `TypeError`). The consistent fix is a string sibling, e.g. `_coerce_name(value, allowed) -> str | None` with an `isinstance(value, str)` guard, assigned like the others: `generated["check_tool"] = _coerce_name(result.get("check_tool"), allowed_tool_names)`. That fixes the crash and keeps one idiom for "hallucinated reference → log + None".

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -465,13 +505,37 @@ async def _generate(instruction: str) -> dict[str, Any]:
...
+        logger.info("Discarding generated check_tool %r: not among the offered tools", check_tool)
+        generated["check_tool"] = None
```
alloy-cicd: **Orphaned companions:** nulling `check_tool` leaves `check_args`, `check_args_exprs`, `cel_expr` and `llm_condition` in the draft — all written against the tool that was just deleted. The frontend applies them independently (`if (result.check_args) …` at SchedulerPage.tsx:368, SchedulerJobDetailPage.tsx:394), and `argsModeFor(args, mcpTools.find(t => t.name === result.check_tool))` then runs with `undefined` for the tool.

Concrete failure: model invents `github_list_issues` with `check_args: {"repo": "x"}` and a `cel_expr` over `result.issues`. The form opens with no tool but pre-filled args and condition; the user picks a real tool from the dropdown, saves, and the job fails on its first run with an argument error — the exact outcome this discard was added to prevent. When the tool is rejected, the whole check block should go with it.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -287,14 +296,45 @@ def _get_scheduler_service(request: Request) -> SchedulerService:
...
+        catalogue = await _list_mcp_tools(request, user)
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.warning("Could not read the tool catalogue for draft generation: %s", exc)
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="Could not read the tool catalogue — try again shortly.",
+        ) from exc
+    candidates = rank_mcp_tools(catalogue.tools, query, _DRAFT_TOOL_CANDIDATES)
```
alloy-cicd, line 341: **Empty candidate set passes through silently.** Two real paths produce `candidates == []`:

1. Admin impersonation — `_list_mcp_tools` returns `MCPToolsResponse(tools=[])` by design (mcp_router.py:309).
2. A query with no lexical overlap — `rank_mcp_tools` drops every tool scoring 0 (e.g. "tell me when CI turns red" shares no token with `github_list_workflow_runs`).

In both cases the paid LLM call still happens with `Available tools:\n[]`, any `check_tool` the model names from memory is discarded at L511, and the user gets either a watch draft with a silently blank check tool or a 503 that misattributes the cause to the model. The old frontend guard (`mcpTools.length === 0` disabling Generate) that used to cover case 1 was removed on both pages, so nothing gates this any more. Consider treating an empty candidate set for a query explicitly — fail with a distinct message, or fall back to an unranked top-N so the ranker is a filter, not a hard gate.

Minor, same function: the `except Exception` branch logs only `logger.warning(..., exc)` — no status code, no traceback — where the sibling path in `grep_mcp_tools` distinguishes `HTTPStatusError`/`ConnectError`/`Timeout` and logs the upstream status. When the gateway starts failing, this is the only line an operator gets.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -465,13 +505,37 @@ async def _generate(instruction: str) -> dict[str, Any]:
...
+        logger.warning(
+            "Draft generation produced nothing usable for query %r (model %s, %d candidate tools, raw keys %s)",
+            data.query,
+            model,
+            len(candidate_tools),
+            sorted(result),
+        )
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="The model produced no usable draft — try again or rephrase the request.",
```
alloy-cicd: **Status-code inconsistency:** the sibling endpoint in this router answers the identical situation with 422 — `generate_condition` raises `HTTP_422_UNPROCESSABLE_ENTITY, detail="The model produced no usable condition — rephrase the request."`. Here the same cause gets 503, which this handler already uses for two genuinely different conditions: missing model config (L448, never recovers on retry) and gateway failure (L466, transient). Three unrelated causes now share one status, so a retry-on-503 layer re-runs a deterministic content failure, and the frontend can't distinguish "rephrase" from "try later" from "admin must configure a model". Matching `generate_condition`'s 422 would fix both the inconsistency and the retry semantics.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -465,13 +505,37 @@ async def _generate(instruction: str) -> dict[str, Any]:
...
             generated["check_args"] = None
 
-    return _build_draft(generated)
+    draft = _build_draft(generated)
+    # A draft with nothing in it is not "the request implied nothing" — it is a
+    # generation that produced no usable output (no JSON in the reply, or none of the
+    # fields asked for). Rendering that as a 200 left the form silently empty and the
+    # logs silent with it. A draft the model filled only partly is a different outcome
+    # and stays a success.
+    if not draft.model_dump(exclude_none=True):
```
alloy-cicd: Minor: `not draft.model_dump(exclude_none=True)` answers "did the generator fill anything?" via a serialization round-trip. It works today because every `ScheduledJobDraft` field defaults to `None`, but the day a draft field gains a non-None default (or is excluded from dumps) the guard silently never fires again, and no test catches it — the suite only exercises the all-None case. Since `_build_draft` already computes the `accepted` dict, returning/checking that (`if not accepted`) — or `if not draft.model_fields_set` — states the intent directly and is immune to serialization changes.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -305,17 +345,17 @@ async def generate_job_draft(
...
-    The sub-agents and delivery channels the model may choose from are read here rather
-    than accepted from the caller, so a generated job can only ever reference something
-    the user can already reach.
+    The tools, sub-agents and delivery channels the model may choose from are read here
+    rather than accepted from the caller, so a generated job can only ever reference
+    something the user can already reach. Tools are additionally ranked against the
+    request and cut to a handful — see `_DRAFT_TOOL_CANDIDATES`.
     """
+    candidate_tools = await _candidate_tools(request, current_user, data.query)
     tools_summary = json.dumps(
```
alloy-cicd, line 391: Nit: `indent=2` on the tool schemas inflates the very token budget the `_DRAFT_TOOL_CANDIDATES` comment is managing — pretty-printed JSON Schema runs roughly 1.3–1.6× the tokens of compact JSON, and the whole prompt is re-sent verbatim on every `_repair_cel` round-trip. `separators=(",", ":")` (or just dropping `indent`) buys back a chunk of that for free; the model doesn't need the whitespace.

alloy-cicd (commented):

The move server-side (catalogue read for the authenticated user, ranked cut, fail-loudly-on-empty) is the right shape, and the new test module covers the headline behaviours well. The findings below are ordered by severity: one crash path, one draft-coherence bug, one silent degraded mode, and a handful of consistency/diagnosability points.

**Correctness**
1. `check_tool not in allowed_tool_names` raises `TypeError` (→ 500) when the model returns a non-hashable value — every neighbouring generated value goes through a type-guarded `_coerce_*` helper, this one doesn't.
2. Discarding a hallucinated `check_tool` leaves its `check_args` / `check_args_exprs` / `cel_expr` in the draft; the form applies them independently, so the user can save a job whose arguments belong to a different tool.
3. An *empty* candidate set is not handled: admin impersonation returns `tools=[]` by design, and a query with no lexical overlap ranks to `[]` — the LLM is still called with `Available tools: []`, and the old `mcpTools.length === 0` button guard that used to cover this was removed on both pages.

**Consistency / diagnosability**
4. The sibling endpoint `generate_condition` answers the identical "model produced nothing usable" case with **422**; here it's 503, the same status already used for missing model config and gateway failure.
5. The create dialog still swallows the error body (`catch { setError('AI generation failed…') }`), so the new 503 details never reach the user on the primary entry point — the detail page uses `formatApiError(e)`.
6. In `gateway_chat_json`, the new warning covers the no-JSON branch but `json.loads` on the greedy `\{.*\}` match can still raise one line below, unlogged, misattributed as "AI generation service unavailable".

Smaller notes inline (emptiness test via `model_dump`, `indent=2` prompt cost, test fixture patching one level too high). Not raised as blocking: the cross-router import of the private `_list_mcp_tools` (a `services/` home for catalogue read + ranking would be cleaner), and the catalogue fetch being serialized in front of the DB reads and the LLM call with no caching — both worth a thought if this endpoint gets traffic.

packages/console-backend/console_backend/services/llm_gateway.py
```diff
@@ -179,6 +179,9 @@ async def gateway_chat_json(
     cleaned = re.sub(r"```(?:json)?\s*", "", text).strip().rstrip("`")
     match = re.search(r"\{.*\}", cleaned, re.DOTALL)
     if not match:
+        # The caller decides what an empty answer means; this is the only place that
+        # knows why it is empty, so the shape of the reply is recorded here.
+        logger.warning("No JSON object in the model's reply (%d chars) for model %s", len(text), model)
```
alloy-cicd, line 184: The new warning covers the no-JSON branch, but the line below can still raise unlogged: `re.search(r"\{.*\}", …, DOTALL)` is greedy, so a prose reply containing two objects (`…option {"job_type": "watch"} or maybe {"job_type": "task"}`) matches from the first `{` to the last `}` and `json.loads` raises `JSONDecodeError`. In `generate_job_draft` that's swallowed by the broad `except Exception` into 503 "AI generation service unavailable" — the wrong cause, and the diagnosability this PR adds never fires for exactly the case it targets. Wrapping the `json.loads` in `try/except JSONDecodeError` → same warning + `return {}` closes the gap and keeps the documented contract ("returns {} when there is no object to be found").

packages/console-frontend/src/pages/SchedulerPage.tsx
```diff
@@ -323,10 +323,7 @@ function CreateJobDialog({
     setAiLoading(true);
     setError(null);
     try {
-      const result = await generateJobDraft(
-        mcpTools as unknown as Record<string, unknown>[],
-        aiQuery,
-      );
+      const result = await generateJobDraft(aiQuery);
```
alloy-cicd, line 327: The new 503 details ("rephrase the request" / "Could not read the tool catalogue — try again shortly") never reach the user here: this handler's `catch` at L419 discards the error object and always shows "AI generation failed. Please fill in the fields manually." The detail page's handler in this same PR does it right (`formatApiError(e)` at SchedulerJobDetailPage.tsx:420). So on the primary entry point, a model miss, a gateway outage and a missing model default all look identical, and the one remedy that works (rephrase) is never suggested — the legibility this PR adds is only half-delivered. Suggest mirroring the detail page's `catch (e) { setError(\`AI generation failed: ${formatApiError(e)}…\`) }`.

packages/console-backend/tests/test_generate_job_draft_endpoint.py
```diff
@@ -0,0 +1,250 @@
...
+    scheduler_service._resolve_timezone = AsyncMock(return_value="Europe/Zurich")
+    app.state.scheduler_service = scheduler_service
+    app.state.delivery_channel_repository = MagicMock(
+        list_all_channels=AsyncMock(return_value=[])
+    )
+    return TestClient(app)
+
+
+@pytest.fixture
+def gateway():
```
alloy-cicd: Two test-fidelity notes:

1. Patching `gateway_chat_json` (the parser) means `test_no_json_in_the_reply_is_a_503` never executes the code path its comment describes — the new `logger.warning("No JSON object in the model's reply…")` at llm_gateway.py:184 is unreached by this whole module. The sibling suite (tests/test_generate_condition.py) patches one level deeper at `gateway_chat` and stubs the raw reply text, which exercises the salvage/parse path for real.
2. The `client` fixture shadows the repo-wide `client` from tests/conftest.py, silently changing its type (async `httpx.AsyncClient` everywhere else, sync `TestClient` here). A local name like `draft_client` avoids the trap for the next person copying assertions between files.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Fixed in 2df3e76a: `check_tool` now goes through `_coerce_name`, the string sibling of `_coerce_id` (isinstance guard, log + None). Test: `test_a_non_string_tool_name_is_dropped_not_a_500`.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Fixed in 2df3e76a: when the tool is rejected, `check_args`, `check_args_exprs`, `cel_expr` and `llm_condition` are dropped with it. Test: `test_a_tool_the_model_invented_goes_with_its_arguments_and_condition`.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Partly taken in 2df3e76a. The catalogue-failure branch now logs the exception type, upstream status and traceback, and an empty candidate set gets its own info line (distinguishing an empty catalogue, e.g. impersonation, from no lexical overlap). Not taken: failing or padding with an unranked top-N. An empty offer is legitimate for a task job, and an unranked slice of the catalogue would be arbitrary tools rather than relevant ones — the model would be steered toward a random subset. If we want a fallback for zero-overlap watch requests, that is a ranking-quality question (e.g. treating `_` as a word boundary in the scorer) worth its own issue rather than this PR. The removed frontend guard only covered the case by accident (it also blocked Generate while the catalogue was loading).

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Agreed and changed in 2df3e76a: 422 with the same wording as `generate_condition`. Note for the author: issue #200 proposed 503 — flagged in the PR body so it can be vetoed.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Fixed in 2df3e76a: `if not draft.model_fields_set` — `_build_draft` constructs the model from the accepted dict, so this is exactly the set of fields the generator filled.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Fixed in 2df3e76a: `separators=(",", ":")`, no indent. The test probe was adjusted to the compact shape.

packages/console-backend/console_backend/services/llm_gateway.py
aartaria: Fixed in 2df3e76a: `json.loads` is wrapped, logs the reply length and error, and returns `{}` per the documented contract. Test: `test_two_objects_in_prose_are_a_422_not_a_gateway_outage`.

packages/console-frontend/src/pages/SchedulerPage.tsx
aartaria: Fixed in 2df3e76a: the create dialog's catch now mirrors the detail page and shows `formatApiError(e)`.

packages/console-backend/tests/test_generate_job_draft_endpoint.py
aartaria: Both taken in 2df3e76a: the no-JSON / two-objects / fenced-valid cases patch `gateway_chat` and feed raw text, so the salvage and both new warnings run for real; the fixture is renamed `draft_client`.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (f641eeaa67c3469790eda0bda0d5d38861b10238..2df3e76a62a193930bc003a96173e9e515b54d8a)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | `TypeError` → 500 on non-string `check_tool` | **verified-fixed** — `_coerce_name` with `isinstance` guard, assigned like its `_coerce_id` sibling; covered by `test_a_non_string_tool_name_is_dropped_not_a_500` |
| 2 | Orphaned `check_args`/`check_args_exprs`/`cel_expr`/`llm_condition` after discarding an invented tool | **verified-fixed** — the whole check block is popped when a provided tool is rejected, and only then (a legitimately tool-less `llm_condition` survives); covered by `test_a_tool_the_model_invented_goes_with_its_arguments_and_condition` |
| 3 | Empty candidate set passes through silently | **partly fixed / partly rejected-by-author** — the diagnosability half is verified-fixed (catalogue-failure log now carries exception type, upstream status and traceback; an empty offer gets its own line distinguishing an empty catalogue from zero overlap, with a test). The fallback half (fail or pad with unranked top-N) was rejected with reasoning I find sound: an empty offer is legitimate for task jobs, and an unranked slice would steer the model toward arbitrary tools. The ranking-quality follow-up (e.g. `_` as a word boundary in the scorer) deserves its own issue as the author suggests. |
| 4 | 503 for a content miss, inconsistent with `generate_condition`'s 422 | **verified-fixed** — now 422 with matching wording; the divergence from issue #200's proposed 503 is flagged in the PR body for the human reviewer to veto |
| 5 | Emptiness test via `model_dump` round-trip | **verified-fixed** — `model_fields_set`, which is exactly the accepted dict since `_build_draft` skips None and invalid values |
| 6 | `indent=2` prompt token cost | **verified-fixed** — compact `separators=(",", ":")`, and the test probe asserts it matches the compact shape |
| 7 | Uncaught `JSONDecodeError` on the greedy salvage in `gateway_chat_json` | **verified-fixed** — wrapped, logged, returns `{}` per the documented contract; exercised for real by `test_two_objects_in_prose_are_a_422_not_a_gateway_outage` on the raw-text path |
| 8 | Create dialog swallowing the error detail | **verified-fixed** — `formatApiError(e)`, mirroring the detail page |
| 9 | Tests patching one level too high; `client` fixture shadowing conftest | **verified-fixed** — `raw_gateway` patches `gateway_chat` so the salvage and both new warnings run for real (plus a fenced-valid-reply case); fixture renamed `draft_client` |

I also checked the fixes themselves for regressions: the always-assigned `generated["check_tool"]` is None-skipped by `_build_draft` so it cannot pollute `model_fields_set`; the companion-drop condition (`coerced is None and result.get("check_tool") is not None`) correctly spares a draft where the model legitimately omitted the tool; and the new 422's string `detail` renders correctly through `formatApiError`. Nothing to flag.

PR #206, opened 2026-09-07
feat: key the orchestrator discovery cache on a per-user entitlement version

### LINKED ISSUE #171: orchestrator: fan discovery-cache invalidation out to every replica
## Summary

The orchestrator's per-user discovery cache (`app/core/discovery_cache.py`) is in-process. Console-backend invalidates it by POSTing `/internal/discovery-cache/invalidate` when a user's entitlements change (tool whitelist, role, bypass rules, group default agents, group→MCP-server access) — but that POST reaches **one replica**. Behind a load balancer the other replicas keep serving the old entry until `AGENT_DISCOVERY_CACHE_TTL` expires.

Until now the TTL was kept at 60 s partly for this reason (and partly because discovered tools embedded bearer tokens — no longer the case since #170: bearers are minted at call time by a per-user provider). With the credential concern gone, the TTL's only remaining jobs are (a) bounding how long a catalogue change made *on the gateway* goes unnoticed and (b) bounding the multi-replica invalidation lag. Raising the TTL to minutes (desirable: it removes ~2 s of cold discovery from most turns in low-traffic environments) widens (b) for **revocations** — grants merely arrive late, but a removed tool/server/agent stays visible on N−1 replicas for up to one TTL.

## Proposal

Deliver invalidations to every replica instead of one, so the TTL can be set on freshness grounds alone:

- Reuse an existing fan-out channel rather than adding infrastructure — e.g. the pub/sub / live-update channel console-backend already uses to push events to frontends, or a lightweight Postgres `LISTEN/NOTIFY` on the shared database every replica already connects to. Each replica subscribes at startup and applies `invalidate_users(user_subs)` / `invalidate_all()` locally.
- Keep the HTTP endpoint as a fallback/manual lever; make the broadcast idempotent (a replica receiving both is fine).
- Optionally also broadcast when an admin changes gateway-facing config through the console, so (a) shrinks too.
- Then raise `AGENT_DISCOVERY_CACHE_TTL` (gitops) to the 5–15 min range.

## Acceptance

- With ≥2 orchestrator replicas, a console-side entitlement change is reflected on the user's next turn on **any** replica, without waiting for the TTL.
- A replica that missed a broadcast (restart, network blip) still converges within the TTL (today's behaviour as the floor).
- Unit test for the subscriber; an integration note on how to verify with two local processes.

## Context

Analysis in #154 / #170: cache key = `(user_sub, groups, sub_agent_config_hash, ENTITLEMENT_POLICY_VERSION)`; group membership and sub-agent config changes already invalidate implicitly through the key; console-side settings changes invalidate the receiving replica immediately; gateway-side changes and other replicas are TTL-bound.

aartaria:
closes #171

## Why

The orchestrator's per-user discovery/registry cache was invalidated by console-backend **pushing** a POST to *one* orchestrator replica whenever it remembered an entitlement had changed. Two gaps:

- the POST reached a single replica (#171);
- a mutation site could simply be forgotten. **Activating a sub-agent for the orchestrator never invalidated anything**, so a newly enabled sub-agent was ignored for a full cache TTL (10 min in prod) — the orchestrator kept delegating to `general-purpose` instead.

## What

**Push → pull.** console-backend exposes `GET /api/v1/auth/me/entitlement-version`: an opaque digest over every row that decides a user's entitlements (role, settings/whitelist/bypass rules, group memberships, group default agents, sub-agent activations, and the activated agents' default version + its content hash). The orchestrator fetches it once per turn, *before* the cache lookups, and folds it into `cache_key`. A changed entitlement makes the stale entry unreachable on the next turn, on every replica, with no broadcast and nothing to remember at a mutation site. On a failed fetch the last known stamp is reused, so a console blip degrades to today's TTL behaviour.

The one entitlement not held in the console DB — group → MCP-gateway server access — bumps a new `users.entitlements_touched_at` column for the group's members (migration 090), which moves the digest the same way.

**Removed as redundant:** the `/internal/discovery-cache/invalidate` endpoint and its `azp` gate, console-backend's `orchestrator_cache` client and its four router call sites, `invalidate_users` / owner-scoped eviction, `get_group_member_subs`. `invalidate_all` and `ENTITLEMENT_POLICY_VERSION` stay as the fleet-wide lever for gateway-side catalogue changes the console cannot see — the TTL's only remaining job.

Cost: one extra in-cluster round trip per turn (single indexed query), versus the ~3 s cold discovery it protects.

## Notes for reviewers

- `sub_agent_config_hash` in the key was never a digest of the user's sub-agent set; it is the playground's "test this exact version" hash and is `None` on a normal turn. Docstrings now say so.
- Follow-up outside this repo: the comment on `AGENT_DISCOVERY_CACHE_TTL` in gitops (`orchestrator.yaml`) still claims console-side changes invalidate immediately via the POST; it should be reworded to reference the entitlement version.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no API consumers of the removed internal endpoint other than console-backend itself; DB migration is additive.


packages/console-backend/console_backend/services/entitlement_version.py
```diff
@@ -0,0 +1,101 @@
...
+          WHERE m.user_id = u.id)               AS group_defaults,
+        (SELECT count(*) || ':' || coalesce(max(a.created_at)::text, '')
+           FROM user_sub_agent_activations a
+          WHERE a.user_id = u.id)               AS activations,
+        (SELECT coalesce(max(sa.updated_at)::text, '') || ':'
+                || coalesce(string_agg(sa.id || '@' || coalesce(sa.default_version::text, '')
+                                       || '#' || coalesce(cv.version_hash, ''), ','
+                                       ORDER BY sa.id), '')
+           FROM sub_agents sa
+           JOIN user_sub_agent_activations a ON a.sub_agent_id = sa.id
```
alloy-cicd: **[correctness] System-owned public sub-agents are invisible to the stamp (INNER JOIN on activations)**

`activated_agents` INNER-JOINs `user_sub_agent_activations`, but `/api/v1/sub-agents/activated` also admits system-owned public agents with **no** activation row (`sub_agent_service.py:277-281`: `usa.sub_agent_id IS NOT NULL OR (sa.owner_user_id = 'system' AND sa.is_public = TRUE)`). Approving a new default version of such an agent bumps `sub_agents.default_version`/`updated_at` (`sub_agent_repository.py:143-154`) but creates activation rows only for group-default members and the owner (`sub_agent_service.py:1650-1663`). For every user who has the agent solely via the implicit predicate, **no term of this query moves**, so their cached discovery keeps serving the old prompt/tools/model until the TTL — contradicting the docstring's "a newly approved version is picked up too" (lines 19-22).

Suggested fix: a second subselect mirroring the endpoint's implicit predicate, e.g. `max(updated_at) || string_agg(id || '@' || default_version || '#' || version_hash)` over `sub_agents WHERE owner_user_id = 'system' AND is_public = TRUE`.

packages/console-backend/console_backend/services/entitlement_version.py
```diff
@@ -0,0 +1,101 @@
...
+           LEFT JOIN sub_agent_config_versions cv
+                  ON cv.sub_agent_id = sa.id AND cv.version = sa.default_version
+          WHERE a.user_id = u.id)               AS activated_agents
```
alloy-cicd: **[correctness] `sub_agent_permissions` (and `catalog_permissions`) are entitlement tables the fingerprint cannot see**

Verified: the only writer of `sub_agent_permissions` (`SubAgentRepository.update_permissions`, `sub_agent_repository.py:277-298`, via `PUT /{id}/permissions`) writes nothing this query reads — it neither deletes members' `user_sub_agent_activations` rows nor calls `touch_group_member_entitlements`. Yet the non-admin predicate behind `/api/v1/sub-agents/activated` (`sub_agent_service.py:344-348`) is `(owned | public | group-shared via sub_agent_permissions) AND activated`, so **unsharing** agent S from group G removes it on a fresh fetch while no member's stamp moves — cached discovery keeps serving S until the TTL. **Re-sharing** to a member who still holds an old activation row is equally invisible. (First-time grants are naturally covered: the user must activate, which moves `activations`.)

Same shape for `catalog_permissions`: `accessible_catalog_ids` rides the cached `User` (`registry.py:349` → `GET /api/v1/catalogs` → `catalog_repository.py:122-126`) and gates the `catalog_search` tool, but neither `catalogs` nor `catalog_permissions` appears here. That hole predates this PR (the push path didn't cover it either), but the docstring now claims the digest covers "everything in this database that decides which tools and sub-agents a user is entitled to".

Fix: one count+`max(created_at)` subselect per table (joined through the user's groups), or `touch_group_member_entitlements` at the mutation sites — and scope the docstring claim to match whatever is covered.

packages/console-backend/console_backend/services/entitlement_version.py
```diff
@@ -0,0 +1,101 @@
...
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
+
+# One round trip, all indexed lookups by user id / group id.
+_VERSION_QUERY = text("""
+    SELECT
+        u.role,
+        u.status::text                          AS status,
+        u.is_administrator,
+        u.updated_at::text                      AS user_updated_at,
```
alloy-cicd: **[efficiency] `u.updated_at` (and `user_settings.updated_at`) make the stamp churn on non-entitlement writes — including every login**

`handle_login_callback` upserts the user on every OIDC login with an unconditional `updated_at = EXCLUDED.updated_at` (`auth_controller.py:222` → `user_service.py:533-558`), and SCIM PUT/PATCH re-syncs do the same for name/email/attributes (`scim_service.py:258-278, 341`). So logging out and back in — or an IdP nightly re-push of the whole user population — moves every affected user's stamp and evicts their discovery + user + embedded-runnable entries for zero entitlement change; each pays the full re-discovery on their next turn. The query already names the entitlement-bearing `users` columns explicitly (`role`, `status`, `is_administrator`, `entitlements_touched_at`), so `u.updated_at` contributes only false positives.

Similarly `settings_updated_at` (lines 47-49) is bumped by **every** settings write — timezone, preferred model, thinking level, phone confirmation — where the deleted push path invalidated only on `{"mcp_tools", "tool_bypass_rules"}`. Hashing those two columns' values instead of the row mtime keeps the stamp on the actual entitlements. (If the coarseness is intentional — some of those fields do live on the cached `User` — worth saying so in the docstring.)

packages/console-backend/console_backend/services/entitlement_version.py
```diff
@@ -0,0 +1,101 @@
...
+           FROM sub_agents sa
+           JOIN user_sub_agent_activations a ON a.sub_agent_id = sa.id
+           LEFT JOIN sub_agent_config_versions cv
+                  ON cv.sub_agent_id = sa.id AND cv.version = sa.default_version
+          WHERE a.user_id = u.id)               AS activated_agents
+    FROM users u
+    WHERE u.id = :user_id
+""")
+
+_TOUCH_GROUP_QUERY = text("""
```
alloy-cicd, line 102: **[convention] Raw `UPDATE users` bypasses the repository/audit pattern**

`packages/console-backend/AGENTS.md:61`: "**CRITICAL: All database write operations (INSERT/UPDATE/DELETE) MUST use the repository pattern to ensure automatic audit logging**" (reinforced at :115, whose ❌ example is exactly this shape). This write is triggered by an admin action and mutates a `users` column with no audit row — the mechanism it replaced was audit-neutral (an outbound HTTP POST). If `entitlements_touched_at` is deliberately exempt as non-auditable bookkeeping, that exemption deserves a sentence in AGENTS.md; otherwise route it through a repository.

Minor, same query: the deleted `get_group_member_subs` filtered `deleted_at IS NULL AND status = 'active'`; this UPDATE touches all member rows. Harmless today, but it's a silent divergence from every other membership query in `UserGroupService`.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -686,15 +692,24 @@ async def execute(
         if sub_agent_config_hash:
             logger.info(f"[CONSOLE] Console mode enabled for sub-agent config hash: {sub_agent_config_hash}")
 
+        # One cheap call per turn: the user's entitlement version. It goes into every per-user
+        # cache key below, so a changed entitlement (activated sub-agent, whitelist, role,
+        # group default, gateway server access, ...) makes the stale entries unreachable on
+        # this turn — on every replica, with no push-based invalidation.
+        entitlement_version = resolve_entitlement_version(
+            user_sub, await self.registry_service.get_entitlement_version(user_token)
```
alloy-cicd: **[performance] The stamp fetch is a serialized, unconditional hop on every turn's TTFT critical path, with a 5s budget**

Before this PR a warm turn made zero console-backend calls; now every turn awaits this GET alone, before the user-cache lookup — one RTT + JWT validation + the fingerprint query per turn, and up to `timeout=5.0` (`registry.py:338`) of dead wait when console-backend is degraded, on turns that would otherwise be pure cache hits.

Options that keep the design: start the fetch as an `asyncio.create_task` right after `caller_sub` extraction so it overlaps the `enqueue_event`/`try_register` awaits and is only awaited here; drop the timeout to ~1s (the fallback to the remembered stamp is benign by construction); and/or debounce per `user_sub` for a few seconds to spare bursty multi-message turns.

Related: a 401/403 from console-backend is treated identically to a transport blip (warn + reuse last stamp), so an entitlement change made while a user's token is rejected is silently masked rather than surfaced — worth distinguishing if that ever matters.

packages/console-backend/console_backend/routers/admin_mcp_gateway_server_access_router.py
```diff
@@ -22,12 +22,14 @@ def get_mcp_gateway_service(request: Request) -> McpGatewayServerAccessService:
...
-    member_subs = await request.app.state.user_group_service.get_group_member_subs(db, group_id)
-    schedule_orchestrator_discovery_cache_invalidation(background_tasks, request, reason, member_subs)
+async def _touch_group_members(db: DbSession, group_id: int) -> None:
+    """Move the members' entitlement version so the orchestrator re-discovers their tools.
+
+    Group → MCP-server access lives in the gateway, so the version fingerprint cannot see
+    the change on its own — this is the one entitlement that needs an explicit bump.
+    """
+    await touch_group_member_entitlements(db, group_id)
+    await db.commit()
```
alloy-cicd: **[robustness] The bump is now blocking-and-fatal after an unrollbackable gateway write**

The replaced mechanism was explicitly best-effort (old `orchestrator_cache.py`: "any failure is logged and swallowed so it can never block the triggering action") and ran as a background task. Now if `grant_server_access` succeeds against the gateway and this UPDATE/commit then fails (lock contention on `users` rows for a large group vs. a concurrent SCIM sync, transient DB error), the endpoint returns 500 even though the grant was applied — the admin retries a grant that already exists, and the bump can still be lost (members stale until the TTL). A try/except-log around the touch restores the old contract; the TTL remains the correctness floor either way.

packages/orchestrator-agent/app/core/discovery_cache.py
```diff
@@ -70,20 +83,52 @@ def cache_key(
...
+        },
         sort_keys=True,
     )
     return hashlib.sha256(payload.encode()).hexdigest()
 
 
+# Last entitlement version successfully fetched per user_sub. Only consulted when the
+# per-turn fetch fails, so a console-backend blip degrades to "reuse the current entry
+# until the TTL" instead of a cold re-discovery (or an error) on every turn.
+_last_entitlement_version: dict[str, str] = {}
```
alloy-cicd: **[minor] `_last_entitlement_version` is the module's only unbounded store**

Every `TtlTokenCache` here is capped (`_DEFAULT_MAX_ENTRIES` + `_evict`, with the rationale spelled out at lines 74-78), but this dict grows one permanent entry per distinct `user_sub` and is cleared only by `invalidate_all()` — which no production code path calls any more (see the note on `config.py`). Two side effects beyond the slow leak: the fallback is unbounded in **age** (a week-old stamp is reused as happily as a fresh one), and a first-turn fetch failure returns `None` silently — the `[ENTITLEMENT-VERSION]` warning that AGENTS.md documents as the diagnostic only fires when a previous stamp exists. A small `TtlTokenCache` with a long TTL (or a capped `OrderedDict`) fixes bound and age in a couple of lines.

packages/orchestrator-agent/app/models/config.py
```diff
@@ -345,14 +349,14 @@ class AgentSettings:
...
-    # additionally bounded by the user token's expiry.
+    # gateway itself, invisible to the console) may go unnoticed. Entitlement changes made
+    # through the console do not wait for it: the per-user entitlement version fetched each
+    # turn is part of the cache key (see discovery_cache). Entries are additionally bounded
+    # by the user token's expiry.
     AGENT_DISCOVERY_CACHE_TTL = _int_env("AGENT_DISCOVERY_CACHE_TTL", 60)
     # Cross-cutting invalidation lever for the discovery/registry caches: bump this (env)
-    # or call discovery_cache.invalidate_all() when a group→server/tool access policy
-    # changes without the user's own groups/config changing.
+    # or call discovery_cache.invalidate_all() for a fleet-wide flush, e.g. after a gateway
```
alloy-cicd: **[minor] `invalidate_all()` lost its only production entrypoint, but is still documented as the ops lever**

The deleted `/internal/discovery-cache/invalidate` endpoint was its only caller; `invalidate_all` is now reachable only from tests, yet this comment (and the function's own docstring at `discovery_cache.py:244-250`) still tell an operator to call it for a fleet-wide flush after a gateway-side catalogue change — the one job the PR explicitly leaves to this lever. The remaining real options are waiting out `AGENT_DISCOVERY_CACHE_TTL` (60s) or bumping `ENTITLEMENT_POLICY_VERSION`, which needs a redeploy. Either restore a locked-down flush endpoint, or update this comment + docstring and consider deleting `invalidate_all` and its tests.

alloy-cicd (commented):

The pull-based entitlement stamp is a real improvement over push invalidation, and the removed-behavior audit checks out: every deleted invalidation trigger (role change, settings whitelist/bypass, group default agents, gateway grant/revoke, group membership) is re-established by the fingerprint; all referenced columns exist and are index-covered; no dangling references to the deleted endpoint or `orchestrator_cache` module remain in code; all three `cache_key` call sites pass the new stamp.

**8 inline comments.** The two that matter most are verified fingerprint coverage gaps: system-owned public sub-agents (INNER JOIN on activations) and `sub_agent_permissions` unshare/re-share — both leave users on stale cached entitlements until the TTL despite the docstring's "a forgotten mutation site cannot leave the cache stale". Next tier: stamp over-churn from row-mtimes (`u.updated_at` moves on every login/SCIM sync), and the per-turn stamp fetch being a serialized 5s-budget hop on the TTFT critical path.

Notes without an inline anchor (files not in the diff):
- `packages/litellm-proxy/custom_logger.py:15` still points its phase-2 TODO at "console-backend's `orchestrator_cache`" as the reference client-credentials implementation — this PR deletes that module.
- `registry.py`'s new `get_entitlement_version` is the fourth copy of the "GET console path with user bearer → warn + fall back" body (after `_fetch_accessible_catalog_ids`, `_fetch_user_role`, `_fetch_user_settings`, which already disagree on timeouts and status checks); a small shared `_get_json(path, token, timeout)` helper would keep the auth/error contract in one place.
- `packages/console-backend/AGENTS.md:184` still says migrations live in `infrastructure/roles/basis/files/ddl/scripts/` — a directory that doesn't exist; the new `090_...sql` correctly went to `sqlmigrations/ddl/`, so the doc line is stale.

Checked and clean: group-default agents are covered (they materialize activation rows, and count+max(created_at) heuristics hold for delete/insert swaps); `user_settings` upsert always bumps `updated_at` on both INSERT and UPDATE paths; auth on the new endpoint works for the orchestrator-forwarded user token; `users.id` typing and the `::text` casts are correct; the migration has both rambler directives; new tests use `pg_session` per house style.

aartaria:
Triage of the automatic review (round 1), addressed in 7348a6fe:

1. **System-owned public agents invisible** — fixed. The served-agents term now takes `activated OR (owner = system AND is_public)`, mirroring the endpoint predicate. Test added.
2. **`sub_agent_permissions` / `catalog_permissions` not covered** — fixed. Both are folded in via the user's groups (count + last change), plus owned catalogs. Docstring now enumerates exactly what is covered. Test added.
3. **Stamp churn on login / any settings write** — fixed. `users.updated_at` dropped (role/status/admin are read explicitly); `user_settings` is hashed by whitelist + bypass-rule *value*, not row mtime. Test added asserting a login re-upsert and a timezone change leave the stamp unchanged.
4. **Raw `UPDATE users` bypasses the audited repository pattern** — kept, documented as the one exemption in console-backend `AGENTS.md`: the column is cache bookkeeping with no business meaning, and the admin action that triggers it is audited on its own. The deleted-members filter difference is harmless (touching a deleted user's row has no effect) and left as is.
5. **Serialized hop on the TTFT path** — partly addressed. The fetch now starts before turn registration and is awaited only once the turn is claimed (cancelled on the steering path), and the timeout is 2 s. It is still one in-cluster round trip per turn by design; that is the cost of pull over push. Debounce and 401-vs-transport distinction: not done — a 401 on the user's own token fails the registry fetch on the same turn anyway.
6. **Bump is blocking-and-fatal after a gateway write** — fixed, best-effort again (logged, never fails an applied grant).
7. **Unbounded last-stamp dict** — fixed: a `TtlTokenCache` (size-capped, 24 h aged); a first-turn fetch failure is now logged as well.
8. **`invalidate_all` dead but documented** — fixed by deleting it and its test; the config comment now names the two real levers (TTL, `ENTITLEMENT_POLICY_VERSION` via redeploy).

Unanchored notes: litellm-proxy TODO reworded; console-backend `AGENTS.md` migrations path corrected. The shared console-GET helper in `registry.py` is a fair refactor but out of scope here.

packages/console-backend/console_backend/routers/admin_mcp_gateway_server_access_router.py
```diff
@@ -22,12 +26,19 @@ def get_mcp_gateway_service(request: Request) -> McpGatewayServerAccessService:
...
+    Group → MCP-server access lives in the gateway, so the version fingerprint cannot see
+    the change on its own — this is the one entitlement that needs an explicit bump.
+    Best-effort: the gateway write it follows cannot be rolled back, so a failure here is
+    logged and the members converge within the orchestrator's cache TTL instead.
+    """
+    try:
+        await touch_group_member_entitlements(db, group_id)
+        await db.commit()
+    except Exception as e:  # noqa: BLE001 — must never fail an already-applied gateway grant
+        logger.warning("Failed to bump entitlement version for members of group %s (%s): %s", group_id, reason, e)
```
alloy-cicd, line 41: **[correctness] The swallowed failure resurfaces at request teardown — add a rollback**

If the UPDATE (or the commit) fails, the session is left in an aborted transaction, and this except swallows that only until the request ends: `get_db_session` (`db/session.py:26-27`) runs `await session.commit()` after the handler yields back, which on an aborted session raises (`PendingRollbackError`), gets caught by the dependency's own except, rolls back, and **re-raises** — so the endpoint still 500s after an applied gateway grant, which is exactly what this try/except was added to prevent. Add `await db.rollback()` in this except (best-effort too, if you like) so the session is clean when the teardown commit runs.

packages/orchestrator-agent/app/core/executor.py
```diff
@@ -637,6 +643,8 @@ async def execute(
                 f"[STEERING] Active stream found for context_id={context_id}, "
                 f"queuing message for running orchestrator (queue depth: {active.message_queue.qsize() + 1})"
             )
+            if entitlement_fetch is not None:
+                entitlement_fetch.cancel()  # steering into a running turn: no cache lookups here
```
alloy-cicd: **[minor] The early fetch task is orphaned on the raise paths before this cancel**

The three steering rejections above (`raise InvalidParamsError()` at lines 628, 635, 641) and the zero-trust raises after the turn is claimed (lines 686, 689) all exit between `create_task` (line 578) and the await at line 702, leaving the task running unobserved. No exception can leak (`get_entitlement_version` never raises), so the only cost is a discarded console-backend request per rejected/failed turn — but a `try/finally` around the registration-to-await span (or cancelling in the reject branches) would make the lifetime explicit and keep rejected steering attempts from generating console traffic.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (f3233d79b423d1b2aeddb05edc890fc140c69e9b..7348a6fe974d93a44fc6d80c5e80f55770d4de51)

1. **System/public sub-agents invisible to the stamp** — ✅ verified-fixed. The `served_agents` term now mirrors the endpoint's predicate exactly (`sa.deleted_at IS NULL AND (EXISTS activation OR (owner_user_id = 'system' AND is_public))`, matching `sub_agent_service.py:287`), so approving a new default version of a system agent moves every user's stamp.
2. **`sub_agent_permissions` / `catalog_permissions` gaps** — ✅ verified-fixed. New `shared_agents`, `shared_catalogs`, `owned_catalogs` subselects over the `ug` CTE. Verified the writers: both permission tables are written delete-all + re-insert (`sub_agent_repository.py:277-298`, `catalog_repository.py:395-402`), so count catches removals and `max(created_at)` catches re-shares and same-count swaps.
3. **Stamp churn on logins/SCIM/settings writes** — ✅ verified-fixed. `u.updated_at` dropped (role/status/admin read explicitly); settings hashed by `mcp_tools`/`tool_bypass_rules` **value**. Both columns are JSONB (ddl/008, ddl/058), so the `::text` rendering is normalized and stable.
4. **Audit-pattern bypass** — ✅ verified-fixed as a documented exemption (console-backend AGENTS.md + module docstring). The minor active/deleted-member filter divergence wasn't taken — fine as an accepted trade-off, not re-raised.
5. **Serialized 5s stamp fetch on the TTFT path** — ✅ verified-fixed. The fetch now starts before turn registration and overlaps `try_register`; timeout cut to 2s; steering path cancels it. (Still one console call per warm turn — no debounce — which is a defensible design choice; not re-raised.) One small orphan-task edge introduced by the fix, inline below.
6. **Blocking-and-fatal group bump** — ⚠️ partially fixed. The try/except is there, but without a rollback the swallowed failure resurfaces at request teardown — inline comment below; this is the one finding that still bites.
7. **Unbounded `_last_entitlement_version`** — ✅ verified-fixed (`TtlTokenCache`, 24h TTL, size-capped; the fetch-failed warning now fires in the no-stamp case too, and the new test pins the bounding policy).
8. **`invalidate_all()` dead but documented** — ✅ verified-fixed (function and its tests deleted; `ENTITLEMENT_POLICY_VERSION`/TTL documented as the honest remaining levers).

The round-1 summary notes (litellm-proxy TODO, migrations doc path, root AGENTS.md wording) are also all fixed. **2 new inline comments on the delta**, both on edges introduced by the fixes themselves.

aartaria:
Round 2 triage, addressed in daa1ee21:

1. **Missing rollback in the best-effort bump** — fixed, `await db.rollback()` in the except so the dependency's teardown commit no longer fails the applied grant.
2. **Orphaned early fetch task on raise paths** — fixed by not starting it early: the only awaits it could overlap are in-memory turn registration, so the overlap was not worth a lifetime to manage. The fetch is awaited inline at the zero-trust site (2 s budget, falls back to the last known stamp).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (7348a6fe974d93a44fc6d80c5e80f55770d4de51..daa1ee211920e3e23ddc16155c685f8523c01c65)

1. **[correctness] Swallowed bump failure resurfaced at request teardown** — ✅ verified-fixed. The except now does `await db.rollback()`, so the session is clean when `get_db_session`'s teardown commit runs and a failed bump can no longer 500 an already-applied gateway grant. The rollback discards nothing else: the gateway write is HTTP, and the bump is the handler's only DB write.
2. **[minor] Early stamp-fetch task orphaned on raise paths** — ✅ verified-fixed, by reverting to an inline await with the author's reasoning that the only awaits the task could overlap (turn registration) are in-memory, so the overlap bought nothing and the lifetime management wasn't earning its keep. That reasoning holds; the round-1 latency concern remains addressed by the 2s timeout and the benign last-known-stamp fallback.

Every previous finding is resolved and the delta (2 files, +5/−11) introduces nothing new. No inline comments this round.

PR #209, opened 2026-09-07
fix(scheduler): stop the model's reasoning from eating the draft's output budget

### LINKED ISSUE #208: scheduler: draft generation is cut off by the model's own reasoning — thinking tokens consume max_tokens, the reply is truncated mid-JSON
## Problem

Follow-up to #200. With the tool catalogue trimmed to 15 candidates, `POST /api/v1/scheduler/generate-job-draft` for *"notify me every time a new bug report is filed"* still fails — now loudly, as intended:

```
WARNING console_backend.services.llm_gateway   No JSON object in the model's reply (94 chars) for model gemini-3.5-flash
WARNING console_backend.routers.scheduler_router Draft generation produced nothing usable for query '…' (model gemini-3.5-flash, 15 candidate tools, raw keys [])
POST /api/v1/scheduler/generate-job-draft 422
```

The ranking worked (`console_list_bug_reports` was offered first). The 94 characters were the **opening of a correct JSON object, cut off**.

### Cause

`gateway_chat` sends `max_tokens=1024` and no `reasoning_effort`. The `chat:low` tier resolves to a model that reasons by default (Gemini 3.5 Flash here), and reasoning tokens count against `max_tokens`. The model spends the budget thinking and is stopped a few tokens into the answer, so there is no closing brace and `gateway_chat_json`'s salvage finds no object.

Reproduced against the proxy with a prompt the size of the real one (~10.9k input tokens):

| call | finish_reason | reasoning tokens | text tokens | parseable |
|---|---|---|---|---|
| as deployed, `max_tokens=1024` | `length` | 979 | 41 | no — cut mid-object |
| `max_tokens=1024`, `reasoning_effort="none"` | `stop` | 0 | 122 | yes, correct tool |
| `max_tokens=4096` | `stop` | 2879 | 185 | yes, correct tool |

Even a 159-token prompt spent 739 reasoning tokens on a 74-token answer, so prompt size is not the trigger; any reasoning model on this tier hits it.

### Two smaller defects alongside

- `gateway_chat_json` logs only the reply *length* when it finds no object. The finish reason and the start of the reply are what make this diagnosable without reproducing it. `gateway_chat` currently discards `finish_reason` entirely.
- The 422 detail says *"rephrase the request"* — wrong advice for a truncated reply; the request was fine.

## Proposal

- Pass `reasoning_effort="none"` on the draft and condition generators, as conversation titling already does: these calls are mechanical JSON-filling and reasoning only burns the output budget. (`drop_params` on the proxy makes it a no-op for models without the parameter.)
- Surface the finish reason: have `gateway_chat` expose it (or log it), and have `gateway_chat_json` log finish reason plus the first ~200 characters of the reply on the no-object / unparseable paths.
- When the reply was cut off (`finish_reason == "length"`), say so in the error detail instead of suggesting a rephrase.

## Acceptance

- [ ] the bug-report request produces a watch on `console_list_bug_reports` on the `chat:low` tier with a reasoning model
- [ ] the draft and condition generators disable reasoning on their gateway calls
- [ ] an unparseable/empty reply logs finish reason and a reply snippet
- [ ] a length-truncated reply is reported as such, not as a request to rephrase

aartaria:
closes ringier-data/nannos#208

## What

Follow-up to #205. With the catalogue trimmed, *"notify me every time a new bug report is filed"* still failed on the `chat:low` tier, now as a 422 telling the user to rephrase. The log showed ranking had worked (`console_list_bug_reports` offered first) and the model had replied with 94 characters of non-JSON.

**Cause:** the tier resolves to a reasoning model (Gemini 3.5 Flash locally). Reasoning tokens count against `max_tokens=1024`; the model spent 979 of them thinking and was stopped 41 tokens into the JSON object. `gateway_chat` discarded `finish_reason`, so the cutoff was indistinguishable from a content miss.

## How

- **Thinking off for the generators.** Draft and condition generation pass `reasoning_effort="none"` (as conversation titling already does): mechanical JSON-filling from a prompt that states the rules. The proxy drops the parameter for models without the control.
- **Finish reason kept.** `gateway_chat` returns `GatewayText`, a `str` subclass carrying `finish_reason`, so every existing caller and every test stubbing it with a literal keeps working. `gateway_chat_json` reads it (tolerating plain strings).
- **Diagnosable log line.** On a no-object or unparseable reply, `gateway_chat_json` logs the finish reason and the first 200 characters of the reply, not just its length.
- **Truncation named.** When the reason there is no object is `finish_reason == "length"`, `gateway_chat_json` raises `GatewayReplyTruncated`; both scheduler generators turn that into a 422 saying the reply was cut off, not a request to rephrase. A complete object that merely stopped at the limit is still returned.

- **Trigger-notification writer** (`scheduler_engine._write_notification`, `max_tokens=256`) also gets `reasoning_effort="none"` after review: plain-text writing, and a cut-off sentence would reach the person verbatim.

Not touched: the watch judge (`watch_evaluator`, `max_tokens=512`, no `reasoning_effort`) has the same exposure on a reasoning model — a cutoff there reads as "condition not met". Judging is semantic, so whether to disable thinking or raise the budget is a separate call; noting it rather than folding it in.

Also on this branch: Andrea's trailing-slash fix for `_console_mcp_url` (1e92666e), with its debugging leftover removed.

## Verification

- Live, against the local proxy with a prompt the size of the real one (~10.9k tokens), through the patched `gateway_chat_json`:
  - thinking on, `max_tokens=1024` → `GatewayReplyTruncated`, log shows `finish_reason=length` and the opening of the object
  - `reasoning_effort="none"` → `job_type=watch`, `check_tool=console_list_bug_reports`, a CEL expression over `prev`
- New tests: `test_llm_gateway.py` (finish reason carried; cutoff raises; complete-at-limit returned; snippet + finish reason logged; plain-string stubs unaffected; `reasoning_effort` forwarded), `test_generate_job_draft_endpoint.py` and `test_generate_condition.py` (thinking off; truncated reply → 422 "cut off").
- console-backend full suite: 1719 passed. `ruff check` clean on touched files.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — No. `gateway_chat` still returns a `str`; `gateway_chat_json` gains an optional parameter and a new exception on a path that previously returned `{}` for a cut-off reply (both callers handle it; the watch judge already catches `Exception`).

### DIFF
--- a/packages/console-backend/console_backend/services/mcp_tool_client.py
+++ b/packages/console-backend/console_backend/services/mcp_tool_client.py
@@ -50,7 +50,7 @@ def _console_mcp_url() -> str:
     Deliberately loopback rather than a configured public URL: the call must reach *this*
     instance, and it has no business leaving the pod to do it.
     """
-    return f"http://127.0.0.1:{os.getenv('API_PORT', '5001')}/mcp"
+    return f"http://127.0.0.1:{os.getenv('API_PORT', '5001')}/mcp/"
 
 
 async def token_for(tool_name: str, user_access_token: str) -> str:
@@ -221,6 +221,7 @@ async def call_tool(
                 "params": {"name": tool_name, "arguments": arguments or {}},
             },
             timeout=timeout,
+            # follow_redirects=True,
         )
         response.raise_for_status()
         data = parse_envelope(response, expect_id=_CALL_ID)
--- a/packages/console-backend/console_backend/services/mcp_tool_client.py
+++ b/packages/console-backend/console_backend/services/mcp_tool_client.py
@@ -221,7 +221,6 @@ async def call_tool(
                 "params": {"name": tool_name, "arguments": arguments or {}},
             },
             timeout=timeout,
-            # follow_redirects=True,
         )
         response.raise_for_status()
         data = parse_envelope(response, expect_id=_CALL_ID)
--- a/packages/console-backend/console_backend/routers/scheduler_router.py
+++ b/packages/console-backend/console_backend/routers/scheduler_router.py
@@ -45,7 +45,7 @@ from ..services.cel_condition import (
     evaluate_arg_exprs,
     validate_cel_expression,
 )
-from ..services.llm_gateway import gateway_chat_json
+from ..services.llm_gateway import GatewayReplyTruncated, gateway_chat_json
 from ..services.scheduler_engine import SchedulerEngine
 from ..services.scheduler_service import _UNSET, SchedulerService
 from ..utils.timezones import resolve_timezone
@@ -117,6 +117,17 @@ _EXPRESSION_RULES = (
     "Every watch needs cel_expr, llm_condition, or both.\n\n"
 )
 
+
+#: Thinking off for the draft and condition generators. Both are mechanical JSON-filling
+#: from a prompt that already states the rules, and reasoning tokens count against
+#: `max_tokens`: on the low tier a reasoning model spent 979 of a 1024 budget thinking
+#: and was cut off 41 tokens into the answer, which read as "no usable draft". The proxy
+#: drops the parameter for models that have no such control.
+_GENERATION_REASONING = "none"
+
+#: What the person sees when the model's reply hit the output budget. Not a request to
+#: rephrase — the request was fine.
+_TRUNCATED_DETAIL = "The model's reply was cut off before it finished — try again."
 
 #: How many times a generated expression that fails to compile or evaluate is sent
 #: back with its error. Two is deliberate: the first retry fixes most syntax slips,
@@ -491,10 +502,13 @@ async def generate_job_draft(
             model=model,
             max_tokens=1024,
             metadata={"user_sub": current_user.sub},  # OIDC subject — the gateway/proxy attributes by sub, not internal id
+            reasoning_effort=_GENERATION_REASONING,
         )
 
     try:
         result = await _generate(prompt)
+    except GatewayReplyTruncated as exc:
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=_TRUNCATED_DETAIL) from exc
     except Exception as exc:
         logger.warning("Watch-param generation via gateway failed: %s", exc)
         raise HTTPException(
@@ -646,6 +660,7 @@ async def generate_condition(
             model=model,
             max_tokens=1024,
             metadata={"user_sub": current_user.sub},
+            reasoning_effort=_GENERATION_REASONING,
         )
 
     notes: list[str] = []
@@ -656,6 +671,8 @@ async def generate_condition(
     for _ in range(1 + _GENERATE_CONDITION_RETRIES):
         try:
             candidate = await _generate(instruction)
+        except GatewayReplyTruncated as exc:
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=_TRUNCATED_DETAIL) from exc
         except Exception as exc:
             logger.warning("Condition generation via gateway failed: %s", exc)
             raise HTTPException(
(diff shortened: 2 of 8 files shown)


packages/console-backend/console_backend/services/mcp_tool_client.py
```diff
@@ -221,6 +221,7 @@ async def call_tool(
                 "params": {"name": tool_name, "arguments": arguments or {}},
             },
             timeout=timeout,
+            # follow_redirects=True,
```
alloy-cicd: Leftover from debugging the `/mcp` 307: the commented-out `# follow_redirects=True,` is dead code now that the trailing slash in `_console_mcp_url` avoids the redirect.

```suggestion
```

alloy-cicd:
Review finding (not on a diff line, so posted here): **the trigger-notification writer has the same reasoning-budget exposure this PR fixes**, on the same tier.

`scheduler_engine.py:456`:
```python
message = await gateway_chat(prompt, model=model, max_tokens=256)
```
The model is `defaults.get("chat:low")` — the exact tier that resolved to a reasoning model and spent 979/1024 tokens thinking in the incident this PR fixes. Here the budget is 256 and no `reasoning_effort` is passed, so on that tier the reply is likely empty (→ silent fallback to the raw-result message) or cut off mid-sentence — and since `finish_reason` is discarded at this call site, a non-empty truncated sentence is sent verbatim to the notification channel. Unlike the watch judge (semantic, deliberately deferred in the PR description), this call is mechanical text-writing, so `reasoning_effort=_GENERATION_REASONING` applies exactly; worth folding in or at least adding to the "not touched" note.

alloy-cicd (commented):

Two findings, both minor:

1. **(inline, below)** `mcp_tool_client.py:224` — commented-out `# follow_redirects=True,` debug leftover, made obsolete by the trailing-slash fix in `_console_mcp_url`; removal suggested in the inline comment.
2. **(PR-level comment, file not in diff)** `scheduler_engine.py:456` — the trigger-notification writer has the same reasoning-budget exposure this PR fixes elsewhere: `chat:low` reasoning model, `max_tokens=256`, no `reasoning_effort`, and `finish_reason` is discarded, so a truncated notification would be sent verbatim.

The core change verified clean: `GatewayText` keeps str-compatibility for all callers and test stubs, the truncation raise is ordered before the generic `except Exception` in both generators, the watch judge intentionally absorbs the new exception, and the complete-object-at-limit case returns before the `finish_reason` check.

packages/console-backend/console_backend/services/mcp_tool_client.py
aartaria: Removed in b5f4067e. (The line came in with the trailing-slash commit that landed on this branch, not with the reasoning-budget change, but it is in the PR diff so it goes.)

aartaria:
Re the PR-level finding on `scheduler_engine.py:456` (trigger-notification writer, same reasoning-budget exposure): taken in b5f4067e — `reasoning_effort="none"`, since two sentences of plain text are mechanical writing and a cut-off sentence would go to the person verbatim. Test asserts the parameter. The watch judge remains deliberately out of this PR (semantic judgement; whether to disable thinking or raise its budget is its own call).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (1e92666e35c06b73d923cf7e1de2733026af3f70..b5f4067e38ff239b32d639dfe3564851b08cd213)

Ledger of round-1 findings:

1. `mcp_tool_client.py:224` commented-out `# follow_redirects=True,` — **verified-fixed** in b5f4067e (line removed).
2. `scheduler_engine.py:456` trigger-notification writer reasoning-budget exposure — **verified-fixed** in b5f4067e: the call now passes `reasoning_effort="none"` (matching the draft/condition generators; the proxy's `drop_params: true` keeps non-reasoning models unaffected), with a rationale comment and a test asserting the kwarg (`test_the_model_writes_it`).

The delta introduces nothing new — every previous finding is resolved and no new issues found. Clean incremental round.

PR #210, opened 2026-09-07
fix(orchestrator-agent): derive the turn budget from the compiled graph

AlexMetsai:
Extracted from #128 at review request, and rewritten rather than cherry-picked so
it answers the second question too: the multiplier is now derived from the graph
instead of being a constant someone has to remember to update.

## The bug

`MAX_RECURSION_LIMIT` was 50, chosen as though it counted model calls. It counts
LangGraph **super-steps**, and every middleware hook is its own graph node, so one
model call costs a whole lap:

```
SteeringMiddleware.before_model            1
A2ATaskTrackingMiddleware.before_model     2
model                                      3
TodoListMiddleware.after_model             4
_PTCToleranceCodeInterpreterMiddleware...  5
RepeatedToolCallMiddleware.after_model     6
ConditionalHumanInTheLoopMiddleware...     7
tools                                      8
```

So `50` was **six model calls per turn**. The orchestrator spends calls on
`write_todos` bookkeeping and filesystem exploration between delegations, so an
ordinary two-delegation request ("look up X and post it to Slack") ran out. Live
traces showed it completing the request correctly and then having the finished
answer discarded and replaced with the "shall I continue?" prompt.

The multiplier always existed as a fact about the middleware stack. It just was
never written down, so the limit was set at an implied exchange rate of 1.

## The fix

The budget is expressed in the unit people reason about —
`ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`, default 25 — and converted per compiled
graph in `app/core/step_budget.py`. `MAX_RECURSION_LIMIT` is gone from
`AgentSettings` entirely; there is nothing left for it to name.

## Can it be derived? Yes

Hook nodes are named `<Middleware>.<hook>`, so how often a node runs is readable
off its name:

| bucket | count | runs |
| --- | --- | --- |
| `.before_model` / `.after_model` | 6 | once per model call |
| `model`, `tools` | 2 | the cycle core |
| `.before_agent` / `.after_agent` | 3 | once per turn |
| `__start__` | 1 | never billed |

Giving `steps_per_model_call = 8`, `base_steps = 3`, and a limit of **203** for 25
model calls. Nothing is hardcoded — the core cost is counted from the
classification too, since langchain only adds a `tools` node when the graph has
tools and a rename would land both in `unclassified`.

Checked against real graph runs rather than asserted, on **both**
structured-output paths:

| model calls | `ToolStrategy` | thinking enabled | derived |
| --- | --- | --- | --- |
| 1 | 10 | 11 | 11 |
| 2 | 18 | 19 | 19 |
| 3 | 26 | 27 | 27 |

Exact for the thinking path, one step generous for `ToolStrategy`. That asymmetry
is deliberate and is the subject of the review round below.

## Why deriving beats a better constant

A literal goes stale the moment a middleware with model hooks is added or
removed, and nothing forces the update. That is the fragility raised in review,
and it is the whole justification.

An earlier version of this description also claimed the constant was *already*
wrong because the PTC middlewares appear and disappear with
`CODE_INTERPRETER_PTC`. **That was false** — see below.

## How it is tested

`tests/test_step_budget.py`, 21 tests. Four do real work; the rest are
configuration.

- **`test_marginal_cost_of_a_model_call_matches_the_derivation`**, parametrized
  over `[None, ThinkingLevel.high]`, runs turns of 1 to 4 model calls, counts
  super-steps, and asserts every marginal equals what was derived from the node
  names. If LangGraph renames a hook or starts running one twice per cycle,
  measurement and derivation diverge and this fails.
- **`test_the_budget_is_never_below_what_a_turn_actually_costs`**, over both paths
  and three turn lengths. This is the invariant that matters: a budget one step
  short is a `GraphRecursionError` fired *after* the answer is composed. It also
  asserts the slack stays under one full cycle, so "round up" cannot drift into
  "guess big".
- **`test_every_node_is_classified`** fails if the compiled graph grows a node
  shape the derivation does not recognise.
- **`test_an_unknown_node_inflates_the_budget_and_warns`** covers the same case at
  *runtime*, which is the one that survives a stack difference between CI and a
  deployment.

## Limitations, stated rather than discovered

- **Node count is an upper bound on a cycle's steps.** A hook that returns early
  still occupies a node; a skipped branch simply does not run. Over-estimating
  gives the model at least the intended number of calls, which is the safe
  direction — under-estimating is the bug being replaced.
- **Unknown nodes are charged at the per-model-call rate**, the most expensive
  assumption, and logged. Counting them as free would shrink the budget toward
  exactly the truncation this fixes.
- **Interrupt and resume are not modelled.** Resuming re-enters nodes, so a turn
  with several HITL approvals costs slightly more than the formula says. At 25
  model calls there is ample headroom, but it is not exact.
- **Assumes one node per super-step.** True for this graph, which has no parallel
  node execution. If that changes, the measured tests above are what catch it.

## What review found

Eight comments, and two were defects in the first commit rather than style
points. Both were found by measuring, and both are fixed:

1. **The PTC justification was factually wrong.**
   `build_code_interpreter_middlewares` unconditionally returns one
   `_PTCToleranceCodeInterpreterMiddleware`; `CODE_INTERPRETER_PTC` changes tool
   exposure inside `eval`, never the node count. Verified by compiling the graph
   both ways — the node sets are identical. The claim had conflated the
   orchestrator's stack with the **general-purpose sub-agent's**, where
   `ToolsetSelectorMiddleware` genuinely is PTC-gated. It appeared in three
   comments, the first commit message and this description; the code comments are
   corrected and staleness alone carries the argument.
2. **`base_steps` was off by one on half the graphs.** It subtracted a step for
   the `tools` node the closing model call skips, which holds only under
   `ToolStrategy`. When `select_response_format` returns `(None, True)` — thinking
   on Anthropic/Bedrock, or Gemini's builtin-tools path — `FinalResponseSchema` is
   bound as a `return_direct` tool and the closing call routes through `tools`
   like any other. Measured at 11/19/27 against a derived 10/18/26: one step
   short, i.e. a `GraphRecursionError` after the answer was composed. The
   subtraction is gone and both paths are now measured. The original test pinned
   `thinking_level=None`, which is how it survived.

Also fixed: `steps_per_model_call` hardcoded the core cost while discarding the
count it had just computed; `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN` had no lower
bound, so `0` bricked every turn with no log pointing at the variable; the
`unclassified` guard existed only in tests while production counted unknown nodes
as free. Two efficiency points too — `recursion_limit_for` classified the graph
twice, and `classify_nodes` used `graph.get_graph().nodes`, which runs langgraph's
edge-discovery simulation just to list names (0.00ms vs 14.27ms per twenty calls
against the compiled `Pregel`'s own `nodes` dict).

`test_fifty_steps_was_six_model_calls` is deleted rather than updated. Review
argued it documented rather than protected, and dropping the `- 1` proved the
point immediately by flipping it to 5. Editing the constant would have asserted a
historical fact that is no longer true of the current stack, so the figure lives
in the module docstring instead.

### Still open

- **The legacy warning is half-misleading.** `MAX_RECURSION_LIMIT` still
  configures sub-agents running in this same process — agent-common's
  `dynamic_agent` reads it at default 75 — so an operator who follows the advice
  and unsets it silently changes every sub-agent's bound. `dynamic_agent.py:117`
  also still references the attribute this PR deletes. Needs a wording change here
  plus a follow-up for porting the model-call budget to the three sibling
  packages, which still have the same 50-steps-is-six-calls truncation one layer
  down.
- **Sub-agent graphs inherit the derived limit**, with a fresh budget per `task`
  dispatch, via langgraph's ambient config propagation. Where a sub-agent
  previously inherited 50 super-steps it now inherits ~203, against a leaner stack
  it was not derived from. I have not verified the propagation claim yet; if it
  holds, the nested runaway ceiling rises materially and sub-agent graphs may need
  their own explicit limit. Worth settling before merge.

## Files

- `app/core/step_budget.py` — new. The classification and the conversion.
- `app/core/graph_factory.py` — computes the limit from the graph it just compiled.
- `app/models/config.py` — `MAX_MODEL_CALLS_PER_TURN` replaces `MAX_RECURSION_LIMIT`,
  with a lower clamp and an upper sanity warning; the legacy name logs a warning
  naming the replacement rather than being ignored in silence.
- `tests/test_step_budget.py` — new.
- `tests/test_orchestrator_thinking_integration.py` — two lines. It mocks
  `create_deep_agent`, and `_create_graph` now inspects the graph it gets back, so
  the mock needs an iterable `nodes`. Called out so it does not read as unrelated
  churn.

`844 passed` locally, no gateway or credentials needed.

## Sequencing with #128

#128 carries an earlier version of this fix with the multiplier as a literal. Once
this lands, that needs reverting there — otherwise the next `main` merge conflicts
on `config.py` and two versions of the same fix sit under review at once.

Worth being explicit that this **inverts the merge order**: "extract it" sounds
independent, but #128's real tier depends on this fix. `chains_analyst_then_slack`
is the scenario that surfaced the bug, dying on `GraphRecursionError` at 50. So
this should merge first.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — behavioural, and intended: the
      orchestrator stops reading `MAX_RECURSION_LIMIT` and reads
      `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN` instead. A deployment setting the old
      name gets a warning and the new default (25 model calls, 203 steps) rather
      than silent truncation at six. Nothing in this repo sets it — checked every
      yaml, template and shell script including
      `example-k8s-deployment/base/orchestrator.yaml` — but the infra repo is worth
      a grep before merge so the warning does not arrive unexplained.

### DIFF
--- a/packages/orchestrator-agent/app/core/graph_factory.py
+++ b/packages/orchestrator-agent/app/core/graph_factory.py
@@ -913,8 +913,8 @@ class GraphFactory:
         # Override deepagents' recursion_limit default of 1000, which is too high to
         # catch a runaway loop. The budget is configured in model calls and converted
         # here, against *this* graph: the multiplier is the middleware stack's per-call
-        # node cost, which differs between configurations (the PTC middlewares come and
-        # go with CODE_INTERPRETER_PTC), so it cannot be a constant.
+        # node cost, so it goes stale the moment a middleware with model hooks is added
+        # or removed — which is why it is counted rather than written down.
         recursion_limit = recursion_limit_for(compiled_graph, self.config.MAX_MODEL_CALLS_PER_TURN)
         compiled_graph = compiled_graph.with_config({"recursion_limit": recursion_limit})
         logger.info(
--- a/packages/orchestrator-agent/app/core/graph_factory.py
+++ b/packages/orchestrator-agent/app/core/graph_factory.py
@@ -72,6 +72,7 @@ from ..models.config import AgentSettings, GraphRuntimeContext
 from ..models.schemas import FinalResponseSchema
 from .file_tools import create_presigned_url_tool
 from .steering_state import get_all_active_subagent_dispatches, get_orchestrator_pending_messages
+from .step_budget import recursion_limit_for
 from .time_tools import create_time_tool
 
 logger = logging.getLogger(__name__)
@@ -909,10 +910,19 @@ class GraphFactory:
             context_schema=GraphRuntimeContext,
             response_format=response_format,
         )
-        # Override deepagents default recursion_limit (1000) with configured value
-        # This prevents infinite loops from reaching the high default limit
-        compiled_graph = compiled_graph.with_config({"recursion_limit": self.config.MAX_RECURSION_LIMIT})
-        logger.info(f"Graph created for model: {model_type} with recursion_limit={self.config.MAX_RECURSION_LIMIT}")
+        # Override deepagents' recursion_limit default of 1000, which is too high to
+        # catch a runaway loop. The budget is configured in model calls and converted
+        # here, against *this* graph: the multiplier is the middleware stack's per-call
+        # node cost, which differs between configurations (the PTC middlewares come and
+        # go with CODE_INTERPRETER_PTC), so it cannot be a constant.
+        recursion_limit = recursion_limit_for(compiled_graph, self.config.MAX_MODEL_CALLS_PER_TURN)
+        compiled_graph = compiled_graph.with_config({"recursion_limit": recursion_limit})
+        logger.info(
+            "Graph created for model: %s with recursion_limit=%d (%d model calls per turn)",
+            model_type,
+            recursion_limit,
+            self.config.MAX_MODEL_CALLS_PER_TURN,
+        )
 
         return compiled_graph
 
--- a/packages/orchestrator-agent/app/core/step_budget.py
+++ b/packages/orchestrator-agent/app/core/step_budget.py
@@ -101,15 +101,3 @@ def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
         max_model_calls,
     )
     return limit
-
-
-def affordable_model_calls(graph: Any, recursion_limit: int) -> int:
-    """Inverse of :func:`recursion_limit_for` -- what a given limit actually buys.
-
-    Exists to make the old failure legible: it answers "50 steps was how many
-    model calls?" (six), which is the number nobody could see at the call site.
-    """
-    per_call = steps_per_model_call(graph)
-    if per_call <= 0:
-        return 0
-    return max((recursion_limit - base_steps(graph)) // per_call, 0)
--- a/packages/orchestrator-agent/app/core/step_budget.py
+++ b/packages/orchestrator-agent/app/core/step_budget.py
@@ -1,0 +1,115 @@
+"""Convert a turn budget in model calls into LangGraph's super-step budget.
+
+LangGraph's ``recursion_limit`` counts **super-steps**, not model calls. Every
+middleware hook is its own graph node, so one model call costs several of them.
+That multiplier is invisible at the call site, which is how the limit came to be
+50 -- a sensible number of model calls and a crippling number of steps. At 50 a
+turn was capped at six model calls, and the orchestrator spends calls on
+write_todos bookkeeping and filesystem exploration between delegations, so an
+ordinary two-delegation request ("look up X and post it to Slack") exhausted the
+budget and the user was asked to continue a task that had already completed
+correctly.
+
+The multiplier is therefore **derived from the compiled graph** rather than
+written down. A hardcoded number is wrong twice over: it goes stale the moment a
+middleware is added, and it cannot be right for more than one configuration --
+the PTC middlewares appear and disappear with ``CODE_INTERPRETER_PTC``, so the
+orchestrator's real stack differs between deployments.
+
+How the derivation works
+------------------------
+Middleware hook nodes are named ``<Middleware>.<hook>``, so the hook type is
+readable off the node name:
+
+- ``.before_model`` / ``.after_model`` run once per **model call**
+- ``.before_agent`` / ``.after_agent`` run once per **turn**
+- ``model`` and ``tools`` are the core of each cycle
+
+Measured against real graph runs (see ``tests/test_step_budget.py``): the
+marginal cost of a model call is exactly ``per_call_hooks + 2``, and a turn of N
+calls costs ``per_turn_hooks - 1 + N * (per_call_hooks + 2)``. The ``- 1`` is
+real, not a fudge: the final model call emits ``FinalResponseSchema`` and ends
+the turn without entering ``tools``, so one cycle is a step short.
+
+Node count is an **upper bound** on the steps a cycle can spend, which is the
+safe direction: a hook that returns early still occupies a node, while a skipped
+branch simply does not run. Over-estimating gives the model at least the intended
+number of calls; under-estimating is the bug this replaces.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+PER_MODEL_CALL_HOOKS = (".before_model", ".after_model")
+PER_TURN_HOOKS = (".before_agent", ".after_agent")
+CORE_CYCLE_NODES = ("model", "tools")
+TERMINAL_NODES = ("__start__", "__end__")
+
+
+def classify_nodes(graph: Any) -> dict[str, list[str]]:
+    """Partition a compiled graph's nodes by how often they run.
+
+    Every node lands in exactly one bucket. ``unclassified`` is the important
+    one: it is empty today, and anything appearing in it means LangGraph grew a
+    node shape this module does not understand -- at which point the derivation
+    below is quietly wrong and the test asserting it is empty fails.
+    """
+    buckets: dict[str, list[str]] = {"per_model_call": [], "per_turn": [], "core": [], "terminal": [], "unclassified": []}
+    for name in graph.get_graph().nodes:
+        if name.endswith(PER_MODEL_CALL_HOOKS):
+            buckets["per_model_call"].append(name)
+        elif name.endswith(PER_TURN_HOOKS):
+            buckets["per_turn"].append(name)
+        elif name in CORE_CYCLE_NODES:
+            buckets["core"].append(name)
+        elif name in TERMINAL_NODES:
+            buckets["terminal"].append(name)
+        else:
+            buckets["unclassified"].append(name)
+    return buckets
+
+
+def steps_per_model_call(graph: Any) -> int:
+    """Super-steps one model call costs in *graph*: its hooks, plus model and tools."""
+    buckets = classify_nodes(graph)
+    return len(buckets["per_model_call"]) + len(CORE_CYCLE_NODES)
+
+
+def base_steps(graph: Any) -> int:
+    """Super-steps a turn costs regardless of how many model calls it makes.
+
+    The per-turn hooks, minus one for the ``tools`` step the closing model call
+    never takes -- it emits the final-response envelope and the turn ends.
+    """
+    return max(len(classify_nodes(graph)["per_turn"]) - 1, 0)
+
+
+def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
+    """LangGraph ``recursion_limit`` that affords *max_model_calls* model calls."""
+    per_call = steps_per_model_call(graph)
+    base = base_steps(graph)
+    limit = base + per_call * max_model_calls
+    logger.info(
+        "Derived recursion_limit=%d from the compiled graph: %d base + %d per model call x %d calls",
+        limit,
+        base,
+        per_call,
+        max_model_calls,
+    )
+    return limit
+
+
+def affordable_model_calls(graph: Any, recursion_limit: int) -> int:
+    """Inverse of :func:`recursion_limit_for` -- what a given limit actually buys.
+
+    Exists to make the old failure legible: it answers "50 steps was how many
+    model calls?" (six), which is the number nobody could see at the call site.
+    """
+    per_call = steps_per_model_call(graph)
+    if per_call <= 0:
+        return 0
+    return max((recursion_limit - base_steps(graph)) // per_call, 0)
--- a/packages/orchestrator-agent/app/core/step_budget.py
+++ b/packages/orchestrator-agent/app/core/step_budget.py
@@ -11,10 +11,8 @@ budget and the user was asked to continue a task that had already completed
 correctly.
 
 The multiplier is therefore **derived from the compiled graph** rather than
-written down. A hardcoded number is wrong twice over: it goes stale the moment a
-middleware is added, and it cannot be right for more than one configuration --
-the PTC middlewares appear and disappear with ``CODE_INTERPRETER_PTC``, so the
-orchestrator's real stack differs between deployments.
+written down: it changes whenever a middleware carrying model hooks is added or
+removed, and nothing forces a hand-maintained constant to be updated with it.
 
 How the derivation works
 ------------------------
@@ -25,16 +23,27 @@ readable off the node name:
 - ``.before_agent`` / ``.after_agent`` run once per **turn**
 - ``model`` and ``tools`` are the core of each cycle
 
-Measured against real graph runs (see ``tests/test_step_budget.py``): the
-marginal cost of a model call is exactly ``per_call_hooks + 2``, and a turn of N
-calls costs ``per_turn_hooks - 1 + N * (per_call_hooks + 2)``. The ``- 1`` is
-real, not a fudge: the final model call emits ``FinalResponseSchema`` and ends
-the turn without entering ``tools``, so one cycle is a step short.
+Deliberately an over-estimate
+-----------------------------
+The budget must never be *below* what a turn really costs -- that is the bug
+being replaced -- so every judgement here rounds up.
 
-Node count is an **upper bound** on the steps a cycle can spend, which is the
-safe direction: a hook that returns early still occupies a node, while a skipped
-branch simply does not run. Over-estimating gives the model at least the intended
-number of calls; under-estimating is the bug this replaces.
+- Node count is an upper bound on a cycle's steps: a hook that returns early
+  still occupies a node, while a skipped branch simply does not run.
+- Nodes this module does not recognise are counted as *per model call*, the most
+  expensive assumption, and logged. Ignoring them would shrink the budget toward
+  exactly the truncation this exists to prevent.
+- ``base_steps`` counts every per-turn hook, even though on some graphs the
+  closing model call ends the turn without entering ``tools`` and one cycle is a
+  step short. Whether it does depends on the structured-output strategy:
+  ``ToolStrategy`` (no thinking) exits without ``tools``, while a graph that
+  binds ``FinalResponseSchema`` as a ``return_direct`` tool -- thinking enabled
+  on Anthropic/Bedrock, or Gemini's builtin-tools path, i.e. when
+  ``select_response_format`` returns ``(None, True)`` -- routes the closing call
+  through ``tools`` like any other. Subtracting the difference would be exact for
+  the first and one step short for the second, and one step short is a
+  ``GraphRecursionError`` fired after the answer is composed. So it is not
+  subtracted. Measured in ``tests/test_step_budget.py`` against both paths.
 """
 
 from __future__ import annotations
@@ -53,13 +62,25 @@ TERMINAL_NODES = ("__start__", "__end__")
 def classify_nodes(graph: Any) -> dict[str, list[str]]:
     """Partition a compiled graph's nodes by how often they run.
 
+    Reads ``graph.nodes`` -- the compiled ``Pregel``'s own dict -- rather than
+    ``graph.get_graph().nodes``, which runs langgraph's edge-discovery simulation
+    to build a drawable graph just to list names. The two differ only by
+    ``__end__``, which is never counted.
+
     Every node lands in exactly one bucket. ``unclassified`` is the important
-    one: it is empty today, and anything appearing in it means LangGraph grew a
-    node shape this module does not understand -- at which point the derivation
-    below is quietly wrong and the test asserting it is empty fails.
+    one: anything in it means LangGraph grew a node shape this module does not
+    understand, so the callers below charge it at the per-model-call rate and say
+    so out loud.
     """
-    buckets: dict[str, list[str]] = {"per_model_call": [], "per_turn": [], "core": [], "terminal": [], "unclassified": []}
-    for name in graph.get_graph().nodes:
+    buckets: dict[str, list[str]] = {
+        "per_model_call": [],
+        "per_turn": [],
+        "core": [],
+        "terminal": [],
+        "unclassified": [],
+    }
+    for raw in graph.nodes:
+        name = str(raw)
         if name.endswith(PER_MODEL_CALL_HOOKS):
             buckets["per_model_call"].append(name)
         elif name.endswith(PER_TURN_HOOKS):
@@ -73,31 +94,45 @@ def classify_nodes(graph: Any) -> dict[str, list[str]]:
     return buckets
 
 
+def _warn_unclassified(buckets: dict[str, list[str]]) -> None:
+    if buckets["unclassified"]:
+        logger.warning(
+            "Unrecognised graph nodes %s are being charged at the per-model-call rate. "
+            "app/core/step_budget.py needs to learn how often they run; until then the "
+            "turn budget is a safe over-estimate rather than a correct one.",
+            sorted(buckets["unclassified"]),
+        )
+
+
+def _steps_per_model_call(buckets: dict[str, list[str]]) -> int:
+    # `core` is counted, not assumed to be 2: langchain only adds a `tools` node
+    # when the graph has tools, and a rename would land both in `unclassified`.
+    return len(buckets["per_model_call"]) + len(buckets["core"]) + len(buckets["unclassified"])
+
+
+def _base_steps(buckets: dict[str, list[str]]) -> int:
+    return len(buckets["per_turn"])
+
+
 def steps_per_model_call(graph: Any) -> int:
     """Super-steps one model call costs in *graph*: its hooks, plus model and tools."""
     buckets = classify_nodes(graph)
-    return len(buckets["per_model_call"]) + len(CORE_CYCLE_NODES)
+    _warn_unclassified(buckets)
+    return _steps_per_model_call(buckets)
 
 
 def base_steps(graph: Any) -> int:
-    """Super-steps a turn costs regardless of how many model calls it makes.
-
-    The per-turn hooks, minus one for the ``tools`` step the closing model call
-    never takes -- it emits the final-response envelope and the turn ends.
-    """
-    return max(len(classify_nodes(graph)["per_turn"]) - 1, 0)
+    """Super-steps a turn costs regardless of how many model calls it makes."""
+    return _base_steps(classify_nodes(graph))
 
 
 def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
-    """LangGraph ``recursion_limit`` that affords *max_model_calls* model calls."""
-    per_call = steps_per_model_call(graph)
-    base = base_steps(graph)
-    limit = base + per_call * max_model_calls
-    logger.info(
-        "Derived recursion_limit=%d from the compiled graph: %d base + %d per model call x %d calls",
-        limit,
-        base,
-        per_call,
-        max_model_calls,
-    )
-    return limit
+    """LangGraph ``recursion_limit`` that affords *max_model_calls* model calls.
+
+    Classifies once: the two components are derived from a single pass, since
+    this runs on every ``_create_graph`` including the cold path that rebuilds
+    per turn.
+    """
+    buckets = classify_nodes(graph)
+    _warn_unclassified(buckets)
+    return _base_steps(buckets) + _steps_per_model_call(buckets) * max_model_calls
--- a/packages/orchestrator-agent/app/models/config.py
+++ b/packages/orchestrator-agent/app/models/config.py
@@ -41,6 +41,10 @@ def _int_env(name: str, default: int) -> int:
 # using the compiled graph, so no multiplier is written down anywhere.
 MAX_MODEL_CALLS_PER_TURN_ENV = "ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN"
 DEFAULT_MAX_MODEL_CALLS_PER_TURN = 25
+MIN_MAX_MODEL_CALLS_PER_TURN = 1
+SANE_MAX_MODEL_CALLS_PER_TURN = 200
+"""Not a limit — only the point above which a value is more likely a typo than an
+intent, and worth a warning because the runaway guard stops being one."""
 
 # Deliberately *not* read any more. `MAX_RECURSION_LIMIT` is a shared env name:
 # ringier-a2a-sdk and agent-runner default it to 50, agent-common's dynamic_agent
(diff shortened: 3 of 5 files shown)


packages/orchestrator-agent/app/core/step_budget.py
```diff
@@ -0,0 +1,115 @@
...
+def affordable_model_calls(graph: Any, recursion_limit: int) -> int:
+    """Inverse of :func:`recursion_limit_for` -- what a given limit actually buys.
+
+    Exists to make the old failure legible: it answers "50 steps was how many
+    model calls?" (six), which is the number nobody could see at the call site.
+    """
+    per_call = steps_per_model_call(graph)
+    if per_call <= 0:
+        return 0
+    return max((recursion_limit - base_steps(graph)) // per_call, 0)
```
aartaria: is it even used beside in tests?

aartaria (approved):
LGTM

packages/orchestrator-agent/app/core/step_budget.py
```diff
@@ -0,0 +1,115 @@
...
+    return len(buckets["per_model_call"]) + len(CORE_CYCLE_NODES)
+
+
+def base_steps(graph: Any) -> int:
+    """Super-steps a turn costs regardless of how many model calls it makes.
+
+    The per-turn hooks, minus one for the ``tools`` step the closing model call
+    never takes -- it emits the final-response envelope and the turn ends.
+    """
+    return max(len(classify_nodes(graph)["per_turn"]) - 1, 0)
```
alloy-cicd: **Confirmed off-by-one on the bind-as-tool structured-output paths.** The `- 1` assumes the closing model call never enters `tools`, but that only holds for the ToolStrategy path. When `select_response_format` returns `(None, True)` — thinking enabled on Anthropic/Bedrock, or Gemini via `has_builtin_tools` (`agent_common/a2a/structured_response.py:119-124`) — `FinalResponseSchema` is bound as a regular `return_direct=True` tool (`graph_factory.py:798-811`), and in langchain 1.3.11 that tool call is a pending call routed through the `tools` node before the return-direct edge exits (`langchain/agents/factory.py`, `_make_model_to_tools_edge` only skips `structured_output_tools`, which is empty on this path).

Measured against the real compiled graph (thinking="high"): 1/2/3-call turns cost **11/19/27** super-steps vs the derived 10/18/26 — and with the derived limit, `GraphRecursionError` fires *after* the final tool executed but before the closing `after_agent` hook, i.e. exactly at the edge of budget, while delivering the already-composed answer. There is no slack elsewhere (the ToolStrategy formula is exact).

`tests/test_step_budget.py` pins `MODEL_TYPE="claude-sonnet-4.5"` with `thinking_level=None`, so only the ToolStrategy variant is validated.

Simplest fix consistent with the module's own "over-estimating is the safe direction" doctrine: drop the `- 1` entirely (one spare step on ToolStrategy graphs), or condition it on `requires_response_tool`. Either way, a measured test for a thinking-enabled graph would pin it.

alloy-cicd (commented):

The direction (budget in model calls, multiplier derived per graph) is sound and the measured tests are a genuinely good guard. Two confirmed correctness issues and one scope gap stand out; the rest are robustness/accuracy cleanups.

**Correctness (verified empirically against the real compiled graph):**
1. Off-by-one on the bind-as-tool structured-output paths (thinking on Anthropic/Bedrock, and Gemini): the final `FinalResponseSchema` call **does** execute the `tools` node, so the derived limit is one super-step short and a full-budget turn dies while delivering the final answer. Details on `base_steps`.
2. Sub-agents inherit the parent's new ~200-step `recursion_limit` (deepagents `task` passes no limit; langgraph propagates the ambient config), each dispatch with a fresh counter over a leaner stack — the runaway ceiling per dispatch rises from 50 super-steps to ~60–100 model calls. Details on `graph_factory.py:919`.
3. `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN=0`/negative is accepted and bricks every turn. Details on `config.py:375`.

**Scope note:** the identical super-step-as-budget bug remains live in agent-runner (`agent/core.py:94`, default 50), agent-common's `dynamic_agent` (default 75) and ringier-a2a-sdk (`agent/langgraph.py:54`, default 50) — the sub-agents this orchestrator delegates to will keep truncating at a handful of model calls, and the new deprecation warning actively tells operators to stop setting a variable those packages still read. Fine to defer, but worth a tracked follow-up.

**Alternative worth weighing before this module calcifies:** the pinned langchain 1.3.11 ships `langchain.agents.middleware.ModelCallLimitMiddleware(run_limit=N, exit_behavior="end")` — it counts actual model calls (no node-name reverse-engineering, immune to hook renames and `wrap_model_call` migrations) and ends the turn at a clean boundary with an AI message instead of a mid-step `GraphRecursionError`. `recursion_limit` could then become a single generous backstop constant, and `step_budget.py` + its 300-line test file disappear. The open question it forces (how a jump-to-end turn without `FinalResponseSchema` renders in `agent.py`'s response pipeline) is arguably the question this PR's error path leaves unanswered anyway. Not blocking — the current approach works — but if the answer is "we considered it and rejected it", that reasoning belongs in the module docstring.

Findings verified against the checked-out head; happy to re-verify in a follow-up round.

packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -909,10 +910,19 @@ def _create_graph(self, model_type: ModelType, thinking_level: Optional[Thinking
...
-        # This prevents infinite loops from reaching the high default limit
-        compiled_graph = compiled_graph.with_config({"recursion_limit": self.config.MAX_RECURSION_LIMIT})
-        logger.info(f"Graph created for model: {model_type} with recursion_limit={self.config.MAX_RECURSION_LIMIT}")
+        # Override deepagents' recursion_limit default of 1000, which is too high to
+        # catch a runaway loop. The budget is configured in model calls and converted
+        # here, against *this* graph: the multiplier is the middleware stack's per-call
+        # node cost, which differs between configurations (the PTC middlewares come and
+        # go with CODE_INTERPRETER_PTC), so it cannot be a constant.
+        recursion_limit = recursion_limit_for(compiled_graph, self.config.MAX_MODEL_CALLS_PER_TURN)
+        compiled_graph = compiled_graph.with_config({"recursion_limit": recursion_limit})
```
alloy-cicd, line 919: **The derived ~200-step limit is inherited by every in-process sub-agent, with a fresh counter per dispatch.** deepagents 0.6.12's `task`/`atask` invoke sub-agent graphs with only `{"configurable": {"ls_agent_type": ...}}` (`deepagents/middleware/subagents.py:691-693, 719-721`) and rely on langgraph's ambient config propagation — which does carry `recursion_limit` from the parent run into the child Pregel (verified through `merge_configs`/`ensure_config` in langgraph 1.2.7). Nothing binds a limit on sub-agent graphs themselves.

So where a sub-agent previously inherited 50 super-steps, it now inherits ~200 — measured against a middleware stack it wasn't derived from. Sub-agent stacks are leaner (fewer hook nodes per call), so 200 steps buys them on the order of 60–100 model calls, and each `task` dispatch gets its own fresh budget: the whole-tree runaway ceiling per dispatch rises ~4x in steps and far more in model calls. The deleted comment's stated purpose ("prevents infinite loops") is materially weakened for the nested tree.

If that's acceptable, a sentence in the comment acknowledging the inheritance would help; if not, the sub-agent graphs need their own explicit `recursion_limit` (e.g. derived the same way against their own stacks, or a modest constant).

packages/orchestrator-agent/app/models/config.py
```diff
@@ -331,8 +368,11 @@ class AgentSettings:
     MAX_RETRIES = 3
     BACKOFF_FACTOR = 3.0
 
-    # Recursion limit configuration (overrides deepagents default of 1000)
-    MAX_RECURSION_LIMIT = int(os.getenv("MAX_RECURSION_LIMIT", "50"))
+    # Turn budget, in model calls (overrides deepagents' recursion_limit default of
+    # 1000). GraphFactory converts this to a super-step limit per compiled graph via
+    # app/core/step_budget.py — the multiplier depends on the middleware stack, so it
+    # is counted from the graph rather than written here.
+    MAX_MODEL_CALLS_PER_TURN = _resolve_max_model_calls_per_turn()
```
alloy-cicd, line 412: **No lower bound: `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN=0` (or negative) bricks every turn.** `_int_env` only guards parse failures, so 0/-1 flow through; `recursion_limit_for` then returns `base_steps` (or a negative number) and every request hits the recursion-limit path on its first super-steps — each user message gets the "I've been working on this task for a while and need to take a break" reply having done zero work, with nothing logged that points at the env var. Given this module already prefers falling back over crashing at import, clamping to `max(value, 1)` (with the same warning style as the malformed case) fits the existing pattern. An upper-bound sanity warning (e.g. a typo'd 2500 → ~20k steps, i.e. no runaway protection) would be cheap in the same place.

packages/orchestrator-agent/app/core/step_budget.py
```diff
@@ -0,0 +1,115 @@
...
+            buckets["terminal"].append(name)
+        else:
+            buckets["unclassified"].append(name)
+    return buckets
+
+
+def steps_per_model_call(graph: Any) -> int:
+    """Super-steps one model call costs in *graph*: its hooks, plus model and tools."""
+    buckets = classify_nodes(graph)
+    return len(buckets["per_model_call"]) + len(CORE_CYCLE_NODES)
```
alloy-cicd: `len(CORE_CYCLE_NODES)` hardcodes the 2 instead of using `len(buckets["core"])`, which is computed two lines up and discarded. That quietly contradicts the module's "derived from the compiled graph, nothing written down" contract: if langchain renames `model`/`tools` (both would land in `unclassified`) or compiles without a `tools` node (langchain only adds it when tools exist), the function still returns `hooks + 2` — a plausible-looking number that's wrong by the real core cost, and the mocked graph in `test_orchestrator_thinking_integration.py` keeps passing because it hardcodes the old names. `len(buckets["core"])` makes the derivation self-consistent with the classification it's built on.

packages/orchestrator-agent/app/core/step_budget.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    return max(len(classify_nodes(graph)["per_turn"]) - 1, 0)
+
+
+def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
+    """LangGraph ``recursion_limit`` that affords *max_model_calls* model calls."""
+    per_call = steps_per_model_call(graph)
+    base = base_steps(graph)
+    limit = base + per_call * max_model_calls
+    logger.info(
```
alloy-cicd: **The `unclassified` guard exists only in the test suite; production silently counts unknown nodes as free.** The docstring calls `unclassified` "the important one", but `steps_per_model_call`/`base_steps` ignore it and `recursion_limit_for` logs without checking it — so any node shape this module doesn't recognize (a future `.before_tools`, an async-named hook, a renamed core node) silently shrinks the budget toward exactly the truncation this PR fixes. The only guard, `test_every_node_is_classified`, compiles one graph variant under CI's env; a stack difference between CI and a deployment (env-gated middleware, a langchain bump in prod images) is invisible to it.

One line here restores the invariant at runtime: `if buckets["unclassified"]: logger.warning(...)` (or raise, since a mis-derived budget is arguably worse than a loud startup failure). Cheap and in the spirit of the module.

Two efficiency nits in the same function, take or leave: (a) it classifies the graph twice (`steps_per_model_call` and `base_steps` each call `classify_nodes`) — classify once and derive both numbers; (b) `classify_nodes` reads names via `graph.get_graph().nodes`, which runs langgraph's `draw_graph` edge-discovery simulation (a bounded pregel re-execution) just to list names — the compiled `Pregel` already exposes `graph.nodes` as a plain dict. Both are paid on every `_create_graph`, including the deliberately-uncached cold-start path where the graph is rebuilt per turn; and the INFO log here duplicates the one at the `graph_factory.py:920` call site.

packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -909,10 +910,19 @@ def _create_graph(self, model_type: ModelType, thinking_level: Optional[Thinking
...
             response_format=response_format,
         )
-        # Override deepagents default recursion_limit (1000) with configured value
-        # This prevents infinite loops from reaching the high default limit
-        compiled_graph = compiled_graph.with_config({"recursion_limit": self.config.MAX_RECURSION_LIMIT})
-        logger.info(f"Graph created for model: {model_type} with recursion_limit={self.config.MAX_RECURSION_LIMIT}")
+        # Override deepagents' recursion_limit default of 1000, which is too high to
+        # catch a runaway loop. The budget is configured in model calls and converted
+        # here, against *this* graph: the multiplier is the middleware stack's per-call
+        # node cost, which differs between configurations (the PTC middlewares come and
```
alloy-cicd: The load-bearing justification here is factually wrong: `build_code_interpreter_middlewares` (`agent_common/core/graph_utils.py`) **unconditionally** returns exactly one `_PTCToleranceCodeInterpreterMiddleware` — `CODE_INTERPRETER_PTC` is passed as a constructor flag that changes tool exposure, never the middleware (or node) count. The compiled graph's node set is identical with PTC on and off. The same claim is repeated in `step_budget.py`'s module docstring and `config.py`'s comment. The real (and sufficient) justification is staleness — the multiplier changes whenever a middleware with model hooks is added or removed — so the comments should say that; as written, a future reader will believe the multiplier is deployment-dependent and, e.g., reject a simpler design on evidence that doesn't exist.

packages/orchestrator-agent/app/models/config.py
```diff
@@ -36,6 +36,43 @@ def _int_env(name: str, default: int) -> int:
...
+    """Model calls allowed per turn, warning if the old shared env var is set.
+
+    The legacy name is not honoured, because inheriting it is the bug. Silence
+    would be worse than either choice, so an operator who set it is told that it
+    no longer applies here and what to set instead.
+    """
+    legacy = os.getenv(LEGACY_RECURSION_LIMIT_ENV)
+    if legacy and legacy.strip():
+        logger.warning(
+            "%s=%s is set but no longer configures the orchestrator: it is a shared "
```
alloy-cicd, line 68: This warning is half-misleading for the deployment it targets: `MAX_RECURSION_LIMIT` still configures the sub-agents running **in this same process** — agent-common's `dynamic_agent` reads it (default 75) for every dynamic sub-agent's bound, and agent-runner (50) / ringier-a2a-sdk (50) read it too. An operator who follows "no longer configures the orchestrator … use ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN instead" and unsets it silently changes every sub-agent's recursion bound. Also worth noting: `dynamic_agent.py:117`'s comment still says it "mirrors the orchestrator's MAX_RECURSION_LIMIT", an attribute this PR deletes. Suggest the warning say the variable remains live for sub-agents, and a follow-up ticket for porting the model-call budget to the three sibling packages — they still have the exact 50-steps≈6-calls truncation this PR diagnoses, which will now resurface one layer down as sub-agents returning partial results.

packages/orchestrator-agent/tests/test_step_budget.py
```diff
@@ -0,0 +1,300 @@
...
+
+def test_fifty_steps_was_six_model_calls():
+    """The reported bug, in the unit that makes it obvious.
+
+    Keeps the finding legible: the old default was not "50 of something
+    generous", it was six model calls, which an ordinary two-delegation turn
+    exceeds. If this number ever climbs to something comfortable, the middleware
+    stack got cheaper and the incident is worth re-reading.
+    """
+    assert affordable_model_calls(_compiled_graph(), 50) == 6
```
alloy-cicd: This pins a magic 6 that is a function of the ambient middleware stack: any PR that adds/removes a `before_model`/`after_model` hook flips it to 5 or 7 and fails with a message about a historical incident rather than the change at hand — and the only way to satisfy it is to edit the constant, so it documents rather than protects. The incident number would live just as legibly in the module docstring or PR description. (Related: `affordable_model_calls` has no production caller — its only consumers are this test and the round-trip test above, which can only fail if the two functions disagree with each other, not with the graph. Both could move into the test module or be dropped.)

packages/orchestrator-agent/app/core/step_budget.py
AlexMetsai: step_budget.py:88

packages/orchestrator-agent/app/models/config.py
AlexMetsai: config.py:375

packages/orchestrator-agent/app/core/step_budget.py
AlexMetsai: step_budget.py:79

packages/orchestrator-agent/app/core/step_budget.py
AlexMetsai: step_budget.py:96

packages/orchestrator-agent/app/core/graph_factory.py
AlexMetsai: graph_factory.py:916

packages/orchestrator-agent/tests/test_step_budget.py
AlexMetsai: test_step_budget.py:241

packages/orchestrator-agent/app/core/graph_factory.py
AlexMetsai: Good catch on the mechanism — it's real, but it doesn't reach our sub-agents, and
the reason is your own point in the `config.py:64` thread.

I reproduced the propagation with plain langgraph, forwarding only the
`configurable` stamp the way `task`/`atask` does. The child loops forever, so the
number of steps it reaches before `GraphRecursionError` *is* the limit that
applied to it:

| child | parent | steps the child got |
|---|---|---|
| unbound, standalone | — | 10007 |
| binds 5 | 203 | **5** |
| unbound | 203 | **203** |
| unbound | 50 | 50 |
| binds 75 | 203 | **75** |

So an *unbound* child does inherit the parent's limit — you're right about that.
But a child that binds its own wins the collision, exactly as the `task`/`atask`
comment claims.

Every sub-agent kind we have is covered:

- **Dynamic local sub-agents** (`general-purpose`, `task-scheduler`,
  `agent-creator`, `skill-assessor`, user-created) bind their own at
  `agent_common/agents/dynamic_agent.py:1522` —
  `.with_config({"recursion_limit": _SUB_AGENT_RECURSION_LIMIT})`, default 75.
  That's the last row above: measured to keep 75 under a 203 parent.
- **`file-analyzer`** is a `FileAnalyzerRunnable` and **Foundry agents** are
  `FoundryLocalAgentRunnable`s — plain runnables, not compiled graphs, so there is
  no super-step budget to inherit.
- **Remote A2A sub-agents** run in another process.

A repo-wide grep finds exactly two `recursion_limit` bindings: that line and
`graph_factory.py:919`. So "nothing binds a limit on sub-agent graphs themselves"
isn't the case, and the nested runaway ceiling is unchanged by this PR.

Worth noting these two comments interact: `MAX_RECURSION_LIMIT` still being live
for sub-agents is what makes my warning misleading (your other point, which is
right and I'm fixing), and it is also what stops them inheriting 203. Fixing the
wording resolves both.

Separately, you were right to be suspicious of that comment in general: it claimed
deepagents defaults to 1000. It's **9,999** (`deepagents/graph.py:880`) — I had
inherited the number from the comment I replaced without checking it. Fixing that
too.

packages/orchestrator-agent/app/models/config.py
AlexMetsai: Reworded. It now leads with "no longer configures the orchestrator", then explicitly says not to unset it, names the sub-agent consumers with their defaults, and points at SUB_AGENT_RECURSION_LIMIT as the way to decouple the two budgets — dynamic_agent already honours that ahead of the shared name, which your comment made me go and check. A test pins the wording, since the wording is the whole fix.

dynamic_agent.py:117 fixed too. It now records the sub-agent bound as its own rather than "mirroring" a deleted attribute, and notes why the with_config beneath it matters: langgraph propagates recursion_limit into a child graph that doesn't set one, so that line is what keeps sub-agents off the orchestrator's limit — the measurement is in the other thread.

Follow-up for porting the model-call budget to agent-runner, agent-common and ringier-a2a-sdk: <ISSUE LINK>. You're right that they still carry the 50-steps-is-six-calls truncation, and that it will now show up one layer down as sub-agents returning partial results.

AlexMetsai: https://github.com/ringier-data/nannos/issues/216

packages/orchestrator-agent/app/models/config.py
```diff
@@ -36,6 +36,79 @@ def _int_env(name: str, default: int) -> int:
...
+            "name used by agent-runner, agent-common and ringier-a2a-sdk with "
+            "different defaults, and inheriting a sub-agent's super-step budget is "
+            "what truncated orchestrator turns. Use %s instead -- it is counted in "
+            "model calls, not super-steps.",
+            LEGACY_RECURSION_LIMIT_ENV,
+            legacy,
+            MAX_MODEL_CALLS_PER_TURN_ENV,
+        )
```
alloy-cicd, line 76: **The legacy warning is half-misleading, and following it changes sub-agent behaviour.**

`MAX_RECURSION_LIMIT` is not only a shared name used by agent-runner, agent-common and ringier-a2a-sdk — in *this* process it still configures the sub-agents the orchestrator dispatches to: `agent-common/agent_common/agents/dynamic_agent.py:117-122` reads it as the fallback for `SUB_AGENT_RECURSION_LIMIT` (default 75), and that value is applied at `dynamic_agent.py:1522`.

Scenario: a deployment exports `MAX_RECURSION_LIMIT=120` to give sub-agents headroom. After this PR the orchestrator logs *is set but no longer configures the orchestrator … Use ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN instead*. The operator does exactly that — sets the new variable, removes the old one — and every local sub-agent silently drops from 120 to 75 super-steps, reintroducing the same mid-task truncation one layer down, with nothing in the logs connecting the two.

The message needs a clause saying the variable still bounds sub-agents in this same process and should be kept if it was set for them. Related: the comment at `dynamic_agent.py:117` (Mirrors the orchestrator MAX_RECURSION_LIMIT) is stale as of this PR and names an attribute that no longer exists.

packages/orchestrator-agent/app/core/step_budget.py
```diff
@@ -0,0 +1,138 @@
...
+def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
+    """LangGraph ``recursion_limit`` that affords *max_model_calls* model calls.
+
+    Classifies once: the two components are derived from a single pass, since
+    this runs on every ``_create_graph`` including the cold path that rebuilds
+    per turn.
+    """
+    buckets = classify_nodes(graph)
+    _warn_unclassified(buckets)
+    return _base_steps(buckets) + _steps_per_model_call(buckets) * max_model_calls
```
alloy-cicd, line 138: **The derived limit has no floor, so a classification that finds nothing produces `recursion_limit=0`.**

The operator-facing input is floored (`MIN_MAX_MODEL_CALLS_PER_TURN`), but the derived output is not. If `compiled_graph.nodes` ever comes back empty or all-terminal — a langgraph/deepagents change that moves node registration off the `Pregel.nodes` dict, or `create_deep_agent` returning a wrapper — this returns `0 + 0 * 25 == 0`, `with_config(recursion_limit=0)` is applied, and **every** turn dies on its first super-step with `GraphRecursionError`, surfacing as *I have been working on this task for a while and need to take a break* having done no work.

That is the same failure the config-side clamp was added to prevent, and the same argument the PR makes for charging unknown nodes at runtime applies here: `test_every_node_is_classified` and `test_the_core_cycle_cost_is_counted_not_assumed` only assert `buckets["core"]` is non-empty in CI, against one compiled variant under CI environment. A floor, or a loud failure when `core` and `per_model_call` are both empty, turns a silent total outage into a startup error.

packages/orchestrator-agent/tests/test_step_budget.py
```diff
@@ -0,0 +1,367 @@
...
+
+    `MAX_RECURSION_LIMIT` is read by agent-runner and ringier-a2a-sdk (default
+    50) and agent-common (75). A deployment pinning 50 — the orchestrator's own
+    former default — used to silently reinstate the truncation, and CI could not
+    catch it: the env var is unset there, so the tests stayed green against the
+    intended budget.
+    """
+    monkeypatch.setenv(LEGACY_RECURSION_LIMIT_ENV, "50")
+
+    assert _resolve_max_model_calls_per_turn() == DEFAULT_MAX_MODEL_CALLS_PER_TURN
```
alloy-cicd, line 306: **This asserts the default without clearing the variable the PR introduces.**

The test sets the legacy name but never clears `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`. Run the suite in a shell or container that exports it — precisely what the PR asks operators to do, and what a developer reproducing a deployment would do — and `_resolve_max_model_calls_per_turn()` correctly returns e.g. 40 while the test fails on `assert 40 == 25`, reporting a false regression in the legacy-env behaviour it is actually meant to pin.

```suggestion
    monkeypatch.delenv(MAX_MODEL_CALLS_PER_TURN_ENV, raising=False)
    monkeypatch.setenv(LEGACY_RECURSION_LIMIT_ENV, "50")

    assert _resolve_max_model_calls_per_turn() == DEFAULT_MAX_MODEL_CALLS_PER_TURN
```

packages/orchestrator-agent/app/core/step_budget.py
```diff
@@ -0,0 +1,138 @@
...
+def steps_per_model_call(graph: Any) -> int:
+    """Super-steps one model call costs in *graph*: its hooks, plus model and tools."""
+    buckets = classify_nodes(graph)
+    _warn_unclassified(buckets)
+    return _steps_per_model_call(buckets)
+
+
+def base_steps(graph: Any) -> int:
+    """Super-steps a turn costs regardless of how many model calls it makes."""
+    return _base_steps(classify_nodes(graph))
```
alloy-cicd, line 126: **`base_steps` is unused, and it is the one entry point that does not warn.**

`recursion_limit_for` is the only conversion production uses, and `tests/test_step_budget.py` imports only `classify_nodes`, `recursion_limit_for` and `steps_per_model_call` — this is left over from the commit that dropped the inverse helper.

It also has a quietly different contract from its two siblings: no `_warn_unclassified`. A later caller that computes the per-turn overhead through it against a graph with unrecognised nodes gets a number with nothing in the logs pointing at the stale classification, which is the exact failure mode the previous review round fixed for production. Either delete it or route it through `_warn_unclassified` like the others.

alloy-cicd (commented):

Summary for the four inline comments posted alongside this review. The direction of the PR — budgeting in model calls and deriving the graph's recursion limit from the compiled node classification — is sound, and it degrades safely: every node-drift direction (renames, removals) pushes nodes into `unclassified`, which is charged at the per-model-call rate, so the estimate errs high rather than low.

**Findings**

1. `app/models/config.py:76` — the legacy `MAX_RECURSION_LIMIT` warning is half-misleading. The variable still bounds every local sub-agent in this same process (`dynamic_agent.py:117-122` uses it as the `SUB_AGENT_RECURSION_LIMIT` fallback, default 75, applied at `dynamic_agent.py:1522`), so an operator who follows the advice and unsets it silently drops each sub-agent to 75 super-steps. The comment at `dynamic_agent.py:117` is also stale.
2. `app/core/step_budget.py:138` — `recursion_limit_for` floors its input via the config clamp but not its output. A graph whose classification finds no core/hook nodes yields a recursion limit of 0, which fails every turn on the first super-step. The CI-only-guard argument the PR makes for unknown nodes applies here too.
3. `tests/test_step_budget.py:306` — `test_the_shared_env_var_no_longer_configures_the_orchestrator` asserts the default without `delenv`-ing `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`, so it fails in any environment that exports the variable this PR introduces.
4. `app/core/step_budget.py:126` — `base_steps()` is dead code (no production or test caller) and is the only public entry point that skips `_warn_unclassified`.

**Checked and cleared**

- Both structured-output paths are genuinely covered: `select_response_format` has exactly two outcomes (a `ToolStrategy`, or none-plus-bind-as-tool), and the `[None, ThinkingLevel.high]` parametrization hits both, so the `base_steps` off-by-one fixed in the last commit cannot recur silently. Gemini's path collapses onto the same bind-as-tool branch.
- `_steps_per_model_call` correctly counts `core` from the classification rather than assuming 2.
- The PR's open question about sub-agents inheriting the derived ~203 does not hold: `create_runnable_config` (`ringier-a2a-sdk/utils/config.py:23`) builds a fresh config with no `recursion_limit` key, and `dynamic_agent.py:1747` spreads that into `standalone_config`, so the sub-agent graph's own `with_config` recursion limit of 75 is what wins. Sub-agents stay at 75 before and after this PR.
- `RepeatedToolCallMiddleware` is configured with `max_repeats=5` (`graph_factory.py:199`), so the measurement tests' three identical `get_current_time` calls stay under the loop-detection threshold — though the helper would trip it past ~6 model calls if the parametrization is extended.
- No stale `MAX_RECURSION_LIMIT` references remain in orchestrator code, tests, docs, yaml or shell scripts. `test_orchestrator_thinking_integration.py` is the only test that patches `create_deep_agent`, and it was correctly updated to expose an iterable `nodes`.
- `graph.nodes` survives `with_config` (confirmed by the pre-existing `tests/integration/test_edge_cases.py:79`, which reads `graph.nodes.keys()` after the same call), so both the new production read and `test_the_graph_is_compiled_with_the_derived_limit` are on solid ground.

Submitted as a comment — the human reviewer decides.

PR #214, opened 2026-09-07
fix(console-backend): restore scheduled-spend attribution, and tell the console's own spend from the agent's

### LINKED ISSUE #213: console-backend: harden gateway_chat for unattended use — attribute scheduled spend, share the JSON salvage, never cut a reply off silently
## Problem

Follow-up to #208, which fixed the draft generator being cut off by its own reasoning tokens. Reviewing every `gateway_chat` call site afterwards showed the same class of exposure elsewhere, plus two smaller gaps.

### 1. Scheduled spend is unattributed

The gateway logs cost only when the call carries `x-litellm-spend-logs-metadata` with a `user_sub`. Four of six callers pass it; the two that don't are exactly the ones that run unattended, on every check or trigger:

| caller | `user_sub` | cadence |
|---|---|---|
| draft / condition generation | yes | on request |
| conversation titling | yes (+ `conversation_id`) | once per conversation |
| catalog summarization | yes (+ `catalog_id`) | per document |
| watch judge (`watch_evaluator`) | **no** | every check of a job with `llm_condition` |
| trigger-notification writer (`scheduler_engine._write_notification`) | **no** | every trigger |

Both have the job owner in hand: the engine fetches the owner's access token by `job.user_id` before evaluating, so the sub is one lookup away.

### 2. Conversation titling duplicates the JSON salvage

`parse_summary_response` re-implements, character for character, the fence-stripping / brace-matching / `json.loads` salvage that `gateway_chat_json` centralised — the duplication that helper was introduced to remove. It therefore also misses the finish-reason logging and the `GatewayReplyTruncated` signal added in #209.

### 3. A cut-off reply is only noticed by the JSON helper

`gateway_chat` returns the text and, since #209, the finish reason — but only `gateway_chat_json` acts on it. Two callers still run with thinking on and a small budget:

- the watch judge: `max_tokens=512`, no `reasoning_effort`. A cutoff reads as "condition not met" and is swallowed by the broad `except`.
- catalog summarization: `max_tokens=200`, no `reasoning_effort`. A cutoff stores half a sentence as the document summary.

## Proposal

- **Attribution:** pass `{"user_sub": <owner sub>, "scheduled_job_id": job.id}` from the watch judge and the notification writer, mirroring what titling does with `conversation_id`.
- **Shared salvage:** have conversation titling call `gateway_chat_json` and keep only its field validation and trimming.
- **Never silent:** `gateway_chat` logs a warning whenever `finish_reason == "length"`, so no caller can be cut off without a trace. Default `reasoning_effort` to `"none"` in the helper — every console-backend call is a mechanical utility call — with callers opting *in* to thinking. The judge is the only arguable case, and a truncated judgement is worse than an unreasoned one.

## Acceptance

- [ ] the watch judge and the notification writer carry `user_sub` (and the job id) on their gateway calls
- [ ] conversation titling parses through `gateway_chat_json`; `parse_summary_response` no longer contains its own salvage
- [ ] a `finish_reason == "length"` reply is logged by `gateway_chat` for every caller
- [ ] thinking is off by default in `gateway_chat`; any caller that wants it says so explicitly

aartaria:
closes ringier-data/nannos#213

Follow-up to #209. Reviewing every `gateway_chat` call site turned up the three things the issue lists, and then a fourth that only became visible once we understood how spend is classified. Grew accordingly — it now touches three packages plus the console frontend.

### 1. Scheduled spend was unattributed — a regression, not an old gap

The issue framed this as "two of six callers never passed a `user_sub`". That's the symptom. Both calls — the watch judge and the trigger-notification writer — **used to run inside the agent**, where the SDK's attribution ContextVars carried the owner and the job id automatically. Moving the watch decision into the scheduler (#166, so a quiet poll stops costing an agent run) took them out of that context, and console-backend set those vars nowhere.

The spend didn't become hard to read — it **left `usage_logs` entirely**. `custom_logger._build_record` returns `None` when there's no `user_sub`, so no record is POSTed at all.

So the dispatch re-enters the scope the agent path never left, rather than threading a parameter. `owner_sub` came out of five signatures, and whatever gateway call this path grows next is attributed without anyone remembering. A scope rather than `set_attribution` because `_tick` dispatches each job in its own task but `run_job_now` awaits the dispatch **inline from a request handler**, where a job id left set would follow the rest of that request.

### 2. Console-backend's own spend was booked as agent spend

`get_usage_by_service` never stored a service — it **derived** one from whichever id was set, with `'orchestrator'` as the else. A console utility call carries no id at all, so drafting a job, generating a condition and naming a conversation all read as agent runs the user never made.

`service` is now a declarable dimension: a ContextVar in the SDK, carried through the proxy's `_build_record` (and opted into its dead-letter allow-list — not PII, and it says whose spend was lost), stored on `usage_logs` by migration **091**. The read path is `COALESCE(service, <the old CASE>)`, so rows written before the column existed and every agent-path row classify exactly as they did. `context_header` deliberately does *not* forward it: the receiver is a different service doing its own work.

### 3. Three callers were billing the wrong identity

`users.id` is the OIDC subject only for people onboarded before the id/sub split. The scheduler holds `scheduled_jobs.user_id`, titling the socket session's `user_id`, catalog sync `catalogs.owner_user_id` (`JOIN users u ON c.owner_user_id = u.id`). The latter two landed only because the ingest falls back to an id lookup — the fallback whose comment reads *"some callers pass owner_user_id"*, which is precisely these.

All three resolve the subject through one helper, `billing_subject`, which **keeps the id as a fallback**: a failed lookup should cost the right shape, not the usage row. The catalog fix reaches its embeddings too, which took the same value.

### 4. The usage table still said "Orchestrator"

Found in local testing, and the most instructive bug here: `/my-logs` builds `UsageLog` **field by field**, so `service` was dropped between a query that selected it and a response model that declared it. The frontend then fell back to a hardcoded `'Orchestrator'` — the same guess-from-what-is-set the backend had just stopped making, reimplemented in TSX.

Fixed at both ends, with a Postgres-backed test over the whole write → list → response path, because a field-by-field constructor drops the next field just as quietly.

### 5. One way to attribute console-backend's own work

Console-backend's own work now runs in an `attribution_scope` everywhere — the scheduler dispatch, catalog sync, titling, and both console request handlers. The argument that settled the request handlers is the one behind the whole branch: forgetting `metadata` on a call added later doesn't misclassify the spend, it *loses* it. `console_web_search` is the documented exception — not this service's own work, but a step in an agent's turn that we host, carrying the caller's forwarded context. The rule lives in `spend_attribution`'s docstring rather than being re-decided per caller.

### 6. The rest of #213

- **Shared salvage:** `parse_summary_response` carried a character-for-character copy of the salvage `gateway_chat_json` centralised, and so missed the finish-reason logging and `GatewayReplyTruncated` from #209. It now takes the parsed object and keeps only its title-specific validation.
- **Never silent:** `gateway_chat` warns whenever `finish_reason == "length"` — truncation had been becoming a half-sentence document summary, or a judgement swallowed as "condition not met". `reasoning_effort` defaults to `"none"`; thinking is opt-in, `None` still means "send nothing". That turns it off for the judge (512 tokens) and catalog summarization (200).

### Two bugs the tests caught in this PR's own changes

- Adding a `service` column silently broke `GROUP BY service`: with a real column of that name, Postgres binds `GROUP BY` to the **input column** rather than the output alias, so it stopped grouping by the `COALESCE` expression. It's `GROUP BY 1` now. Only the database-backed test could see this.
- The owner lookup rides the session the rest of the dispatch runs on, so swallowing its error without `rollback()` left the session refusing every subsequent statement — a best-effort lookup that would have killed the run.

### Notes for the reviewer

- **Deploy ordering.** The proxy deploys separately, and both orders degrade gracefully: an old proxy drops the new field and rows fall back to derivation (today's behaviour); a new proxy against an old backend sends `service: None`. Migration 091 must land before the ingest writes the column.
- **SDK regen drift.** Regenerating the frontend SDK also picks up `/auth/me/entitlement-version` from #206, which merged without regenerating it. Unrelated to this change, but it is what the documented command produces.
- **UI naming.** `Agent` → `Source` (the column holds a sub-agent, a scheduled job, *or* a service), and `Direct API Calls` → `Calls outside a conversation`.

### Testing

- console-backend `uv run pytest tests/`: **1759 passed**, including Postgres-backed tests for the classification and the read path.
- ringier-a2a-sdk **392 passed** (2 skipped); litellm-proxy **42 passed**.
- Dependents of the changed SDK module: agent-common **23 passed**, orchestrator-agent content builder **32 passed**. (`agent-common/tests/test_foundry_cost_version.py` can't collect in a fresh worktree — `foundry_sdk` isn't installed — pre-existing and unrelated.)
- console-frontend: `tsc` clean, `vite build` clean, and `UsagePage` lint is one warning *better* than base (the label maps moved to module scope).
- `ruff check` reports the same 22 pre-existing findings as `main`.

## Checklist


alloy-cicd (commented):

The attribution/threading design is sound and the tests are thorough. Seven findings, ranked; only the first is a functional bug, the rest are behavior-change blast radius, log noise, lost test coverage, and duplication.

1. **`_owner_sub` swallows a DB error without rolling back the shared session** — a failed lookup poisons the session used by the rest of the dispatch, breaking the function's own "must not stop the job" contract (scheduler_engine.py:501).
2. **The `reasoning_effort="none"` default flip silently reaches two untouched call sites** — the watch judge (the one semantic-judgment caller, with no test pinning it) and catalog summarization — and a provider that accepts the param but rejects the value `"none"` turns both into silent fallbacks (llm_gateway.py:143).
3. **The salvage tests deleted from test_conversation_summary.py are not actually "covered there"** — test_llm_gateway.py has no fence/prose/malformed-JSON tests (test_conversation_summary.py:19).
4. **A truncated JSON reply now warns 2–3 times for one event**, and the new warning also fires when the object is still salvaged (llm_gateway.py:198).
5. **The owner-sub lookup runs on every dispatch but is consumed on only two narrow paths**, and the justifying docstring claim ("already reads the owner to mint the access token") is inaccurate (scheduler_engine.py:204).
6. **The attribution-metadata ternary is duplicated and drops `scheduled_job_id` when the sub is missing** (watch_evaluator.py:312).
7. **Stale docstring**: gateway_chat's callers no longer all run "outside any sub-agent / scheduled-job context" (llm_gateway.py:166).

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -465,6 +489,22 @@ async def _write_notification(self, job: ScheduledJob, outcome: WatchOutcome | N
...
+    async def _owner_sub(self, db: Any, job: ScheduledJob) -> str | None:
+        """The OIDC subject of the job's owner, or None when it cannot be read.
+
+        None is not a failure: an unattributed LLM call is worse accounting than an
+        attributed one, but a job must not stop running because a lookup came back empty.
+        """
+        try:
+            result = await db.execute(text("SELECT sub FROM users WHERE id = :user_id"), {"user_id": job.user_id})
+            sub = result.scalar_one_or_none()
+        except Exception:
```
alloy-cicd: **Correctness — the swallowed DB error leaves the shared session in a failed transaction.**

This `db` is the same session `_dispatch_job` goes on to hand to the watch evaluator (`ModelDefaultsRepository.get_all`) and `_build_message_args`. After a failed `execute`, SQLAlchemy raises `PendingRollbackError` on every subsequent statement until `rollback()` is called — so the docstring's contract ("a job must not stop running because a lookup came back empty") holds for an *empty* result but not for a *failed* query: the dispatch dies a few statements later with an unrelated-looking error. And since this SELECT now runs on every dispatch (see the companion comment at line 204), job types that never consume `owner_sub` — plain task jobs, CEL-gate-only watches — gained this failure mode too.

Fix: `await db.rollback()` in this `except` (or run the lookup in its own short-lived session).

Verification hook for the next round: a test where `db.execute` raises for the `SELECT sub` query and the dispatch still evaluates/dispatches would pin this.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -195,13 +195,21 @@ async def _dispatch_job(self, job: ScheduledJob, run_id: int | None = None) -> N
                     )
                     return
 
+                # Who the job runs as, in the gateway's vocabulary. The proxy attributes
+                # spend by OIDC subject and logs nothing without one, and both LLM calls
+                # on this path (the watch judge, the notification writer) run unattended
+                # on a schedule — exactly where cost accumulates unseen. `job.user_id` is
+                # the internal user id, which is not the subject for anyone onboarded
+                # after the id/sub split, so it is looked up rather than passed through.
+                owner_sub = await self._owner_sub(db, job)
```
alloy-cicd: **Placement/efficiency — resolved unconditionally on every dispatch, but consumed on only two narrow paths** (the LLM judge for `llm_condition` watches, and `_write_notification` for notify-only watches whose author left the message empty). Task jobs — the common case — pay a users-table round trip per dispatch, concurrently per tick, for a value nothing reads.

Also, the justifying docstring in `watch_evaluator.evaluate` ("the engine resolves it once per dispatch (it already reads the owner to mint the access token)") is inaccurate: `SchedulerTokenService.get_access_token` reads `user_offline_tokens`, never `users` — this is a genuinely new query, not a reuse.

Two cheaper options:
- Decode the sub from the token already in hand: `access_token` is a Keycloak JWT whose `sub` claim is exactly this value — `jwt.decode(access_token, options={"verify_signature": False})["sub"]` (precedent: auth_router.py:168). Zero I/O and it cannot drift from the identity the job actually runs as.
- Or resolve lazily at the two consumers, via the existing `UserService.get_user` (user_service.py:66) rather than a second bespoke `users` query that can drift from the canonical one (e.g. if soft-delete filtering is added there).

packages/console-backend/console_backend/services/llm_gateway.py
```diff
@@ -140,7 +140,7 @@ async def gateway_chat(
     model: str,
     max_tokens: int = 1024,
     metadata: dict | None = None,
-    reasoning_effort: str | None = None,
+    reasoning_effort: str | None = "none",
```
alloy-cicd, line 154: **Behavior-change blast radius — the default flip silently reaches the two call sites that pass nothing:**

1. **The watch judge** (watch_evaluator.py:308) — the one caller doing *semantic judgment* ("an attendee looks external to the company") rather than mechanical JSON-filling, with 512 tokens of headroom. A reasoning model that previously deliberated now answers with thinking off; a borderline verdict can flip, unattended, with no diff line touching that file and no test pinning the judge's `reasoning_effort`. If intended (the docstring's "judge a stated condition" suggests it is), an explicit `reasoning_effort="none"` + test at the judge would record the decision.
2. **Catalog summarization** (catalog/sync.py:1965) — likely an improvement at max_tokens=200, but also unreviewed here.

Two secondary risks worth a thought:
- `drop_params: true` only drops the param for models that don't support it at all. A model that accepts `reasoning_effort` but rejects the value `"none"` (e.g. o-series accepts only low|medium|high) now 400s on these two previously-working paths — and both fail *silent* (summary falls back to `Document: <name>`, judge fails closed so the watch never fires).
- The escape hatch inverts intuition: `reasoning_effort=None` now means *more* reasoning than the default. A future caller writing `None` as the "safe/off" choice re-enables thinking on a small budget — the exact truncation this PR series exists to prevent. A named sentinel for "omit the param" would avoid repurposing `None`.

packages/console-backend/console_backend/services/llm_gateway.py
```diff
@@ -183,7 +188,21 @@ async def gateway_chat(
...
     content = choice.get("message", {}).get("content")
-    return GatewayText(content or "", finish_reason=choice.get("finish_reason"))
+    finish_reason = choice.get("finish_reason")
+    if finish_reason == "length":
+        # Logged here, for every caller, rather than in `gateway_chat_json` alone: a reply
+        # that was cut off looks like a short one to a prose caller too — a half-sentence
+        # stored as a document summary, a truncated judgement read as "condition not met".
+        # The callers differ in what they do about it; none of them should have to notice
+        # it for themselves, and before this it went unlogged on four of the six paths.
+        logger.warning(
```
alloy-cicd, line 205: **Log noise — one truncation event now warns 2–3 times, and this warning also fires on successful calls.**

On the JSON path: this new warning fires, then `gateway_chat_json`'s existing "No JSON object … finish_reason=length" warning logs the same model/length/snippet again before raising `GatewayReplyTruncated`, and the caller logs a third time (conversation_summary's "titling call failed"). Log-based alerting counts one budget cutoff as three warnings per poll on a busy judged watch.

Also, when `finish_reason == "length"` but a complete object is still salvaged (the reply was cut exactly after the closing brace), `gateway_chat_json` returns the dict and the call *succeeds* — yet this "cut off" warning fired anyway.

Suggestion: keep this warning (it's the one that covers prose callers) and have `gateway_chat_json` skip its own shape-warning when `finish_reason == "length"` — just raise, since everything its log line adds was recorded here.

packages/console-backend/console_backend/services/watch_evaluator.py
```diff
@@ -292,7 +301,16 @@ async def _judge(
...
-            parsed = await gateway_chat_json(prompt, model=model, max_tokens=512)
+            # The gateway logs cost only for a call that names a subject, and this one
+            # runs unattended on every poll of every judged watch — the spend that most
+            # needs attributing was the spend it recorded nothing about. The job id rides
+            # along so the bill can be read per watch, not just per owner.
+            parsed = await gateway_chat_json(
+                prompt,
+                model=model,
+                max_tokens=512,
+                metadata={"user_sub": owner_sub, "scheduled_job_id": job.id} if owner_sub else None,
```
alloy-cicd: **Duplication — this ternary exists character-for-character here and in `scheduler_engine._write_notification`, and it throws away `scheduled_job_id` when `owner_sub` is None** even though the job id is known and would let the unattributed bucket still be grouped per watch (`_gateway_headers` already strips None values, so passing the dict unconditionally may be all it takes — depending on whether the proxy tolerates sub-less metadata).

Bigger-picture (optional): the SDK already has the canonical mechanism for exactly this — `ringier_a2a_sdk.cost_tracking.attribution` (`attribution_scope` / `attribution_header`, ContextVar-based), and llm_gateway already imports from `ringier_a2a_sdk`. A scope around the dispatch would delete the owner_sub threading through five signatures and auto-attribute any *future* LLM call on this path — with the current optional-`owner_sub=None` plumbing, the next call site that forgets to thread it compiles, tests green, and silently reintroduces the invisible-spend bug this PR fixes.

packages/console-backend/tests/test_conversation_summary.py
```diff
@@ -16,49 +16,42 @@
 # ---- parsing ---------------------------------------------------------------
 
 
-def test_parses_plain_json():
-    parsed = cs.parse_summary_response('{"title": "Campaign 42 pacing", "summary": "Why campaign 42 under-delivered."}')
-    assert parsed == ("Campaign 42 pacing", "Why campaign 42 under-delivered.")
+# Digging the object out of fences and prose belongs to `gateway_chat_json` and is
```
alloy-cicd, line 37: **Test coverage — "covered there" is not currently true.**

test_llm_gateway.py has no test for fence-stripping (` ```json … ``` `), prose-wrapped objects, empty replies, non-dict JSON (`[1, 2, 3]`), or the `json.JSONDecodeError` branch (`'{"broken": '`) of `gateway_chat_json` — the only remaining salvage coverage is the prose-with-no-object case. The cases deleted here were the only unit coverage of those branches; a regression in the fence/brace regexes (llm_gateway.py:243-244) would pass the whole suite while titling, the watch judge, and condition generation silently start reading `{}`.

Suggestion: port the deleted cases (fenced+chatter, plain JSON, empty, `[1,2,3]`, broken JSON) into `TestGatewayChatJson` so the claim in this comment becomes true.

packages/console-backend/console_backend/services/llm_gateway.py
```diff
@@ -150,10 +150,15 @@ async def gateway_chat(
...
+    utility call (title a conversation, summarize a document, judge a stated condition)
+    running on the cheap tier with a small `max_tokens`, and reasoning tokens count
+    against that budget — a reasoning model spends it thinking and is cut off partway
+    into the answer. A caller that genuinely wants the model to reason says so; passing
+    ``None`` omits the parameter entirely and leaves the model to reason as it normally
+    would. The proxy runs `drop_params: true`, so a model that takes no such param is
+    unaffected either way.
 
     Note: the canonical attribution-header builder lives in agent-common
     (`attribution.attribution_header`, used by the chat client + embeddings adapter). It is
```
alloy-cicd: **Stale docstring (minor), two lines below this hunk (166-168):** "gateway_chat's only callers (watch-param generation, catalog summarization) run outside any sub-agent / scheduled-job context" — this PR makes the scheduler's watch judge and notification writer callers, from a scheduled-job context, passing `scheduled_job_id` in metadata. The rationale for not importing agent-common's attribution builder no longer matches the call sites.

aartaria:
Round 1 triage — 4 of 7 addressed in 3a9632a7, 3 discarded with reasons.

**Fixed**

1. **`_owner_sub` left the shared session in a failed transaction** — confirmed and the one real bug. The lookup rides the session the rest of `_dispatch_job` runs on, so swallowing the error broke the very promise it was written for: the next statement would have raised `PendingRollbackError` and killed the run, on every job type. It now rolls back, and that discards nothing — the token refresh before it commits its own writes (`scheduler_token_service`). Covered by two tests: the rollback happens, and a rollback that fails too is still not fatal.
3. **Salvage lost its unit coverage** — right, `test_llm_gateway.py` had no fence/prose tests. Restored where the salvage now lives: fenced, fenced-with-chatter, plain, plus the no-object replies (empty, prose, malformed, a bare array, two objects in one reply).
5 (part) / 7. **Inaccurate docstrings** — both corrected. `gateway_chat`'s note no longer claims its callers run outside any scheduled-job context, and `evaluate`'s no longer claims the engine "already reads the owner" (it fetches a token by user id; the subject is a separate read).

**Discarded**

2. **Blast radius of the `reasoning_effort` default flip** — this is the change the issue asks for, and the judge is the case it weighed by name: a truncated judgement is swallowed as "condition not met", which is worse than an unreasoned one. It is no longer unpinned — there is now a test asserting the judge runs with thinking off and saying why. On a provider rejecting the *value*: the proxy runs `drop_params: true`, and two callers have been sending exactly `"none"` in production since #209, so the value is not new, only its reach. `None` meaning "let the model decide" is the documented escape hatch, not an accident.
4. **Duplicate truncation warnings on the JSON path** — the two lines state different facts: one that the output budget ran out, one that nothing parseable came back. Collapsing them would put the cutoff knowledge back inside `gateway_chat_json`, which is the coupling this PR exists to remove. Firing when a complete object was still salvaged is intended too — the reply *was* cut off, and whatever followed the object was lost.
6. **The metadata ternary is in two places** — two one-line expressions in two modules; a shared helper for that is more indirection than it removes, and `attribution_scope` lives in agent-common, which console-backend deliberately does not depend on. Dropping `scheduled_job_id` when there is no sub is deliberate: the proxy logs nothing for a call that names no subject, so job-id-only metadata would be a header nobody reads.
5 (rest). **Where the lookup runs** — it stays unconditional: one primary-key read on an already-open session, and with the rollback in place it can no longer harm the dispatch. Decoding the JWT trades an authoritative read for a parsing dependency, and `UserService.get_user` selects fifteen columns and builds a `User` to yield one string.

Full suite after the fixes: **1742 passed**.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (e79ef395296e2da5b0935a414a42df7d5010a56d..3a9632a718d3ab902462b464be91657d353722b0)

Every previous finding is resolved or rejected with reasoning I accept, and the delta introduces nothing new. No inline comments this round.

1. **Shared session left in a failed transaction after `_owner_sub` error** — **verified-fixed** in 3a9632a7: `db.rollback()` in the except (itself guarded), plus two tests (`test_a_failed_lookup_rolls_the_shared_session_back`, `test_a_rollback_that_fails_too_is_still_not_fatal`). I re-verified the commit message's "discards nothing" claim: on the dispatch path the token service only *reads* this session (`_load_encrypted_blob` SELECT; the offline-token store/revoke commits live in other flows), so the rollback cannot lose pending writes.
2. **`reasoning_effort="none"` default-flip blast radius** — **partially addressed / remainder rejected-by-author, accepted**: the judge's thinking-off is now pinned by `test_the_judge_does_not_think_on_its_own_budget` with the rationale in its docstring, which was the substantive ask. The provider-rejects-"none" and `None`-inversion points were rejected with fair grounds (`drop_params: true`, the value shipping in production since #209, the escape hatch documented).
3. **Salvage lost its unit coverage** — **verified-fixed**: `TestSalvagingTheObject` covers fenced/fenced-with-chatter/plain extraction and the no-object contract (empty, prose, malformed, bare array, and the greedy two-objects case).
4. **Duplicate truncation warnings** — **rejected-by-author, accepted**: the two lines do state different facts (budget ran out vs. nothing parseable), and collapsing them would re-couple what the PR decouples.
5. **Unconditional owner lookup + inaccurate docstring** — docstring **verified-fixed** (`evaluate` no longer claims the engine "already reads the owner"); the placement itself **rejected-by-author, accepted** — one primary-key read on an open session, now harmless with the rollback in place.
6. **Duplicated metadata ternary / dropped `scheduled_job_id`** — **rejected-by-author, accepted** (job-id-only metadata is a header the proxy never reads). One factual footnote, not a re-raise: `attribution_scope` lives in `ringier_a2a_sdk.cost_tracking.attribution`, which llm_gateway already imports from — not agent-common. It doesn't change the outcome, just worth knowing if the threading ever grows past two call sites.
7. **Stale `gateway_chat` docstring** — **verified-fixed**.

A clean incremental round — nothing blocking from this reviewer; the human reviewer decides.

aartaria:
Round 2 came back clean — no new findings, and the four round-1 fixes verified against the new head.

One correction to my own round-1 triage, on discard 6: I wrote that `attribution_scope` "lives in agent-common, which console-backend deliberately does not depend on". Wrong on both halves — it lives in `ringier_a2a_sdk.cost_tracking.attribution`, and console-backend does depend on that package (`llm_gateway` imports `LazyClient` from it).

The discard still stands, for a better reason. That mechanism is ContextVars stamped onto the header by an httpx event hook installed on the clients `create_model` builds; `llm_gateway` has its own plain pooled client with no such hook, so adopting it here means wiring the hook into that client rather than passing a dict — well past what this PR is for. The explicit `metadata=` argument is what all six console-backend call sites already use, and this PR's job was to make two of them stop passing nothing.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (3a9632a718d3ab902462b464be91657d353722b0..0865f4fbe4b6d760710e5be8d7d3725c561d354f)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger** (round-1 numbering; round 2 raised no findings of its own):

1. Shared-session poisoning — **verified-fixed** in round 2, unchanged here.
2. `reasoning_effort` default-flip blast radius — **rejected-by-author (accepted in round 2)**, judge pin still in place.
3. Salvage test coverage — **verified-fixed** in round 2, unchanged here.
4. Duplicate truncation warnings — **rejected-by-author (accepted in round 2)**.
5. Unconditional owner lookup — **rejected-by-author (accepted in round 2)**; the corrected docstring gained a further improvement here (`_owner_sub` now states the id/sub split rationale).
6. Duplicated attribution ternary / dropped `scheduled_job_id` / SDK scope suggestion — **verified-fixed**: 0865f4fb reverses the earlier discard and adopts `attribution_scope` around the dispatch plus the canonical `attribution_header` in `_gateway_headers`, deleting the `owner_sub` threading from five signatures and both copies of the ternary. Notably it takes a cleaner path than either option I sketched — the SDK builder merged at the header-build site, not the httpx event hook — so the four per-call sites keep their explicit `metadata` unchanged.
7. Stale `gateway_chat` docstring — **verified-fixed** in round 2; superseded text removed entirely here.

**Delta verification (what I checked, so it's on record):**
- `attribution_header(**overrides)` ignores unknown keys and `None` values and lets overrides win — and all explicit `metadata` keys at the four per-call sites (`user_sub`, `conversation_id`, `catalog_id`) are in the SDK's `_FIELDS`, so nothing is silently dropped by the switch to the canonical builder. Empty attribution stamps no header, matching the old behavior (pinned by `test_nothing_to_attribute_stamps_no_header`).
- `llm_gateway._client` is a plain `httpx.AsyncClient` with no `_stamp_metadata` event hook, so there is no double-stamping path where the hook could overwrite the merged header.
- console-backend sets no ambient attribution vars anywhere else, so the scope cannot collide with request-handler state; `run_job_now`'s inline await is covered by `test_the_scope_does_not_outlive_the_dispatch`, and the scope-vs-`set_attribution` choice is documented for exactly that reason.
- An unresolved owner now yields an ambient `{scheduled_job_id}` with no `user_sub` — the proxy drops that record just as it dropped the old `metadata=None` call, so behavior is unchanged (pinned by `test_an_unresolvable_owner_still_dispatches`).
- The tests deleted with the parameter (`test_the_judgement_is_billed_to_the_job_owner`, the two `_write_notification` billing tests) are superseded by `TestAttributionScope` + `TestAttributionOnTheWire`, which cover the same guarantees at the right level.

A clean incremental round — nothing blocking from this reviewer; the human reviewer decides.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -195,41 +198,59 @@ async def _dispatch_job(self, job: ScheduledJob, run_id: int | None = None) -> N
...
+                # (`custom_logger._build_record` needs a user_sub) and the spend left
+                # `usage_logs` altogether. `scheduled_job_id` is half the point:
+                # `usage_repository` classifies by it, so without it this recurring overhead
+                # reads as the user's own 'orchestrator' spend — and the scope says
+                # `service` outright rather than leaving that to be inferred from an id.
+                #
+                # A scope, not `set_attribution`: `_tick` dispatches each job in its own
+                # task, but `run_job_now` awaits this inline from a request handler, where
+                # a job id left set would follow the rest of that request.
+                owner_sub = await resolve_user_sub(db, job.user_id, context=f"job {job.id}")
```
alloy-cicd: **Minor / consistency — the scheduler is now the only path without the internal-id fallback.**

This round gives titling and catalog sync a deliberate degradation: `user_sub or user_id`, on the stated grounds that "a wrong-shaped-but-resolvable value beats no usage row at all" (the gateway batch endpoint resolves `get_user_by_sub(...) or get_user(...)`, so the internal id still lands the row). The scheduler keeps `user_sub=owner_sub` with no fallback, so when this lookup fails, the proxy drops the record and the judge/notification spend vanishes from `usage_logs` entirely — the exact outcome the fallback was introduced to avoid, on the one path that runs unattended where nobody would notice.

`job.user_id` is the same internal `users.id` as the other two callers', so `user_sub=owner_sub or job.user_id` would behave identically. If the asymmetry is deliberate (e.g. not wanting a non-subject in LiteLLM's own spend table for the recurring path), a one-line comment here saying so would keep the next reader from "fixing" it either way — `test_an_unresolvable_owner_still_dispatches` currently pins the no-fallback behavior without saying whether it's a choice.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (0865f4fbe4b6d760710e5be8d7d3725c561d354f..d1e431bd5997140e46f220b302022324471f85dd)

All prior findings remain settled; the delta is a substantial new feature commit (the `service` dimension, subject resolution for two more callers, catalog-worker scope) and it holds up. One minor consistency question inline — nothing blocking.

**Ledger** (round-1 numbering; rounds 2–3 raised nothing new):
1. Shared-session poisoning — **verified-fixed**; the rollback semantics now live in the shared `resolve_user_sub` (spend_attribution.py) with its tests moved there intact, including rollback-on-failure and rollback-failure-not-fatal.
2. `reasoning_effort` default flip — **rejected-by-author (settled round 2)**, untouched here.
3. Salvage test coverage — **verified-fixed (round 2)**, untouched.
4. Duplicate truncation warnings — **rejected-by-author (settled round 2)**.
5. Unconditional owner lookup — **rejected-by-author (settled round 2)**; the lookup now also feeds the `service` scope, strengthening the case for resolving it per dispatch.
6. Attribution threading / SDK scope — **verified-fixed (round 3)** and extended here: the scope now carries `service`, catalog sync/reindex get the same treatment, and `_owner_sub` is generalized into `spend_attribution.resolve_user_sub` reused by all three paths.
7. Stale docstring — **verified-fixed (round 2)**.

**Delta verification (on record for the human reviewer):**
- **The `service` value actually flows end to end** — this was the one place the feature could silently no-op, and it doesn't: SDK `_FIELDS` gained `service` (so scope and explicit metadata both stamp it), the proxy's `_build_record` carries it, `GatewayUsageLogCreate` inherits the new `UsageLogCreate.service` field (so `model_dump()` keeps it through the gateway batch endpoint), `batch_log_usage` passes dicts through untouched, and `create_usage_log` inserts it. Dead-letter safe keys updated too.
- **`GROUP BY 1` is the correct fix**, not a style choice: with `service` now both a real column and the output alias, Postgres binds `GROUP BY service` to the input column, and the COALESCE/CASE inputs would no longer be grouped — a hard error the author says a database-backed test caught (`test_usage_service_classification.py` exercises the real query).
- **`context_header` popping `service`** is the right call — the callee is a different service doing its own work — and it's test-pinned in the SDK.
- **The fallback classification is backward-compatible by construction**: NULL `service` derives exactly as before (test-pinned for pre-existing rows and agent-path rows), and `test_the_service_names_match_what_the_usage_view_derives` guards the string coupling between `spend_attribution` and the SQL fallback.
- **The titling/catalog identity fix is sound**: both resolved subjects fall back to the internal id, which the gateway batch endpoint genuinely resolves (`get_user_by_sub(...) or get_user(...)` in usage_router), so a failed lookup degrades to the legacy shape rather than losing the row. The catalog worker wraps setup→run→teardown in the scope, so `sync.py`'s explicit per-call metadata merges with (rather than replaces) the ambient `service`.
- The only caller of `maybe_summarize_conversation` (app.py) was updated for the removed `user_sub` parameter; `text` and the other imports survive the `_owner_sub` removal.

One inline comment below on a consistency gap in the new fallback pattern. Submitted as COMMENT — the human reviewer decides.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Not deliberate — you're right, and it's fixed in the next commit.

My reasoning was "don't regress a caller that currently works": titling and catalog pass the internal id today and it resolves, so switching them to a subject must not cost them the row; the scheduler passed nothing, so `None` wasn't a regression. That's true but it's the wrong test. The question isn't whether a path gets worse than it was, it's whether the row survives — and on the one unattended path, losing it is precisely the bug this PR opened with.

Rather than add `owner_sub or job.user_id` at a third call site, the trade now lives once in `spend_attribution.billing_subject`: prefer the subject, fall back to the internal id, and document why in the one place instead of three duplicated comments. `resolve_user_sub` stays as the primitive that can return `None` (it's the one with the rollback contract and its own tests). All three callers use `billing_subject`.

The scheduler test that asserted an unresolvable owner dispatches with no `user_sub` now asserts it dispatches billing the internal id, and there's a parametrized test for both unreadable cases (no row, failed lookup).

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (d1e431bd5997140e46f220b302022324471f85dd..80adf7a8878d2bc9f5a710dbc62ace90558071af)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger:**
- Round-4 finding (scheduler as the only path without the internal-id fallback) — **verified-fixed** in 80adf7a8, and the fix went one better than the suggestion: rather than a third `sub or user_id` at the call site, the trade lives once in `spend_attribution.billing_subject` (subject preferred, internal id as fallback, rationale in its docstring), now used by all three paths — scheduler, titling, catalog sync/reindex. The author's thread reply confirms the asymmetry was unintentional.
- Rounds 1–3 findings — unchanged from prior ledgers: 1, 3, 6, 7 verified-fixed; 2, 4, 5 rejected-by-author with reasoning accepted in round 2.

**Delta verification:** `billing_subject` always returns a value (`resolve_user_sub(...) or user_id`), so every scope now carries a `user_sub` and no path can lose the usage row to the proxy's no-subject discard; `resolve_user_sub` stays the None-returning primitive holding the rollback contract and its tests, untouched. All three call sites switched to the helper, the redundant fallback comments collapsed into its docstring, and the tests were updated coherently — including `test_an_unresolvable_owner_still_dispatches`, which now pins the fallback billing (`user_sub == job.user_id`) instead of pinning its absence, and the parametrized `test_an_unreadable_subject_bills_the_internal_id_rather_than_nobody` covering both the no-subject and failed-lookup cases.

A clean incremental round — nothing blocking from this reviewer; the human reviewer decides.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Round 5 is clean and this is verified fixed — thanks, the finding was right and my "don't regress a caller that already works" test was the wrong one.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 6 (80adf7a8878d2bc9f5a710dbc62ace90558071af..9ca6567fe72f45676e6e1157c6ce98df19223096)

Every previous finding is resolved and the delta introduces nothing new. No inline comments this round.

**Ledger:** unchanged from round 5 — the round-4 finding (scheduler missing the internal-id fallback) stays **verified-fixed** via `billing_subject`; round-1 findings 1, 3, 6, 7 verified-fixed; 2, 4, 5 rejected-by-author with reasoning accepted in round 2. Rounds 2, 3, and 5 raised nothing.

**Delta verification (read-path fix, handler scopes, frontend):**
- **The `/my-logs` drop is genuinely the last gap on the read path**: `list_usage_logs` selects `u.*` (so the new column was already coming back), and the only field-by-field `UsageLog(...)` construction in the codebase is the one in usage_router now carrying `service=`. The pg-backed `test_the_declared_service_survives_the_whole_read_path` covers write → list → response model, which is the right level for a bug of this shape.
- **The two console handlers' `attribution_scope`** wraps every gateway call including `_repair_cel`'s repair attempts and the condition generator's retry loop (test-pinned across attempts); the sync contextmanager restores the ContextVars in a finally, so the `raise HTTPException` paths can't leak the scope — and `test_the_scope_does_not_outlive_the_request` pins that too. Nothing else inside either scope makes an outbound attributed call, so no spend can be accidentally captured by the caller's scope.
- **`console_web_search` keeping per-call metadata is the right exception** — it bills a forwarded caller, not this service's own work — and the scope-vs-metadata rule now documented in `spend_attribution`'s module docstring matches what the code does everywhere.
- **Frontend**: `serviceLabel` falls back to 'Orchestrator' for null/unknown, mirroring the backend derivation for legacy rows; the constants moved to module scope (a real fix for the chart's `useMemo` dependency); 'Console' added to labels/colors and the chart description; the 'Agent'→'Source' and 'Direct API Calls'→'Calls outside a conversation' renames match what the column now holds. The regenerated SDK types carry `service` (plus the unrelated #206 entitlement-version pickup the author flagged in the commit message).

A clean incremental round — nothing blocking from this reviewer; the human reviewer decides.

PR #215, opened 2026-09-07
fix(ptc): stop rate-limited tool calls from burning a sub-agent's recursion budget

### LINKED ISSUE #211: A sub-agent burns its whole recursion budget retrying rate-limited tool calls: 403 quota errors classify as auth, gateway throttling as system_error, and eval-wrapped retries are invisible to loop detection
### Summary

A sub-agent given a read-only research task ("find whether an issue already exists for X")
spent its **entire** recursion budget retrying tool calls that were failing with provider
rate limits, and died with `Recursion limit of N reached without hitting a stop condition`.
The parent turn then returned `task_state='failed'`.

The model was *not* misreading the errors — it read them correctly (its own interpreter
comments say `// Rate limited again. Let's wait and do a single search`) — but it has no way
to actually wait, and nothing in the stack stops it. Three independent defects combine:

1. a provider rate limit is classified as `auth`, not `transient`;
2. the gateway's own rate-limit wording is classified as `system_error`;
3. retries issued from inside the PTC code interpreter are invisible to loop detection.

Net effect: ~40 model turns of the same three searches, ~100 failed tool calls, and a
sub-agent that hit the wall about 20 seconds *after* it had already found the answer.

### Observed sequence

Two consecutive `task` delegations to the same sub-agent, each running the loop until its
budget was gone (`langgraph_step` 3 → 198, then 205 → 399). Representative window, one
`eval` call per line:

```
13:18:48  // Rate limited again. Let's wait and do a single search      -> 403 rate limit
13:18:54  // Single search after rate limit reset                       -> 403 rate limit
13:18:58  (same query, comment removed)                                 -> 403 rate limit
13:19:01  (same query)                                                  -> 403 rate limit
13:19:04  (same query)                                                  -> 403 rate limit
13:19:08  (same query)                                                  -> {"count": 0, "items": []}
```

Five wasted turns to get one answer, and that shape repeats for every query the model tried.
Error tally over the failing window: **62** provider `403 ... rate limit exceeded` and **56**
gateway `Streamable HTTP error: Error POSTing to endpoint: too many requests`.

A secondary aggravator: the interpreter fans host calls out in parallel (`Promise.all` over
several reads in the same millisecond), one succeeds and the rest get `too many requests` —
so the agent is partly rate-limiting itself.

### Root cause 1 — a `403` rate limit is classified as `auth`

`classify_error` tests `_AUTH_PATTERNS` **before** `_TRANSIENT_PATTERNS`
(`packages/orchestrator-agent/app/middleware/error_classification_middleware.py:79-82`), and
`_AUTH_PATTERNS` matches the bare token `403` (line 38). Providers that signal quota
exhaustion with `403 ... rate limit exceeded` therefore classify as `auth`, even though the
text also matches `_TRANSIENT_PATTERNS` (line 33):

```python
>>> classify_error("failed to search: 403 API rate limit exceeded for user ID …")
'auth'      # both patterns match; auth is checked first
```

Consequences: `ToolRetryMiddleware`, which is documented to handle transient errors, never
engages; and the model is handed an auth-flavoured hint for a quota problem, which is not
something re-authentication can fix.

### Root cause 2 — the gateway's rate-limit wording matches no pattern

The MCP gateway surfaces its own throttling as
`Streamable HTTP error: Error POSTing to endpoint: too many requests` — no `429`, no
`rate limit`. `_TRANSIENT_PATTERNS` misses it, every other pattern misses it, and it falls
through to the `_looks_like_error` default:

```python
>>> classify_error("Error: Streamable HTTP error: Error POSTing to endpoint: too many requests")
'system_error'
```

`system_error` is the label the trace actually shows on the escalated failure. So the single
most common error in this incident was reported to the model as an unexpected crash rather
than as backpressure.

### Root cause 3 — `eval`-wrapped retries are invisible to loop detection

`RepeatedToolCallMiddleware` hashes the whole argument dict
(`packages/agent-common/agent_common/middleware/loop_detection_middleware.py:161-172`), and
for the PTC interpreter that dict is the **program text**. Changing a single comment line
yields a different hash, so `max_repeats` (identical args) never fires. And `eval` is listed
in `dispatch_tools` (`packages/agent-common/agent_common/core/graph_utils.py:1676`), which
exempts it from `max_tool_repeats` — deliberately, and for a good reason, but it means the
per-tool-name cap is off too.

The middleware's own recorded state confirms it saw nothing: ~39 consecutive `eval` entries,
**every hash distinct**, while roughly half of them re-issued the same underlying search.
With both thresholds inert, `force_stop_after` can never trigger either. The only remaining
backstop is `_SUB_AGENT_RECURSION_LIMIT`
(`packages/agent-common/agent_common/agents/dynamic_agent.py:118`), whose comment already
names this exact scenario: *"prevents runaway loops when the model ignores tool errors and
keeps retrying"*. It worked — it is the symptom, not the bug.

### Suggested direction

- **Order classification by specificity, not by pattern list order.** A rate-limit match
  should beat a bare-status-code auth match. Narrow `_AUTH_PATTERNS` so `403` alone does not
  win over an explicit `rate limit` / quota phrase in the same message.
- **Teach the transient pattern the gateway's wording** (`too many requests`, `too_many_requests`),
  so gateway backpressure is `transient` rather than `system_error`.
- **Make rate limits non-retryable for the model.** Honour `Retry-After` /
  `x-ratelimit-reset` where the transport exposes it and back off below the model, or return
  a terminal *"unavailable for Ns — do not retry, report back"* result. The model cannot
  sleep; asking it to "retry after a moment" guarantees an immediate retry.
- **Hash the inner call, not the interpreter program.** Loop detection should see the host
  tool calls a program makes (name + args), not the source text wrapping them. Today the PTC
  path is a blind spot for every loop guard we have.
- **Cap host-call concurrency from the interpreter** (a small semaphore) so a `Promise.all`
  cannot manufacture its own 429s.
- **Add a consecutive-failure budget** to the sub-agent loop: after N failed tool calls of
  the same kind, force a structured report-back instead of spending the remaining budget.

### Notes

- Diagnosed from a production trace; identifiers, repository names and operator details are
  intentionally omitted here.
- `classify_error` snippets above were verified by running the two regexes in this repo
  against the observed error strings.

aartaria:
closes #211

### What broke

A research sub-agent burned its entire recursion budget re-issuing the same searches while the provider answered `403 API rate limit exceeded` and the MCP gateway answered `too many requests`. The model read the errors correctly — it just cannot wait inside `eval`, and nothing in the stack stopped it. Three defects lined up (details and evidence in #211).

### What changes

**1. Error classification by specificity** — `packages/orchestrator-agent/app/middleware/error_classification_middleware.py`

`_AUTH_PATTERNS` used to match a bare `403` and was checked before `_TRANSIENT_PATTERNS`, so `403 … rate limit exceeded` classified as `auth` (re-authenticating cannot fix a quota). The gateway's `too many requests` matched nothing and fell through to `system_error`. Order is now: unambiguous auth (`401`, `unauthorized`, `need-credentials`, …) → rate-limit phrase → bare `403`/`forbidden` → other transient. `too many requests` is a transient pattern.

| message | before | after |
|---|---|---|
| `403 API rate limit exceeded for user ID …` | auth | **transient** |
| `Streamable HTTP error: … too many requests` | system_error | **transient** |
| `HTTP Error 401: rate limit on auth endpoint` | auth | auth |
| `403 Forbidden: insufficient scope` | auth | auth |

**2. Identical inner calls are counted where the real call is** — `packages/agent-common/agent_common/middleware/ptc_guard.py`

`RepeatedToolCallMiddleware` hashes the `eval` call, i.e. the *program text*; a changed comment line is a new hash, and `eval` is exempt from `max_tool_repeats` by design. So a program re-issuing `tools.x(sameArgs)` every turn was invisible to every loop guard. The PTC guard now applies **the stack's own `RepeatedToolCallMiddleware` instance** to every inner call — the same way it already reuses `ConditionalHumanInTheLoopMiddleware`'s bypass rules, scorer and threshold for HITL. `evaluate()` (lifted out of `aafter_model`, which now calls it too) is the single rule: same `max_repeats`/`max_tool_repeats`/`window_size`/`dispatch_tools`, same `BLOCKED: …` wording via `blocked_message()`, recorded in the middleware's own checkpointed `tool_call_history` under `eval:<tool>` — so configuring the middleware configures the PTC path. One deliberate exception: `evaluate(..., program_call=True)` skips the same-tool cap, which counts model turns — a program iterating over N distinct items is one turn (the same reasoning `dispatch_tools` applies to `task`/`eval`), and the namespaced key keeps a bulk loop from inflating the same-tool count of later direct calls. The instance is threaded through `build_code_interpreter_middlewares(loop_detection=…)` in both stacks. A blocked inner call comes back to the program as a `repeated_call` payload (a value, not an exception). No loop middleware in the stack means no loop check inside `eval` either — exactly as no risk scorer means no HITL. The guard runs on the sandbox bridge thread and cannot see graph state, so `awrap_tool_call` seeds the `eval` turn from `tool_call_history` and writes it back with the `ToolMessage` as a `Command` update, only when it changed.

**3. A rate-limited inner call is returned, not thrown**

Previously the `ToolException` escaped the guard as a bare throw the program would catch and retry. It now comes back as a `rate_limited` payload (same shape as the approval/rejection payloads) naming the tool, carrying up to 300 chars of the provider's text, and telling the model not to retry but to report back. Within a turn the payload is cached like any other result, so the provider is not hit again for the same call.

**4. Prompt guidance** — `dynamic_tool_dispatch.py`: `transient` now says a rate limit is the one transient error that must never be retried.

### Not in this PR (follow-ups from #211)

- A concurrency cap on host calls issued from `eval` (`Promise.all` fan-out manufactured some of the 429s).
- A consecutive-*failure* budget for the sub-agent loop; this PR only catches *identical* repeats and terminal rate limits.
- Parent and sub-agent are separate graphs with separate state, so their histories never interact; an interrupt resumed on another pod continues from the checkpointed `tool_call_history`.

### Verification

- `packages/agent-common`: **1000 passed** (`uv run pytest`), incl. 9 new unit tests in `test_ptc_guard.py` and a real-`eval` e2e in `test_ptc_guard_e2e.py` asserting the payload comes back as a value (no `<error>`) and the inner call lands in the middleware's checkpointed `tool_call_history` next to the `eval` entry the middleware recorded itself.
- `packages/orchestrator-agent`: **828 passed**, incl. 3 new classifier tests using the two incident strings verbatim (identifiers scrubbed).
- `ruff check` clean on all touched files; `ruff format --check` differences on touched files are pre-existing on `main` and left alone (only my own new lines are formatted).

## Checklist


packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -418,6 +519,27 @@ async def _guarded(runtime: ToolRuntime = None, **kwargs: Any) -> Any:  # type:
         if turn is not None and call_key in turn.results:
             return turn.results[call_key]
 
+        # 1b. The same call, re-issued across ``eval`` runs of this agent run, past
+        #     the identical-arguments threshold: refuse it. This is the loop the
+        #     model-boundary detector cannot see (it only sees ``eval``) — see
+        #     ``PTC_INNER_CALL_MAX_REPEATS``.
+        repeats = record_ptc_inner_call(thread_id, call_key)
```
alloy-cicd: **[correctness — confirmed] The repeat counter counts attempts, not executions, and HITL approval rounds burn it — an approved call can be refused.**

`record_ptc_inner_call` runs before the recorded-decision branch, before the approval-pending return, and before `_execute`, while `_PTC_INNER_CALLS` survives interrupt replays and `turn.results` does not (`begin_ptc_turn`/`end_ptc_turn` wrap every `awrap_tool_call` invocation, including replays — `graph_utils.py:1112/1150`). One approval round for a single high-risk call costs **3 increments**: initial eval run (pending, count 1) → interrupt; resume replays the node with a fresh turn (count 2); decision-honoring run increments to 3 *before* reaching the `approve` branch. A second approval of the same call later in the run reaches count 6 > 5 and returns `repeated_call` for a call the human just approved. Rejected and approval-pending calls likewise burn budget on every replay, and failed executions (timeouts) count the same as successes — 5 transient failures make the 6th, would-succeed attempt refused with a message claiming the args "produce the same result".

Related scope mismatch: the counter is per-run but the dedup cache is per-eval-turn, so legitimate once-per-eval-turn repeats — re-reading the same config file at the top of each program (fresh REPL per `mode="call"`), or polling a status endpoint once per eval turn where the result genuinely changes — accumulate to a refusal on the 6th turn. This half is exercised by the new tests, so it's a deliberate trade-off, but the "same arguments produce the same result" premise is false for time-varying tools.

**Suggested fix:** record the count only for calls that actually reach `_execute` (after the cache/decision/bypass/pending branches), so HITL round-trips and approval payloads stop consuming the budget; consider whether time-varying tools need a carve-out or a cached-result response instead of an error.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1007,6 +1008,18 @@ def after_model(self, state: Any, runtime: Any) -> dict[str, Any] | None:
     async def aafter_model(self, state: Any, runtime: Any) -> dict[str, Any] | None:
         return self._exposure_state_update()
 
+    def after_agent(self, state: Any, runtime: Any) -> dict[str, Any] | None:
+        # The guard's identical-inner-call counts (``ptc_guard._PTC_INNER_CALLS``) live
+        # for one agent run; drop them with the REPL slot so the next run starts clean.
+        update = super().after_agent(state, runtime)
+        clear_ptc_inner_calls(resolve_ptc_thread_id(runtime))
```
alloy-cicd: **[correctness — confirmed] `_PTC_INNER_CALLS` is only cleared here, but `after_agent` is the graph's exit node — it never runs when a run raises, is cancelled (client disconnect), or suspends at a HITL interrupt the user never resumes.**

Unlike `_PTC_TURNS`, which is popped in `awrap_tool_call`'s `finally` (line 1149-1150), abandoned entries persist for the life of the process. The concrete harm isn't the (small) memory leak — it's that stale counts carry into the **next turn on the same `thread_id`**: turn 1 records 4 identical calls and dies at an abandoned interrupt; turn 2's second legitimate identical call hits repeats 6 > 5 and is refused with a message claiming "6 times during this run". That contradicts the counter's documented per-run scope.

**Suggested fix:** also override `before_agent`/`abefore_agent` to call `clear_ptc_inner_calls(resolve_ptc_thread_id(runtime))` after `super()` — the entry node runs unconditionally at every fresh start, and it does *not* re-run on interrupt resume, so counts still correctly survive a mid-run HITL round-trip. (Don't move the clear into `awrap_tool_call`'s `finally` — counts must survive across eval calls within a run.)

packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -391,22 +496,18 @@ async def _execute(
         try:
             return await inner.arun(_inject_for_inner(inner, kwargs, runtime))
         except ToolException as exc:
+            # A quota error is terminal for this run: the model cannot wait inside
+            # ``eval`` and every immediate retry fails the same way. Return it, like
+            # the approval/rejection payloads, so the program sees a value naming
+            # the cause instead of a bare throw it will catch and retry (#211).
+            if is_rate_limit_error(str(exc)):
```
alloy-cicd, line 581: **[correctness — confirmed] Unanchored `429` in `_RATE_LIMIT_PATTERN` turns unrelated tool errors into a *terminal* rate-limited payload.**

`is_rate_limit_error` does an unanchored search, so `429` matches inside any longer number: `ToolException("GET .../issues/4291: 404 Not Found")` returns the `rate_limited` payload telling the model to stop calling the tool for the rest of the run, and the real error survives only as a 300-char `detail`. `rate.?limit` also matches header noise like `x-ratelimit-remaining: 4999` echoed in non-quota errors. On the orchestrator classifier a false positive merely says "transient"; here it swallows the exception and is terminal, so the blast radius is larger.

**Suggested fix:** token-bound the status code (`\b429\b`); since this deliberately mirrors the orchestrator's `_RATE_LIMIT_PATTERNS`, fix both via one shared pattern (see the comment at line 146).

packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -118,6 +121,79 @@ def rejection_payload(tool_name: str, reason: str = "") -> dict[str, str]:
...
+    "run; this call was not executed. The same arguments produce the same result. Do NOT "
+    "retry them — use what you already have, try a substantially different approach, or "
+    "report back."
+)
+
+#: Provider / gateway wording for quota exhaustion. Mirrors the orchestrator's
+#: ``_RATE_LIMIT_PATTERNS`` (error classification): ``403 ... rate limit exceeded``
+#: is a quota problem, not an authorization one, and ``too many requests`` is how the
+#: MCP gateway words its own throttling.
+_RATE_LIMIT_PATTERN = re.compile(r"rate.?limit|429|too.many.requests", re.IGNORECASE)
```
alloy-cicd: **[reuse — confirmed] Third byte-identical copy of `rate.?limit|429|too.many.requests`; the docstring even says it "mirrors" the orchestrator's copy.**

The same alternation now lives here, in the orchestrator's `_RATE_LIMIT_PATTERNS` (error_classification_middleware.py:46), and dead inside its `_TRANSIENT_PATTERNS` (line 34). orchestrator-agent already depends on agent-common (pyproject.toml:34, editable path at :67) and agent-common has no back-imports, so the orchestrator can import one shared predicate — `is_rate_limit_error` is already public here (or promote it to a neutral shared module).

Drift risk: the next provider wording ("quota exceeded", "secondary rate limit") gets added to one copy only, and the PTC path then returns a terminal `rate_limited` payload while the model-boundary path classifies the same text `system_error` (or vice versa) — the exact divergence this PR exists to prevent, invisible to both test suites.

packages/orchestrator-agent/app/middleware/error_classification_middleware.py
```diff
@@ -31,15 +31,30 @@
...
 )
 
+# Quota exhaustion. Checked *between* the two auth tiers below: a provider that
+# spells its quota error ``403 ... rate limit exceeded`` must not classify as
+# ``auth`` on the strength of the bare status code (re-authenticating cannot fix
+# a quota), while a real ``401`` on a rate-limited auth endpoint stays ``auth``.
+# ``too many requests`` is how the MCP gateway words its own throttling — no
+# ``429``, no ``rate limit`` — and without it that error fell through to
+# ``system_error``, i.e. was reported to the model as an unexpected crash.
+_RATE_LIMIT_PATTERNS = re.compile(r"rate.?limit|429|too.many.requests", re.IGNORECASE)
```
alloy-cicd: **[correctness — confirmed, empirically tested] Unanchored `429` plus the new tier order flips genuine 403 permission errors to `transient`.**

`'403 Forbidden: you do not have permission. request-id 8F4297AA'` → **transient** on this branch (`429` matches inside `4297`), **auth** on main — the model now retrys a hard permission error forever. Incidental prose mentions ('See docs on rate limits for allowed usage.') flip the same way. Verified with the exact regexes.

Same class, pre-existing: `_AUTH_PATTERNS`' bare `401` (line 50) still classifies a real quota error carrying `401` inside a request id (`8B4A:1401:2F9C`) as `auth` — the very misclassification this PR targets, left half-fixed since it rides the substring route rather than the 403 route.

**Suggested fix:** word-boundary the status codes — `\b429\b` here, `\b401\b` in `_AUTH_PATTERNS`, `\b403\b` in `_FORBIDDEN_PATTERNS`. Empirically `\b429\b` rejects `4297` and `\b401\b` rejects `1401` while both still match real mentions.

packages/orchestrator-agent/app/middleware/error_classification_middleware.py
```diff
@@ -75,9 +90,14 @@ def classify_error(content: str) -> str | None:
     if classification:
         return classification
 
-    # Pattern-based classification
+    # Pattern-based classification, most specific first. A bare ``403`` only
+    # counts as auth when nothing in the message says "quota".
     if _AUTH_PATTERNS.search(content):
         return "auth"
+    if _RATE_LIMIT_PATTERNS.search(content):
```
alloy-cicd, line 101: **[correctness — confirmed, low severity] The specificity ordering only exists in the text tier: `_classify_json` runs first and still maps `statusCode` 403 unconditionally to `auth`.**

`classify_error` returns `_classify_json`'s answer before this line runs, and `_classify_json` (line 148, untouched by this PR) never looks at wording — so `{"statusCode": 403, "message": "API rate limit exceeded"}` still classifies `auth`, while the identical wording as plain text now classifies `transient`. The actual #211 incident errors were plain-text `ToolException` strings, so this doesn't resurrect the incident — but the classifier is now internally inconsistent for a shape `_classify_json` was built to handle, and no test pins 403-JSON either way.

**Suggested fix:** in `_classify_json`'s 403 branch, check the content against `_RATE_LIMIT_PATTERNS` and return `"transient"` on a match (401 stays unconditional), plus one JSON-shaped regression test.

packages/orchestrator-agent/app/middleware/error_classification_middleware.py
```diff
@@ -31,15 +31,30 @@
 
 # Conservative patterns — start narrow, expand based on actual bug reports
 _TRANSIENT_PATTERNS = re.compile(
-    r"timeout|timed?\s*out|rate.limit|429|503|502|504|5\d{2}\s+(?:internal|bad gateway|service unavailable)",
+    r"timeout|timed?\s*out|rate.?limit|429|too.many.requests"
```
alloy-cicd: **[simplification — confirmed] The rate-limit alternatives in `_TRANSIENT_PATTERNS` are now unreachable.**

`classify_error` checks `_RATE_LIMIT_PATTERNS` (line 97) before `_TRANSIENT_PATTERNS` (line 101), both return `"transient"`, and the alternations are byte-identical; `_TRANSIENT_PATTERNS` has no other consumer in the repo (grep: definition + one check). So the `rate.?limit|429|too.many.requests` slice here is dead code that invites a maintainer to widen quota wording in the regex whose name says "transient" and observe no behaviour change.

**Suggested fix:** drop the rate-limit alternatives from `_TRANSIENT_PATTERNS` and let `_RATE_LIMIT_PATTERNS` be the single owner of quota wording.

packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -118,6 +121,79 @@ def rejection_payload(tool_name: str, reason: str = "") -> dict[str, str]:
...
+
+#: Identical inner calls tolerated per run before the guard refuses to execute them.
+#: Mirrors ``RepeatedToolCallMiddleware(max_repeats=...)`` in ``graph_utils``. That
+#: middleware hashes the ``eval`` call, whose arguments are the *program text* — a
+#: changed comment line is a new hash — so a program re-issuing the same
+#: ``tools.x(sameArgs)`` on every turn is invisible to it. The real call is only
+#: known here, so the identical-arguments check has to live here too (#211).
+PTC_INNER_CALL_MAX_REPEATS = int(os.getenv("PTC_INNER_CALL_MAX_REPEATS", "5"))
```
alloy-cicd: **[nit] `PTC_INNER_CALL_MAX_REPEATS=0` refuses *every* inner call rather than disabling the guard.**

The check is `repeats > PTC_INNER_CALL_MAX_REPEATS` after incrementing, so the first call yields `1 > 0` and no PTC tool ever executes inside eval. An operator reaching for `0` to mean "off" gets total breakage; disabling actually means a large value, and nothing documents that. Suggest clamping (`max(1, ...)`) or a docstring note. Also worth a line in packages/agent-common/AGENTS.md's PTC section, which documents the sibling knob (`PTC_INLINE_RENDER_THRESHOLD`, default 40) but not this one or the new terminal `rate_limited`/`repeated_call` payload contract.

alloy-cicd (commented):

Reviewed the PTC inner-call guard and the rate-limit classification changes. The #211 incident itself (plain-text `403 API rate limit exceeded` / `too many requests`) is correctly fixed and regression-tested. I found **2 high-impact issues** in the new repeat counter's lifecycle (HITL approval rounds burn the budget and can refuse a just-approved call; state leaks/goes stale when a run doesn't end via `after_agent`), **2 correctness issues** in the new rate-limit matching (unanchored `429` substring misclassifies unrelated errors on both the classifier and PTC paths), **1 gap** (the JSON tier still maps 403-with-quota-wording to `auth`), and **2 maintainability points** (the same regex now exists in three places, one of them dead; a `0` value for the new env knob refuses every call). Details inline.

Checked and *not* flagged: sub-agent/thread-id counter collisions (sub-agents get distinct `{conversation_id}::{subagent_type}` thread ids and concurrent same-agent dispatch is refused); the `int(os.getenv(...))` import-time parsing (matches house style for every sibling knob); returning the rate-limit payload as a value instead of raising (the eval-propagation rationale in the module docstring holds); the module-dict-vs-ContextVar choice (the `run_coroutine_threadsafe` reasoning is real).

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Fixed in dd2f8af9 — the count moved into `_execute`, so only calls that reach the inner tool are counted; pending approvals, interrupt replays and recorded decisions leave it alone (new test `test_hitl_round_trips_do_not_consume_the_repeat_budget` walks the exact 3-increment scenario). The time-varying-tools carve-out I'm leaving as a deliberate trade-off: the threshold mirrors `RepeatedToolCallMiddleware(max_repeats=5)`, which applies the same rule to direct calls, and a per-tool exemption list is a bigger design than this bug warrants.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Fixed in dd2f8af9 — `before_agent`/`abefore_agent` overrides clear the counts on entry as well (it does not re-run on an interrupt resume, so a mid-run approval round-trip still keeps its counts). Test: `test_before_agent_clears_counts_left_by_an_abandoned_run`.

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Fixed in dd2f8af9 — `\b429\b`, and the pattern is now the single `RATE_LIMIT_PATTERN` in agent-common. Added `issues/4291` and `8F4297AA` as negative cases in `test_is_rate_limit_error_ignores_other_failures`.

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Fixed in dd2f8af9 — one public `RATE_LIMIT_PATTERN` / `is_rate_limit_error` in `agent_common.middleware.ptc_guard`; the orchestrator classifier imports it and `_TRANSIENT_PATTERNS` lost its copy.

packages/orchestrator-agent/app/middleware/error_classification_middleware.py
aartaria: Fixed in dd2f8af9 — `\b429\b`, `\b401\b`, `\b403\b`. `test_status_codes_are_word_bounded` covers your two examples (`8F4297AA` stays auth, `8B4A:1401:2F9C` with quota wording is transient).

packages/orchestrator-agent/app/middleware/error_classification_middleware.py
aartaria: Fixed in dd2f8af9 — the 403 branch of `_classify_json` now returns `transient` when the content matches the rate-limit pattern; 401 stays unconditional. Test: `test_json_403_with_quota_wording_is_transient`.

packages/orchestrator-agent/app/middleware/error_classification_middleware.py
aartaria: Fixed in dd2f8af9 — the quota alternatives are gone from `_TRANSIENT_PATTERNS`; `_RATE_LIMIT_PATTERNS` (the shared pattern) is the only owner.

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Fixed in dd2f8af9 — clamped to `max(1, …)` with a docstring note that there is no off value, and documented in AGENTS.md's PTC section as item 5 together with the `rate_limited`/`repeated_call` payload contract.

packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -118,6 +121,84 @@ def rejection_payload(tool_name: str, reason: str = "") -> dict[str, str]:
...
+    "report back."
+)
+
+#: Provider / gateway wording for quota exhaustion: ``403 ... rate limit exceeded`` is
+#: a quota problem, not an authorization one, and ``too many requests`` is how the
+#: MCP gateway words its own throttling. The status code is word-bounded so a request
+#: id or an issue number containing ``429`` is not a rate limit. This is the single
+#: definition — the orchestrator's error classification imports it — so the PTC path
+#: and the model-boundary path can never disagree on what a rate limit looks like.
+RATE_LIMIT_PATTERN = re.compile(r"rate.?limit|\b429\b|too.many.requests", re.IGNORECASE)
```
alloy-cicd: **[minor — residual from round 1, finding 3] `rate.?limit` still matches `x-ratelimit-*` header echoes.**

The `\b429\b` fix covers the digit half, but the wording half of the round-1 comment remains: empirically on this head, `is_rate_limit_error('403 Forbidden: insufficient scope (x-ratelimit-remaining: 4999)')` and `('404 Not Found; headers: X-RateLimit-Reset=…')` both return `True` — `rate.?limit` matches inside `x-ratelimit`/`X-RateLimit`, so a non-quota error echoing standard rate-limit headers still becomes a terminal `rate_limited` payload on the PTC path (and now, via the shared pattern, classifies `transient` in the orchestrator too).

**Suggested fix:** a negative lookbehind, e.g. `(?<![\w-])rate.?limit` — it rejects `x-ratelimit-remaining`/`X-RateLimit-Reset` while still matching `rate limit`, `rate-limit`, and `ratelimit` at a word start (verified against the incident strings and the existing positive test cases). Fine as a follow-up if you'd rather not touch the pattern again this round.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (923f7d75d5f9010622cb9ffdfdc428a4c7abcf32..dd2f8af9c26137124724531f84ce62bbebc37120)

**Disposition of round-1 findings:**

| # | Finding | Disposition |
|---|---|---|
| 1 | HITL approval rounds burn the repeat counter | **verified-fixed** — count moved into `_execute`; `test_hitl_round_trips_do_not_consume_the_repeat_budget` walks the exact 3-increment scenario (pending 0, replay 0, approved execution 1, rejection still 1). The time-varying-tools carve-out was **rejected-by-author** with reasoning (mirrors `RepeatedToolCallMiddleware`'s rule for direct calls) — accepted, not re-raised. |
| 2 | `_PTC_INNER_CALLS` leak / stale counts when a run never reaches `after_agent` | **verified-fixed** — `before_agent`/`abefore_agent` overrides clear on entry; `test_before_agent_clears_counts_left_by_an_abandoned_run` covers it. |
| 3 | Unanchored `429` terminal on the PTC path | **mostly fixed** — `\b429\b` with my two examples as negative tests. One residual, see the single inline comment. |
| 4 | Triplicated rate-limit regex | **verified-fixed** — one public `RATE_LIMIT_PATTERN`/`is_rate_limit_error` in agent-common, imported by the orchestrator classifier. |
| 5 | Unanchored classifier digits flip 403→transient / 401 residual | **verified-fixed** — `\b429\b`, `\b401\b`, `\b403\b`; `test_status_codes_are_word_bounded` pins both round-1 example strings. |
| 6 | JSON tier still mapped 403-with-quota-wording to `auth` | **verified-fixed** — `_classify_json`'s 403 branch applies `is_rate_limit_error`; 401 stays unconditional; JSON regression test added. |
| 7 | Dead rate-limit alternatives in `_TRANSIENT_PATTERNS` | **verified-fixed** — dropped, single owner. |
| 8 | `PTC_INNER_CALL_MAX_REPEATS=0` footgun / AGENTS.md gap | **verified-fixed** — clamped to `max(1, …)`, documented in AGENTS.md item 5 with the payload contract. |

**Delta review:** I read the full 923f7d75..dd2f8af9 diff, confirmed all four `_execute` call sites pass the new keyword args, and ran both touched suites at head (`test_ptc_guard.py` + `test_ptc_guard_e2e.py`: 69 passed; `test_error_classification_middleware.py`: 26 passed). No new correctness or security issues introduced by the fixes. The one inline comment below is the residual half of round-1 finding 3, minor.

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Fixed in 7e6cc9dd — `(?<![\w-])rate.?limit`; both of your header-echo strings are negative cases in `test_is_rate_limit_error_ignores_other_failures`, and `rate-limit`/`ratelimit` at a word start stay positive.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1080,9 +1132,26 @@ async def awrap_tool_call(self, request: Any, handler: Any) -> Any:
             return await handler(request)
         runtime = getattr(request, "runtime", None)
         thread_id = resolve_ptc_thread_id(runtime)
-        if not self._ptc_enabled or self._ptc_risk_scorer is None:
-            return await self._run_eval_with_guidance(request, handler, thread_id)
+        # One turn per ``eval`` execution, guarded or not: it is how the guard (on
+        # the bridge thread) reaches the run's inner-call counts, seeded here from
+        # the checkpointed state and written back with the result.
+        turn = begin_ptc_turn(thread_id)
```
alloy-cicd: **[minor — pre-existing hazard, surface widened by this refactor] Two concurrent `eval` calls on one thread stomp each other's turn — and the inner-call counts now ride on the turn.**

`begin_ptc_turn` *replaces* `_PTC_TURNS[thread_id]` and `end_ptc_turn` pops it unconditionally, so if the model ever emits two `eval` tool calls in one assistant message (nothing in the graph forbids parallel tool calls, and the tool node runs them concurrently), eval B's `begin` swaps the turn out from under eval A: A's inner calls bump B's counts and share B's results/decisions, and whichever `finally` runs first pops the turn, leaving the other's remaining inner calls with `turn is None` — where the repeat guard is now deliberately inert — and its count write-back computed against a stale seed (last writer wins, since the state channel replaces the dict).

This collision predates the PR for the guarded HITL path (pending/decisions had the same exposure), but before b20562ae the unguarded path never touched `_PTC_TURNS` and the counts lived elsewhere; now both are on the shared slot. If concurrent evals are impossible by construction somewhere I didn't find, a one-line comment saying so at `begin_ptc_turn` would close this; otherwise keying `_PTC_TURNS` by `(thread_id, tool_call_id)` makes it airtight. Fine as a follow-up.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (dd2f8af9c26137124724531f84ce62bbebc37120..b20562ae4df9fa500acb443196c40fd2836e459b)

**Disposition of previous findings:**

| Finding | Disposition |
|---|---|
| Round-2 residual: `rate.?limit` matched `x-ratelimit-*` header echoes | **verified-fixed** in 7e6cc9dd — `(?<![\w-])rate.?limit`, both header-echo strings as negative tests, `rate-limit`/`ratelimit` at a word start kept positive. |
| Round-1 #1 (HITL rounds burned the counter) | **verified-fixed**, and the property survives the b20562ae refactor — `test_hitl_round_trips_do_not_consume_the_repeat_budget` was rewritten for the state-seeded turns and still walks the 3-phase accounting. The time-varying-tools carve-out stays **rejected-by-author** (accepted, not re-raised). |
| Round-1 #2 (leak / stale counts when a run never reached `after_agent`) | **verified-fixed**, now structurally: counts live in checkpointed graph state (`ptc_inner_call_counts`), `before_agent` resets to `{}`, and the module dict plus its hand-rolled lifecycle are gone entirely. An interrupt resume on another pod now *continues* the count instead of restarting it — strictly better than what I asked for. |
| Round-1 #3–#8 | **verified-fixed** (unchanged from round 2; the refactor keeps `\b429\b`, the shared `RATE_LIMIT_PATTERN`, the JSON-tier 403 rule, the clamp, and the AGENTS.md item, which was updated for the new state mechanics). |

**Delta review (7e6cc9dd + b20562ae):** I traced the new turn lifecycle end-to-end — `awrap_tool_call` now begins a turn on every path, seeds `turn.inner_call_counts` from state, and writes changed counts back as a `Command` beside the `ToolMessage`; counts bumped in an eval run that ends at an interrupt are deliberately not persisted, which on resume makes the replay's re-executions idempotent w.r.t. counting (fail-open, correct). No leftover references to the removed helpers. Both touched suites pass at head (`test_ptc_guard.py` + `test_ptc_guard_e2e.py`: 73; `test_error_classification_middleware.py`: 26, exercising the shared pattern's new lookbehind through the classifier).

No new correctness or security regressions found. The single inline comment below is a minor pre-existing concurrency hazard whose surface the refactor widens — flagged for a follow-up decision, not blocking.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Agreed it is real and pre-existing — and after tracing it, not closable from the guard alone: `langchain_quickjs` keys the REPL slot itself per `thread_id` and resets it after every call in `mode="call"`, so two concurrent evals already collide one layer below `_PTC_TURNS`. Re-keying the collector by `(thread_id, tool_call_id)` would fix our bookkeeping while the sandbox underneath still gets reset mid-run. The fix has to be a single-eval-per-step guarantee (or a per-call slot upstream); filed as #217 with the three layers and options laid out. In this PR (7826c720) I documented the inherited assumption at `begin_ptc_turn` as you suggested.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 4 (b20562ae4df9fa500acb443196c40fd2836e459b..f96911e638c7f1abe270cdece594a73fadb96fa0)

**Disposition of previous findings:**

| Finding | Disposition |
|---|---|
| Round-3 inline: concurrent `eval` calls stomp the per-thread turn | **rejected-by-author with reasoning, accepted** — the author traced it deeper (langchain_quickjs's REPL slot collides per-thread one layer below, so re-keying the collector alone wouldn't help), documented the inherited constraint at `begin_ptc_turn` (7826c720), and filed the single-eval-per-step guarantee as a follow-up. Verified the comment is in place; not re-raised. |
| Rounds 1–2 (all 9) | **remain verified-fixed** through the f96911e6 refactor — execution-point judgment (HITL rounds still don't consume budget), the shared `RATE_LIMIT_PATTERN` with lookbehind, the classifier orderings and JSON-tier rule are all untouched or preserved by the rewritten tests. The `PTC_INNER_CALL_MAX_REPEATS` knob (round-1 #8's clamp) is now **gone entirely**, superseded by the middleware's own configuration — a strictly cleaner resolution. |

**Delta review (7826c720 + f96911e6):** the unification is well built — `evaluate()` is genuinely the code `aafter_model` runs, the state channel is declared identically on both schemas so they merge, the seed/write-back deep-copies before the reducer applies, and blocked-call recording keeps the escalation semantics. All 114 tests in the three touched suites pass at head. The PR checklist still advertises the removed `PTC_INNER_CALL_MAX_REPEATS` env knob — worth updating the breaking-change line to name the middleware configuration instead.

One new finding, inline below: unifying the policy quietly puts inner calls under the **`max_tool_repeats` (same-tool, different-args) cap**, which the production configs set to 10 with no inner-tool exemption — and iterating one tool over a list of distinct items is the canonical PTC pattern the `dispatch_tools` comment itself defends for `eval`.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1564,6 +1629,23 @@ def build_common_middleware_stack(
     # own tags, so in-process sub-agent LLM calls are billed to the sub-agent
     # (not the orchestrator) regardless of which dispatch path invoked them.
     middleware: list = [GatewayAttributionMiddleware(), ContinueOnTruncationMiddleware()]
+    # One loop-detection policy for the stack. It sits near the end of the stack
+    # (below) and is also handed to the code interpreter so the PTC guard applies
+    # the same rule, with the same history, to the calls a program makes inside
+    # ``eval`` — which the model boundary only ever sees as one ``eval`` call.
+    loop_detection = RepeatedToolCallMiddleware(
+        max_repeats=5,
+        max_tool_repeats=10,
```
alloy-cicd, line 1742: **[correctness — new in f96911e6] The same-tool cap now binds inner calls: a legitimate PTC bulk loop over 11+ distinct items is blocked at item 11, and the tool stays blocked for the rest of the thread.**

`evaluate()` applies both rules, and inner tools are not in `dispatch_tools`, so inside `eval` a plain `for (const id of ids) await tools.getCampaign({id})` over 11+ **different** ids hits `same_tool_count > 10` and gets `repeated_call` — this comment's own rationale for exempting `task`/`eval` ("legitimately called many times with *different* arguments") applies verbatim to a program-driven inner loop, which is the pattern PTC exists for. Under the previous head (b20562ae) only identical-args repeats were judged, so this is a behavior regression for bulk programs, not a restoration of missing coverage.

Two knock-ons make it worse than a one-item refusal:
1. Blocked entries are appended without the window trim (deliberately, for escalation), so once a tool's history passes the cap, **every** later call to it — different args, direct or via `eval`, this thread — is also blocked; the tool is effectively dead for the conversation.
2. The first *direct* call to that tool afterwards is judged at the model boundary with `repeat_count` ≈ the whole history (e.g. 16), and `repeat_count - max_repeats >= force_stop_after` (16−5≥3) force-stops the run immediately — a single legitimate bulk program primes a hard stop.

The identical-args rule (`max_repeats=5`) inside `eval` is exactly right and is what #211 needed. **Suggested fix:** judge inner calls against the identical-args rule only (e.g. an `evaluate(..., include_same_tool=False)` flag or a guard-side skip of the `same_tool` verdict), or give the PTC judgment a separate, much higher per-tool cap — keeping one policy object but acknowledging that "N distinct calls" means loop at the model boundary and means ordinary iteration inside a program.

aartaria: Confirmed and fixed in 6d8bfba0 — this was a regression from the unification. `evaluate(..., program_call=True)` now skips only the same-tool cap (the rule counts model turns; a program iterating over N distinct items is one turn — the same reasoning the config applies to `task`/`eval` via `dispatch_tools`), while `max_repeats`, window, hashing, filters and wording stay shared. Program calls are recorded under `eval:<tool>` (`ptc_history_key`) so a bulk loop cannot inflate the same-tool count of later direct calls — your knock-ons 1 and 2 are covered by that namespacing, and `test_loop_policy_configuration_governs_the_ptc_path` now runs a 12-item loop then checks a direct call is not blocked. Trade-off stated in the docstring: identical-args sharing *across* the two paths is given up; #211 was identical args across eval turns, still fully covered. PR body checklist updated to drop the removed knob.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 5 (f96911e638c7f1abe270cdece594a73fadb96fa0..6d8bfba0e8085c12429722b7d1631fb78d27c505)

**Disposition of previous findings:**

| Finding | Disposition |
|---|---|
| Round-4 inline: the same-tool cap bound inner calls — bulk iteration blocked at item 11, tool then dead for the thread, force-stop primed for the next direct call | **verified-fixed** in 6d8bfba0. `evaluate(..., program_call=True)` skips exactly the same-tool cap and nothing else (identical-args rule, window, hashing, filters, wording stay shared), and program calls are recorded under their own `eval:<tool>` history key, which closes both knock-ons: a bulk loop can no longer block later direct calls nor prime the force-stop. Verified against the new tests — the boundary still trips the cap for the same 8 distinct calls a program is allowed, the identical-args rule still binds inside `eval` at the configured threshold, and the namespacing keeps the direct-call count untouched. The cost — identical-args history no longer shared between the direct and program paths — is named and reasoned in the commit (#211 was identical args *across eval turns*, which stays covered); that disposition is sound. |
| Round-3 inline (concurrent-eval turn stomping) | unchanged: **rejected-by-author with reasoning** (inherited from the per-thread REPL slot one layer down), documented at `begin_ptc_turn`, follow-up filed. |
| Rounds 1–2 (all findings) | **remain verified-fixed**; nothing in this delta touches them. |

**Delta review (6d8bfba0):** every previous finding is resolved and the delta introduces nothing new. The exemption is scoped tightly (`_check_for_loop` skips only the `max_tool_repeats` clause for `program_call`), `applies_to` still filters on the bare tool name while history is namespaced, blocked-entry escalation semantics are preserved within the `eval:` namespace, AGENTS.md item 5 documents the exception and the trade-off, and the PR checklist no longer advertises the removed env knob. All 115 tests in the three touched suites pass at head. No inline comments this round.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 6 (6d8bfba0e8085c12429722b7d1631fb78d27c505..0719f823e345856100bd75cc922fb48339c5c10f)

**Disposition of previous findings:** every finding from rounds 1–5 remains resolved; nothing in this delta touches them. (Round 3's concurrent-eval point stays rejected-by-author with reasoning and a filed follow-up; round 4's same-tool-cap regression stays verified-fixed via `program_call=True` + the `eval:<tool>` namespace.)

**Delta review (0719f823):** every previous finding is resolved and the delta introduces nothing new. This commit hardens the history write-back that round 3's summary noted as a logged-and-dropped edge: `_with_tool_call_history` now merges every mergeable `Command.update` shape (dict, `(channel, value)` pair sequence, `None`) while preserving `goto`/`graph`/`resume`, and raises a `TypeError` — fail-closed, never silently dropping a loop-guard record — for the shapes the eval handler never produces (annotated state object, bare root value). The unchanged-history pass-through still short-circuits regardless of shape. Verified the merge logic against all five shapes, and the new `test_history_write_back_merges_every_command_shape_and_never_drops` covers each one including the raise and the pass-through. All 116 tests in the three touched suites pass at head. No inline comments this round.

From the automated reviewer's side the PR has been stable for two consecutive clean rounds; the approve/merge decision rests with the human reviewer.

PR #220, opened 2026-09-09
feat(embed): sub-agent owned by embedded apps

thecopy:
**Feature: Agent Type Embedded**
This adds a new agent type: Embedded. This agent's configuration (prompt, skills etc) is comfing from a HTTPS URL. This URL contains the system prompt and skills; e.g. https://www.myapp.com/.well-known/agent-skills/index.json

It follows https://github.com/cloudflare/agent-skills-discovery-rfc

Embedded agent types also contain a list of `azp`. When an access token is used to connect to the webscoekt chat API in Console Backend, the `azp` claim is checked, and the subagent with a matching `azp` will be selected.

**Feature: Embed SDK Dev: Show Auth details**
Adds access token details in dev pane

**Bug fix**
Sub-agent names are validated where they are typed (`SubAgentCreate`, `SubAgentUpdate`, `AutomatedSubAgentConfig`): letters, digits, `-`, `_`, must  start with a letter, max 64. The rule already existed in agent-common, but only bit later, inside the orchestrator.

## Checklist


packages/console-backend/tests/test_embed_binding_router.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+
+@pytest.mark.asyncio
+async def test_whole_version_operations_are_refused_on_bound_sub_agents():
+    detail = await rejected(request_with(binding()), None)
+    assert (
+        detail is not None
+        and "the version" in detail
+        and "https://riad.example" in detail
```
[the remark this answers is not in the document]
thecopy: test file

packages/console-backend/app.py
```diff
@@ -1580,7 +1596,32 @@ async def _resolve_socket_user_via_cookie(environ: dict[str, Any]) -> str | None
...
+
+def _apply_embed_scope(metadata: dict[str, Any], socket_session: SocketSession) -> int | None:
+    """Replace any client-sent execute-only target with the connection's bound sub-agent.
+
+    The orchestrator runs ``executeOnlySubAgentId`` from the message metadata as the
+    top-level graph (ADR-0004). Only console-backend may set it, from the azp binding
+    made at connect (ADR-0006) — a page cannot pick a sub-agent it was not bound to.
+    Returns the bound id, or None for console sessions and unbound tokens.
+    """
+    metadata.pop("executeOnlySubAgentId", None)
```
alloy-cicd, line 1616: **[high | fail-open scope] Unbound azp silently runs the full orchestrator inside the host page.**

`_apply_embed_scope` unconditionally pops the client-sent `executeOnlySubAgentId`/`subAgentId`, and stamps a replacement only when the token's azp has a binding. When the azp is unbound — binding not yet created, deleted, an azp typo in the admin form, or the swallowed exception in the connect path (~line 1684: "a binding hiccup must not reject the login", which also covers a transient DB error or a 60s stale negative cache entry right after the admin created the binding) — the turn is forwarded with **no** execute-only target: the generic routing orchestrator runs inside the third-party page with all of the user's activated agents and tools, with no error or UI signal anywhere. Pre-PR, the client-declared id was validated by the orchestrator and a wrong id **failed closed**.

This is also a deploy-order break: old embed-sdk clients (the RIAD cockpit still sets `subAgentId` in `rcplus-alloy-cockpit-frontend/app/src/nannos/config.ts`) have their claim discarded the moment this backend deploys, silently flipping every existing embedded deployment to full-orchestrator mode until an admin creates a binding per host azp.

Suggestion: for bearer-token (embedded) connections, treat "no binding resolved" as fail-closed — reject the turn (or the connect) with an explicit error the host page can surface — and keep fail-open only for genuine console sessions.

packages/console-backend/app.py
```diff
@@ -1628,6 +1671,20 @@ async def _resolve_socket_user_via_token(token: str) -> str | None:
...
+        # orchestrator's /activated list contains the agent on their very first turn.
+        embedded_sub_agent_id: int | None = None
+        embed_service = getattr(sio.app_instance.state, "embed_binding_service", None)  # type: ignore[attr-defined]
+        azp = claims.get("azp")
+        if embed_service is not None and isinstance(azp, str) and azp:
+            try:
+                embedded_sub_agent_id = await embed_service.bind_connection(db, user=user, azp=azp)
+                if embedded_sub_agent_id is not None:
+                    await db.commit()
+            except Exception:  # noqa: BLE001 — a binding hiccup must not reject the login
```
alloy-cicd, line 1693: **[medium | fail-open on error] `bind_connection` failure degrades a BOUND host to unbound instead of erroring.**

Any exception here (transient DB error, or a stale 60s negative entry in `EmbedBindingService._azp_cache` on this pod right after the binding was created) lets the connection come up with `embedded_sub_agent_id=None`. For an embedded surface, unbound is a privilege *widening*, not a degraded mode — see the comment at `_apply_embed_scope`. Related: the guards behind `getattr(app.state, "embed_binding_service", None)` (also `sub_agent_router.py:77`, `:437`, `conversation_router.py:101`) mean an init-order regression silently disables all embed enforcement; prefer a hard dependency that raises when the service is absent.

packages/console-backend/console_backend/routers/conversation_router.py
```diff
@@ -94,6 +96,13 @@ async def get_conversations_by_user(
             # Exclude playground conversations (those with sub_agent_config_hash set)
             conversations = [c for c in conversations if c.sub_agent_config_hash is None]
 
+        # Embedded hosts (ADR-0006): the scope comes from the bearer token's azp when that
+        # azp is bound to a sub-agent, never from the query string alone.
+        embed_service = getattr(request.app.state, "embed_binding_service", None)
+        if embed_service is not None:
+            bound = await embed_service.sub_agent_id_for_azp(await get_client_id_from_request(request))
```
alloy-cicd, line 103: **[high | data exposure] Unbound azp returns the user's ENTIRE conversation list to the third-party page.**

The new SDK no longer sends `embedded_sub_agent_id` (removed from `conversations-store.ts` `loadList`), so scoping relies entirely on this azp lookup. When `sub_agent_id_for_azp` returns `None` (binding missing, deleted, or not yet visible through the 60s cache), no filter is applied and the host page's JavaScript receives every conversation of the user — console titles and other applications' conversations — exactly the disclosure the docstring says must not happen. The old client always sent the param, so the unbound state is a regression.

Also worth a look while here:
- `DELETE /{conversation_id}` and `PATCH /{conversation_id}` in this router take no embed scope at all, so a host bearer token can delete/rename the user's console conversations today. Consider a shared FastAPI dependency that resolves the bound sub-agent once and applies to all conversation/message routes.
- Minor: `get_client_id_from_request` re-runs a full JWT signature validation on a token `require_auth_or_bearer_token` validated milliseconds earlier (verified: no per-token cache, claims not stashed). Stashing the validated claims/azp on `request.state` would drop the second validation on this polled endpoint.

packages/console-backend/console_backend/services/embed_binding_service.py
```diff
@@ -0,0 +1,561 @@
...
+    async def bind_connection(
+        self, db: AsyncSession, user: User, azp: str | None
+    ) -> int | None:
+        """Called from the socket token path. Activates the user for the bound sub-agent and
+        records the sighting. Returns the sub-agent id to stamp on the socket session, or None
+        when the token's `azp` is not bound."""
+        sub_agent_id = await self.sub_agent_id_for_azp(azp, db)
+        if sub_agent_id is None or azp is None:
+            return None
+        await self._sub_agents.repo.bulk_activate_sub_agent(
```
alloy-cicd, line 487: **[high | verified] EMBED activation alone does not make the bound agent reachable — embedded turns fail for every non-owner user.**

The orchestrator builds `local_subagents` from `GET /api/v1/sub-agents/activated` → `get_accessible_sub_agents(..., activated_only=True)`, whose non-admin WHERE clause (`sub_agent_service.py:344-348`) requires owned OR `is_public` OR group-assigned, with the activation filter only ANDed on top. `create_bound_sub_agent` creates the agent private and owned by the calling admin, and `bulk_activate_sub_agent` writes only a `user_sub_agent_activations` row — which never satisfies the owned/public/group disjunction. So for a regular cockpit user the agent is absent from `local_subagents` and every embedded turn fails with "This assistant isn't available for your account" (`executor.py:930-952`); only the binding-creating admin can actually chat. The comment above ("the orchestrator's /activated list contains the agent on their very first turn") is not delivered by any code in this PR.

Fix options: make embed-bound agents visible via the activation row (extend the WHERE clause for EMBED activations), or create them `is_public=True`, or group-assign — but pick one deliberately, since it defines the exposure of bound agents in the console UI too.

packages/console-backend/console_backend/services/sub_agent_service.py
```diff
@@ -2987,6 +2987,105 @@ async def resolve_imported_skills_bulk(self, db: AsyncSession, sub_agents: list[
...
+            mcp_tools = list(baseline.mcp_tools) if baseline and baseline.mcp_tools else []
+        model: str | None = None
+        if model_tier is None and baseline is not None:
+            model = baseline.model
+            model_tier = baseline.model_tier.value if baseline.model_tier else None
+        if enable_thinking is None and thinking_level is None and baseline is not None:
+            enable_thinking = baseline.enable_thinking
+            thinking_level = baseline.thinking_level
+
+        new_version = (existing.current_version or 0) + 1
```
alloy-cicd: **[high | verified] `publish_managed_version` can collide with a soft-deleted version number and wedge the sync loop permanently.**

`new_version = (existing.current_version or 0) + 1` — but the normal update path allocates `SELECT COALESCE(MAX(version), 0) + 1` (line ~974) precisely because soft-deleted versions keep their number under `UNIQUE(sub_agent_id, version)` (ddl/005), and `update_current_version_to_previous` lowers `current_version` after a draft delete. Reproduction: agent has approved v1, user creates draft v2 and deletes it (`current_version` back to 1, v2 row soft-deleted); admin binds the agent → first sync inserts version 2 → `IntegrityError`. Because `_apply_definition` only stores the new revision after a successful publish and `sync_all` rolls back and logs on any exception, **every subsequent sync pass retries the same colliding insert forever**. Use the same MAX(version)+1 allocation here.

packages/console-backend/console_backend/routers/sub_agent_router.py
```diff
@@ -43,6 +51,62 @@ def get_sub_agent_service(request: Request) -> SubAgentService:
...
+def get_embed_binding_service(request: Request) -> EmbedBindingService:
+    """Get the embed binding service (ADR-0006) from app state."""
+    return request.app.state.embed_binding_service
+
+
+#: Fields the host ALWAYS publishes for a bound sub-agent (ADR-0006).
+_HOST_MANAGED_FIELDS = ("description", "system_prompt", "skills")
+
+
+async def _reject_if_embed_bound(
```
alloy-cicd, line 63: **[high | verified] The embed-bound invariant is enforced only on four router endpoints — `DELETE /sub-agents/{id}` and service-internal writes bypass it.**

1. `DELETE /{sub_agent_id}` (line ~496) calls `delete_sub_agent` with no `_reject_if_embed_bound`. The delete is soft (`repo.delete(..., soft=True)`), so the `ON DELETE CASCADE` on `sub_agent_embed_bindings` never fires. Verified consequences: the sync loop keeps publishing versions into the deleted agent (`get_sub_agent_by_id` has no `deleted_at` filter); `bind_connection` still activates users and stamps sockets from `sub_agent_embed_binding_azps`; embedded turns then fail because `/activated` filters `deleted_at IS NULL`; and `_refuse_taken_azps` keeps the azp claimed, blocking any replacement binding until someone finds the orphaned row.
2. Version writes that converge in `SubAgentService.update_sub_agent` skip the guard entirely — e.g. `skill_activation_service` calls `update_skill_hash_in_config`/`remove_skill_from_config`, so editing/deleting a registry skill row of a bound agent writes an un-synced version that the next sync clobbers.

Consider enforcing the "host owns the content" invariant inside `SubAgentService` (single sanctioned writer: `publish_managed_version`), with the router 409 as a friendly translation — then delete either cascades the binding or is refused.

packages/console-backend/console_backend/routers/sub_agent_router.py
```diff
@@ -43,6 +51,62 @@ def get_sub_agent_service(request: Request) -> SubAgentService:
...
+    service = getattr(request.app.state, "embed_binding_service", None)
+    if service is None:
+        return
+    binding = await service.get_binding(db, sub_agent_id)
+    if binding is None:
+        return
+    if data is None:
+        touched = ["the version"]
+    else:
+        touched = [f for f in _HOST_MANAGED_FIELDS if getattr(data, f) is not None]
```
alloy-cicd, line 101: **[medium | API ergonomics] Guard refuses on field PRESENCE, not on actual change — and the ownership matrix is duplicated in the frontend.**

`getattr(data, f) is not None` 409s any client that round-trips the full config, even when the host-owned values are unchanged. The frontend compensates by hand-omitting fields per save (`SubAgentDetailPage`: `system_prompt: isEmbedBound ? undefined : ...`, plus the hostOwns* trio), re-encoding the backend's host-ownership matrix in TS; any other client (the MCP `console_update_sub_agent` tool, scripts) gets spurious 409s on bound agents even when only a Nannos-side field changed. Comparing against the current approved version and refusing only real diffs of host-published fields would keep the matrix in one place and let clients simply disable inputs.

packages/console-backend/console_backend/services/well_known_agent.py
```diff
@@ -0,0 +1,683 @@
...
+    fetched_at: datetime
+    ttl_seconds: int
+
+
+def origin_of(url: str) -> str:
+    parts = urlsplit(url)
+    return f"{parts.scheme}://{parts.netloc}".lower()
+
+
+def compute_revision(
```
alloy-cicd, line 177: **[high | verified] `compute_revision` ignores index.json-only fields — host changes to tools/description/model tier never sync, even via force refresh.**

The hash covers only `framing_version | prompt_digest | sorted(skill_digests)` — i.e. AGENT.md and SKILL.md bytes. Changes to `x-nannos-agent` fields (`name`, `description`, `tools`, `model_tier`, `thinking_level`) or per-skill index descriptions leave the revision unchanged, so `_apply_definition` takes the `revision == binding.revision` branch and only bumps `fetched_at` — no version is published. `force=True` doesn't help: it only bypasses the fetch TTL cache; the revision-equality skip runs unconditionally, so the freshly fetched (changed) definition is discarded while the admin UI reports a successful sync. Include a digest of the normalized index metadata in the revision hash.

alloy-cicd (commented):

Reviewed the full diff (backend embed-binding service + well-known client, socket/send scope, conversation router, orchestrator registry guard, embed-sdk, console-frontend). The core design — server-side azp→sub-agent binding replacing client-declared scope — is sound, but the **unbound-azp state consistently fails open**, and several sync/lifecycle edges are unhandled.

**High severity (verified against the code):**
1. Unbound/failed azp binding silently runs the **full orchestrator inside the third-party page** (`app.py`), including a deploy-order break for existing hosts still sending `subAgentId`.
2. Unbound azp returns the user's **entire conversation list to the host page** (`conversation_router.py`).
3. EMBED activation alone doesn't make a private bound agent visible to non-owner users — **every embedded turn fails** for regular users (`embed_binding_service.py`).
4. `publish_managed_version` can collide with soft-deleted version numbers → **sync loop wedges permanently** (`sub_agent_service.py`).
5. `DELETE /sub-agents/{id}` isn't embed-guarded → zombie binding, permanently claimed azp (`sub_agent_router.py`).
6. `compute_revision` ignores index.json-only fields → host changes to tools/description/model tier **never sync**, even via force refresh (`well_known_agent.py`).

**Medium:** stale `definition` shown after base_url change; per-pod 60s azp cache with negative caching and no cross-pod invalidation; presence-based (not change-based) 409 on PATCH; `isReadOnly` narrowing in the SDK; host-name resolution from `window.location.origin`; raw-SQL writes bypassing the AuditedRepository rule in `packages/console-backend/AGENTS.md`.

**Minor:** name-slug not validated against `SUB_AGENT_NAME_RE`; sequential skill fetches; a 4th frontmatter parser; duplicated binding form with typos.

Details inline (18 comments). Findings 1–6 and the base_url/cache/JWT claims were adversarially re-verified against the checked-out head; the rest come from direct diff analysis. Submitted as COMMENT — approval is left to the human reviewer.

packages/console-backend/console_backend/services/embed_binding_service.py
```diff
@@ -0,0 +1,561 @@
...
+            definition = await self._client.fetch(data.base_url, force=True)
+        except WellKnownFetchError as e:
+            raise EmbedBindingError(
+                f"Could not read the host definition at {data.base_url}{WELL_KNOWN_INDEX_PATH}: {e}"
+            ) from e
+
+        # The row name is the derived one: it doubles as the orchestrator's task-tool
+        # identifier, so it cannot hold the host's display name verbatim.
+        sub_agent_id = await self._sub_agents.create_managed_sub_agent(
+            db, actor, name=definition.agent.sub_agent_name
```
alloy-cicd, line 229: **[medium | verified] Changing `base_url` keeps the OLD host's `definition` — admin UI attributes the previous app's agent/skills to the new URL.**

The `ON CONFLICT DO UPDATE` sets `base_url` and NULLs `revision` when the URL changed, but `definition` is not in the SET list. If the immediate sync of the new URL fails (`sync_binding`'s failure path writes only `last_error`/`last_error_at` and never raises), `_row_to_binding` renders `agent`/`skills` from the stale `definition` under the new `base_url` — potentially forever if the new host stays broken — and `_reject_if_embed_bound` keeps enforcing the old host's ownership matrix. Clear `definition` alongside `revision` when `base_url` changes (or render an explicit "not yet synced" state when `revision IS NULL`).

packages/console-backend/console_backend/services/embed_binding_service.py
```diff
@@ -0,0 +1,561 @@
...
+        # azp -> (monotonic expiry, sub_agent_id or None)
+        self._azp_cache: dict[str, tuple[float, int | None]] = {}
+        # sub_agent_id -> last error text already logged at WARNING
+        self._logged_errors: dict[int, str] = {}
+
+    # ------------------------------------------------------------------ reads
+
+    async def get_binding(
+        self, db: AsyncSession, sub_agent_id: int
+    ) -> EmbedBinding | None:
```
alloy-cicd, line 115: **[medium | verified] Per-pod 60s azp cache (including negative entries) with no cross-pod invalidation.**

`_azp_cache` is a per-process dict; `upsert`/`delete` only `clear()` the local pod. In a multi-replica deployment: an admin creates a binding on pod A and immediately tests — the socket connect lands on pod B holding a cached `None`, and (because unbound fails open, see `app.py`) the test runs full-orchestrator with an unscoped conversation list. Conversely, after `DELETE /embed-binding`, other pods keep activating users and stamping turns onto the un-bound agent for up to 60s, violating the UI's "users are no longer activated automatically" confirmation. Bounded by the TTL, but combined with the fail-open unbound behavior the window matters; if unbound becomes fail-closed this drops to a plain freshness nit worth a comment.

packages/console-backend/console_backend/services/embed_binding_service.py
```diff
@@ -0,0 +1,561 @@
...
+            raise EmbedBindingError(
+                "azp already bound to another sub-agent: " + ", ".join(conflicts)
+            )
+
+    async def _write_binding_rows(
+        self, db: AsyncSession, actor: User, sub_agent_id: int, data: EmbedBindingUpsert
+    ) -> None:
+        """Insert or replace the binding row and its azp rows. No fetch, no commit."""
+        now = datetime.now(timezone.utc)
+        await db.execute(
```
alloy-cicd: **[medium | conventions] All embed-binding writes bypass the repository/audit pattern required by `packages/console-backend/AGENTS.md`.**

"CRITICAL: All database write operations (INSERT/UPDATE/DELETE) MUST use the repository pattern to ensure automatic audit logging" — but this service does raw `db.execute(text(...))` for the binding upsert (here), the azp rows replace (~line 298), `delete_binding` (~line 315), and the definition/revision update (~line 396). Creating, retargeting, or deleting an embed binding is an admin action with real authorization impact (it decides which JWT azp maps to which agent), yet it produces no `audit_logs` row. The timestamp-only bookkeeping UPDATEs (`fetched_at`/`last_seen_at`) arguably fit the documented cache-bookkeeping exemption, but the binding mutations and the definition write do not — an `AuditedRepository` for `sub_agent_embed_bindings` would cover them.

Also in this file, minor: `is_bound` (line ~124) has no callers (as does `WellKnownAgentClient.invalidate`) — dead API surface on a brand-new class; and `create_bound_sub_agent` re-implements `sync_binding`'s tail (get/apply/refetch) where it could end with `return await self.sync_binding(...)` like `upsert_binding` does.

packages/embed-sdk/src/transport/conversations-store.ts
```diff
@@ -468,11 +469,14 @@ export class ConversationsStore {
     return this.contextKeys.get(id);
   }
 
-  /** A conversation owned by ANOTHER embedded surface renders read-only here. */
+  /** A conversation an embedded surface owns renders read-only in the console:
+   *  its turns assume a live host page with registered objects. An embedded
+   *  surface's own list is server-scoped to its sub-agent, so there every
+   *  stamped conversation is its own and stays editable. */
   isReadOnly(id: string): boolean {
```
alloy-cicd, line 476: **[medium] `isReadOnly` narrowed from per-conversation ownership to a surface-level boolean.**

Previously an embedded surface compared the conversation's `embeddedSubAgentId` against its own scope; now `!this.opts.embedded` makes ALL embedded-stamped conversations editable on any embedded surface. That is only safe if the server list is guaranteed scoped — and it is not when the token's azp is unbound (see `conversation_router.py`): the unscoped list can contain another application's conversations, which old code rendered read-only and new code lets the user send turns into from the wrong host page. If the server-side unbound case becomes fail-closed this is fine; otherwise keep a client-side ownership check as a floor.

packages/embed-sdk/src/core/index.ts
```diff
@@ -257,22 +260,39 @@ export class NannosCore {
...
+  }
+
+  /**
+   * The display name of the agent this host publishes (ADR-0006): read from the
+   * host's OWN origin, `/.well-known/agent-skills/index.json` →
+   * `x-nannos-agent.name`. Same-origin, unauthenticated, cached for the page's
+   * life. Null outside a browser, when the host publishes no definition, or on
+   * any error — callers fall back to the A2A handshake's agent name.
+   */
+  resolveHostAgentName(fetchImpl: typeof fetch = (...args) => fetch(...args)): Promise<string | null> {
```
alloy-cicd, line 282: **[medium] `resolveHostAgentName` reads the well-known index from `window.location.origin`, but the binding's `base_url` need not equal the SPA origin.**

The binding's `base_url` is admin configuration (e.g. an API service at `https://riad-api.alloy.ch`) while the widget runs on the SPA origin (`https://riad.alloy.ch`); if the SPA doesn't serve `/.well-known/agent-skills/index.json` on its own origin (or its static server 404s), `hostAgentName` stays null and the header falls back to the A2A handshake name ("Orchestrator Agent") — the mislabeling the removed `resolveSubAgentName` (backend lookup by id) existed to fix. Consider having the backend expose the bound agent's display name to the SDK (it knows it from the binding) instead of re-fetching the index from the page origin.

packages/console-backend/console_backend/services/well_known_agent.py
```diff
@@ -0,0 +1,683 @@
...
+
+class WellKnownSkill(BaseModel):
+    name: str
+    description: str
+    body: str = Field(description="SKILL.md body, frontmatter stripped")
+    url: str
+    digest: str
+
+
+def to_sub_agent_name(published_name: str) -> str:
```
alloy-cicd, line 113: **[medium] `to_sub_agent_name` never validates its output against the actual authority, `SUB_AGENT_NAME_RE`.**

It re-encodes the name rule with private constants (`MAX_NAME_LENGTH`, `_NAME_SEPARATORS`, `_LEADING_NON_LETTER`) instead of asserting the produced slug matches `SUB_AGENT_NAME_RE`/`validate_sub_agent_name` from `models/sub_agent.py` (a module it already imports from). `create_managed_sub_agent` writes the derived name straight to the DB row, so if the canonical pattern is ever tightened, this path can again produce invalid names that bypass validation — recreating exactly the broken-registry-row failure this PR's `registry.py` guard exists to mitigate (the agent silently disappears per-user). One `assert SUB_AGENT_NAME_RE.match(slug)` (or reuse of the validator) closes the loop. Related: the same regex now exists in three hand-synced copies (agent-common `a2a/models.py`, `models/sub_agent.py`, frontend `subAgentName.ts`) — declaring it as `pattern=` on the pydantic Field would push it into OpenAPI/`types.gen.ts`, and a cross-package equality test would catch drift.

packages/console-backend/console_backend/services/well_known_agent.py
```diff
@@ -0,0 +1,683 @@
...
+            base_url, index_url, origin, agent_meta["prompt"]["url"], "agent"
+        )
+        prompt_bytes, _ = await self._get(
+            base_url, prompt_url, origin, MAX_SKILL_FILE_SIZE_BYTES, "agent"
+        )
+        _verify_digest(base_url, "agent", prompt_bytes, agent_meta["prompt"]["digest"])
+        prompt_body = _parse_prompt(base_url, prompt_bytes)
+
+        skills: list[WellKnownSkill] = []
+        for entry in skill_entries:
```
alloy-cicd, line 364: **[minor | perf] Skill files are fetched sequentially — up to `MAX_SKILL_FILES=20` awaited GETs, each with a 10s timeout, inside admin HTTP requests (probe, create/PUT binding) and the sync loop.**

A slow host serializes its latency 20×; the URLs are independent and already same-origin-checked, so `asyncio.gather` (optionally under a small semaphore) turns N round-trips into ~1 of wall time before the digest checks.

packages/console-backend/console_backend/services/well_known_agent.py
```diff
@@ -0,0 +1,683 @@
...
+    material = "|".join([framing_version, prompt_digest, *sorted(skill_digests)])
+    return hashlib.sha256(material.encode("utf-8")).hexdigest()[:16]
+
+
+def version_hash_for(revision: str) -> str:
+    """The `sub_agent_config_versions.version_hash` a synced revision gets (`wk` + 10 hex)."""
+    return f"wk{revision[:10]}"
+
+
+def split_frontmatter(text: str) -> tuple[dict[str, Any], str]:
```
alloy-cicd, line 221: **[minor | reuse] This is the repo's 4th SKILL.md frontmatter parser.**

`_strip_skill_frontmatter` (`sub_agent_service.py:78`), `_extract_description_from_frontmatter` (`skill_registry_service.py:631`) and `_extract_description` (`skill_sources/github.py:459`) already exist — the legacy ones use a naive `find("---", 3)` that breaks on `---` inside YAML values, while this one is regex + `yaml.safe_load`. The same SKILL.md can round-trip differently depending on which service touches it. This new, correct implementation is the one that should live in a shared util and replace the others (fine as a follow-up, but worth a tracking note). Same story for `_SKILL_NAME` (line ~79), which subtly differs from `_validate_skill_name` in `skills_registry_router.py:1061` — two authorities for what a valid skill name is.

packages/console-frontend/src/components/subagents/EmbeddedAgentForm.tsx
```diff
@@ -0,0 +1,263 @@
...
+            <CardHeader>
+              <CardTitle>Application</CardTitle>
+              <CardDescription>
+                The application publishes this agent under <code>/.well-known/agent-skills/</code>. Nannos reads name,
+                description, system prompt and skills from there and keeps them in sync. They cannot be edited here.
+              </CardDescription>
+            </CardHeader>
+            <CardContent className="space-y-4">
+              <div className="space-y-2">
+                <Label htmlFor="embed-base-url">Base URL *</Label>
```
alloy-cicd, line 93: **[minor] Duplicates `EmbedBindingForm` (EmbedBindingDialog.tsx) — same Base URL + azp field pair for the same `EmbedBindingUpsert` body — and the copy has typos.**

Only `parseAzps` was extracted; the hint text already drifts between the two, and this copy ships "When a token is **recieved**" and "this **sug-agent** will be activated" verbatim in the admin UI. Extracting the field pair into the shared `embedBinding.ts` module (or reusing `EmbedBindingForm` with a slot for the probe button) keeps the binding contract in one place. Related nit in `SubAgentCreatePage.tsx` (~line 84): the "Choose Agent Type" chooser is rendered twice (page-level embedded branch + inside `SubAgentForm`) with two type states; the page copy never passes `disabled`, so cards stay clickable while the create mutation is pending.

packages/orchestrator-agent/app/core/registry.py
```diff
@@ -424,82 +424,94 @@ def _to_user(
...
-                            scopes=cv.foundry_scopes or [],
-                            version=cv.foundry_version,
+                        logger.debug(
+                            f"Added Foundry local sub-agent '{local_subagents[-1].model_dump_json()}' for user sub {user_sub}"
                         )
-                    )
-                    logger.debug(
-                        f"Added Foundry local sub-agent '{local_subagents[-1].model_dump_json()}' for user sub {user_sub}"
-                    )
+            except ValidationError as e:
```
alloy-cicd, line 520: **[minor | design note] The ValidationError skip is a good stopgap, but the misdiagnosis layer it works around is still in place.**

As the new comment itself notes, `get_user`'s blanket `except Exception → return None` (~line 318) is what turned one bad row into "user not found" and a broken every-turn experience. The next non-ValidationError defect in per-user conversion (a KeyError from a schema change, a TypeError in placeholder resolution) takes the whole user down the same way, disguised again. Consider narrowing `get_user`'s handling (infrastructure errors retryable, data errors degrade per-agent via a small `_convert_sub_agent(sa) -> config | None` helper wrapping the ladder), plus a one-off report of existing `sub_agents` rows violating the new name rule — the guard skips them silently.

packages/console-backend/app.py
thecopy: **Kept fail-open, by decision.** On the socket a bearer token is only ever an embedded host; the user is authenticated, and Keycloak client registration is the authorization gate — an unbound azp gives the user nothing beyond what the console gives them.

What changed in 93e360b9: an unbound azp now logs a WARNING at connect, so an azp typo or a not-yet-created binding is visible instead of silent. ADR-0006 records the rationale.

The deploy-order claim does not apply: no deployed cockpit (`main`, `v1.96.0`) carries the nannos integration at all; only the spike branch does, and it already uses the new SDK. To revisit if a third-party host is ever bound.

_— reply drafted with Claude Code_

packages/console-backend/app.py
thecopy: Not changed — same decision as the `_apply_embed_scope` thread. An error in `bind_connection` still connects unbound; the exception is logged and the unbound state is now logged at WARNING (93e360b9). The `getattr` guards stay for now.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/routers/conversation_router.py
thecopy: Unbound stays unfiltered, by the same decision as on `app.py`: the list is the authenticated user's own, and the only host is first-party. Bound tokens are scoped exactly as before.

Deferred as follow-ups: embed scope on conversation `DELETE`/`PATCH`, and stashing the validated claims on `request.state` to drop the second JWT validation.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/embed_binding_service.py
thecopy: **Fixed in 93e360b9.** An `embed` activation row is now an access grant in `get_accessible_sub_agents` (non-admin branch gains `OR usa.activated_by = 'embed'`), and such agents get `effective_permission = "read"`.

`tests/test_embed_activation_access.py`: a non-owner without groups does not see the private bound agent, and sees it in the activated list right after the embed activation.

Consequence accepted: users who arrived through the host also see the agent in their console list. ADR-0006 updated — its sentence about `activated_only` was wrong.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/sub_agent_service.py
thecopy: **Fixed in 93e360b9:** `publish_managed_version` allocates `COALESCE(MAX(version), 0) + 1`, same as `update_sub_agent`. `test_publish_managed_version_skips_numbers_held_by_deleted_drafts` reproduces the draft-v2-then-deleted case against Postgres and gets version 3.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/routers/sub_agent_router.py
thecopy: **Point 1 fixed in 93e360b9:** `DELETE /sub-agents/{id}` calls `_reject_if_embed_bound(..., deleting=True)` and returns 409 "remove the embed binding first" while bound; the console's delete button is disabled with the same hint. Test added in `tests/test_embed_binding_router.py`.

Point 2 (service-internal writes via `skill_activation_service`) deferred — moving the guard into `SubAgentService` is the right shape, out of scope here.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/routers/sub_agent_router.py
thecopy: Deferred. Agreed that a change-based comparison would let clients round-trip the config and would remove the frontend `hostOwns*` trio; tracked as a follow-up.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/well_known_agent.py
thecopy: **Fixed in 93e360b9:** `compute_revision` now hashes a canonical JSON of framing version + agent block (name, description, organization, tools, model_tier, thinking_level, prompt digest) + skills sorted by name (name, description, digest).

Tests: `test_index_metadata_is_part_of_the_revision` (each field moves the hash) and `test_revision_tracks_index_only_changes_end_to_end` (same AGENT.md/SKILL.md bytes, different tool list → new revision).

Note: existing bindings publish one extra version after deploy.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/embed_binding_service.py
thecopy: **Fixed in 93e360b9:** the `ON CONFLICT` now clears `definition` together with `revision` when `base_url` changes (`repositories/embed_binding_repository.py`). ADR-0006 schema section updated.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/embed_binding_service.py
thecopy: Not changed. With fail-open kept on purpose (see the `app.py` thread) this is the freshness nit: a binding created on pod A is unknown on pod B for up to 60 s (and warned in the log), a deleted one lingers up to 60 s elsewhere. Acceptable for now.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/embed_binding_service.py
thecopy: **Fixed in 93e360b9:** binding create/replace/remove go through the new `EmbedBindingRepository(AuditedRepository)` and land in `audit_logs` as an UPDATE of the sub-agent (entity `sub_agent`, id = sub_agent_id) with the binding before/after — no new entity type or migration needed. `delete_binding` now takes the actor.

Sync bookkeeping (`revision`, `definition`, `fetched_at`, `last_error`, `last_seen_at`, `azps_seen`) stays a direct write; documented in the repository docstring and ADR-0006. `is_bound` / `invalidate` and the `create_bound_sub_agent` tail: left as is.

_— reply drafted with Claude Code_

packages/embed-sdk/src/transport/conversations-store.ts
thecopy: Not changed, accepted with the fail-open decision: the list an embedded surface receives is the authenticated user's own, and the only host is first-party. If unbound ever becomes fail-closed this is moot; if a third-party host is bound we revisit both together.

_— reply drafted with Claude Code_

packages/embed-sdk/src/core/index.ts
thecopy: **Addressed in 93e360b9:** the connect handshake now carries `embeddedAgent` (sub-agent id, published name, description, organization, revision), resolved from the binding on console-backend (`_embedded_agent_info` in `app.py`); the composer shows it via `useEmbeddedAgent`. The label no longer depends on the page origin.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/well_known_agent.py
thecopy: **Fixed in 93e360b9:** `to_sub_agent_name` validates its output with `SUB_AGENT_NAME_RE.fullmatch` and falls back to `FALLBACK_AGENT_NAME` otherwise, so a tightened rule can never reach a row unvalidated. `test_to_sub_agent_name_output_is_always_accepted_by_the_console_models` covers the property.

The three regex copies / `pattern=` on the pydantic field: deferred.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/well_known_agent.py
thecopy: Deferred. `asyncio.gather` under a small semaphore is the obvious change; out of scope for this PR.

_— reply drafted with Claude Code_

packages/console-backend/console_backend/services/well_known_agent.py
thecopy: Deferred as a follow-up: consolidate on `split_frontmatter` (regex + `yaml.safe_load`), replace the three legacy parsers, and keep one skill-name authority.

_— reply drafted with Claude Code_

packages/console-frontend/src/components/subagents/EmbeddedAgentForm.tsx
thecopy: Typos fixed in 93e360b9 ("received", "sub-agent"). Extracting the shared field pair, and the double type chooser / missing `disabled` on `SubAgentCreatePage`: deferred.

_— reply drafted with Claude Code_

packages/orchestrator-agent/app/core/registry.py
thecopy: Deferred; agreed on the direction (narrow `get_user`'s `except`, per-agent conversion helper). Out of scope here.

_— reply drafted with Claude Code_

PR #221, opened 2026-09-09
fix(console-frontend): page through the admin tool risk scores table

aartaria:
No linked issue — spotted while reading the admin page.

## What

The **Tool Risk Scores** admin page fetched a fixed first page (`limit=100, offset=0`) and ignored the `total` the API returns, so beyond 100 scored tools the table silently showed a truncated list with nothing to indicate rows were missing.

The gap was UI-only: `GET /api/mcp/tools/risk-scores` already takes `limit`/`offset` and returns `total`, and the orchestrator's `ToolRiskCache` refresh already pages through it correctly.

- `page` state, 20 rows per page, offset derived from the page; query key includes the page.
- Reuses the existing `components/admin/Pagination` control (same pattern as UsersPage / AuditPage), fed by `total`.
- Invalidating the last row of a trailing page steps back one page so you don't land on an empty view. The existing `invalidateQueries({ queryKey: ['adminToolRiskScores'] })` calls still match by prefix.
- Drops a `sort: 'updated_at:desc'` query param the backend never declared and silently ignored — `tool_risk_repository.get_scores_paginated` orders by `updated_at DESC` unconditionally, so ordering is unchanged.

## Verification

`tsc --noEmit` and `eslint` on the touched file are both clean. There is no test file for this page; behaviour change is presentational.

_Authored by Claude Code (agent) on @aartaria's machine._

## Checklist

### DIFF
--- a/packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
+++ b/packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
@@ -19,6 +19,7 @@ import {
   DialogTitle,
 } from '@/components/ui/dialog';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
+import { Pagination } from '@/components/admin/Pagination';
 import { client } from '@/api/generated/client.gen';
 
 interface RiskFactor {
@@ -49,7 +50,7 @@ interface UpsertPayload {
 async function fetchRiskScores(limit = 100, offset = 0): Promise<{ items: ToolRiskScore[]; total: number }> {
   const res = await client.get({
     url: '/api/mcp/tools/risk-scores',
-    query: { limit, offset, sort: 'updated_at:desc' },
+    query: { limit, offset },
   });
   return res.data as { items: ToolRiskScore[]; total: number };
 }
@@ -74,6 +75,8 @@ function getRiskBadge(score: number) {
 }
 
 const ALL_ACTIONS = ['approve', 'edit', 'reject'] as const;
+
+const PAGE_SIZE = 20;
 
 /** Form state for add/edit dialog. */
 interface FormState {
@@ -115,10 +118,11 @@ export function ToolRiskScoresPage() {
   const [editingExisting, setEditingExisting] = useState(false);
   const [form, setForm] = useState<FormState>(emptyForm());
   const [jsonError, setJsonError] = useState<string | null>(null);
+  const [page, setPage] = useState(1);
 
   const { data, isLoading, isRefetching } = useQuery({
-    queryKey: ['adminToolRiskScores'],
-    queryFn: () => fetchRiskScores(100, 0),
+    queryKey: ['adminToolRiskScores', page],
+    queryFn: () => fetchRiskScores(PAGE_SIZE, (page - 1) * PAGE_SIZE),
   });
 
   const deleteMutation = useMutation({
@@ -126,6 +130,10 @@ export function ToolRiskScoresPage() {
       deleteRiskScore(toolName, serverSlug),
     onSuccess: () => {
       toast.success('Risk score invalidated — tool will be re-scored on next use');
+      // Removing the last row of a trailing page would leave us past the end.
+      if ((data?.items?.length ?? 0) === 1 && page > 1) {
+        setPage(page - 1);
+      }
       queryClient.invalidateQueries({ queryKey: ['adminToolRiskScores'] });
       setConfirmDelete(null);
     },
@@ -197,6 +205,7 @@ export function ToolRiskScoresPage() {
   };
 
   const scores = data?.items ?? [];
+  const total = data?.total ?? 0;
 
   return (
     <div className="flex flex-col gap-6 p-4 pb-8">
@@ -323,6 +332,8 @@ export function ToolRiskScoresPage() {
           </table>
         </div>
       )}
+
+      {!isLoading && <Pagination page={page} limit={PAGE_SIZE} total={total} onPageChange={setPage} />}
 
       {/* Detail Dialog */}
       <Dialog open={!!selectedScore} onOpenChange={() => setSelectedScore(null)}>


packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
```diff
@@ -324,6 +333,8 @@ export function ToolRiskScoresPage() {
         </div>
       )}
 
+      {!isLoading && <Pagination page={page} limit={PAGE_SIZE} total={total} onPageChange={setPage} />}
```
alloy-cicd, line 336: **Page changes unmount the table and this pagination bar.** With `page` in the query key and no `placeholderData`, every visit to a not-yet-cached page has `isLoading === true` (v5: `isPending && isFetching`), so the table swaps to `TableSkeleton` and this `!isLoading &&` gate unmounts the pager — the Next button the user just clicked vanishes from under the cursor and the layout collapses, inviting a mis-aimed second click. Flipping back to an already-seen page also refetches every time (default `staleTime: 0`).

Suggested shape: add `placeholderData: keepPreviousData` (from `@tanstack/react-query`; v5.90 is in `package.json`) to the `useQuery`, and render `<Pagination …/>` unconditionally — it already returns `null` when `total === 0` (`Pagination.tsx:16-18`), which is how UsersPage:321 and AuditPage:243 use it. The two changes together keep rows, total, and the pager mounted during fetches.

packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
```diff
@@ -115,17 +118,22 @@ export function ToolRiskScoresPage() {
...
+    queryFn: () => fetchRiskScores(PAGE_SIZE, (page - 1) * PAGE_SIZE),
   });
 
   const deleteMutation = useMutation({
     mutationFn: ({ toolName, serverSlug }: { toolName: string; serverSlug: string }) =>
       deleteRiskScore(toolName, serverSlug),
     onSuccess: () => {
       toast.success('Risk score invalidated — tool will be re-scored on next use');
+      // Removing the last row of a trailing page would leave us past the end.
+      if ((data?.items?.length ?? 0) === 1 && page > 1) {
```
alloy-cicd, line 134: **The `items.length === 1` guard only covers one shrink path, and the block double-fetches.**

1. Out-of-range pages are still reachable: another admin (or a second tab) deletes rows, or Refresh runs after the count shrank — the guard reads the stale `data` closure and doesn't fire, the refetch at the old offset returns `items: []` with `total > 0`, and the UI shows the absolute "No tool risk scores found. Scores are generated automatically…" copy above a footer reading e.g. "Showing 41 to 40 of 40 results / Page 3 of 2" (`Pagination` does no clamping either). A derived clamp covers every shrink path at once — `RateCardsPage.tsx:185-191` already does `const totalPages = Math.max(1, Math.ceil(total / PAGE_SIZE))` + `useEffect(() => { if (page > totalPages) setPage(1) })`, which would replace this special case entirely.

2. `setPage(page - 1)` is a queued state update, so the synchronous prefix `invalidateQueries` that follows still sees the observer on `['adminToolRiskScores', page]` and (v5 default `refetchType: 'active'`) fires a network request for the page being abandoned; the new page then fetches again on mount. Two round trips per trailing-page delete, one discarded. The derived clamp avoids this too, since the page transition and refetch collapse into one.

packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
```diff
@@ -115,17 +118,22 @@ export function ToolRiskScoresPage() {
   const [editingExisting, setEditingExisting] = useState(false);
   const [form, setForm] = useState<FormState>(emptyForm());
   const [jsonError, setJsonError] = useState<string | null>(null);
+  const [page, setPage] = useState(1);
 
   const { data, isLoading, isRefetching } = useQuery({
-    queryKey: ['adminToolRiskScores'],
-    queryFn: () => fetchRiskScores(100, 0),
+    queryKey: ['adminToolRiskScores', page],
```
alloy-cicd, line 124: **Upsert success never resets `page`, so the saved row is invisible from page ≥ 2.** The backend orders by `updated_at DESC` unconditionally (`tool_risk_repository.get_scores_paginated`), so a created **or just-edited** score always lands at the top of page 1. `upsertMutation.onSuccess` (~line 147) only invalidates: a user on page 3 who saves gets the success toast, page 3 refetches, and their row is gone from view — an edit even makes the row they were looking at disappear from the current page. Everything below also shifts by one, so the window looks scrambled. `setPage(1)` in the upsert `onSuccess` fixes both cases.

packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
```diff
@@ -49,7 +50,7 @@ interface UpsertPayload {
 async function fetchRiskScores(limit = 100, offset = 0): Promise<{ items: ToolRiskScore[]; total: number }> {
   const res = await client.get({
     url: '/api/mcp/tools/risk-scores',
-    query: { limit, offset, sort: 'updated_at:desc' },
+    query: { limit, offset },
```
alloy-cicd, line 53: Reuse note: the generated TanStack layer already exposes `listRiskScoresApiMcpToolsRiskScoresGetOptions` / `…GetQueryKey` (and `…GetInfiniteOptions`) in `api/generated/@tanstack/react-query.gen.ts`, typed by the response model — the sibling admin pages build their queries from those factories (e.g. `listUsersApiV1AdminUsersGetOptions` in UsersPage:47). Hand-rolling the `['adminToolRiskScores', page]` key + raw `client.get` means invalidations issued via the generated key factory elsewhere won't hit this cache, and it's exactly what let the inert `sort:` param sit here unnoticed — the typed `query` object (only `limit`/`offset`) would have rejected it at compile time. Not blocking, but worth converting while this file is open.

packages/console-frontend/src/pages/admin/ToolRiskScoresPage.tsx
```diff
@@ -49,7 +50,7 @@ interface UpsertPayload {
 async function fetchRiskScores(limit = 100, offset = 0): Promise<{ items: ToolRiskScore[]; total: number }> {
```
alloy-cicd, line 50: Minor: `limit = 100, offset = 0` defaults are now dead — the sole caller always passes `PAGE_SIZE` and a computed offset — and the `100` contradicts `PAGE_SIZE = 20` declared below. A future call site written as `fetchRiskScores()` would silently reintroduce the unpaginated 100-row fetch this PR removes, with no type error. Dropping the defaults (`limit: number, offset: number`) makes that impossible.

alloy-cicd (commented):

The pagination wiring is sound at the API level (the backend's `limit`/`offset` contract, the 1-based `Pagination` component, and the prefix-matching `invalidateQueries` calls all check out, and dropping the inert `sort` param is a correct no-op — FastAPI never declared it and the repository hardcodes `ORDER BY updated_at DESC`). The findings are all UX/state edge cases that paging newly makes reachable, plus two smaller cleanups. Details inline; roughly in severity order:

1. Every page change unmounts the table **and the pagination control itself** (no `placeholderData: keepPreviousData`, plus the `!isLoading &&` gate).
2. The `items.length === 1` back-off only covers one shrink path — any other way the row count drops leaves the user on an out-of-range page showing the absolute "No tool risk scores found" empty state while `total > 0`; the same block also double-fetches on delete.
3. Saving from the Add/Edit dialog never resets `page`, and the backend sorts by `updated_at DESC`, so the created/edited row always lands on page 1 — invisible to a user sitting on page ≥ 2.
4. The hand-rolled query key/fetch bypasses the generated `listRiskScoresApiMcpToolsRiskScoresGet*` factories the sibling admin pages use.
5. `fetchRiskScores`' dead `limit = 100` default now contradicts `PAGE_SIZE`.

Minor notes not worth an inline thread: the query destructures no `isError`, so a failed fetch on page > 1 renders the empty-state copy and (with `total` falling back to 0) hides the pager, leaving no way back but a reload — pre-existing gap, but paging makes it a trap. The Refresh button's `isRefetching` is false for a first visit to a page-scoped key, so it doesn't spin/disable during those loads (`isFetching` would). And `ORDER BY updated_at DESC` has no tiebreaker, so batch re-scoring that stamps identical timestamps can duplicate/skip rows across page boundaries — a backend nicety (`updated_at DESC, tool_name`) outside this diff.

PR #222, opened 2026-09-09
perf(agent-common): run fast-model utility calls with thinking off

aartaria:
closes #

## What

`create_model` sends `reasoning_effort` only when a caller passes a `thinking_level`, so the utility calls riding `get_default_fast_model()` sent nothing and inherited the **provider** default — dynamic thinking on `gemini-3.5-flash`, the current `chat:low`.

Measured on LangSmith (dev project, 2026-09-09), `tool-call-summarize` took **2.7–25.5 s, median ~8 s**, to write one sentence about one `ls` call. All of it sits between the model emitting the tool call and the HITL approval card the user is waiting for. The slowest run produced exactly this, in 25.5 s:

> "I am going to look at the list of files and folders in the main system directory."

Its invocation params carried `model: gemini-3.5-flash`, `stream: false`, and no `reasoning_effort` at all.

Same class of bug as the `gateway_chat` fix in #209 — which is why console-backend's `llm_gateway` already defaults to `reasoning_effort="none"`. agent-common had no way to express it: `ThinkingLevel` deliberately has no `none` member, since the app models thinking-off as a toggle rather than a tier.

## How

- **`REASONING_OFF`** + keyword-only `create_model(reasoning_effort=…, max_tokens=…)` overrides, so a caller can state "off" instead of staying silent.
- **`create_fast_model()`** — cheap tier, thinking off, no streaming, short output cap. For the utility calls: summaries, risk scores, watch conditions, chunk descriptions.
- **`max_tokens_for_effort(REASONING_OFF)` now returns `None`.** It previously fell through to the unknown-tier fallback and would have quietly requested `_DEFAULT_REASONING_MAX_TOKENS` for a call that does no thinking at all — a trap for the first caller to pass `"none"`.
- **The summarizer uses it, and bounds its own tail**: a 5 s `asyncio.wait_for` past which the card renders raw args (the already-documented `None` fallback). `TimeoutError` is logged at warning rather than as an exception — a slow gateway is expected weather, but a systematic regression should stay visible.

- **Truncation is named, not swallowed.** The short cap is a risk this PR introduces, and with `with_structured_output` a reply stopped at `max_tokens` comes back as invalid JSON — logged, before this, as a generic "summarization failed". console-backend's `gateway_chat` / `GatewayText` already learned that lesson (*"went unlogged for a month"*), so the summarizer requests `include_raw=True` and splits the two cases: `finish_reason == "length"` warns that the batch hit the output cap, anything else warns *unreadable* with the parse error. The raw-args fallback is unchanged in both, and the unwrapping sits inside the existing `try` — the envelope's shape is the provider's business, and a caller mid-`interrupt()` must not receive an exception from a display-only nicety.
- **The duplicated policy is cross-referenced.** `REASONING_OFF` and `_FAST_MODEL_MAX_TOKENS` hold the same values as `gateway_chat`'s defaults for the same reasons; console-backend keeps its own copy because it imports neither agent-common nor langchain (the way it also duplicates the gateway Bearer-key default). One decision, two homes — the comments now say so on both sides.

## Deliberately not in scope

Six other fast-model call sites stay on plain `create_model`: `hitl_resume.py`, `tool_risk_scorer.py`, `indexing_store.py`, `dynamic_tool_dispatch.py`, `toolset_selector.py`, `file_analyzer.py`. Risk scoring, toolset selection and file analysis may lean on reasoning for quality, so switching them is a judgement call, not a mechanical one — happy to do it in a follow-up if we think the quality cost is nil.

The structural fix — render the approval card immediately and patch the summary in afterwards — is a bigger change and a worse fit for Slack (the card is a posted message, so it means a `chat.update` per summary, and a card that rewrites itself a second later reads as a glitch). Worth filing separately.

## Testing

- 4 new tests pinning `create_fast_model`'s contract (thinking off, no streaming, capped output), 2 for the timeout fallback, 1 for `REASONING_OFF`, and 3 for the unusable-reply paths (truncation names the cap; an unreadable reply names the parse error and specifically does *not* claim truncation; a missing `raw` still falls back cleanly).
- Full agent-common suite: **995 passed**.
- `ruff check --line-length 120` clean on all touched files.

Not verified against a live gateway — the latency win is inferred from the traces above plus the removal of the thinking default, so worth a look at a fresh `tool-call-summarize` trace after this deploys.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no; `create_model`'s new params are keyword-only with defaults that preserve today's behaviour.

---
🤖 Written by Claude Code (Opus 5) in a session driven by @aartaria.

### DIFF
--- a/packages/agent-common/agent_common/core/model_factory.py
+++ b/packages/agent-common/agent_common/core/model_factory.py
@@ -244,6 +244,12 @@ def assert_gateway_configured() -> None:
 # this runtime mapping never disagree.
 _NON_PORTABLE_EFFORT: dict[str, str] = {"minimal": "low", "xhigh": "high"}
 
+# LiteLLM's explicit "no extended thinking" value, portable across providers (the gateway
+# drops it for models with no reasoning to disable). NOT a `ThinkingLevel` member: the app
+# models thinking-off as a toggle rather than a tier, so it can only be requested through
+# `create_model(reasoning_effort=...)` — which is what `create_fast_model` does.
+REASONING_OFF = "none"
+
 
 def get_reasoning_effort(
     thinking_level: ThinkingLevel | None, model_type: ModelType | None = None
@@ -293,8 +299,11 @@ def max_tokens_for_effort(effort: str | None) -> int | None:
     Returns ``None`` when no effort is active (reasoning off) — callers should leave
     ``max_tokens`` unset so the gateway default applies and we don't lower a non-reasoning
     model's own output cap. See ``_MAX_TOKENS_BY_EFFORT`` for the rationale.
+
+    ``REASONING_OFF`` counts as no effort: it is the explicit off switch, so it must not
+    raise the ceiling the way a real tier does (it has no thinking budget to make room for).
     """
-    if not effort:
+    if not effort or effort == REASONING_OFF:
         return None
     return _MAX_TOKENS_BY_EFFORT.get(effort, _DEFAULT_REASONING_MAX_TOKENS)
 
@@ -304,12 +313,23 @@ def create_model(
     thinking_level: ThinkingLevel | None = None,
     callbacks: list | None = None,
     streaming: bool = True,
+    *,
+    reasoning_effort: str | None = None,
+    max_tokens: int | None = None,
 ) -> BaseChatModel:
     """Create a gateway-backed chat model for the given alias.
 
     `thinking_level` becomes `reasoning_effort` and is always forwarded; the gateway
     drops it for models that don't support reasoning (`drop_params`), so no per-model
     capability check is needed here.
+
+    `reasoning_effort` overrides that mapping with a raw LiteLLM value. It exists for
+    ``REASONING_OFF`` — `ThinkingLevel` has no `none` member (the app models thinking-off as
+    a toggle, not a tier), so without this there was no way to say "off" and the call
+    inherited the PROVIDER default instead. See `create_fast_model`.
+
+    `max_tokens` overrides the effort-derived output ceiling for callers that know their
+    answer is short.
     """
     # Gateway-aware subclass: preserves reasoning_content that base ChatOpenAI (>=1.2) drops.
     ChatOpenAI = _gateway_chat_openai_cls()
@@ -325,7 +345,7 @@ def create_model(
         callbacks = [cb for cb in callbacks if type(cb).__name__ != "CostTrackingCallback"] or None
 
     model_kwargs: dict = {}
-    effort = get_reasoning_effort(thinking_level, model_type)
+    effort = reasoning_effort if reasoning_effort is not None else get_reasoning_effort(thinking_level, model_type)
     if effort:
         model_kwargs["reasoning_effort"] = effort
 
@@ -333,7 +353,7 @@ def create_model(
     # whole (low) gateway-default max_tokens and truncate the answer. Unset for reasoning-off
     # turns (leave the gateway default). ContinueOnTruncationMiddleware raises this further
     # per-retry when a turn still gets cut off. See max_tokens_for_effort.
-    max_tokens = max_tokens_for_effort(effort)
+    max_tokens = max_tokens if max_tokens is not None else max_tokens_for_effort(effort)
 
     # NOTE: Native Gemini server-side web search (googleSearch) is NOT enabled here.
     # Enabling it for a tool-using deep agent requires `include_server_side_tool_invocations`
@@ -736,6 +756,45 @@ def get_default_fast_model() -> ModelType | None:
     Returns ``None`` only when no chat default is configured at all; runtime callers that must
     actually run a model pair this with require_default_model()."""
     return _default_alias_for("chat:low") or get_default_model()
+
+
+# One sentence per tool call, one risk score, one condition verdict — generous for a batch
+# of a few summaries, low enough that a model ignoring "ONE short sentence" cannot stretch
+# the call. Utility work must never inherit the reasoning tiers' 8k-32k ceiling.
+_FAST_MODEL_MAX_TOKENS = 1024
+
+
+def create_fast_model(
+    model_type: ModelType,
+    *,
+    max_tokens: int = _FAST_MODEL_MAX_TOKENS,
+    callbacks: list | None = None,
+) -> BaseChatModel:
+    """A cheap-tier model configured for LOW LATENCY: thinking off, no streaming, short cap.
+
+    For the utility calls that ride `get_default_fast_model()` — tool-call summaries, risk
+    scoring, watch-condition checks, chunk descriptions. They want the cheap tier's speed,
+    not its reasoning, and several of them sit on a user-visible critical path.
+
+    Thinking off is the whole point. `create_model` sends `reasoning_effort` only when a
+    caller passes a `thinking_level`, so these calls sent nothing and inherited the PROVIDER
+    default — dynamic thinking on Gemini 3.x flash, the current `chat:low`. Measured on
+    tool-call summaries (LangSmith, 2026-09-09): 2.7-25.5 s, median ~8 s, to produce one
+    sentence about one `ls` call — all of it in front of the HITL approval card. Same failure
+    as the `gateway_chat` fix in #209, which is why console-backend's `llm_gateway` already
+    defaults to `reasoning_effort="none"`; agent-common had no way to express it.
+
+    Streaming is off because every one of these callers awaits the whole answer before doing
+    anything with it (structured output, a score, a verdict) — a token stream buys nothing
+    and costs a per-chunk callback trip.
+    """
+    return create_model(
+        model_type,
+        callbacks=callbacks,
+        streaming=False,
+        reasoning_effort=REASONING_OFF,
+        max_tokens=max_tokens,
+    )
 
 
 def get_default_indexing_model() -> ModelType | None:
--- a/packages/agent-common/agent_common/core/model_factory.py
+++ b/packages/agent-common/agent_common/core/model_factory.py
@@ -257,9 +257,7 @@ _NON_PORTABLE_EFFORT: dict[str, str] = {"minimal": "low", "xhigh": "high"}
 REASONING_OFF = "none"
 
 
-def get_reasoning_effort(
-    thinking_level: ThinkingLevel | None, model_type: ModelType | None = None
-) -> str | None:
+def get_reasoning_effort(thinking_level: ThinkingLevel | None, model_type: ModelType | None = None) -> str | None:
     """Map the app `thinking_level` to LiteLLM's `reasoning_effort`.
 
     low/medium/high are accepted by every reasoning provider and pass through. The
@@ -336,7 +334,20 @@ def create_model(
 
     `max_tokens` overrides the effort-derived output ceiling for callers that know their
     answer is short.
+
+    Both overrides are validated rather than normalized: an empty `reasoning_effort` would
+    take the override branch and then fail the `if effort:` test below, sending nothing and
+    silently inheriting the provider default — the exact regression `REASONING_OFF` exists to
+    prevent, with no error to notice. A non-positive `max_tokens` would reach the gateway as
+    a 400 or an empty completion. ValueError, not assert: asserts vanish under `-O`.
     """
+    if reasoning_effort is not None and not reasoning_effort:
+        raise ValueError(
+            f"reasoning_effort must be a LiteLLM effort value or None, not {reasoning_effort!r} — "
+            f"pass REASONING_OFF ({REASONING_OFF!r}) to turn thinking off."
+        )
+    if max_tokens is not None and max_tokens <= 0:
+        raise ValueError(f"max_tokens must be positive when given, not {max_tokens!r}")
     # Gateway-aware subclass: preserves reasoning_content that base ChatOpenAI (>=1.2) drops.
     ChatOpenAI = _gateway_chat_openai_cls()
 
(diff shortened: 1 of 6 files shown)


packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -738,6 +764,48 @@ def get_default_fast_model() -> ModelType | None:
     return _default_alias_for("chat:low") or get_default_model()
 
 
+# One sentence per tool call, one risk score, one condition verdict — generous for a batch
+# of a few summaries, low enough that a model ignoring "ONE short sentence" cannot stretch
+# the call. Utility work must never inherit the reasoning tiers' 8k-32k ceiling. Same value
+# as `llm_gateway.gateway_chat`'s default for the same reason (see REASONING_OFF on why that
+# module keeps its own copy). A caller whose answer outgrows it should raise it explicitly
+# rather than discover the truncation: with thinking off, the cap only binds real output.
+_FAST_MODEL_MAX_TOKENS = 1024
```
alloy-cicd: **Batch truncation: now diagnosed, still not prevented.** 74ae1291 makes a truncated batch visible in the log (`finish_reason=length`, naming the cap) — that's the right first half. But this 1024-token cap still applies to the *whole* batched structured-output call in `summarize_action_requests`, and the batch is unbounded: `attach_summaries` (via `conditional_hitl.py` and `graph_utils.py`) passes every non-self-evident pending tool call in the turn. The pre-PR code had the gateway default (~4k) of headroom here.

The new comment says "a caller whose answer outgrows it should raise it explicitly" — but the summarizer is exactly such a caller and doesn't: a turn interrupting with ~15–20 parallel tool calls (fewer in a token-heavy language) exceeds 1024 output tokens and **every** card falls back to raw args, precisely the large approval batches where the prose matters most. Now it's at least logged correctly, but the user experience regresses relative to pre-PR.

Suggest the summarizer pass a batch-aware cap, e.g. `create_fast_model(alias, max_tokens=max(_FAST_MODEL_MAX_TOKENS, N * len(calls)))` — `create_fast_model` already supports the override.

*Verify later:* the summarizer call site scales `max_tokens` with `len(calls)` (or bounds the batch).

packages/agent-common/agent_common/core/tool_call_summarizer.py
```diff
@@ -101,11 +145,15 @@ async def summarize_action_requests(
     if not calls:
         return []
 
-    from agent_common.core.model_factory import create_model, get_default_fast_model, require_default_model
+    from agent_common.core.model_factory import create_fast_model, get_default_fast_model, require_default_model
 
     try:
-        model = create_model(get_default_fast_model() or require_default_model(), streaming=False)
-        structured_model = model.with_structured_output(ToolCallSummaries)
+        model = create_fast_model(get_default_fast_model() or require_default_model())
```
alloy-cicd: **Model resolution is outside the timed section, and it can block the event loop.** The `_SUMMARY_TIMEOUT_SECONDS` comment promises a "hard ceiling on how long a card may wait", but `create_fast_model(get_default_fast_model() or require_default_model())` runs before `asyncio.wait_for` starts counting — and both resolvers call the synchronous `_refresh_if_stale`, which on a cold or expired-and-cold cache blocks the event loop on urllib (2 s fetch timeout) or `cond.wait_for` up to `_COLD_WAIT=3 s` (model_factory.py ~463–502).

So the first HITL interrupt after process start (or after a gateway blip poisons the defaults cache) can spend ~3–5 s in synchronous resolution *plus* the 5 s ainvoke budget — ~10 s card latency despite the 5 s ceiling — and because the block is synchronous, it stalls every other coroutine in the process, not just this card.

Fix: hoist the model construction into the timed section (wrap the construction + ainvoke in an inner coroutine and `wait_for` that), so a cold-cache stall consumes the summary budget and falls back to raw args.

*Verify later:* model construction/resolution is inside the `wait_for` scope.

packages/agent-common/agent_common/core/tool_call_summarizer.py
```diff
@@ -123,12 +171,34 @@ async def summarize_action_requests(
...
+
+        # Inside the try on purpose: the envelope's shape is the provider's business, and
+        # this is a best-effort path whose caller is mid-``interrupt()``. Anything
+        # unexpected here degrades to raw args like every other failure, rather than
+        # raising into the HITL request.
+        result: ToolCallSummaries | None = envelope.get("parsed")
+        if result is None:
+            _log_unusable_reply(envelope, len(calls))
+            return None
+    except TimeoutError:
```
alloy-cicd, line 238: **This branch catches more than the `wait_for` budget.** Since Python 3.10+, builtin `TimeoutError` is also `socket.timeout` — so a transport-level timeout raised *inside* the `ainvoke` pipeline (e.g. a socket op timing out at 0.3 s) lands here and gets logged as "exceeded 5.0s" with no traceback, even though the budget never elapsed. That points an on-call engineer at gateway latency when the actual problem is a client-side transport error, and it downgrades what used to be a `logger.exception` with a stack to a bare warning.

Cheap hardening: measure elapsed time around the `wait_for` and include it in the log (or only phrase the message as budget expiry when `elapsed >= _SUMMARY_TIMEOUT_SECONDS`).

*Verify later:* the timeout log reports actual elapsed time (or distinguishes budget expiry from transport timeout).

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -325,15 +351,15 @@ def create_model(
         callbacks = [cb for cb in callbacks if type(cb).__name__ != "CostTrackingCallback"] or None
 
     model_kwargs: dict = {}
-    effort = get_reasoning_effort(thinking_level, model_type)
+    effort = reasoning_effort if reasoning_effort is not None else get_reasoning_effort(thinking_level, model_type)
```
alloy-cicd, line 365: **Falsy-but-not-None overrides silently degrade.** Two sharp edges in the new public parameters:

- `reasoning_effort=""` (e.g. wired from `os.getenv(..., "")`) is not `None`, so it takes the override branch and skips `get_reasoning_effort` — then fails `if effort:` below, so *nothing* is sent and the call silently inherits the provider default (dynamic thinking): the exact regression class this PR eliminates, with no error.
- `max_tokens=0` passes the `is not None` guard at line 362 and is forwarded as `max_tokens=0` to the gateway, producing a 400 or an empty completion instead of falling back to the effort-derived/gateway default.

A guard at the top (`assert reasoning_effort != ""` / `assert max_tokens is None or max_tokens > 0`, or normalizing `"" → None`) would fail these at the boundary instead of at request time.

*Verify later:* empty-string effort and non-positive max_tokens are rejected or normalized.

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -738,6 +764,48 @@ def get_default_fast_model() -> ModelType | None:
...
+
+def create_fast_model(
+    model_type: ModelType,
+    *,
+    max_tokens: int = _FAST_MODEL_MAX_TOKENS,
+    callbacks: list | None = None,
+) -> BaseChatModel:
+    """A cheap-tier model configured for LOW LATENCY: thinking off, no streaming, short cap.
+
+    For the utility calls that ride `get_default_fast_model()` — tool-call summaries, risk
```
alloy-cicd: **The caller list overclaims — only the summarizer was switched.** Of the utility callers this docstring names:

- **Risk scoring** — `tool_risk_scorer.py:298` still calls `create_model(get_default_fast_model() or require_default_model(), streaming=False)`: it gates the approval card and keeps inheriting provider-default dynamic thinking (the measured 2.7–25.5 s).
- **Reply classification** (not named, but the closest sibling) — `hitl_resume.py:246` has the identical bare pattern; it sits between the user's typed "yes" and the resume, with no timeout guard at all.
- **Watch-condition checks** live in console-backend's `gateway_chat` (already `reasoning_effort="none"` per #209), not on `get_default_fast_model()`.
- **Chunk descriptions** ride `get_default_indexing_model()` (`indexing_store.py:431`), also unconverted (and per-chunk, so the latency multiplies by corpus size).

Suggest either converting `tool_risk_scorer.py:298` and `hitl_resume.py:246` in this PR (same package, same HITL critical path this PR is about), or trimming the docstring to what's actually covered and tracking the rest. (orchestrator-agent has three more hand-rolled copies — `toolset_selector.py:343`, `dynamic_tool_dispatch.py:1597`, `file_analyzer.py:370` — fine as a follow-up.)

*Verify later:* the two agent-common sites call `create_fast_model`, or the docstring no longer claims them.

packages/agent-common/tests/test_tool_call_summarizer.py
```diff
@@ -173,3 +177,98 @@ async def run(_: object):
...
+
+        with (
+            patch("agent_common.core.model_factory.create_fast_model", return_value=_SlowModel()),
+            patch("agent_common.core.model_factory.get_default_fast_model", return_value="fast"),
+            patch.object(tcs, "_SUMMARY_TIMEOUT_SECONDS", 0.01),
+        ):
+            assert await tcs.summarize_action_requests([("ls", {"path": "/x"}, "list files")]) is None
+
+    @pytest.mark.asyncio
+    async def test_a_timeout_leaves_the_action_request_untouched(self):
```
alloy-cicd: **This test never exercises the timeout path.** It patches `summarize_action_requests` itself to `AsyncMock(return_value=None)`, so it only re-tests the pre-existing None-fallback branch of `attach_summaries` — byte-similar to `test_a_failed_summary_leaves_the_request_untouched` above. If the timeout handler were changed to return `[]` or partial results (or to mutate args before timing out), this test would still pass while a regression stamps the half-populated cards its own docstring warns about.

Either delete it (the slow-model test already proves timeout → `None`, and the older test proves `None` → untouched) or make it real: reuse the `_SlowModel` fixture so the actual timeout path feeds `attach_summaries`.

*Verify later:* the test is removed or exercises `summarize_action_requests` for real.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (31656f401f72d76f1bd781c2250acb003ae5e609..74ae1291a3cf8bf954560b034029917f436c97c3)

**Note on round 1:** the round-1 review was analyzed but its submission to GitHub failed mid-flight during a GitHub API outage (connection resets) and never landed — there are no prior review threads on this PR. The commit 74ae1291 was therefore made independently of it. This review carries the round-1 findings that are still valid at the current head, plus the round-2 delta check.

**Ledger of round-1 findings (never delivered — dispositions verified against 74ae1291):**
1. `_FAST_MODEL_MAX_TOKENS=1024` caps the whole unbounded summary batch — **partially addressed independently**: 74ae1291 adds `include_raw=True` + `_log_unusable_reply`, so a truncated batch is now correctly diagnosed in the log (`finish_reason=length`, naming the cap) instead of surfacing as a generic parse failure. The cap itself is still batch-blind, so a large batch still loses *all* summaries — inline comment below.
2. Model resolution outside the `wait_for` budget — **still-open** (inline).
3. `except TimeoutError` also catches transport timeouts and mislabels them as budget expiry — **still-open** (inline).
4. Falsy-but-not-None overrides (`reasoning_effort=""`, `max_tokens=0`) silently degrade — **still-open** (inline).
5. `create_fast_model` docstring overclaims its callers; `tool_risk_scorer.py:298` and `hitl_resume.py:246` still inherit provider-default thinking — **still-open** (inline).
6. `test_a_timeout_leaves_the_action_request_untouched` never exercises the timeout path — **still-open** (inline).

**Round-2 delta review (31656f40..74ae1291): no new issues.** The envelope unwrapping is correctly placed inside the existing `try` (a malformed envelope degrades to raw args rather than raising into the `interrupt()`), `_log_unusable_reply` handles a missing/None `raw` safely, the `finish_reason` lookup matches the OpenAI-compatible gateway's `response_metadata`, the new tests import `types` correctly and pin both log branches plus the missing-envelope case, and the REASONING_OFF/`gateway_chat` cross-references resolve the round-1 duplication note. Good fix.

Human reviewer decides — submitted as comment only.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in eb80e888: `_summary_token_budget(call_count)` scales from the (now public) `FAST_MODEL_MAX_TOKENS` floor at 160 tokens per pending call, and the summarizer passes it via the `max_tokens` override. Small batches keep the fast-model default; a 20-call batch gets 3200. Agreed on the direction — the flat cap failed backwards, hitting hardest exactly where the prose matters most. Four tests cover it, including one asserting the scaled budget actually reaches the model (the scaling is worthless if the call site drops it).

packages/agent-common/agent_common/core/tool_call_summarizer.py
aartaria: Fixed in eb80e888, but not the suggested way — the suggested fix would not have worked. Wrapping the construction in `wait_for` bounds nothing here: `_refresh_if_stale` runs `_run()` in the calling thread on a cold cache (2 s urllib) or blocks a cold waiter on `cond.wait_for` up to `_COLD_WAIT` (3 s), and `asyncio.wait_for` can only cancel at an await point — synchronous code never yields, so the card would still have waited and the loop would still have stalled. What actually fixes both halves is `asyncio.to_thread`: resolution+construction now run in `_build` off the loop, wrapped in `_invoke` so one awaitable covers construction and request, and that is what `wait_for` bounds. The thread still runs to completion after a timeout (nothing can stop it) but finishes into a discarded result instead of holding the card. Diagnosis was right and worth catching — the "hard ceiling" comment was overclaiming.

packages/agent-common/agent_common/core/tool_call_summarizer.py
aartaria: Fixed in eb80e888: the branch measures `time.monotonic()` around the call and logs actual elapsed alongside the budget, tagging it either `budget expired` or `transport timeout, not the budget` on the `elapsed >= budget` comparison. Confirmed the premise — `socket.timeout` has been an alias of builtin `TimeoutError` since 3.10, so a socket-level timeout inside the request does land here.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in eb80e888: `create_model` now raises `ValueError` for an empty `reasoning_effort` (naming `REASONING_OFF` in the message) and for a non-positive `max_tokens`. Went with exceptions rather than the suggested asserts, since asserts vanish under `-O` and these are exactly the cases that must not pass silently. Both are new parameters with no existing callers, so raising is safe. Two tests cover them.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in eb80e888 by trimming the docstring, not by converting the callers. It now names the one converted caller (`tool_call_summarizer`), lists `tool_risk_scorer` / `hitl_resume` / `indexing_store` as still inheriting provider-default thinking with the reason they were left (converting them is a quality judgement — reasoning may be buying something real in a score or a classification — not a mechanical swap), and drops the watch-condition claim, which you are right about: those run through console-backend `gateway_chat`, thinking-off since #209. Keeping them out of this PR deliberately, per the scope note in the description; happy to take the two agent-common sites in a follow-up if we decide the quality cost is nil.

packages/agent-common/tests/test_tool_call_summarizer.py
aartaria: Fixed in eb80e888: the test now drives the real `summarize_action_requests` through the shared `_slow_model()` fixture and `attach_summaries`, so the actual timeout path feeds the fallback — renamed `test_a_timed_out_batch_leaves_the_action_request_untouched`. Correct catch: as written it stubbed out the very function under test and duplicated the older None-fallback test, so a handler that started returning `[]` or mutating args before giving up would have gone unnoticed.

Issue #217, opened 2026-09-08
PTC: two concurrent eval calls on one thread collide — per-thread REPL slot, turn collector and HITL batching all assume one eval per step


aartaria:
### Summary

Nothing guarantees that a model step emits at most one `eval` tool call, yet every piece of per-`eval` bookkeeping on the PTC path is keyed by `thread_id` alone. If two `eval` calls land in one assistant message the tool node runs them concurrently and they stomp each other. Surfaced by the automatic review of #215 (round 3); pre-existing, widened there by moving the inner-call counts onto the same slot.

### Where the collision happens

Bottom-up:

1. **`langchain_quickjs` REPL slot** — `sync_eval`/`async_eval` resolve `thread_id` and call `middleware._repl_for_eval(thread_id)`; in `mode="call"` the `finally` does `reset_repl(thread_id)`. Two concurrent evals share one sandbox and the first to finish resets it under the other.
2. **`ptc_guard._PTC_TURNS`** — `begin_ptc_turn(thread_id)` *replaces* the thread's turn and `end_ptc_turn` pops it unconditionally. Eval B's `begin` swaps the turn out from under eval A: A's inner calls bump B's `inner_call_counts`, share B's `results`/`decisions`/`pending`, and whichever `finally` runs first leaves the other with `turn is None` (repeat guard inert) and a count write-back computed against a stale seed — last writer wins, since the state channel replaces the dict.
3. **HITL batching** — `take_ptc_pending(thread_id)` drains one collector for the thread, so pending approvals from two evals would be batched into whichever `interrupt()` fires first.

(2) and (3) could be re-keyed by `(thread_id, tool_call_id)` — the guard does receive the outer `eval` runtime, so a per-call identity is available on both sides — but that only fixes the bookkeeping while (1) still collides underneath. The fix has to be a **single-`eval`-per-step guarantee**, or an upstream per-call REPL slot.

### Options

- **Constrain the model.** Bind `eval` so parallel tool calls are disabled when it is present (provider `parallel_tool_calls=False` where supported), or have the model-boundary hook split a multi-`eval` message into sequential steps. Cheapest and closes all three layers.
- **Serialize at the tool node.** A per-thread lock around `eval` execution in `_PTCToleranceCodeInterpreterMiddleware.awrap_tool_call`. Keeps concurrency for other tools; the two evals then run back-to-back with correct turns and a correct REPL reset between them.
[...]

PR #224, opened 2026-09-09
fix(hitl): give every approval ask its own id, not a hash of tool+args

### LINKED ISSUE #223: HITL: a repeated approved call is never re-asked, and the turn hangs forever
## Summary

Approving a risk-gated tool call inside `eval` and then letting the program call it again leaves the turn parked on "Working…" forever. The second approval card never renders, so no decision can be submitted and the run waits on a human who was never asked.

Reproduced 3x in the console. Pre-existing on `main`.

## Reproduction

In the console chat, with `CODE_INTERPRETER_PTC=1` and a tool whose risk score is above the approval threshold (`ls` is scored Critical/100%):

> Make exactly 6 separate eval calls, one after another, never two in the same eval. In eval call number N the code must be exactly these two lines: `// attempt N` then `await tools.ls({path: '/'})`.

The first card renders and can be approved. `ls` then executes for real. The model issues the second `eval` call, the backend correctly raises a fresh interrupt — and nothing appears in the UI.

## Cause

One string is doing two incompatible jobs.

For calls gated inside the code interpreter, the wire `_call_id` was `p.call_key` = `sha256(tool + sorted args)` (`agent_common/middleware/ptc_guard.py`). That key exists so the guard can recognise *the same call* across an `eval` re-run and replay the human's decision instead of asking twice — content-derivation is exactly right for that job.

Clients need the opposite property: one id per *question asked*. `TurnSession.handle` (`packages/embed-sdk/src/transport/turn-session.ts`) suppresses an `input-required` prompt whose ids it has already answered, to absorb a snapshot redelivered by a subscribe racing a resume. So a genuine second ask, arriving under the id just answered, is classified as a replay and dropped.

The backend is correct throughout — every `input-required` emission carried a complete `action_requests` payload.

Contributing to it: `_call_id` was documented in **none** of the three runtime copies of the extension list, so its uniqueness properties were never stated and the SDK inferred the wrong one. Clients that only echo the field back (Slack, Google Chat, email) were unaffected, but they read the same undocumented field.

The normal (non-PTC) HITL path was never affected — it uses the model's `tool_call["id"]`, which is already unique per call.

## Fix

- The wire id becomes `ask_id(call_key, ask_scope, ask_round)`. Both scopes are required: a PTC turn spans one `eval` invocation and serves an identical call from `turn.results`, so two asks always mean two `eval` calls — which a per-turn counter alone cannot separate. `ask_scope` is the model's `eval` tool call id; `ask_round` separates successive interrupts within one `eval`. Both are replay-stable, so no checkpointed mapping is needed.
- `call_key` stays internal and content-derived, with its docstring stating what it is (a per-turn decision/result memo key) and what it is not (the durable per-user bypass policy).
- Decision matching accepts either the ask id or a bare `call_key`, so a client build that predates ask ids is still matched by id.
- Positional alignment is now used only when no usable ids arrive at all. A *partial* id match used to fall through to it silently, which could apply a decision to a call the human never saw.
- The `_call_id` invariant is documented where it is emitted and in the `a2a-extensions.json` registry, and `hitlDecisionTypes` joins the pinned vocabularies with both language copies asserted against it.

aartaria:
closes ringier-data/nannos#223

Approving a risk-gated call inside `eval` and then letting the program call it again parked the turn on "Working…" forever — the second approval card never rendered, so no decision could be submitted and the run waited on a human who was never asked. Pre-existing on `main`; reproduced 3x in the console.

## Cause

One string was doing two incompatible jobs. For calls gated inside the code interpreter the wire `_call_id` was `p.call_key` = `sha256(tool + sorted args)`, whose purpose is to let the guard recognise *the same call* across an `eval` re-run and replay the human's decision instead of asking twice — content-derivation is right for that. Clients need the opposite: one id per *question asked*. `TurnSession.handle` suppresses an `input-required` prompt whose ids it already answered (to absorb a snapshot redelivered by a subscribe racing a resume), so a genuine second ask under the id just answered was classified as a replay and dropped.

The backend was correct throughout — every `input-required` emission carried a complete `action_requests` payload. The normal (non-PTC) HITL path was never affected: it uses the model's `tool_call["id"]`, already unique per call.

Contributing to it: `_call_id` appeared in **none** of the three runtime copies of the extension list, so its uniqueness properties were never stated and the SDK inferred the wrong one.

## What changed

- **`ptc_guard.ask_id(call_key, ask_scope, ask_round)`** is now the wire id. Both scopes are needed: a PTC turn spans one `eval` invocation and serves an identical call from `turn.results`, so two asks always mean two `eval` calls — which a per-turn counter alone cannot separate (both would be round 0). `ask_scope` is the model's `eval` tool call id, `ask_round` separates successive interrupts within one `eval`. Both are replay-stable (the tool call id is checkpointed on the AI message and the node re-runs from the top), so **no checkpointed mapping is required**.
- `call_key` stays internal, with its docstring stating what it is — an ephemeral per-turn decision/result memo key — and what it is not: the durable per-user bypass policy (`context.tool_bypass_rules`), which is a different mechanism.
- Decision matching accepts the ask id **or** a bare `call_key`, so a client build predating ask ids is still matched by id rather than dropped to positional alignment.
- **Positional alignment is now used only when no usable ids arrive at all.** A *partial* id match previously fell through to it silently, which could pair a decision with a call the human never saw — approving one path while another executes.
- The `_call_id` invariant is documented where it is emitted (`orchestrator app/core/a2a_extensions.py`) and in the `a2a-extensions.json` registry, which is the one artifact all four clients' authors share. Also documented: `_risk_metadata`, the decision `id`, and the `bypass*` triple, none of which were in the docs. `_summary` is deliberately left unspecified — it may yet be computed asynchronously, and a contract promising it in the initial request would be wrong the moment it is.
- `hitlDecisionTypes` joins `clientActionKinds`/`activityLogKinds` as a pinned vocabulary, with both language copies (`hitl_resume.HITL_DECISION_TYPES`, `approval-codec.HITL_DECISION_TYPES`) asserted against it. It has teeth: the PTC path now logs an unknown decision type before failing it closed.

## Testing

- `test_second_eval_call_about_the_same_call_gets_a_distinct_ask_id` — end-to-end through a real `eval`: approve, repeat the call in a new `eval`, assert the two asks carry different ids and the same call key. Verified it **fails** with the old id restored (`assert 'safe_read:379f257c1f25d1ac' != 'safe_read:379f257c1f25d1ac'`).
- `turn-session-answered.test.ts` — three tests pinning the suppression rule from the client side: same ask suppressed, new ask rendered, and a new ask still rendered when the answered id is a bare call key (mid-deploy).
- `test_hitl_decision_type_conformance.py` + an added case in `extensions.test.ts` — the registry pins.
- Suites: agent-common 969 pass, orchestrator HITL/resume 115 pass, embed-sdk 423 pass, embed-sdk typecheck clean.

Verified in the running console: six sequential `eval` calls each raise their own card, each is approvable, and all six execute — previously card 1 was approved and the turn then hung.

Pre-existing failures untouched and unrelated, confirmed by reverting on the same tree: `test_foundry_cost_version.py` errors on a missing `foundry_sdk`, and `test_thinking_blocks_round_trip.py`'s 9 failures also fail on clean `main`. Separately, `npm run lint` in embed-sdk fails because `eslint` is not installed anywhere in the repo — pre-existing, left alone.

## Note for reviewers

⚠️ This branch and #215 (`bugfix/issue-211-ratelimit-loop`) both rewrite `_build_ptc_hitl_request` and `_apply_ptc_decisions`. Whichever merges second will need a manual conflict resolution. The two fixes are independent — this bug predates #215 and is not caused by it.

---

🤖 Authored by Claude Code (Opus 5) in an interactive session with @aartaria, who directed the investigation and reviewed the design.

## Checklist

The wire format of `_call_id` changes shape (it now carries a scope suffix). No client parses it — every consumer treats it as an opaque string and echoes it back — and the server accepts the old bare-`call_key` form for one deploy, so this is not a breaking change in practice.


packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1211,15 +1229,42 @@ def _apply_ptc_decisions(
...
-        use_by_id = bool(by_id) and all(p.call_key in by_id for p in pending)
+        matched: dict[str, dict[str, Any]] = {}
+        for p in pending:
+            found = by_id.get(ask_id(p.call_key, ask_scope, ask_round)) or by_id.get(p.call_key)
+            if found is not None:
+                matched[p.call_key] = found
+        # Positional zip ONLY when the client sent no usable ids at all. Falling back
+        # on a PARTIAL match would silently pair a decision with a different call than
+        # the human answered for — an approval landing on a call they never saw.
+        use_by_id = len(matched) == len(pending)
```
alloy-cicd: **Correctness — partial id match falls back to the positional zip it's meant to prevent.**

The executor deliberately produces *mixed* decision lists: `_decisions_for_interrupt` (executor.py:271-272) substitutes an id-less `{"type": "reject"}` for any action_request whose id it couldn't match. One stale/mangled id in an N-call batch therefore shrinks `matched`, flips `use_by_id` to False, and the whole batch — including decisions with *correct* ask ids — is zipped positionally at line 1256. `decisions` is in checkpointed action_request order while `pending` is in the re-run's registration order, which the comment above (and `ask_id`'s docstring) says can differ for parallel eval calls: a real approval can land on a call the user rejected or never saw, and the placeholder reject lands on the approved call.

The comment at 1243-1245 says falling back on a partial match is wrong — and the code then does it. Since every unmatched entry is already a reject by construction, the safe branch is: apply `matched` by id and reject the remainder, reserving the positional zip for the no-ids-at-all legacy case.

alloy-cicd (commented):

The per-ask id design is sound and the docs/tests pinning the registry vocabulary are welcome. Eight review angles ran and every candidate finding was adversarially verified against the code before posting; several plausible-sounding claims were dropped on verification (notably: the bare-`call_key` compat arm in `_apply_ptc_decisions` **is** reachable for pre-deploy checkpoints and works as documented, and an id-less eval tool call cannot actually reach an ask, so the `ask_scope or ""` degradation is not a live bug).

**What survived verification** — 7 inline comments:

- **2 correctness issues** in `graph_utils.py`: the positional fallback on a *partial* id match (the executor's id-less reject placeholders defeat `use_by_id` and can land an approval on the wrong call), and the thread-scoped pending list vs. per-eval-call `ask_scope` mismatch under parallel `eval` tool calls (regresses the id-matching the PR introduces).
- **1 contract inconsistency**: the newly documented "unknown decision type ⇒ rejection" rule holds only on the PTC path; the native path raises `ValueError` and kills the turn.
- **4 smaller items**: registry wording that contradicts the SDK's (id, kind) suppression rule; the zod envelope/decode not derived from the new `HITL_DECISION_TYPES`; the log-only vocabulary check; duplicate ask-id derivation with test-only defaults and a test gap (no test exercises the primary ask-id matching arm).

**Not inline** (files outside this diff / minor):
- The "unique per ask, not per (tool, args)" invariant is restated in full in **7 places** in this PR (registry, orchestrator + console `a2a_extensions.py`, `extensions.ts`, `demux.ts`, `turn-session.ts`, `ptc_guard.ask_id`, `graph_utils` inline comment). The registry and `ptc_guard.ask_id` earn their copies; consider trimming the rest to pointers before they drift.
- Heads-up: client-slack's multi-HITL widget base64-packs one `{id, name, detail, …}` per pending call into a single Slack button `value` (`taskResponseHandler.ts:863`) with no guard against Slack's 2000-char cap, and `_call_id` is not in its `HIDDEN_KEYS`, so the longer ask ids land in the payload twice. Pre-existing headroom problem (~4–8 pending calls), made slightly worse here — worth a follow-up truncation guard on the Slack side.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1097,6 +1098,14 @@ async def awrap_tool_call(self, request: Any, handler: Any) -> Any:
         if self._ptc_enabled and not self._ptc_tools_by_thread.get(thread_id):
             self._ptc_tools_by_thread[thread_id] = tuple(self._collect_ptc_tools(request))
         turn = begin_ptc_turn(thread_id)
+        # Identity of the questions this ``eval`` asks, as the client sees them (see
+        # ptc_guard.ask_id): the model's tool call id scopes them to this ``eval``
+        # invocation, the counter separates successive rounds within it. Replay-safe
+        # without any checkpointed map — the tool call id is checkpointed on the AI
+        # message, and the node re-runs from the top so the counter is rebuilt in
+        # lockstep with the interrupts it names.
+        ask_scope = str((getattr(request, "tool_call", None) or {}).get("id") or "")
```
alloy-cicd: **Correctness — thread-scoped pending list vs. per-eval-call `ask_scope` under parallel `eval` calls.**

PTC pending state is keyed by thread only (`_PTC_TURNS[thread_id]`; `take_ptc_pending` drains the whole list regardless of which eval registered each item), but `ask_scope` is *this one wrapper's* tool_call id. If the model emits two parallel `eval` calls c1/c2 (a supported scenario — see the Gemini parallel-tool-call tests, and `begin_ptc_turn` at line 1100 already clobbers the shared state per wrapper), whichever wrapper drains the mixed list stamps every ask `…@c1:0`; on the resume replay the *other* wrapper can drain those items, compute `ask_id(key, "c2", 0)`, miss both the scoped and bare-key lookups (the client echoed `…@c1:0`), and fall to the positional zip — or die on the count-mismatch ValueError. Pre-PR, the scope-free `call_key` matched regardless of which wrapper applied it, so this specifically regresses the id-matching arm. Consider keying pending items by the eval tool-call id that registered them rather than per thread.

Nit, same line: `tool_call` is already bound at line 1079 in this scope — `str(tool_call.get("id") or "")` suffices, and a debug log when the id is empty would make the "scope must be unique per invocation" invariant loud instead of silent.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1211,15 +1229,42 @@ def _apply_ptc_decisions(
...
+        # to the WRONG call (e.g. approve `/memories` lands on `/`). The client echoes
+        # the ask id we stamped (``ask_id(call_key, ask_scope, ask_round)``); a client build that
+        # predates ask ids echoes the bare ``call_key``, which is still unambiguous
+        # because a call key appears at most once in ``pending``. Accept either, so a
+        # mid-deploy client is matched by id rather than dropped to positional.
         by_id = {d["id"]: d for d in decisions if isinstance(d, dict) and "id" in d}
-        use_by_id = bool(by_id) and all(p.call_key in by_id for p in pending)
+        matched: dict[str, dict[str, Any]] = {}
+        for p in pending:
+            found = by_id.get(ask_id(p.call_key, ask_scope, ask_round)) or by_id.get(p.call_key)
```
alloy-cicd, line 1263: **Maintainability — the ask id is derived twice, the new params default to the unsafe shape, and no test covers the primary matching arm.**

1. This expression must stay byte-identical to the stamping in `_build_ptc_hitl_request` (line 1181); if they ever diverge, matching silently degrades to positional with only a log warning. Deriving the id once (e.g. stamped on the pending record, or returned by the builder) removes the coupling.
2. The `ask_scope=""` / `ask_round=0` defaults on both helpers exist only for two pre-existing unit tests (test_ptc_guard.py:482/501); production always passes both. A defaulted call silently produces the scopeless `key@:0` shape this PR removes — make them required keyword-only and update the two tests.
3. Test gap: the only unit test of this matcher (`test_apply_ptc_decisions_matches_by_id_not_position`, test_ptc_guard.py:455) passes bare call keys, so it exercises the legacy `by_id.get(p.call_key)` fallback arm — and the new e2e test resumes with an id-less `{"type": "approve"}` and only asserts the two ask ids differ. Nothing asserts that a decision echoing the full ask id actually matches. Worth one test on the primary arm before the fallback is someday removed.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1211,15 +1229,42 @@ def _apply_ptc_decisions(
...
+                "falling back to positional alignment",
+                len(pending) - len(matched),
+                len(pending),
+            )
 
         for i, p in enumerate(pending):
-            decision = by_id[p.call_key] if use_by_id else decisions[i]
+            decision = matched[p.call_key] if use_by_id else decisions[i]
             dtype = decision.get("type")
+            if dtype not in HITL_DECISION_TYPES:
```
alloy-cicd: **This check is log-only, and membership doesn't mean "honored".**

The `if dtype not in HITL_DECISION_TYPES` block changes no control flow — the `if dtype == "approve" … else` below already rejects everything non-approve. And `edit` *is* in `HITL_DECISION_TYPES`, so it never triggers this warning yet is also rejected by the same else branch. A future maintainer adding a type to the registry + constant will see the conformance tests pass while the server still rejects it. Suggest either folding this into the else branch (warn on anything that isn't `approve`/`reject`, noting PTC can't honor `edit`) or moving the vocabulary normalization to the shared funnel (see the comment on hitl_resume.py).

Minor hardening while here: a malformed list like `{"decisions": ["approve"]}` passes untouched through `_extract_hitl_decisions` and `structural_decisions` and reaches the positional path as a string — `decision.get` (line 1257, pre-existing) raises AttributeError, and this new frozenset test raises TypeError on an unhashable `type` value. Line 1285 already guards with `isinstance(decision, dict)`; the same guard here would make unreadable decisions reject instead of crash the eval node.

packages/agent-common/agent_common/core/hitl_resume.py
```diff
@@ -86,8 +86,24 @@ def authorization_verdict(payload: Any) -> tuple[str | None, str]:
     return None, message
 
 
+#: Decision types a client may send back on the human-in-the-loop extension.
+#: Pinned to the repo-root ``a2a-extensions.json`` registry (see
+#: ``tests/test_hitl_decision_type_conformance.py``), whose copy in
+#: ``embed-sdk approval-codec.ts`` must agree. Anything outside this set is
+#: treated as a rejection rather than trusted — an unreadable answer must never
```
alloy-cicd: **The documented fail-closed rule only holds on the PTC path.**

"Anything outside this set is treated as a rejection" (also stated in a2a-extensions.json) is enforced solely in `_apply_ptc_decisions`. On the native path, `structural_decisions` returns client-sent decisions untouched and `conditional_hitl.py:392/655` hand them to langchain's `HumanInTheLoopMiddleware._process_decision`, which **raises `ValueError`** for a type it doesn't know — and also for a known type not in that tool's `allowed_decisions` (e.g. `edit` sent to a default approve/reject gate). A client-sent `{"type": "banana"}` to a native-gated tool kills the turn with a stack trace instead of the documented safe rejection. Since both paths funnel through `decisions_from_resume`/`structural_decisions`, normalizing unknown types to `{"type": "reject", "message": …}` here would establish the rule once for both paths (and would let the PTC-local check be simplified).

a2a-extensions.json
```diff
@@ -20,5 +20,12 @@
...
   "activityLogKinds": [
     "note"
+  ],
+  "_hitlCallIdComment": "Identity rule of the human-in-the-loop extension, stated here because every client depends on it and none can derive it. Each action_request carries args._call_id identifying ONE ASK: unique per question, NOT per (tool, args). Asking about the same tool with the same arguments twice \u2014 a program repeating a call in a later step \u2014 yields two DIFFERENT ids. Clients echo it back as the decision's id, and a client that suppresses prompts it has already answered (embed-sdk TurnSession, absorbing a snapshot redelivered by a subscribe racing a resume) must key that on this id alone. Deriving an identity from the tool name and args instead makes a genuine second ask look like a replay: the card is dropped, no decision can be submitted, and the turn hangs forever \u2014 the failure this rule prevents. Producers: agent_common ptc_guard.ask_id for calls gated inside the code interpreter, the model's tool_call id on the normal path (conditional_hitl). The server's own per-turn memo key IS content-derived (ptc_guard._call_key) and deliberately never ships.",
```
alloy-cicd: **"must key that on this id alone" contradicts the SDK's own suppression rule.**

`demux.ts` (comment retained in this PR) documents that a risk-gated `client_action` approval and its round-trip request deliberately **share one `_call_id`**, which is why `turn-session.ts` keys the answered-map on (id, kind) — keying on the id alone was precisely a dropped-card/parked-turn bug there. A future client (client-slack and client-google-chat read `args._call_id` today but don't yet suppress) implementing this sentence verbatim re-creates it. Suggest: "…must key that on this id (plus the prompt kind, where the client also negotiates the client-action extension — see embed-sdk `demux.ts` `approvalPrompt`), never on an identity derived from tool name and args."

packages/embed-sdk/src/transport/approval-codec.ts
```diff
@@ -41,10 +41,19 @@ export function wireCallId(partId: string): string {
...
 
-/** A single HITL decision as the backend consumes it (aligned by `id` = `_call_id`). */
+/**
+ * Decision types the backend accepts on the human-in-the-loop extension. Pinned
+ * to the repo-root `a2a-extensions.json` registry (see `extensions.test.ts`),
+ * whose copy in agent-common `core/hitl_resume.py` HITL_DECISION_TYPES must
+ * agree — the server treats an unknown type as a rejection, so a type added on
+ * one side only silently blocks the call it was meant to allow.
+ */
+export const HITL_DECISION_TYPES = ['approve', 'reject', 'edit'] as const;
```
alloy-cicd, line 51: **Derive the envelope schema and decode mapping from this constant, or the pin doesn't cover the codec.**

`envelopeSchema` below (line 74) restates the literals — `z.enum(HITL_DECISION_TYPES)` works since this is an `as const` readonly tuple — and `decodeApproval` (line 138) hardcodes the mapping a third time (`e.type === 'edit' ? 'edit' : 'reject'`). The new conformance test pins only this exported const, so a 4th registry type would pass both conformance tests while `safeParse` failure silently degrades the decision to a plain approve/reject and line 138 maps it to `reject`. Deriving both from the const makes the registry pin actually reach the wire codec.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Fixed in 794d2375: position is now used only when the client sends no ids at all; with any id present, unmatched calls are rejected via `_UNMATCHED_DECISION` (with a message the model can act on) rather than borrowing a neighbour by position. Confirmed your reading of `executor._decisions_for_interrupt:272` — the id-less `{"type": "reject"}` padding is exactly what made a partial match reachable. Regression test added: `test_apply_ptc_decisions_rejects_unmatched_instead_of_going_positional`.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Fixed in 794d2375, though not by keying pending per eval call — the guard runs on the sandbox bridge thread and only has the thread id, so it cannot look a turn up by tool call id without plumbing that through the bridge (that is #217 territory). Instead the scope now lives on `_PTCTurnState.ask_scope`, seeded at `begin_ptc_turn`, so stamping and matching read the same value whichever wrapper drains the mixed list. That removes the race you identified; parallel `eval` calls on one thread remain broken because they clobber the shared turn state, which is #217 and is now called out in the field docstring rather than papered over. Nit applied: reuses the `tool_call` already bound above, plus a debug log when the id is empty.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Partly applied in 794d2375. Added the missing coverage — `test_apply_ptc_decisions_matches_scoped_ask_ids` exercises the primary arm with a non-zero `ask_round` and decisions ordered opposite to `pending`; you were right that every existing matcher test hit only the bare-`call_key` fallback. I kept the parameter defaults: `_build_ptc_hitl_request` and `_apply_ptc_decisions` are private statics with one production caller each, both of which now pass `turn.ask_scope` explicitly, and the defaults keep the unit tests readable without a fixture. Left the two `ask_id` derivations as they are — they sit in different phases (stamping vs matching) and hoisting a shared local across the interrupt boundary would obscure that they must agree by construction rather than by variable reuse.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Fixed in 794d2375. The logging moved into the branch where the block actually happens and now tells the two cases apart: `edit` logs at info as a known type this path deliberately cannot honor, an unrecognised type logs at warning as a client speaking a vocabulary this build lacks. Also hardened the non-dict case you flagged — it would have raised `AttributeError` on `.get` and taken the turn with it; it now becomes an explicit rejection with a readable message.

packages/agent-common/agent_common/core/hitl_resume.py
aartaria: Doc corrected in 794d2375, normalisation deliberately not done here. The docstring now states both behaviours explicitly — PTC blocks and logs, the native path raises `ValueError` in langchain`s `_process_decision` — and says that both refuse to execute (the safety property) but only one leaves the agent able to explain itself. You are right that the shared funnel is where they should converge, but changing native-path behaviour is a behavioural change to a path this PR does not otherwise touch, so I have left it as a noted gap rather than growing the diff. Happy to file it as a follow-up if you would rather it not live only in a docstring.

a2a-extensions.json
aartaria: Fixed in 794d2375, in both places the claim appeared. The registry now says the suppression must key on this id, parenthesising that the embed SDK pairs it with the prompt kind because a risk-gated `client_action` call and the round-trip request that follows it share one id. The orchestrator docstring says the same and draws the actual line: pair the id with whatever you like — just not with the call`s own content.

packages/embed-sdk/src/transport/approval-codec.ts
aartaria: Fixed in 794d2375 for the schema: `envelopeSchema.type` is now `z.enum(HITL_DECISION_TYPES)`, so a vocabulary that drifts from the registry can no longer pass the conformance test while the wire codec silently drops the new type. Left `decodeApproval`s `e.type === "edit" ? "edit" : "reject"` as it is: that line is not restating the vocabulary but narrowing on the AI SDK`s approved/not-approved split, and collapsing an unknown non-approve type to `reject` is the fail-closed behaviour we want client-side. Worth revisiting if a fourth type ever lands, which the pin will now make visible.

PR #234, opened 2026-09-11
fix(agent-common): stop two eval calls in one step from colliding

### LINKED ISSUE #217: PTC: two concurrent eval calls on one thread collide — per-thread REPL slot, turn collector and HITL batching all assume one eval per step

aartaria:
closes ringier-data/nannos#217

## What was broken

Nothing guarantees a model step emits at most one `eval` tool call, and `ToolNode` runs the calls of one assistant message concurrently — but every piece of per-`eval` bookkeeping on the PTC path is keyed by `thread_id` alone. Two `eval` calls in one message:

1. **shared a QuickJS context** — `_repl_for_eval(thread_id)` hands both the same REPL and, in `mode="call"`, the first to finish runs `reset_repl` in its `finally`, closing the context the other is still executing in;
2. **shared one `_PTC_TURNS` entry** — `begin_ptc_turn` *replaces* the thread's turn, so the second call swapped the first's out from under it;
3. **shared one HITL collector** — `take_ptc_pending` drains per thread, so both evals' approvals landed in whichever `interrupt()` fired first.

(1) surfaces as `ValueError: already closed` (or `InvalidHandleError`, depending on race timing) raised **out of the tool node** — the whole agent turn dies. In production that is a user-visible hard failure of the conversation, not a degraded tool result.

## The fix

**Serialize, don't re-key.** `ptc_guard.serialized_eval` holds a per-`thread_id`
mutex across the whole `begin_ptc_turn` → `end_ptc_turn` span in `awrap_tool_call`.
It closes all three layers at once, which re-keying the two collectors by
`(thread_id, tool_call_id)` cannot — the REPL still collides underneath — and unlike
`parallel_tool_calls=False` it does not disable parallel tool calls for *every* tool.
Only the `eval` path enters the gate.

It holds a `threading.Lock`, not an `asyncio.Lock`, keyed by `thread_id` alone — the
same key the protected resources use. Per-event-loop keying is not enough: two
parallel `task` dispatches of the same sub-agent share one `thread_id`
(`{context_id}::{checkpoint_ns}`) but reach `eval` via `LocalA2ARunnable.invoke` →
`asyncio.run`, a fresh loop each. The lock is acquired non-blockingly and polled
between awaits rather than blocking the loop or going through `asyncio.to_thread`,
which would park an executor worker per waiter and can deadlock against the
summarizer's own `to_thread` call.

**An incremental write-back.** With the sandbox collision gone, both eval tasks write
the `tool_call_history` channel in one superstep. The issue predicted "last writer
wins"; the reality is worse — LangGraph's default `LastValue` **raises**
`InvalidUpdateError`, killing the turn. The eval path now emits a tagged *delta*
(`history_delta`: the hashes it appended plus the window to apply after) and
`merge_tool_call_history` appends it. Every other writer keeps emitting whole values
and keeps plain replace semantics, so the model-boundary path is untouched by
construction rather than by case analysis.

> ⚠️ The reducer must be the **last** entry in the `Annotated` metadata — LangGraph
> detects it as `callable(metadata[-1])`, so placing it first is silently ignored,
> with no error, returning prod to the exact `InvalidUpdateError` this fixes. Both
> schemas now share one exported `ToolCallHistory` alias so the rule lives in one
> place.

**Approvals are never hidden.** Serializing means each eval raises its *own*
interrupt, so a turn can end with several pending — and the orchestrator rendered
only `interrupts[-1]` while the resume path replicates a blanket approve across all
of them. Approving the card you saw authorised the call you did not.
`AgentStreamResponse.interrupt_value` now folds co-pending approval asks into one
card (each request keeps its `_call_id`, so decisions still route back correctly),
including when a non-approval pause is pending alongside them. Auth and
client-action pauses are never folded in — they need their own round trip.

## Why not just upgrade upstream

Checked first, on the theory that deepagents/quickjs may have hit this already. It does not help:

- `langchain-quickjs` **0.3.7** (already inside our `<0.4` pin) does rework slot keying — `_resolve_thread_id` is gone, replaced by a state-carried `_quickjs_slot_id`. But it is still **one slot per agent run**, minted in `before_agent`: two evals in the same superstep read the same state, get the same slot. `reset_repl` is byte-for-byte unchanged and there is no locking anywhere in the middleware.
- It is also not a drop-in — 0.3.7 imports `TracePolicy`/`omit_payload` from `langchain.agents.middleware.types`, which our pinned langchain lacks (verified: `ImportError` on install).
- Decisively, **layers 2 and 3 are our own code** (`_PTC_TURNS`, `take_ptc_pending`). No upstream version can fix our thread-keyed dicts.
- `deepagents` 0.7.x is a major bump out of our `>=0.6.0,<0.7.0` pin and owns neither the REPL slot nor `ptc_guard` — orthogonal, and not something to bundle into a bugfix.

## Verification

- `tests/test_ptc_parallel_eval.py` — two `eval` calls in one scripted assistant
  message: both results correct with their own sandbox, one interrupt per eval
  (resumed id-keyed, as `executor._build_interrupt_resume_map` does for real), the
  write-back reflecting both, and cross-loop exclusion driven through two
  `asyncio.run` loops on separate threads. Plus reducer unit cases, including a
  regression test that drives 25 real updates out of `RepeatedToolCallMiddleware
  .evaluate` past a 10-entry window and asserts the history stays bounded.
- `orchestrator-agent/tests/test_responses.py` — the approval fold, `_call_id`
  routing, and the mixed auth+approvals case.
- **Every fix was checked by breaking it.** Without the lock both e2e tests fail with
  `already closed`; without the delta both raise `InvalidUpdateError`; the cross-loop
  test fails against per-loop keying; the bounded-history test fails against the
  reducer this PR originally carried.
- `tests/manual/qa_217_parallel_eval.py` — a manual QA harness with per-layer
  diagnostics and a `sequential` control, sharing its agent builder with the pytest
  module so it cannot rot. Reproduced on `origin/main` @ `1c422039` before fixing.
- agent-common **1035 passed**, orchestrator-agent **1036 passed**. The 9 failures in
  `tests/test_thinking_blocks_round_trip.py` are pre-existing — confirmed identical on
  the base commit with my changes stashed. `tests/test_foundry_cost_version.py` cannot
  be collected locally (`foundry_sdk` is an optional extra plain `uv sync` skips), also
  pre-existing.

### Review rounds

Two rounds of automatic review, 12 findings, all triaged on their threads. Round 1
caught a genuine regression in the first reducer (shape inference could not tell a
window trim from a concurrent extension, so history grew unbounded on the ordinary
single-writer path) and the HITL bypass; round 2 caught a narrower variant of the
bypass in the fix itself. One finding was declined with reasoning on the thread
(holding the gate across the 5s-bounded summarizer, because releasing early lets a
sibling eval install its turn and the unwinding `finally` delete it).

Known follow-up, not addressed here: `dynamic_tool_dispatch` still truncates a
sub-agent's own co-pending interrupts to `sub_interrupts[0]`. That truncates rather
than auto-approves — the second stays pending and surfaces next turn — and folding
only the display without folding the resume would be worse, so it wants its own change.

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no. `merge_tool_call_history` changes how concurrent updates to an existing private channel combine; single-writer steps behave exactly as before.

---

🤖 Authored by Claude Code (Opus 5) driving this workspace on behalf of @aartaria — the commit and this description are agent-written and human-reviewed. GitHub does not badge App attribution onto pull requests, so this note is the only signal.

### DIFF
--- a/packages/agent-common/agent_common/core/graph_utils.py
+++ b/packages/agent-common/agent_common/core/graph_utils.py
@@ -87,7 +87,10 @@ from agent_common.middleware.conversation_context_tools_middleware import (
     ConversationContextToolsMiddleware,
 )
 from agent_common.middleware.gateway_attribution_middleware import GatewayAttributionMiddleware
-from agent_common.middleware.loop_detection_middleware import RepeatedToolCallMiddleware
+from agent_common.middleware.loop_detection_middleware import (
+    RepeatedToolCallMiddleware,
+    merge_tool_call_history,
+)
 from agent_common.middleware.prompt_caching import LiteLLMPromptCachingMiddleware
 from agent_common.middleware.ptc_guard import (
     PTC_CODE_INTERPRETER_TOOL_NAME,
@@ -96,6 +99,7 @@ from agent_common.middleware.ptc_guard import (
     clear_ptc_pending,
     end_ptc_turn,
     resolve_ptc_thread_id,
+    serialized_eval,
     take_ptc_pending,
     wrap_tool_for_ptc,
 )
@@ -413,13 +417,16 @@ class _PTCExposureState(REPLState):
 
     ptc_exposed_tool_names: NotRequired[Annotated[list[str], PrivateStateAttr]]
 
-    #: The same channel ``LoopDetectionState`` declares (identical type, so the two
-    #: schemas merge). Declared here because this middleware writes it: the PTC guard
-    #: judges inner calls with the stack's ``RepeatedToolCallMiddleware`` against this
-    #: history — the guard runs on the sandbox bridge thread with no access to graph
-    #: state, so ``awrap_tool_call`` seeds the ``eval`` turn from it and writes the
-    #: turn's history back with the tool result. One history, one rule, both paths.
-    tool_call_history: NotRequired[Annotated[dict[str, list[str]], PrivateStateAttr]]
+    #: The same channel ``LoopDetectionState`` declares — type AND reducer must stay
+    #: identical here or the two schemas will not merge. Declared here because this
+    #: middleware writes it: the PTC guard judges inner calls with the stack's
+    #: ``RepeatedToolCallMiddleware`` against this history — the guard runs on the
+    #: sandbox bridge thread with no access to graph state, so ``awrap_tool_call``
+    #: seeds the ``eval`` turn from it and writes the turn's history back with the
+    #: tool result. One history, one rule, both paths. The reducer is what lets a step
+    #: with two ``eval`` calls write back both tasks' records instead of raising
+    #: ``InvalidUpdateError`` — see ``merge_tool_call_history`` and #217.
+    tool_call_history: NotRequired[Annotated[dict[str, list[str]], PrivateStateAttr, merge_tool_call_history]]
 
 
 def _code_interpreter_ptc_enabled() -> bool:
@@ -1156,34 +1163,39 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
             return await handler(request)
         runtime = getattr(request, "runtime", None)
         thread_id = resolve_ptc_thread_id(runtime)
-        # One turn per ``eval`` execution, guarded or not: it is how the guard (on
-        # the bridge thread) reaches the loop middleware's ``tool_call_history``,
-        # seeded here from the checkpointed state and written back with the result.
-        # Deep-copied: the guard appends, and the state's lists must not be mutated
-        # before the reducer applies the returned update.
-        #
-        # The turn also carries the scope of every ask id it mints (see
-        # ptc_guard.ask_id): the model's tool call id, which scopes this ``eval``
-        # invocation's questions and is checkpointed on the AI message, so it is
-        # stable across the replay. It is owned by the turn rather than read from
-        # whichever wrapper drains ``pending``, so two parallel ``eval`` calls
-        # sharing a thread still stamp and match with one scope.
-        turn = begin_ptc_turn(thread_id, str(tool_call.get("id") or ""))
-        if not turn.ask_scope:
-            # Not fatal — ask ids stay unique per round, and per turn is enough for
-            # a single eval call. Logged because it silently weakens the guarantee
-            # for a thread that later runs a second eval call.
-            logger.debug("[PTC] eval tool call carries no id; ask ids for this turn are scope-less")
-        state = getattr(request, "state", None) or {}
-        seed: dict[str, list[str]] = {k: list(v) for k, v in (state.get(TOOL_CALL_HISTORY_STATE_KEY) or {}).items()}
-        turn.tool_call_history = {k: list(v) for k, v in seed.items()}
-        try:
-            if not self._ptc_enabled or self._ptc_risk_scorer is None:
-                result = await self._run_eval_with_guidance(request, handler, thread_id)
-                return self._with_tool_call_history(result, turn, seed)
-            return await self._run_guarded_eval(request, handler, thread_id, turn, seed)
-        finally:
-            end_ptc_turn(thread_id)
+        # Serialize ``eval`` per thread. A model step may emit two ``eval`` calls and
+        # ``ToolNode`` runs them concurrently, but the REPL slot upstream, the turn
+        # below and the HITL collector are all keyed by ``thread_id`` alone and assume
+        # one ``eval`` in flight (#217). The gate spans the whole begin/end turn span,
+        # not just the eval run, because the turn is one of the things being protected.
+        # Only the ``eval`` path reaches here, so other tools keep their concurrency.
+        async with serialized_eval(thread_id):
+            # One turn per ``eval`` execution, guarded or not: it is how the guard (on
+            # the bridge thread) reaches the loop middleware's ``tool_call_history``,
+            # seeded here from the checkpointed state and written back with the result.
+            # Deep-copied: the guard appends, and the state's lists must not be mutated
+            # before the reducer applies the returned update.
+            #
+            # The turn also carries the scope of every ask id it mints (see
+            # ptc_guard.ask_id): the model's tool call id, which scopes this ``eval``
+            # invocation's questions and is checkpointed on the AI message, so it is
+            # stable across the replay.
+            turn = begin_ptc_turn(thread_id, str(tool_call.get("id") or ""))
+            if not turn.ask_scope:
+                # Not fatal — ask ids stay unique per round, and per turn is enough for
+                # a single eval call. Logged because it silently weakens the guarantee
+                # for a thread that later runs a second eval call.
+                logger.debug("[PTC] eval tool call carries no id; ask ids for this turn are scope-less")
+            state = getattr(request, "state", None) or {}
+            seed: dict[str, list[str]] = {k: list(v) for k, v in (state.get(TOOL_CALL_HISTORY_STATE_KEY) or {}).items()}
+            turn.tool_call_history = {k: list(v) for k, v in seed.items()}
+            try:
+                if not self._ptc_enabled or self._ptc_risk_scorer is None:
+                    result = await self._run_eval_with_guidance(request, handler, thread_id)
+                    return self._with_tool_call_history(result, turn, seed)
+                return await self._run_guarded_eval(request, handler, thread_id, turn, seed)
+            finally:
+                end_ptc_turn(thread_id)
 
     async def _run_guarded_eval(
         self, request: Any, handler: Any, thread_id: Any, turn: Any, seed: dict[str, list[str]]
--- a/packages/agent-common/agent_common/core/graph_utils.py
+++ b/packages/agent-common/agent_common/core/graph_utils.py
@@ -89,8 +89,8 @@ from agent_common.middleware.conversation_context_tools_middleware import (
 from agent_common.middleware.gateway_attribution_middleware import GatewayAttributionMiddleware
 from agent_common.middleware.loop_detection_middleware import (
     RepeatedToolCallMiddleware,
+    ToolCallHistory,
     history_delta,
-    merge_tool_call_history,
 )
 from agent_common.middleware.prompt_caching import LiteLLMPromptCachingMiddleware
 from agent_common.middleware.ptc_guard import (
@@ -418,16 +418,16 @@ class _PTCExposureState(REPLState):
 
     ptc_exposed_tool_names: NotRequired[Annotated[list[str], PrivateStateAttr]]
 
-    #: The same channel ``LoopDetectionState`` declares — type AND reducer must stay
-    #: identical here or the two schemas will not merge. Declared here because this
-    #: middleware writes it: the PTC guard judges inner calls with the stack's
+    #: The same channel ``LoopDetectionState`` declares, via the one shared alias —
+    #: the annotations must match exactly or the two schemas will not merge, and the
+    #: reducer must stay last inside it. Declared here because this middleware writes it: the PTC guard judges inner calls with the stack's
     #: ``RepeatedToolCallMiddleware`` against this history — the guard runs on the
     #: sandbox bridge thread with no access to graph state, so ``awrap_tool_call``
     #: seeds the ``eval`` turn from it and writes the turn's history back with the
     #: tool result. One history, one rule, both paths. The reducer is what lets a step
     #: with two ``eval`` calls write back both tasks' records instead of raising
     #: ``InvalidUpdateError`` — see ``merge_tool_call_history`` and #217.
-    tool_call_history: NotRequired[Annotated[dict[str, list[str]], PrivateStateAttr, merge_tool_call_history]]
+    tool_call_history: NotRequired[ToolCallHistory]
 
 
 def _code_interpreter_ptc_enabled() -> bool:
@@ -1174,11 +1174,14 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
             return await handler(request)
         runtime = getattr(request, "runtime", None)
         thread_id = resolve_ptc_thread_id(runtime)
-        # Serialize ``eval`` per thread. A model step may emit two ``eval`` calls and
-        # ``ToolNode`` runs them concurrently, but the REPL slot upstream, the turn
-        # below and the HITL collector are all keyed by ``thread_id`` alone and assume
-        # one ``eval`` in flight (#217). The gate spans the whole begin/end turn span,
-        # not just the eval run, because the turn is one of the things being protected.
+        # Serialize ``eval`` per thread. Two paths reach one ``thread_id`` at once: a
+        # model step that emits two ``eval`` calls (``ToolNode`` runs them as separate
+        # tasks of one superstep), and two parallel ``task`` dispatches of the same
+        # sub-agent, which share ``{context_id}::{checkpoint_ns}`` but each get their
+        # own event loop from ``asyncio.run``. The REPL slot upstream, the turn below
+        # and the HITL collector are all keyed by ``thread_id`` alone and assume one
+        # ``eval`` in flight (#217). The gate spans the whole begin/end turn span, not
+        # just the eval run, because the turn is one of the things being protected.
         # Only the ``eval`` path reaches here, so other tools keep their concurrency.
         async with serialized_eval(thread_id):
             # One turn per ``eval`` execution, guarded or not: it is how the guard (on
--- a/packages/agent-common/agent_common/core/graph_utils.py
+++ b/packages/agent-common/agent_common/core/graph_utils.py
@@ -89,6 +89,7 @@ from agent_common.middleware.conversation_context_tools_middleware import (
 from agent_common.middleware.gateway_attribution_middleware import GatewayAttributionMiddleware
 from agent_common.middleware.loop_detection_middleware import (
     RepeatedToolCallMiddleware,
+    history_delta,
     merge_tool_call_history,
 )
 from agent_common.middleware.prompt_caching import LiteLLMPromptCachingMiddleware
@@ -1051,15 +1052,22 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
         return self._exposure_state_update()
 
     @staticmethod
-    def _with_tool_call_history(result: Any, turn: Any, seed: dict[str, list[str]]) -> Any:
-        """Attach the turn's ``tool_call_history`` to the ``eval`` result as a state update.
+    def _with_tool_call_history(result: Any, turn: Any) -> Any:
+        """Attach this ``eval``'s ``tool_call_history`` additions to the result.
 
-        The guard extended ``turn.tool_call_history`` for every inner call it judged;
+        The guard recorded every inner call it judged on ``turn.history_appends``;
         this is the only place those entries become durable. When the program made no
-        judged call the history equals ``seed`` (what the state already holds) and the
-        result passes through untouched — no state write. Otherwise a ``ToolMessage``
-        is wrapped in a ``Command`` carrying both the message and the history, and a
-        ``Command`` from the handler gets the history merged into its update.
+        judged call there is nothing to append and the result passes through
+        untouched — no state write. Otherwise a ``ToolMessage`` is wrapped in a
+        ``Command`` carrying both the message and the update, and a ``Command`` from
+        the handler gets the update merged into its own.
+
+        The update is an incremental *delta*, not the whole history: a model step may
+        emit two ``eval`` calls, which ``ToolNode`` runs as two tasks of one
+        superstep, so both write this channel. Deltas commute, so the reducer applies
+        both whatever order it sees them in and caps the window once at the end,
+        instead of each eval trimming against its own partial view — see
+        ``merge_tool_call_history`` and #217.
 
         A guard record must never be dropped silently — that would fail the loop
         guard open for exactly the call it just judged. ``Command.update`` may be a
@@ -1068,9 +1076,12 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
         value) cannot be merged without knowing the schema, and the ``eval`` handler
         never produces them, so they raise instead of being papered over.
         """
-        history = {k: list(v) for k, v in turn.tool_call_history.items()}
-        if history == seed:
+        if not turn.history_appends:
             return result
+        history = history_delta(
+            {k: list(v) for k, v in turn.history_appends.items()},
+            dict(turn.history_caps),
+        )
         if not isinstance(result, Command):
             return Command(update={"messages": [result], TOOL_CALL_HISTORY_STATE_KEY: history})
         update = result.update
@@ -1187,18 +1198,23 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
                 # for a thread that later runs a second eval call.
                 logger.debug("[PTC] eval tool call carries no id; ask ids for this turn are scope-less")
             state = getattr(request, "state", None) or {}
-            seed: dict[str, list[str]] = {k: list(v) for k, v in (state.get(TOOL_CALL_HISTORY_STATE_KEY) or {}).items()}
-            turn.tool_call_history = {k: list(v) for k, v in seed.items()}
+            # The guard's READ view of the history: what the loop rule judges inner
+            # calls against. Deep-copied so the guard's appends cannot mutate the
+            # state's lists in place. What gets written BACK is the delta the guard
+            # records alongside it, not this dict.
+            turn.tool_call_history = {
+                k: list(v) for k, v in (state.get(TOOL_CALL_HISTORY_STATE_KEY) or {}).items()
+            }
             try:
                 if not self._ptc_enabled or self._ptc_risk_scorer is None:
                     result = await self._run_eval_with_guidance(request, handler, thread_id)
-                    return self._with_tool_call_history(result, turn, seed)
-                return await self._run_guarded_eval(request, handler, thread_id, turn, seed)
+                    return self._with_tool_call_history(result, turn)
+                return await self._run_guarded_eval(request, handler, thread_id, turn)
             finally:
                 end_ptc_turn(thread_id)
 
     async def _run_guarded_eval(
-        self, request: Any, handler: Any, thread_id: Any, turn: Any, seed: dict[str, list[str]]
+        self, request: Any, handler: Any, thread_id: Any, turn: Any
     ) -> Any:
         """The HITL loop for a guarded ``eval`` — see ``awrap_tool_call``."""
         runtime = getattr(request, "runtime", None)
@@ -1221,7 +1237,7 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
             result = await self._run_eval_with_guidance(request, handler, thread_id)
             pending = take_ptc_pending(thread_id)
             if not pending:
-                return self._with_tool_call_history(result, turn, seed)
+                return self._with_tool_call_history(result, turn)
 
             # Decisions arrive already aligned 1:1 with `pending`. The orchestrator's
             # resume path (executor._build_interrupt_resume_map) replicates the single
--- a/packages/agent-common/agent_common/core/graph_utils.py
+++ b/packages/agent-common/agent_common/core/graph_utils.py
@@ -1064,10 +1064,11 @@ class _PTCToleranceCodeInterpreterMiddleware(CodeInterpreterMiddleware):
 
         The update is an incremental *delta*, not the whole history: a model step may
         emit two ``eval`` calls, which ``ToolNode`` runs as two tasks of one
-        superstep, so both write this channel. Deltas commute, so the reducer applies
-        both whatever order it sees them in and caps the window once at the end,
-        instead of each eval trimming against its own partial view — see
-        ``merge_tool_call_history`` and #217.
+        superstep, so both write this channel. Deltas with equal caps commute, so the
+        reducer applies both whatever order it sees them in instead of each eval
+        trimming against its own partial view — see ``merge_tool_call_history`` and
+        #217 (mixed caps on one key can differ by a single entry; that docstring says
+        when and why it is acceptable).
 
         A guard record must never be dropped silently — that would fail the loop
         guard open for exactly the call it just judged. ``Command.update`` may be a
--- a/packages/agent-common/agent_common/middleware/loop_detection_middleware.py
+++ b/packages/agent-common/agent_common/middleware/loop_detection_middleware.py
@@ -62,6 +62,55 @@ _STOPPED_BEFORE_EXECUTION = (
 )
 
 
+def merge_tool_call_history(
+    current: dict[str, list[str]] | None,
+    update: dict[str, list[str]] | None,
+) -> dict[str, list[str]]:
+    """Reducer for the ``tool_call_history`` channel.
+
+    Normally one writer updates this per step and the update simply wins. The
+    exception is a model step that emits two ``eval`` tool calls: ``ToolNode`` makes
+    each a task of the *same* superstep, both read the same checkpointed history as
+    their seed, and both write back their own extension of it. Without a reducer
+    that is not a silent last-writer-wins — LangGraph's default ``LastValue``
+    channel *raises* ``InvalidUpdateError: can receive only one value per step``,
+    which fails the whole agent turn (#217).
+
+    Per key, three cases, in the order tested:
+
+    * **Unchanged / first write** — take the update.
+    * **A window trim** (``RepeatedToolCallMiddleware.evaluate`` drops the oldest
+      entries past ``window_size``): the update is a *suffix* of what we hold, so it
+      is shorter by construction. Take it, or the trim would be undone forever.
+    * **Concurrent extension** — two eval tasks grew the same seed. Keep the common
+      prefix once and append each task's own tail, so neither eval's calls vanish
+      from the loop guard's memo.
+
+    Two eval programs whose divergent tails *begin* with the identical arg hash
+    (the same tool called with the same args from both) collapse to one entry: the
+    common prefix absorbs it. That under-counts the repeat by one, which is the
+    safe direction — the guard blocks late rather than early — and it cannot happen
+    at the model boundary, where there is only ever one writer per step.
+    """
+    if not current:
+        return dict(update or {})
+    if not update:
+        return dict(current)
+    merged = {k: list(v) for k, v in current.items()}
+    for key, new in update.items():
+        old = merged.get(key)
+        if old is None or old == list(new):
+            merged[key] = list(new)
+        elif len(new) < len(old) and list(new) == old[-len(new) :]:
+            merged[key] = list(new)
+        else:
+            shared = 0
+            while shared < min(len(old), len(new)) and old[shared] == new[shared]:
+                shared += 1
+            merged[key] = [*old, *new[shared:]]
+    return merged
+
+
 @dataclass(frozen=True)
 class LoopVerdict:
     """Outcome of ``RepeatedToolCallMiddleware.evaluate`` for one prospective call.
@@ -85,7 +134,7 @@ class LoopDetectionState(AgentState):
     Similar to ToolCallLimitMiddleware but tracks both same-args and same-tool patterns.
     """
 
-    tool_call_history: NotRequired[Annotated[dict[str, list[str]], PrivateStateAttr]]
+    tool_call_history: NotRequired[Annotated[dict[str, list[str]], PrivateStateAttr, merge_tool_call_history]]
     """Per-tool history of argument hashes. Format:
     {
         "tool_name": ["args_hash1", "args_hash2", ...],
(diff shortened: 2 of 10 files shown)


alloy-cicd (commented):

The serialization approach is sound (the gate is correctly released across `interrupt()`, and sub-agent thread-id suffixing avoids re-entrant deadlock), and the e2e tests are genuinely valuable. However, verification found **two confirmed high-severity issues** and several smaller ones:

1. **Reducer trim branch is dead code → unbounded history growth and spurious loop blocks** (`loop_detection_middleware.py:104`). Confirmed by simulation: +`window_size` entries per step once the window saturates; a legitimate `task`/`eval` gets falsely BLOCKED and the run force-stopped. This regresses the *single-writer* path, not just the parallel case.
2. **Two co-pending interrupts are now the sanctioned outcome of one turn, but the orchestrator renders only `interrupts[-1]`, and blanket clients (Slack) apply one Approve to the hidden interrupt too** — an unseen high-risk call can execute on an approval given for a different one. Confirmed empirically against the PR's own agent builder.
3. The per-loop gate does not cover a reachable same-process cross-loop path (parallel `task` dispatch of the same dynamic agent via `asyncio.run` threads), and the code comment mischaracterizes that gap as needing a cross-process lock.

Plus smaller notes on gate lifetime/simplification, the summarizer LLM call held under the gate, the duplicated reducer annotation, and manual-QA script drift. Details inline.

Checked and found fine: gate release across `GraphInterrupt`/resume replay; no re-entrant deadlock (sub-agents use suffixed thread ids); `_ptc_default` fallback only reachable without a checkpointer, where serialization is arguably correct; `ptc_exposed_tool_names` has a single writer per step, so it needs no reducer; the sibling-eval undercount is acknowledged in the docstring and fails in the safe direction.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
```diff
@@ -62,6 +62,55 @@
...
+    if not current:
+        return dict(update or {})
+    if not update:
+        return dict(current)
+    merged = {k: list(v) for k, v in current.items()}
+    for key, new in update.items():
+        old = merged.get(key)
+        if old is None or old == list(new):
+            merged[key] = list(new)
+        elif len(new) < len(old) and list(new) == old[-len(new) :]:
```
alloy-cicd: **[critical] This trim branch is unreachable for every production writer, so once a key's window saturates, history grows without bound and inflates repeat counts.**

`RepeatedToolCallMiddleware.evaluate` appends **then** trims (`history = [*tool_history, args_hash]` → `history[-window_size:]`, lines 295-297), so a saturated-window update has `len(new) == len(old)` — head dropped, new hash appended. `len(new) < len(old)` is never true, `new` is never a suffix of `old`, so every trimmed write falls into the concurrent-extension branch below with `shared == 0`, returning `old + new`.

Simulated with the prod config from `graph_utils.py` (`window_size=10`, `max_repeats=5`, `force_stop_after=3`, `dispatch_tools={"task","eval"}`): on the 11th call the stored list goes 10 → 20 entries and grows +10 per step forever (20/30/40/…/210 after 30 steps). Consequences:
- unbounded `tool_call_history` in every checkpoint on any long thread;
- duplicated hashes inflate `tool_history.count(args_hash)` (line 350): a `task`/`eval` argument set genuinely used ≤4 times gets BLOCKED as "called 7 times with identical arguments" (reproduced at step 12 of the simulation), and `force_stop_after=3` then ends the run. `task`/`eval` and `eval:<tool>` program keys are exactly the ones exempt from the per-tool cap, so they are the ones that legitimately reach the window and hit this.

Note this fires on the ordinary single-writer path — no parallel evals needed — so it's a regression of the common case introduced by the reducer.

Related: even the intended two-eval merge is never re-trimmed to `window_size` (line 110 appends unboundedly; the reducer has no access to the window), which pushes lists past saturation one step early.

Suggested direction: make the trim recognizable — e.g. treat `len(new) == len(old)` with `new[:-1] == old[1:]` (rotation) as a replace, or more robustly detect "update's head is a suffix of current" (`new[:-k]` is a suffix of `old` for the k appended entries) and replace rather than concatenate; alternatively have writers emit deltas and apply the window at read time in `evaluate`, so the reducer never has to guess trim-vs-extension.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
```diff
@@ -62,6 +62,55 @@
...
+    their seed, and both write back their own extension of it. Without a reducer
+    that is not a silent last-writer-wins — LangGraph's default ``LastValue``
+    channel *raises* ``InvalidUpdateError: can receive only one value per step``,
+    which fails the whole agent turn (#217).
+
+    Per key, three cases, in the order tested:
+
+    * **Unchanged / first write** — take the update.
+    * **A window trim** (``RepeatedToolCallMiddleware.evaluate`` drops the oldest
+      entries past ``window_size``): the update is a *suffix* of what we hold, so it
```
alloy-cicd: This docstring claim — the trimmed update "is a *suffix* of what we hold, so it is shorter by construction" — is false for the actual writer: `evaluate` appends the new hash before trimming, so the update is an equal-length rotation (`old[1:] + [new_hash]`), never a strict suffix. See the main comment on the trim branch below.

Minor wording nit further down (lines 89-93): "under-counts the repeat by one" understates it — fully identical tails collapse entirely (two evals each making the same call 3× persist 3 hashes, not 6). The direction is still "blocks late", so the design point stands; just worth stating the real bound.

packages/agent-common/tests/test_ptc_parallel_eval.py
```diff
@@ -0,0 +1,221 @@
...
+        # First write / nothing held yet.
+        ({}, {"t": ["a"]}, {"t": ["a"]}),
+        (None, {"t": ["a"]}, {"t": ["a"]}),
+        # A single writer appending: the update simply wins.
+        ({"t": ["a"]}, {"t": ["a", "b"]}, {"t": ["a", "b"]}),
+        # Unchanged key passes through.
+        ({"t": ["a"]}, {"t": ["a"]}, {"t": ["a"]}),
+        # The loop guard's sliding-window trim must NOT be undone: the update is a
+        # suffix of what we hold, so it replaces rather than merges.
+        ({"t": ["a", "b", "c"]}, {"t": ["b", "c"]}, {"t": ["b", "c"]}),
```
alloy-cicd, line 211: This parametrized "window trim" case (`{"t": ["a","b","c"]}` → `{"t": ["b","c"]}`) encodes a strictly-shorter suffix that `evaluate` never produces — allowed calls emit equal-length windows (append-then-trim), blocked calls emit longer-by-one untrimmed lists. So the suite is green while the real trim shape (`["h0".."h9"]` → `["h1".."h9","x"]`) concatenates to 20 entries.

A regression test should drive the update pair straight out of `RepeatedToolCallMiddleware.evaluate` with the window exceeded and assert `len(merged[key]) <= window_size`.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1156,34 +1163,39 @@ async def awrap_tool_call(self, request: Any, handler: Any) -> Any:
...
-            return await self._run_guarded_eval(request, handler, thread_id, turn, seed)
-        finally:
-            end_ptc_turn(thread_id)
+        # Serialize ``eval`` per thread. A model step may emit two ``eval`` calls and
+        # ``ToolNode`` runs them concurrently, but the REPL slot upstream, the turn
+        # below and the HITL collector are all keyed by ``thread_id`` alone and assume
+        # one ``eval`` in flight (#217). The gate spans the whole begin/end turn span,
+        # not just the eval run, because the turn is one of the things being protected.
+        # Only the ``eval`` path reaches here, so other tools keep their concurrency.
+        async with serialized_eval(thread_id):
```
alloy-cicd, line 1187: **[high] This change makes two co-pending HITL interrupts the sanctioned outcome of one turn, but the downstream approval surface still assumes one — and blanket clients auto-approve the hidden one.**

Verified empirically against this PR's own agent builder (langgraph 1.2.7): one `ainvoke` with two high-risk evals returns `state.interrupts == 2` in a single round-trip (eval A's `GraphInterrupt` releases the gate; eval B's separate Send task then acquires it and raises its own interrupt in the same superstep — exactly what `test_two_evals_in_one_step_each_get_their_own_approval` asserts with `sum(asked) == 2`).

Downstream (pre-existing code, but this PR is what makes the state reachable on the eval path — previously it hard-crashed before any unseen approval could exist):
- `orchestrator-agent/app/core/agent.py:1328-1333` renders only `final_state.interrupts[-1].value`, so eval A's approval request is never shown to anyone.
- On resume, `executor.py:270-275` defaults: an id-keyed client silently **rejects** the unseen interrupt's calls; a blanket client replicates the single decision to every co-pending interrupt — and the Slack client sends exactly that shape (`client-slack/src/listeners/actions/hitlButton.ts:104`, `decisions: [{type: 'approve'}]`, no ids). Net effect: the user approves eval B's request and eval A's unseen high-risk call executes on that approval.
- Same truncation on the sub-agent path: `dynamic_tool_dispatch.py:2574-2585` surfaces/resumes `sub_interrupts[0]` only.

Suggest either rendering all pending interrupts (and having blanket decisions apply only to the surfaced one), or — as a stopgap within this PR — merging co-pending PTC asks back into one interrupt payload so the one-interrupt client assumption stays true.

packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -333,12 +336,80 @@ def resolve_ptc_thread_id(runtime: Any) -> str:
...
+    #: Callers inside :func:`serialized_eval` for this key — holder and waiters
+    #: alike. The gate is dropped when it reaches zero, so the registry does not
+    #: grow one permanent entry per conversation for the life of the process.
+    users: int = 0
+
+
+#: Live ``eval`` gates, keyed by ``(event loop, thread_id)``. The loop is part of
+#: the key because an ``asyncio.Lock`` binds to the loop that first awaits it, so a
+#: lock reused across loops (a sync ``invoke`` that spins up its own) would raise
+#: rather than serialize. Same-thread evals on *different* loops are not the race
```
alloy-cicd: **[medium] The "different loops … needs a cross-process lock" framing is inaccurate, and the cross-loop gap is reachable in the mainline deployment.**

The resources this gate protects are all *process-local* and keyed by `thread_id` alone: `_PTC_TURNS`, the pending collector, and the QuickJS REPL slot registry. The gate, however, is keyed per `(id(loop), thread_id)` — so two event loops in one process driving the same `thread_id` get zero serialization.

That scenario is real, not hypothetical: `dynamic_tool_dispatch.py:2030-2036` invokes local dynamic sub-agents synchronously from ToolNode executor threads via `LocalA2ARunnable.invoke` → `asyncio.run(...)` (`a2a/base.py:151`) — a fresh loop per dispatch — and the sub-agent thread id `{context_id}::dynamic-{name}` is stable per conversation+agent, not per call. Two parallel `task` calls to the *same* dynamic agent in one step → two threads, two loops, one thread_id → the exact #217 REPL `already closed` collision this PR fixes, plus turn clobbering, with no gate in the way.

Either extend the gate to serialize cross-loop within the process (e.g. key by `thread_id` with a loop-aware gate or a threading-level lock around the REPL acquisition), or at minimum correct this comment to name the actually-reachable path so the residual risk is visible.

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -1156,34 +1163,39 @@ async def awrap_tool_call(self, request: Any, handler: Any) -> Any:
...
+                # for a thread that later runs a second eval call.
+                logger.debug("[PTC] eval tool call carries no id; ask ids for this turn are scope-less")
+            state = getattr(request, "state", None) or {}
+            seed: dict[str, list[str]] = {k: list(v) for k, v in (state.get(TOOL_CALL_HISTORY_STATE_KEY) or {}).items()}
+            turn.tool_call_history = {k: list(v) for k, v in seed.items()}
+            try:
+                if not self._ptc_enabled or self._ptc_risk_scorer is None:
+                    result = await self._run_eval_with_guidance(request, handler, thread_id)
+                    return self._with_tool_call_history(result, turn, seed)
+                return await self._run_guarded_eval(request, handler, thread_id, turn, seed)
```
alloy-cicd: **[low, latency] The gate is held across `attach_summaries`, an LLM round-trip.** Inside `_run_guarded_eval`, `take_ptc_pending` has already drained before `attach_summaries(...)` is awaited and `interrupt()` raised — so a sibling eval queued on the gate also waits out the summarizer call on the first pass of every new ask round (bounded by `_SUMMARY_TIMEOUT_SECONDS = 5.0`, and the replay pass short-circuits, so this is a bounded cost, not a hang).

If you do move the summarize+interrupt outside the gate, note one trap: `end_ptc_turn` pops `_PTC_TURNS` by thread key, not turn identity — releasing the gate before `interrupt()` would let eval B install its turn, and A's unwinding `finally` would then pop B's live turn. The pop needs to be identity-guarded (`pop only if the registry still holds this turn`) first. Given the 5s bound, also fine to leave as-is with a comment.

packages/agent-common/agent_common/middleware/ptc_guard.py
```diff
@@ -333,12 +336,80 @@ def resolve_ptc_thread_id(runtime: Any) -> str:
...
+    candidate fixes that is wholly ours: re-keying the two collectors by
+    ``(thread_id, tool_call_id)`` still leaves the REPL colliding underneath, and
+    constraining the model (``parallel_tool_calls=False``) would disable parallel
+    calls for *every* tool, not just ``eval``. Concurrency for other tools is
+    untouched — this gate is only ever entered on the ``eval`` path.
+
+    The two evals therefore run back-to-back, each with its own turn and its own
+    REPL reset in between, which is what the bookkeeping already assumes.
+    """
+    key = (id(asyncio.get_running_loop()), thread_id)
```
alloy-cicd: **[low, robustness/simplification] The `id(loop)`-keyed, manually-refcounted registry has an unrecoverable (if unlikely) stale-entry mode, and there's a simpler in-repo precedent.**

If a `serialized_eval` frame is ever torn down without its `finally` running (a suspended frame kept strongly referenced past loop close — not a pattern in today's teardown paths, which we traced as safe, but nothing enforces that), the entry survives with `users >= 1` forever. `id()` of a freed loop is readily reused (the `asyncio.run`-per-dispatch pattern in `a2a/base.py:151` frees and reallocates loops constantly), and a new loop at the same address then inherits an `asyncio.Lock` bound to the dead one → `RuntimeError: … bound to a different event loop` on every subsequent eval for that thread id, unrecoverable because `users` never reaches 0.

`OrchestratorCookieCache` (`console-backend/console_backend/utils/orchestrator_cookie_cache.py:59-61`) already solves per-key lock lifetime with a `WeakValueDictionary[key, asyncio.Lock]`: the context-manager local and each waiter's frame are the strong refs, so the entry disappears exactly when the last user leaves — deleting `_EvalGate`, the `users` counter, the try/finally pop, and the invariant comments in one move, and making a leaked frame keep only its own gate alive. Alternatively, keying by `thread_id` alone with the loop stored on the gate and replace-on-mismatch also works, but then the eviction must be identity-guarded (`pop only if the dict still holds this gate`).

packages/agent-common/agent_common/core/graph_utils.py
```diff
@@ -413,13 +417,16 @@ class _PTCExposureState(REPLState):
...
+    #: The same channel ``LoopDetectionState`` declares — type AND reducer must stay
+    #: identical here or the two schemas will not merge. Declared here because this
+    #: middleware writes it: the PTC guard judges inner calls with the stack's
+    #: ``RepeatedToolCallMiddleware`` against this history — the guard runs on the
+    #: sandbox bridge thread with no access to graph state, so ``awrap_tool_call``
+    #: seeds the ``eval`` turn from it and writes the turn's history back with the
+    #: tool result. One history, one rule, both paths. The reducer is what lets a step
+    #: with two ``eval`` calls write back both tasks' records instead of raising
+    #: ``InvalidUpdateError`` — see ``merge_tool_call_history`` and #217.
+    tool_call_history: NotRequired[Annotated[dict[str, list[str]], PrivateStateAttr, merge_tool_call_history]]
```
alloy-cicd: **[low, maintainability] This `Annotated[...]` must stay byte-identical to the one in `loop_detection_middleware.py:137`, and its correctness depends on the reducer being the *last* metadata entry — a rule currently documented only in `tests/manual/README.md`.**

LangGraph detects the reducer as `callable(metadata[-1])`; put it first (or append another marker after it) and it is silently ignored — no error, and prod returns to the exact `InvalidUpdateError` this PR fixes.

Suggest exporting one alias from `loop_detection_middleware` (e.g. `ToolCallHistory = Annotated[dict[str, list[str]], PrivateStateAttr, merge_tool_call_history]`) and using it in both schemas — one definition, no ordering trap. Per `packages/agent-common/AGENTS.md`'s own "Maintaining These Instructions" rule, the reducer-last gotcha (and the new eval-serialization contract) also belong in that file rather than only in a manual-QA README.

packages/agent-common/tests/manual/qa_217_parallel_eval.py
```diff
@@ -0,0 +1,335 @@
+#!/usr/bin/env python
```
alloy-cicd: **[nit] This script is a near-verbatim fork of `test_ptc_parallel_eval.py`** (`_ProbeArgs`, `_ScriptedModel`, `_PROGRAM`, the agent builder, the resume-map construction — duplicated with asserts turned into prints). It isn't run in CI, so any change to the middleware constructor, `wrap_tool_for_ptc`, or the resume shape rots it silently. Consider importing the shared builder/helpers from the test module, or replacing the script with a README pointer to `uv run pytest tests/test_ptc_parallel_eval.py -s`.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
aartaria: Confirmed and fixed in 101d57ff — thank you, this was a real regression of the ordinary single-writer path, worse than the bug the PR set out to fix. I reproduced your simulation: prod config, one tool called 16 times, 70 entries left in a 10-entry window.

Rather than teach the reducer the rotation shape, the merge is now incremental and the writer declares its intent. The eval path emits a tagged delta (`history_delta`: the hashes it appended plus the window to apply after) and `merge_tool_call_history` appends it; every other writer keeps emitting whole values and keeps plain replace semantics, so the model-boundary path is untouched by construction rather than by case analysis. Deltas commute and the cap is applied once at the end instead of per writer against a partial view.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
aartaria: Fixed in 101d57ff — the false claim is gone with the branch it described, and the docstring now states why shape comparison is not decidable at all (append-then-trim makes a saturated update an equal-length rotation).

On the under-count: also fixed rather than just re-worded. With deltas there is no common-prefix absorption, so two evals making the identical call now persist two entries. Added a test pinning exactly that.

packages/agent-common/tests/test_ptc_parallel_eval.py
aartaria: Fixed in 101d57ff. The parametrized cases now cover the real contract (whole value replaces, delta appends-then-caps, deltas commute, a blocked cap of `None` stays unbounded), and there is a regression test built the way you suggested — it drives updates straight out of `RepeatedToolCallMiddleware.evaluate` for 25 steps past a 10-entry window and asserts the merged history stays at 10. That test fails against the reducer this PR originally had.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Confirmed and fixed in 101d57ff — agreed this is the serious one, and you are right that the PR is what makes the state reachable.

`AgentStreamResponse.interrupt_value` now folds co-pending approval asks into a single card instead of rendering `interrupts[-1]`. The clients already render N action requests in one card and each request keeps its own `_call_id`, so `_build_interrupt_resume_map` still routes decisions back to the interrupt that raised them — and a blanket approve is now legitimate, because the user saw everything it authorises. Auth and client-action pauses are deliberately never folded in: they need their own round trip, so a mixed set still renders the last one and leaves the rest pending. Tests in `orchestrator-agent/tests/test_responses.py` cover the merge, the id routing, and the mixed case.

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Confirmed and fixed in 673a65f3 — you were right on both counts, the framing and the reachability. I traced it: `get_thread_id` defaults to `{context_id}::{checkpoint_ns}`, so two parallel `task` dispatches of the *same* sub-agent in one conversation resolve one thread_id while `LocalA2ARunnable.invoke` -> `asyncio.run` gives each its own loop. Two gates, one QuickJS slot, crash unchanged.

The gate now holds a `threading.Lock` keyed by `thread_id` alone — the same key the protected resources use. It is acquired non-blockingly and polled between awaits rather than blocking the loop or going through `asyncio.to_thread`: a worker per waiter can deadlock, since the holder itself needs a worker for the summarizer's `to_thread` call. The misleading cross-process comment is gone. New test drives two `asyncio.run` loops on separate threads against one thread_id and asserts no overlap; it fails against the old per-loop keying.

packages/agent-common/agent_common/middleware/ptc_guard.py
aartaria: Fixed as a side effect of 673a65f3. The gate no longer keys on `id(loop)` at all, so the recycled-address mode you describe — a new loop inheriting an `asyncio.Lock` bound to a dead one — is gone. The registry is keyed by `thread_id` and guarded by its own `threading.Lock`, since it is now reached from several threads rather than several tasks on one loop.

The refcount stays, but its failure mode is now benign: a leaked entry means a dict entry that is never dropped, holding an unheld lock that still works correctly, rather than an unusable lock. I kept it over `WeakValueDictionary` because the lifetime I need is "someone is inside or waiting", which is what the count expresses directly; a weak reference would tie it to Python object liveness instead.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Fixed in 673a65f3 — the two declarations are now one exported alias, `ToolCallHistory`, used by both schemas. Your suggestion exactly; thank you.

The ordering rule is documented on the alias itself rather than only in the manual README, since that is where anyone would be editing it.

packages/agent-common/tests/manual/qa_217_parallel_eval.py
aartaria: Fixed in 2182fffa rather than argued with — you are right that a fork not run in CI rots silently.

`build_eval_agent` now takes `parallel=True|False` (the `False` case is the script's control scenario) and lives in the pytest module; the script imports it along with `eval_results`. What stays in the script is what is genuinely manual-QA-specific: the three scenarios, the per-layer diagnostics, and the documented pre-fix output. I kept it as a script rather than replacing it with a `pytest -s` pointer because the diagnostics and the recorded before/after are the point of it for a human reproducing the bug.

packages/agent-common/agent_common/core/graph_utils.py
aartaria: Not applying, but recording the analysis — thank you for the `end_ptc_turn` warning, which is the part I would have walked into.

Holding the gate across `attach_summaries` is deliberate. It is bounded by `_SUMMARY_TIMEOUT_SECONDS = 5.0`, only on the first pass of an ask round, and only when a *second* eval is queued behind an approval — a narrow case whose alternative is a correctness hazard, exactly the one you describe: `end_ptc_turn` pops `_PTC_TURNS` by thread key, not by turn identity, so releasing the gate before `interrupt()` lets eval B install its turn and A's unwinding `finally` then deletes B's. Trading a bounded 5s wait for that is not a good trade while the turn registry is keyed the way it is.

Worth revisiting if `_PTC_TURNS` is ever re-keyed by `(thread_id, tool_call_id)` — at that point the pop becomes identity-safe and the summarize+interrupt can move outside the gate.

packages/orchestrator-agent/app/models/responses.py
```diff
@@ -83,6 +83,41 @@ def auth_required(
...
+        their own round trip and cannot be folded into an approval card. The last one
+        renders, as before, and the rest stay pending for the next turn.
+        """
+        pending = list(interrupts or ())
+        if not pending:
+            return {}
+        values = [getattr(i, "value", None) for i in pending]
+        approvals = [v for v in values if isinstance(v, dict) and v.get("action_requests")]
+        if len(approvals) < 2 or len(approvals) != len(values):
+            return values[-1] if values[-1] is not None else {}
```
alloy-cicd, line 122: **[medium] Residual variant of the round-1 bypass: a mixed set with ≥2 approvals abandons the fold, and a blanket approve still replicates to the hidden approval.**

Reproduced empirically against this head: pending `[approval(delete_prod_db), auth, approval(send_email)]` — this branch returns `values[-1]`, so the card shows only `send_email`; a blanket `decisions: [{type: 'approve'}]` resume then hands the hidden `delete_prod_db` interrupt `{'decisions': [{'type': 'approve'}]}` via the replication at `executor.py:273-275`. Same round-1 mechanism, now needing 3+ co-pending interrupts with 2 approvals and a non-approval ordered before an approval — narrower, but the PR's premise (one approval per serialized eval) makes multi-interrupt supersteps routine, and `AuthErrorDetectionMiddleware` wraps eval too, so two eval approvals + one auth-erroring sibling in one assistant message is a legal superstep.

For contrast, the two-interrupt mixed cases are fail-safe (verified: auth-rendered/approval-hidden rejects the hidden approval via `structural_decisions`; approval-rendered/auth-hidden resumes the auth with the raw query). Only the ≥2-approvals mixed shape leaks.

Options: fold the approvals whenever there are ≥2 of them regardless of non-approval siblings (render approvals + leave the non-approval pending), or make the blanket-replication path refuse interrupts whose requests weren't on the rendered card. At minimum the docstring above ("the rest stay pending for the next turn") is inaccurate for hidden approvals under a blanket approve — they don't stay pending, they get the replicated decision.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
```diff
@@ -62,6 +62,85 @@
...
+      model-boundary writer in ``RepeatedToolCallMiddleware`` has always got. That
+      writer runs once per step and is left completely alone by this reducer.
+    * **A delta** (:func:`history_delta`) — append the writer's own hashes, then
+      apply its cap. This is what the ``eval`` path emits, because a model step may
+      emit two ``eval`` calls and ``ToolNode`` makes them two tasks of the *same*
+      superstep. Both then write this channel, and without a reducer that is not a
+      silent last-writer-wins: ``LastValue`` *raises* ``InvalidUpdateError: can
+      receive only one value per step`` and fails the whole agent turn (#217).
+
+    Deltas commute, so two evals in one step produce ``current + A + B`` whatever
```
alloy-cicd: **[nit, doc accuracy]** "Deltas commute" holds only for equal caps. Two same-key deltas with different caps — reachable when two evals in one step call the same tool and one is blocked (cap `None`) while the other is allowed (cap `window_size`), both judging against the same superstep-start seed — are order-dependent because each delta's cap is applied immediately rather than once at the end: simulated allowed-then-blocked → 11 entries vs blocked-then-allowed → 10. Impact is ±1 entry / one step of escalation timing, so no code change needed — just scope the claim ("deltas with equal caps commute; mixed caps may differ by up to one entry") here and at the matching claim in `graph_utils.py`.

packages/agent-common/tests/manual/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+- `parallel-hitl` logs `NoDefaultModelError: No default chat model is configured`
+  from the tool-call summarizer. Expected — no model is configured here, so the
+  approval prompt falls back to raw args. It does not affect the verdict.
+
+### What the fix is
+
+Two parts, both required — each was disabled alone and the harness re-run to confirm:
+
+- **`ptc_guard.serialized_eval`** — a per-`(event loop, thread_id)` mutex held across
```
alloy-cicd: **[nit, stale doc]** This still describes the fix as "a per-`(event loop, thread_id)` mutex" — the design 673a65f3 replaced within this same PR (the gate is now a per-`thread_id` `threading.Lock` polled from any event loop, precisely because per-loop keying missed the cross-loop `task`-dispatch path). Anyone reproducing from this README gets the superseded — and previously-flagged-as-insufficient — description. Same stale wording survives in the PR description.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (164bee2acfc216e20cfdc067ea795d704924a9a3..2182fffa07d4c4ebbc42a73e8702eee044f9e1cc)

**Disposition of round-1 findings** (each re-verified against the new head, by simulation or execution where the original was):

| # | Finding | Disposition |
|---|---|---|
| 1 | Reducer trim branch dead → unbounded growth, spurious blocks | **verified-fixed** (101d57ff) — re-ran the round-1 simulation: history stays at 10 where it previously reached 70; blocked-call escalation feeding `force_stop_after` intact; delta tag is JSON-safe and can't persist as the channel value |
| 2 | Hidden co-pending interrupt auto-approved by blanket clients | **verified-fixed for the pure-approvals case** (101d57ff) — traced the id-keyed and blanket round trips end to end, and the auth-rendered/approval-hidden two-interrupt case correctly rejects. One narrower residual hole in mixed sets remains — see the new inline comment |
| 3 | Per-loop gate misses the cross-loop same-thread_id path | **verified-fixed** (673a65f3) — re-ran the new regression test against the old per-loop keying and it fails as claimed; poll loop yields properly, acquire/release pairing survives `GraphInterrupt`, registry mutations are all under the guard with no await |
| 4 | Gate held across `attach_summaries` | **rejected-by-author** with sound reasoning (bounded 5s, narrow case, and releasing early is unsafe while `end_ptc_turn` pops by thread key) — accepted, not re-raised |
| 5 | `id(loop)` refcount registry stale-lock hazard | **verified-fixed** (side effect of 673a65f3) — key no longer contains a loop id; a leaked refcount now means an idle dict entry, not a poisoned lock |
| 6 | False "shorter by construction" docstring / undercount wording | **verified-fixed** — and the undercount itself is gone (deltas count duplicate calls) with a test pinning it |
| 7 | Reducer test covered an impossible trim shape | **verified-fixed** — new evaluate-driven 25-step regression test; confirmed it fails against the round-1 reducer |
| 8 | Duplicated `Annotated` / reducer-must-be-last trap | **verified-fixed** — single exported `ToolCallHistory` alias used by both schemas, ordering rule documented on the alias |
| 9 | Manual QA script fork drift | **verified-fixed** (2182fffa) — script imports the shared builder, runs end-to-end, still not collected by pytest |

**Delta review:** one residual (narrower) variant of finding 2 in the mixed-interrupt case, plus two doc-accuracy nits — inline below. Nothing else in the delta reached the bar: the threaded gate's fairness (no FIFO) and the unreachable same-key re-entrancy were checked and are not report-worthy.

**Follow-up material (not blocking, noted once):** the sub-agent path `dynamic_tool_dispatch.py:2575/2584` still surfaces only `sub_interrupts[0]` — pre-existing, untouched by this PR, and `interrupt_value` is the natural shared fix; worth a ticket. The PR description also still describes the superseded round-1 design ("per-(event loop, thread_id) mutex", suffix-trim reducer semantics) — worth refreshing before merge since it's the durable record.

packages/orchestrator-agent/app/models/responses.py
aartaria: Confirmed and fixed in b050ad91 — you are right, and the hole was in my own fix rather than in the original code. Reproduced your case: `[approval, auth, approval]` has an approval last, so the mixed-set bail-out rendered one and left the other invisible.

The `len(approvals) != len(values)` guard is gone. The fold now applies whenever two or more approvals are pending, whatever else is pending with them — approvals can never hide behind each other. A non-approval left pending this way is not resumable by an approve (the resume map hands it the user's reply) and renders on the next turn. Test covers the three-interrupt ordering you used.

packages/agent-common/agent_common/middleware/loop_detection_middleware.py
aartaria: Fixed in b050ad91 — claim scoped, no code change, exactly as you suggested. Both the reducer docstring and the matching one in `graph_utils` now say deltas with *equal* caps commute, and spell out the mixed-cap case: an allowed call (cap `window_size`) and a blocked one (cap `None`) on the same key can differ by a single entry depending on order, costing one entry of escalation timing rather than a lost record.

packages/agent-common/tests/manual/README.md
aartaria: Fixed in b050ad91 — both the README and the PR description now describe the per-`thread_id` `threading.Lock`, including *why* per-loop keying was insufficient (the cross-loop `task`-dispatch path), so the superseded design is not left as a trap for anyone reproducing from it.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (2182fffa07d4c4ebbc42a73e8702eee044f9e1cc..b050ad916edd51d3308d46875ae27247e4d1a96f)

**Disposition of round-2 findings** — all three verified-fixed in b050ad91:

| # | Finding | Disposition |
|---|---|---|
| 1 | Mixed set with ≥2 approvals abandoned the fold; blanket approve replicated to the hidden approval | **verified-fixed** — the `len(approvals) != len(values)` guard is gone, the fold now applies whenever ≥2 approvals are pending, and the new test pins the exact `[approval, auth, approval]` ordering from the round-2 reproduction. Traced the resume path at this head: both folded approvals receive the blanket decision (both were on the rendered card), and the hidden non-approval falls to the non-HITL branch and resumes with the raw query — the fail-safe direction verified empirically in round 2. `tests/test_responses.py`: 12 passed |
| 2 | "Deltas commute" overstated for mixed caps | **verified-fixed** — both docstrings (reducer and `graph_utils`) now scope the claim to equal caps and spell out the mixed-cap ±1-entry case accurately |
| 3 | Stale per-`(event loop, thread_id)` gate description in the manual README | **verified-fixed** — README and PR description both describe the per-`thread_id` `threading.Lock` including why per-loop keying was insufficient |

**Delta review:** every previous finding is resolved and the delta (one commit, 53 insertions) introduces nothing new. The fold change was checked for edges — a single approval beside an auth still renders the auth (covered by test), client-action interrupts are correctly excluded from the approval filter, and per-interrupt decision alignment is unaffected. Both touched suites pass at this head (`test_responses.py` 12 passed, `test_ptc_parallel_eval.py` 13 passed). No inline comments this round.

Still on the follow-up list from earlier rounds (non-blocking, unchanged): the sub-agent path `dynamic_tool_dispatch.py` `sub_interrupts[0]` truncation (pre-existing, `interrupt_value` is the natural shared fix), and the deferred gate-across-summarizer latency note, to revisit if `_PTC_TURNS` is ever re-keyed by `(thread_id, tool_call_id)`.

PR #235, opened 2026-09-11
feat(gateway): failover chains per chat tier, and make retries reach production

### LINKED ISSUE #204: AWS Bedrock outage — ServiceUnavailableException causes hard failure in Nannos
## Problem
Users are hitting a `ServiceUnavailableException` from AWS Bedrock: "Bedrock is unable to process your request."

This was first reported via Nannos, then confirmed directly from the AWS Bedrock console using **Claude Sonnet 5**. When Bedrock is unavailable, Nannos fails hard with no fallback.

## Evidence
- Error type: `ServiceUnavailableException`
- Message: "Bedrock is unable to process your request."
- Affected model: Claude Sonnet 5

## Proposed Solution
Investigate and implement **LiteLLM routing** for automatic graceful degradation to another provider or frontier model when Bedrock is unavailable:
https://docs.litellm.ai/docs/routing

This would prevent hard failures for end users during provider outages.

## Reported by
Erik Jonsson Thorén & Andrea Artaria

aartaria:
closes #204

**Authored by Claude Code** (agent) in a design-grilling session with @aartaria. GitHub does not badge App attribution onto pull requests, so this line is the only signal — the commits and issue comments carry it in the audit record.

## What this is

Nannos already routes every LLM call through a LiteLLM proxy, so the issue's "investigate LiteLLM routing" is not greenfield — what was missing is routing *policy*. A Bedrock `ServiceUnavailableException` reached users as a hard failure because the gateway neither retried nor had anywhere to fall back to.

Failover is declared as **tier groups**: the existing chat tiers (`chat:low`, `chat`, `chat:premium`) gain an ordered list of aliases instead of exactly one. The head is the tier's default and keeps today's behaviour byte-for-byte; the tail is the failover chain, which console-backend projects onto the proxy via `POST /fallback`. That endpoint is DB-backed (`store_model_in_db`), so a chain lives beside the model registry and cannot drift against the aliases it names — unlike a chain written into `config.yaml`, which in Nannos deliberately carries no model knowledge at all. **LiteLLM executes the routing; the app only declares it.**

Two scope decisions, both deliberate:

- **Within-tier only.** A tier never chains into another tier: availability and cost/quality are separate axes, and an outage must not silently move a user between them.
- **Chat tiers only, enforced rather than documented.** A failed-over embedding call writes vectors from a *different embedding space* into the same pgvector index. Both sides are pinned to `EMBEDDING_DIMENSION`, so they insert cleanly and then silently degrade similarity search for every document embedded during the outage, permanently. For a chat turn another provider's answer beats an error; for an embedding write an error beats a corrupted index.

## Two further findings, both fixed here

**1. Retries and cooldowns had no single home.** `litellm_settings`/`router_settings` are config-only, and this repo ships no `config.yaml` — so the values existed once in the `scripts/start-local.sh` heredoc and again in whatever config each deployment mounts, kept in step only by comments asserting they matched. **`num_retries: 0` disables failover outright** — a chain is only tried once a deployment's retries are exhausted — so a deployment left there has failover on paper and none in practice. Settings now live in one committed, consumed file (`packages/litellm-proxy/litellm-settings.yaml`) that start-local.sh cats and deployments are expected to copy verbatim; `custom_logger.py` warns at startup when retries are disabled or no cooldown settings are present.

**2. `cache_control` stripping was a denylist**, so it failed open: any provider not named kept Anthropic markers. That already affected OpenAI-format deployments, and a *fallback* onto one would have carried them too — the safety net tearing at the exact moment it was deployed. It is now an allowlist keyed on the **resolved deployment** (ADR-0008). Review found the same argument applies to **`thinking_blocks`**, which travel inline in the already-in-flight request where nothing app-side can react, so those are stripped by the same hook.

**3. Same-alias multi-deployment was rejected as the mechanism.** `bedrock/eu.anthropic.*` is already a cross-region inference profile, so a second Bedrock deployment under one alias buys almost nothing; and one alias spanning two provider families makes `get_model_provider()` a coin flip, silently breaking thinking-block replay and cache-marker handling. Alias-to-alias chains keep `litellm_provider`, `model_info` and rate cards truthful by construction.

## Deployment note

Failover needs both halves, and only one of them is in this repo. The *chain* is DB-backed and configured in the console; the *behaviour that makes it fire* (`num_retries`, `allowed_fails`, `cooldown_time`) is config-only and must be present in the config each deployment mounts. **A deployment whose config still has `num_retries: 0` gains the mechanism from this PR but no actual failover** — the startup warning added here is what makes that state visible.

## Docs

- **ADR-0008** — provider-shape normalization belongs at the resolved deployment, not the requested alias.
- **AGENTS.md** — Model Gateway vocabulary gains *tier group*, and separates **failover** (runtime) from **alias degradation** (registry-time); both were previously called "graceful degradation" and share no mechanism.

## Review

Three automated rounds, 19 findings, all fixed and replied to on their threads; round 3 is clean. Round 1 caught a real build break (the candidate picker read a field `AvailableModel` does not have) that my own typecheck had missed, because the root `tsconfig.json` is solution-style — `tsc --noEmit -p tsconfig.json` checks zero files and exits 0. `tsc -b` is the gate, and the claim below reflects it.

One known cosmetic gap, left as-is: a *rejected* add (400, chain unchanged) does not trigger the length-keyed Select remount, so the failed alias stays in the trigger. The error toast and the unchanged list show the real state.

## Testing

- `packages/litellm-proxy`: **63 pass** (20 new — allowlist inversion incl. Bedrock geographic profiles, the `thinking_blocks` strip and its copy-on-write property, the unresolvable-provider warning, and the startup check).
- `packages/console-backend`: **1966 pass, 3 fail**. All three failures are pre-existing on `origin/main` and untouched here: `test_scheduler_engine.py::TestAttributionScope` references a `_make_job` helper that is defined nowhere (this PR touches no scheduler files). New coverage: 20 tier-group service tests plus **10 real-DB `pg_session` tests** for `replace_fallbacks` — ordered round-trip, a reorder reusing positions against the DEFERRABLE constraint, the same alias across two tiers, clearing, and the audit row's `before`/`after`.
- `console-frontend`: `tsc -b` clean for the changed files (19 errors remain repo-wide, all pre-existing in untouched files from the unbuilt `@nannos/embed-sdk` workspace dep); eslint clean on changed files. SDK regenerated from the exported OpenAPI spec.
- `ruff check` on console-backend: 19 errors before and after — this change adds none.

**Not verified end to end against a live proxy.** The `POST /fallback` projection and the actual failover path have not been exercised against a running gateway with two providers registered; that needs a stack with a second provider.

## Follow-up filed

- #237 — deleting a model leaves it as a tier's *default* (pre-existing; interacts with alias degradation, so deliberately not folded in here).

## Checklist
- [ ] Does this PR introduce a breaking change? 🚨 — no. Tiers with no chain configured behave exactly as today; the `cache_control` allowlist changes behaviour only for deployments that were already being sent markers they do not define.


alloy-cicd (commented):

The design is sound (DB-backed chains beside the registry, head-preserving tier groups), but three findings are load-bearing:

1. **The frontend build is broken** — `FailoverChains.tsx` reads `m.model_name` off `AvailableModel`, which has no such field; `tsc -b` fails, and even transpiled the candidate list is always empty, so no fallback can ever be added from the UI.
2. **`optional=True` on `_request` does not suppress the raise** — it only lowers the log level. LiteLLM's `DELETE /fallback/{model}` 404s when no entry exists, so re-pointing a tier default whose old head had no proxy-side chain aborts the reprojection (spurious warning, and in the drifted case the new head is left with no chain), and clearing an already-clear chain 502s after the DB write committed.
3. **`gateway_mismatch` cannot represent "unknown"** — a proxy that is down or erroring renders every tier as verified-in-sync, which is the exact silent-no-failover state the feature exists to surface.

Beyond the inline comments, two notes with no anchorable diff line:

- `example-k8s-deployment/base/litellm-proxy.yaml` (untouched by this PR) still inlines the old `litellm_settings` — no `modify_params`, no `router_settings`/`allowed_fails`/`cooldown_time` — and boots warning-free, since `resilience_warnings()` only checks `num_retries == 0` and config-file fallbacks. The single-home story holds only for local dev; detail in the comment on `litellm-settings.yaml`.
- The "embeddings must never fail over" rationale is copy-pasted in 8 places (AGENTS.md, DDL 095, two spots in `FailoverChains.tsx`, `model-gateway.ts`, test docstrings, proxy README, `_require_chat_tier`), and the "prose parity drifts" argument in 4. A future policy change has to find them all; one canonical home (the service guard's docstring or the ADR) plus one-line pointers would be safer.

Submitted as COMMENT; the human reviewer decides. This session will be resumed for follow-up rounds.

packages/console-frontend/src/components/admin/FailoverChains.tsx
```diff
@@ -0,0 +1,208 @@
...
+    mutationFn: ({ role, fallbacks }: { role: string; fallbacks: string[] }) =>
+      setTierFailoverChain(role, fallbacks),
+    onSuccess: () => {
+      toast.success('Failover chain updated on the gateway');
+      queryClient.invalidateQueries({ queryKey: ['gateway-tier-groups'] });
+    },
+    onError: (e: unknown) => toast.error(`Failover chain not saved: ${String(e)}`),
+  });
+
+  const candidates = (available ?? []).map((m) => m.model_name).filter(Boolean) as string[];
```
alloy-cicd: **Build break / feature-disabling bug.** `AvailableModel` has no `model_name` field — the generated type exposes `value`/`label`/`provider` (`types.gen.ts:260`). `tsc -b` fails here with TS2339 (`build` is `tsc -b && vite build`), and if transpiled anyway every entry maps to `undefined`, `.filter(Boolean)` empties `candidates`, and every tier permanently shows the disabled "No other models registered" button — no fallback can be added from the UI at all.

There is an existing hook for this exact query: `useAvailableModels()` in `src/config/models.ts:102` (same `['available-models']` key, tuned `staleTime`, maps `m.value`). Bonus: `/api/v1/models` filters to chat-mode models, which is exactly the property you want for chain candidates.

packages/console-frontend/src/components/admin/FailoverChains.tsx
```diff
@@ -0,0 +1,208 @@
...
+  });
+
+  const mutation = useMutation({
+    mutationFn: ({ role, fallbacks }: { role: string; fallbacks: string[] }) =>
+      setTierFailoverChain(role, fallbacks),
+    onSuccess: () => {
+      toast.success('Failover chain updated on the gateway');
+      queryClient.invalidateQueries({ queryKey: ['gateway-tier-groups'] });
+    },
+    onError: (e: unknown) => toast.error(`Failover chain not saved: ${String(e)}`),
```
alloy-cicd: Two issues in this error path:

1. The backend commits the DB write **before** projecting to the gateway (`replace_fallbacks` commits at `model_defaults_repository.py:159`) and returns 502 with detail "Chain saved, but the gateway rejected it". For that case this toast ("not saved") is factually wrong, and since `onError` doesn't invalidate `['gateway-tier-groups']`, the UI keeps showing the old chain while the DB holds the new one.
2. The api wrapper throws the raw `{detail: ...}` body (`model-gateway.ts:229`), so `String(e)` renders `[object Object]` — the backend's carefully worded 400/502 details never reach the admin. `getErrorMessage()` in `src/lib/utils.ts:12` (already used by SecretsVaultList, EmbedBindingPanel, …) handles exactly this shape.

packages/console-frontend/src/components/admin/FailoverChains.tsx
```diff
@@ -0,0 +1,208 @@
...
+ *     so there is exactly one place a tier's primary model is chosen.
+ *   - drift against the live proxy is shown as a warning. A chain that exists only in the
+ *     console is a failover that will not happen, which is the precise failure this feature
+ *     exists to prevent.
+ *
+ * Chat tiers only: the backend refuses embedding roles, because a failed-over embedding call
+ * writes vectors from another embedding space into the same index and quietly degrades search.
+ */
+
+const TIER_LABEL: Record<string, string> = {
```
alloy-cicd: `TIER_LABEL` is a second role→label map rendering on the same page as the existing `roleLabel()` (`ModelGatewayPage.tsx:232`), with different wording: the model cards say "low tier"/"premium tier" while this card says "Low"/"Premium" for the same tier, one screen apart. A future tier has to be added in both places or this card prints the raw role key. Suggest lifting `roleLabel` into a shared module (keyed on the `DefaultRole` union so a new role is a compile error) and consuming it here.

packages/console-frontend/src/components/admin/FailoverChains.tsx
```diff
@@ -0,0 +1,208 @@
...
+  candidates,
+  onSave,
+  saving,
+}: {
+  group: TierGroup;
+  candidates: string[];
+  onSave: (role: string, fallbacks: string[]) => void;
+  saving: boolean;
+}) {
+  const [adding, setAdding] = useState(false);
```
alloy-cicd: Minor UX/state: `adding` makes this a two-click interaction ("Add fallback" only swaps in a closed Select; the user must then open it — no `open`/`defaultOpen` is passed), and it's reset only in `onValueChange`, so an abandoned add leaves the empty Select trigger stuck in place of the button. Rendering the Select unconditionally (disabled when `selectable.length === 0`, placeholder "Add fallback") deletes the state and both branches.

packages/console-backend/console_backend/services/model_gateway_service.py
```diff
@@ -277,6 +277,49 @@ async def delete_model(self, model_id: str) -> None:
...
+            "POST",
+            "/fallback",
+            json={
+                "model": model_name,
+                "fallback_models": list(fallback_models),
+                "fallback_type": "general",
+            },
+        )
+
+    async def delete_fallbacks(self, model_name: str) -> None:
```
alloy-cicd, line 319: **`optional=True` doesn't do what this docstring claims.** `_request` only downgrades the log level (`log = logger.debug if optional else logger.error`, line 176) and still raises `ModelGatewayError` on any HTTP error. LiteLLM v1.90.0's `DELETE /fallback/{model}` returns **404** when no entry exists — exactly the "missing entry" case described here as a non-error.

Consequences:
- In `reproject_tier_group` (`model_defaults_service.py:150`), `delete_fallbacks(previous_head)` 404s whenever the old head had no proxy-side entry (a tier with an empty chain — the common launch state), aborting **before** `set_fallbacks(new_head, chain)`. Every such default change returns the spurious "chain was not re-declared" warning, and when the stored chain is non-empty (drift), the new head genuinely gets no chain on the gateway.
- Saving an empty chain when the proxy already holds none: `set_fallbacks(head, [])` → `delete_fallbacks` → 404 → HTTP 502 to the admin **after** the DB write committed.

Fix: have `_request` return `{}` instead of raising on 404 when `optional=True`, or catch-and-swallow the 404 here.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
```diff
@@ -544,3 +572,87 @@ async def delete_model(model_id: str, request: Request, user: User = Depends(req
...
+    gateway = get_model_gateway_service(request)
+    groups = await defaults_service.get_all_tier_groups(db)
+    chains = await defaults_service.repository.get_all_fallbacks(db)
+
+    out: list[TierGroup] = []
+    for role in CHAT_TIER_ROLES:
+        models = groups.get(role, [])
+        head = models[0] if models else None
+        stored = chains.get(role, [])
+        mismatch: list[str] | None = None
```
alloy-cicd: **"Unknown" and "in sync" are the same value.** The docstring above (line 591) promises that an unreachable proxy leaves the state "unknown, which is not the same as 'in sync'" — but on `ModelGatewayError` this `mismatch` stays `None`, identical to verified-in-sync, and `FailoverChains.tsx` badges only on `gateway_mismatch != null`. With the proxy down (precisely when an admin is checking failover), every tier renders healthy.

Needs a tri-state — e.g. `gateway_state: Literal['in_sync','drifted','unknown']`, or `gateway_mismatch` plus a `gateway_reachable` flag — and a distinct render in the frontend. Note the interaction with the `delete_fallbacks`/404 issue: the drift case "stored chain non-empty, proxy holds nothing" makes `get_fallbacks` raise, so today it lands here as silent "unknown" instead of showing as the mismatch it is.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
```diff
@@ -544,3 +572,87 @@ async def delete_model(model_id: str, request: Request, user: User = Depends(req
...
+    """Every chat tier with its failover chain, plus any drift against the live proxy.
+
+    The proxy is read per tier rather than trusted: a chain that exists only in our table is
+    a failover that will not happen, and that is exactly the failure this feature exists to
+    prevent, so it is surfaced instead of assumed away. A proxy that cannot be reached leaves
+    ``gateway_mismatch`` unset (unknown), which is not the same as "in sync".
+    """
+    defaults_service = get_model_defaults_service(request)
+    gateway = get_model_gateway_service(request)
+    groups = await defaults_service.get_all_tier_groups(db)
```
alloy-cicd, line 619: Duplicate read + layering: `get_all_tier_groups(db)` already merged the chains (it calls `repository.get_all_fallbacks` internally, `model_defaults_service.py:75`), so this router→`service.repository` call re-reads the same table and creates a second derivation path for the same fact — `stored` is exactly `models[1:]`. Dropping this call keeps the repository private to the service and removes a DB round-trip per page load.

Same function, minor: the per-tier `await gateway.get_fallbacks(head)` at line 606 serializes three live proxy reads — worst case 3× the 10s client timeout on a hung proxy. `asyncio.gather(..., return_exceptions=True)` would overlap them while keeping per-tier error handling.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
```diff
@@ -526,14 +529,39 @@ async def set_default(
...
+    except ModelGatewayError as e:
+        # The gateway's own message is logged, never returned: it can carry transport detail
+        # about an internal-only service, and the caller needs the *state* ("your chain is not
+        # live") rather than the cause, which an admin reads from the logs.
+        warning = (
+            "Default saved, but its failover chain was not re-declared on the gateway. "
+            "Re-save the chain to retry; see the console-backend logs for the gateway's response."
+        )
+        logger.error("Failed to reproject failover chain for role=%s after default change: %s", body.role, e)
+    return {"status": "ok", "model_name": alias, "default_for": body.role, "warning": warning}
```
alloy-cicd, line 566: This `warning` is unreachable by any UI: `setGatewayModelDefault` (`model-gateway.ts:193`) destructures only `error` and returns void, and both callers (`ModelGatewayPage.defaultMutation`, `WebSearchSettings`) toast plain success. A failed reprojection is therefore invisible to the admin except via the /tiers drift badge on the next load — which is itself blind when the proxy is unreachable (see the `gateway_mismatch` comment). Either wire the warning into the frontend toast, or drop the field and make the drift badge the single surface; two half-wired surfaces is the worst of both.

packages/console-backend/console_backend/services/model_defaults_service.py
```diff
@@ -52,3 +53,116 @@ async def set_default(self, db: AsyncSession, actor: User, role: str, model_alia
...
+            return
+        defaults = await self.get_all(db)
+        head = defaults.get(role)
+        if previous_head and previous_head != head:
+            still_in_use = any(defaults.get(other) == previous_head for other in CHAT_TIER_ROLES)
+            if not still_in_use:
+                await gateway.delete_fallbacks(previous_head)
+        if not head:
+            return
+        chain = await self.repository.get_fallbacks(db, role)
```
alloy-cicd, line 172: **Promoting an alias that is already in the tier's chain projects a self-referencing chain.** `set_default`/`upsert_default` never touch `model_tier_fallbacks`, so this reads the unmodified chain and projects `set_fallbacks(new_head, chain_containing_new_head)` — e.g. chain `[gpt, vertex]` on head `claude`, promote `gpt`, and the gateway now "fails over" from gpt to gpt, burning a fallback hop on the provider just found down. Afterwards, PUT /tiers validation (`seen = {head}`, lines 103-110) rejects any edit whose body still contains the new head, so adding a fallback or removing a *different* one 400s until the duplicated head itself is removed (the UI's remove button on that alias does repair it, but nothing tells the admin that).

Either strip the new head from the stored chain on promotion (persist + audit it) or reject the promotion with a clear 400.

packages/console-backend/console_backend/services/model_defaults_service.py
```diff
@@ -52,3 +53,116 @@ async def set_default(self, db: AsyncSession, actor: User, role: str, model_alia
...
+        seen: set[str] = {head}
+        for alias in aliases:
+            if alias in seen:
+                raise ValueError(
+                    f"'{alias}' appears twice in the '{role}' tier group; a chain must not "
+                    f"revisit a model it has already tried."
+                )
+            seen.add(alias)
+
+        registered = {m.get("model_name") for m in await gateway.list_models()}
```
alloy-cicd: This validates registration only, not mode: `gateway.list_models()` entries carry `model_info.mode` — the same field `models_router._to_available` uses to keep embeddings out of the chat picker — but it's unused here, so a direct API call can store and project an embedding alias as a chat fallback (the UI can't produce this, so API-level only). During a chat-provider outage the failover request would then hard-fail against an embedding deployment. `_require_chat_tier` enforces the invariant on the tier role but is silent on chain members; rejecting aliases whose mode is not chat closes it in one line.

packages/console-backend/sqlmigrations/ddl/095_create_model_tier_fallbacks.sql
```diff
@@ -0,0 +1,32 @@
...
+-- Chat tiers only, enforced in the service layer as well: failing an embedding call over to
+-- another model writes vectors from a different embedding space into the same pgvector index
+-- (both sides are pinned to EMBEDDING_DIMENSION, so they insert cleanly) and silently
+-- degrades similarity search for every document embedded during the outage, permanently.
+CREATE TABLE model_tier_fallbacks (
+    role TEXT NOT NULL,
+    -- chat-tier role: 'chat' (standard), 'chat:low', 'chat:premium'
+    alias TEXT NOT NULL,
+    -- 1-based rank within the tier's chain; the head (rank 0) is model_defaults.model_alias
+    position INTEGER NOT NULL CHECK (position >= 1),
```
alloy-cicd, line 24: No FK and no cleanup path: `DELETE /models/{model_id}` (`admin_model_gateway_router.py:567`, untouched by this PR) removes the deployment from the gateway but leaves matching rows in this table and leaves the proxy-side chains referencing the alias. A deleted alias thus stays in a tier's failover chain — breaking exactly when the primary is down — while the console keeps listing it; the registration check in `set_failover_chain` only rejects it on the *next* manual edit. Suggest `delete_model` also delete rows here and reproject affected tiers (or at minimum, the mismatch/drift surface should flag chain members that are no longer registered).

packages/console-backend/tests/test_model_tier_groups.py
```diff
@@ -0,0 +1,208 @@
...
+- a tier's chain is keyed proxy-side on its *head* alias, so re-pointing a tier's default has
+  to move the chain and drop the old head — unless that head still serves another tier.
+"""
+
+import pytest
+from console_backend.services.model_defaults_service import ModelDefaultsService
+from console_backend.services.model_gateway_service import ModelGatewayError
+
+
+class _FakeRepo:
```
alloy-cicd, line 20: Console-backend AGENTS.md mandates real-DB tests for mutations and audit logging ("Real Database Testing Over Mocking", "Use `pg_session` fixture", "Tests must verify audit logging behavior"), but the only coverage of the new `replace_fallbacks` — a DELETE+INSERT under a DEFERRABLE unique constraint, plus an audit row — is this `_FakeRepo` one-liner. The failure modes the rule names go untested: position reuse on reorder against the deferred `(role, position)` constraint, the `(role, alias)` PK collision, and whether the audit row is written with `before`/`after`. `tests/test_model_defaults_repository.py` already uses `pg_session` and is the natural home.

packages/litellm-proxy/custom_logger.py
```diff
@@ -164,22 +164,39 @@ def _sanitize_response_format(data: dict) -> None:
...
-    prefixes = _CACHE_CONTROL_STRIP_RULES.get(provider or "")
-    if prefixes is None:
-        return False
-    return not prefixes or model.startswith(prefixes)
+    """True unless the resolved deployment speaks Anthropic's ``cache_control`` idiom.
+
+    Unknown or undeterminable providers strip: losing a cache discount is recoverable and
+    shows up in billing, while forwarding markers a provider rejects fails the call.
+    """
+    keep_markers = _CACHE_CONTROL_KEEP_RULES.get(provider or "")
```
alloy-cicd, line 210: The allowlist inversion also inverted the *unresolvable-provider* case: the old denylist returned False (keep markers) when the provider couldn't be derived; this returns True and strips. The docstring and ADR-0008 present that as a deliberate fail-closed choice, so not contesting it — but two follow-ups: (a) no test covers the provider-`None` path (bare `eu.anthropic.claude-…` model string, no slash prefix, no `custom_llm_provider`) — the new tests all supply one or the other; (b) deployments registered before the console's `family/`-pinning existed, or written directly to the proxy DB, can hit that path and silently lose prompt caching on Anthropic models — a 5-10× input-cost regression with no signal. A warning log when markers are stripped for an unresolvable provider on an Anthropic-looking model would make it visible.

docs/adr/0008-provider-shape-normalization-at-the-resolved-deployment.md
```diff
@@ -0,0 +1,49 @@
...
+  had to change.
+
+## Consequences
+
+- A provider that honours Anthropic-style `cache_control` but is missing from the allowlist
+  silently loses prompt caching, which shows up as a cost regression rather than an error. The
+  allowlist has to be maintained as providers are added; the boot-time gateway settings check is
+  where that omission is meant to become visible.
+- `thinking_blocks` replay remains app-side and alias-keyed
+  (`_thinking_round_trip_provider`). It is not covered by this ADR's mechanism. Extended thinking
```
alloy-cicd: "Extended thinking is dropped on a fallback turn rather than replayed" is accurate only for an *app-side* model switch on a later turn. For the gateway-side failover this PR introduces, the failover happens inside a single request that already carries inline `thinking_blocks` (injected alias-keyed in `agent_common/core/model_factory.py:171-194`), and the proxy normalizes only `cache_control` — nothing drops the blocks, so a mid-conversation Anthropic→non-Anthropic failover forwards them to a provider that may reject unknown message fields, failing the very request the chain was meant to save. Worth correcting this sentence now, and considering a proxy-side `thinking_blocks` strip parallel to the `cache_control` one as a follow-up.

packages/litellm-proxy/litellm-settings.yaml
```diff
@@ -0,0 +1,62 @@
...
+# `general_settings` and NO `model_list`:
+#
+#   - `general_settings` names environment variables that legitimately differ per
+#     environment (local reads LITELLM_DATABASE_URL, k8s reads DATABASE_URL), so it stays
+#     with each consumer.
+#   - `model_list` is deployment-specific and, in Nannos, is not file-based at all: models
+#     are registered at runtime from the console and stored in the proxy DB
+#     (`store_model_in_db`). See the package README.
+#
+# It exists because these settings have no single home otherwise: they are config-only, and
```
alloy-cicd, line 15: The drift this header warns about is already present in-repo: `example-k8s-deployment/base/litellm-proxy.yaml`'s ConfigMap (lines 97-106, untouched by this PR) still carries the old `litellm_settings` — no `modify_params`, no `router_settings` at all (no `allowed_fails`/`cooldown_time`) — and it boots warning-free, because `resilience_warnings()` only flags `num_retries == 0` and config-file fallbacks. Anyone copying the example gets a proxy that never benches a failing deployment — close to the #204 failure this PR fixes. Suggest updating the example (or generating its ConfigMap from this file), and having the boot check also warn when cooldown settings are absent while DB-declared chains exist.

packages/console-frontend/src/components/admin/FailoverChains.tsx
aartaria: Fixed in cb42b210: now uses `useAvailableModels()` and maps `m.value`. You are right on both counts — `tsc -b` failed, and the candidate list would have been empty regardless. Worth recording why I missed it: I verified with `tsc --noEmit -p tsconfig.json`, and the root tsconfig is solution-style (`"files": []` + references), so that command typechecks zero files and exits 0. `tsc -b` is the real gate.

packages/console-frontend/src/components/admin/FailoverChains.tsx
aartaria: Fixed in cb42b210, both parts. The toast now uses `getErrorMessage(e)` and no longer asserts "not saved" — the write may well have committed — and `onError` invalidates `gateway-tier-groups` so the row repaints with its real drift state instead of showing a chain that is no longer what is stored.

packages/console-frontend/src/components/admin/FailoverChains.tsx
aartaria: Fixed in cb42b210: `roleLabel` now lives in `api/model-gateway.ts`, keyed on `Record<DefaultRole, string>` so a new role is a compile error, and both `ModelGatewayPage` and this card consume it. The local copy is gone.

packages/console-frontend/src/components/admin/FailoverChains.tsx
aartaria: Fixed in cb42b210: the Select renders unconditionally (disabled when nothing is selectable, with the placeholder carrying the "No other models registered" wording), and the `adding` state is deleted along with both branches.

packages/console-backend/console_backend/services/model_gateway_service.py
aartaria: Fixed in cb42b210. Confirmed `optional=True` only lowers the log level. Rather than change its semantics — other callers rely on the raise to fall through to the next source — `ModelGatewayError` now carries `status_code`, and `delete_fallbacks`/`get_fallbacks` swallow 404 specifically: for these two routes absent is the desired end state, not a failure.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
aartaria: Fixed in cb42b210: `TierGroup` now carries `gateway_state: Literal["in_sync","drifted","unknown"]`, set to `unknown` on `ModelGatewayError` or when a tier has no head, and `FailoverChains` renders a distinct "Gateway unreachable" badge with its own copy. Your note about the interaction was right too: with `get_fallbacks` now returning `[]` on 404 instead of raising, the "stored chain non-empty, proxy holds nothing" case reports as `drifted` rather than disappearing into `unknown`.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
aartaria: Fixed in cb42b210: the `service.repository` read is gone (`stored` is `models[1:]`, as you say), and the per-tier proxy reads go through `asyncio.gather(..., return_exceptions=True)` with the per-tier handling preserved — so a hung proxy costs one client timeout, not three.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
aartaria: Fixed in cb42b210 by wiring it rather than dropping it. `setGatewayModelDefault` now returns the body, and `defaultMutation.onSuccess` raises a `toast.warning` when `warning` is set. Agreed that two half-wired surfaces was the worst option — the drift badge alone would have left the admin with no signal until they happened to open the failover card.

packages/console-backend/console_backend/services/model_defaults_service.py
aartaria: Fixed in cb42b210: `reproject_tier_group` strips the newly-promoted head from the chain and persists the correction through `replace_fallbacks` (so it is audited), rather than only projecting the repair. Chose repair over a 400 because promotion is a legitimate action and the duplicate is our bookkeeping, not the admin's mistake. Covered by `test_promoting_an_alias_already_in_the_chain_removes_it_from_the_chain`.

packages/console-backend/console_backend/services/model_defaults_service.py
aartaria: Fixed in cb42b210: chain members are now checked against the chat-mode aliases from `list_models()` (treating an absent `mode` as chat, since `/model/info` omits it for plain chat deployments). Tests cover both the rejection and the absent-mode case.

packages/console-backend/sqlmigrations/ddl/095_create_model_tier_fallbacks.sql
aartaria: Fixed in cb42b210 with the stronger of the two options: `delete_model` reads the alias before deleting, then calls a new `drop_alias_from_chains`, which removes it from every chain and reprojects the affected tiers. Best-effort and logged — the deployment is gone either way, so a cleanup failure must not report the delete as failed. No FK was added: the chains reference a gateway-side alias, not a local row, so there is nothing to point a constraint at.

packages/console-backend/tests/test_model_tier_groups.py
aartaria: Fixed in cb42b210. You are right that this violated the repo rule, and the mocked test covered none of the failure modes the rule exists for. Six real-DB tests added to `test_model_defaults_repository.py` on `pg_session`: ordered round-trip, a reorder that reuses positions (which only commits because the constraint is DEFERRABLE), the same alias in two tiers against the (role, alias) PK, clearing, `get_all_fallbacks`, and the audit row's `before`/`after`.

packages/litellm-proxy/custom_logger.py
aartaria: Both addressed in cb42b210. (a) Added `test_unresolvable_provider_on_an_anthropic_model_warns` and a negative case, covering the bare `eu.anthropic.claude-…` path with neither prefix nor `custom_llm_provider`. (b) That exact case now logs a warning naming the model and telling the operator to register it with an explicit provider prefix — you are right that a silent 5-10x input-cost regression is not an acceptable way to fail closed, even when failing closed is the right direction.

docs/adr/0008-provider-shape-normalization-at-the-resolved-deployment.md
aartaria: Fixed in cb42b210, and this was the most useful finding in the round. The sentence was true for an app-side switch on a later turn and simply wrong for the gateway-side failover this PR introduces. Rather than only correct the prose, I took the follow-up: `_strip_thinking_blocks` now runs in the same deployment hook, on the same predicate, with the same copy-on-write discipline — so the Anthropic attempt keeps its blocks and the fallback attempt does not carry a field the fallback provider may reject. The ADR consequence now describes that, and notes the app-side decision stays correct because only the *serving* provider can differ.

packages/litellm-proxy/litellm-settings.yaml
aartaria: Fixed in cb42b210. The example ConfigMap now carries the `litellm_settings`/`router_settings` blocks verbatim from the shared file (with a comment saying to keep them in step and preferring generation), including `modify_params`, `allowed_fails` and `cooldown_time`. The boot check also warns when neither cooldown setting is present — gated on the config actually having a `litellm_settings` block, so an empty or unrelated dict stays quiet. I did not gate it on "DB-declared chains exist" as suggested: the check is a pure function over the parsed config with no DB access at boot, and the warning is just as true for a deployment that has not declared a chain yet.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (defde8ee4fe8021562d9d9b780578d8df4a8de5d..cb42b2103fd5c891c5b39b97073f0d57e773e76a)

**Ledger of round-1 findings — all 15 verified-fixed, none rejected:**

| Finding | Disposition |
|---|---|
| `m.model_name` on `AvailableModel` (build break) | verified-fixed — `useAvailableModels()` + `m.value`; `tsc -b` now passes on this file (remaining errors are the pre-existing embed-sdk workspace-link noise) |
| `optional=True` still raises; 404 aborts reprojection | verified-fixed — `ModelGatewayError.status_code`, 404 swallowed in `delete_fallbacks`/`get_fallbacks` specifically; semantics of `optional` untouched for other callers |
| `gateway_mismatch` conflates unknown/in-sync | verified-fixed — tri-state `gateway_state` end to end, distinct "Gateway unreachable" badge; the drift case "stored non-empty, proxy empty" now reports `drifted` since `get_fallbacks` returns `[]` on 404 |
| Promoted head inside its own chain | verified-fixed — head stripped, persisted via `replace_fallbacks` (audited), covered by test |
| Chain members not validated as chat-mode | verified-fixed (but see the new `"mode": null` comment below) |
| `delete_model` leaves stale chain entries | verified-fixed — `drop_alias_from_chains`, best-effort with logging (but see the new 500-regression comment below) |
| Reprojection `warning` discarded | verified-fixed — returned body + `toast.warning` |
| Error toast wrong for 502-after-commit / `[object Object]` | verified-fixed — `getErrorMessage` + invalidate on error |
| No real-DB/audit tests for `replace_fallbacks` | verified-fixed — six `pg_session` tests incl. deferred-constraint reorder and audit before/after |
| ADR thinking_blocks sentence | verified-fixed — and upgraded: proxy-side `_strip_thinking_blocks` in the same hook, copy-on-write, tested |
| Provider-None cache strip silent + untested | verified-fixed — warning log + positive/negative tests |
| Example k8s ConfigMap drifted | verified-fixed — verbatim blocks + boot warning when both cooldown settings are absent |
| Duplicate fallbacks read + serialized proxy calls | verified-fixed — `stored = models[1:]`, `asyncio.gather` |
| Duplicate role-label map | verified-fixed — shared `roleLabel` keyed on the `DefaultRole` union |
| `adding` state (two clicks, stuck trigger) | verified-fixed — Select rendered unconditionally (but see the new stale-value comment below) |

**The delta introduces 4 new issues** (2 correctness, 2 minor), inline below. Non-blocking follow-up, noted once: deleting a model that is a tier's **default** still leaves `model_defaults` pointing at the gone alias — pre-existing behavior predating this PR, and `drop_alias_from_chains` correctly scopes itself to chains; flagging it only so it can become a ticket rather than a surprise.

packages/console-backend/console_backend/services/model_defaults_service.py
```diff
@@ -52,3 +55,172 @@ async def set_default(self, db: AsyncSession, actor: User, role: str, model_alia
...
+        if unknown:
+            raise ValueError(
+                f"Not registered on the gateway: {', '.join(sorted(unknown))}. "
+                f"Register a model before routing traffic to it."
+            )
+        # Mode, not just existence: _require_chat_tier guards the tier's ROLE, but said nothing
+        # about the chain's members, so an API caller could route chat traffic onto an embedding
+        # deployment — which fails hard at the moment the chain is finally needed.
+        chat_aliases = {
+            m.get("model_name") for m in deployments if (m.get("model_info") or {}).get("mode", "chat") == "chat"
```
alloy-cicd: `.get("mode", "chat")` treats a present-but-`null` mode as non-chat: the default only applies when the key is **absent**, so `{"model_info": {"mode": null}}` yields `None != "chat"` → rejected as "Not chat models". That breaks parity with `_to_available` (`models_router.py:54`: `if info.get("mode") and info.get("mode") != "chat"`), whose falsy-tolerant guard is itself evidence that gateway entries do carry `mode: null` — so the UI picker can offer exactly the candidate this check then 400s. The new test covers "no `model_info` at all" but not this shape.

Suggest `(m.get("model_info") or {}).get("mode") or "chat"` (mirroring `_to_available`'s semantics), plus a test with `model_info: {"mode": None}`.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
```diff
@@ -526,21 +530,165 @@ async def set_default(
...
+async def delete_model(model_id: str, request: Request, db: DbSession, user: User = Depends(require_admin)):
+    """Remove a model from the gateway. The Rate Card is left for historical billing.
+
+    The alias is also dropped from any tier's failover chain: leaving it there would have the
+    gateway fail over to a model it no longer serves, breaking at exactly the moment the primary
+    is down. Read the alias *before* deleting — afterwards the deployment is gone and there is
+    nothing left to map the id to a name.
+    """
+    gateway = get_model_gateway_service(request)
+    model = await gateway.get_model_by_id(model_id)
```
alloy-cicd: Regression in the fix: `get_model_by_id` → `list_models` → `_request`, which raises `ModelGatewayError("Gateway unreachable")` on a transport failure (`model_gateway_service.py:205`) — and this call sits **before** the `try`. With a cold cache and an unreachable proxy, DELETE now 500s where it previously returned a clean 502. Move this lookup inside the `try` (or wrap it separately); since the cleanup is already best-effort, a failed lookup could also just skip the chain cleanup with a log line rather than fail the delete.

packages/console-frontend/src/components/admin/FailoverChains.tsx
```diff
@@ -0,0 +1,217 @@
...
+export function FailoverChains() {
+  const queryClient = useQueryClient();
+
+  const { data: groups, isLoading } = useQuery({
+    queryKey: ['gateway-tier-groups'],
+    queryFn: listTierGroups,
+  });
+  // The shared hook: same query key, tuned staleTime, and /api/v1/models already filters to
+  // chat-mode models — exactly the candidates a chat tier may fall back to.
+  const { models: available } = useAvailableModels();
```
alloy-cicd: One property of `useAvailableModels()` works against this consumer: it falls back to the static `MODEL_OPTIONS` list when the gateway list is empty or unreachable ("pickers always render something", `config/models.ts`). Here that means the add-fallback picker offers aliases that are not registered on the gateway — precisely during an outage, next to the "Gateway unreachable" badge — and picking one 400s "Not registered on the gateway". The backend guard holds, so this is minor, but for a failover chain an empty candidate list is the truthful render. Consider exposing the raw (un-fallback'd) list from the hook, or filtering candidates against something gateway-sourced.

packages/console-frontend/src/components/admin/FailoverChains.tsx
```diff
@@ -0,0 +1,217 @@
...
+          </li>
+        ))}
+      </ol>
+
+      {/* Always rendered: gating it behind an "Add fallback" button made this a two-click
+          interaction (the button only swapped in a closed Select) and stranded an empty
+          trigger whenever an add was abandoned. */}
+      <Select
+        disabled={saving || selectable.length === 0}
+        onValueChange={(alias) => onSave(group.role, [...chain, alias])}
```
alloy-cicd, line 143: Minor: the Select is uncontrolled, so after a pick Radix keeps the value — the trigger then displays the just-added alias instead of the "Add a fallback…" placeholder, even though that alias has left `selectable` (it's now in the chain). Passing `value=""` to keep it controlled-empty, or keying the Select on `chain.length`, resets it after each add.

packages/console-backend/console_backend/services/model_defaults_service.py
aartaria: Fixed in bd68e1b2: `((m.get("model_info") or {}).get("mode") or "chat") == "chat"`. Correct catch — I reached for `.get(key, default)` without checking whether null was a real value in this data, and `_to_available`'s falsy guard is exactly the evidence that it is. Covered by `test_an_explicit_null_mode_is_treated_as_chat`.

packages/console-backend/console_backend/routers/admin_model_gateway_router.py
aartaria: Fixed in bd68e1b2: the lookup moved inside the existing `try`, so an unreachable proxy is a 502 again. I had put it outside to make the ordering ("read the alias before the delete removes it") visually obvious and gave no thought to what it raises on a cold cache.

packages/console-frontend/src/components/admin/FailoverChains.tsx
aartaria: Fixed in bd68e1b2 — and this partly walks back the round-1 suggestion, so worth being explicit. The component now issues the same query with the same `[available-models]` key and staleTime rather than going through the hook, so the cache is still shared and there is no extra request, but it does not inherit the static `MODEL_OPTIONS` fallback. You are right that offering unregistered aliases beside a "Gateway unreachable" badge is the wrong render; an empty picker is the honest one, and the backend guard stays as the backstop rather than the first line of defence.

packages/console-frontend/src/components/admin/FailoverChains.tsx
aartaria: Fixed in bd68e1b2: the Select is remounted on chain change (`key={chain.length}`), so the trigger returns to its placeholder after each add. Kept it uncontrolled — a controlled empty value is awkward in Radix Select, and the remount is the smaller change.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (cb42b2103fd5c891c5b39b97073f0d57e773e76a..bd68e1b268bad622b5bd4bfa740f9c1fe6083d3e)

**Ledger of round-2 findings — all 4 verified-fixed, none rejected:**

| Finding | Disposition |
|---|---|
| `mode: null` rejected as non-chat (`.get("mode", "chat")`) | verified-fixed — `(... .get("mode") or "chat")`, covered by `test_an_explicit_null_mode_is_treated_as_chat` with exactly the `model_info: {"mode": None}` shape; all 20 tier-group tests pass |
| `delete_model` 500s on unreachable proxy (lookup outside `try`) | verified-fixed — `get_model_by_id` moved inside the existing `try`, 502 restored |
| Static `MODEL_OPTIONS` fallback feeding unregistered candidates | verified-fixed — direct `useQuery` on the same `['available-models']` key/staleTime (cache still shared, no extra request) without inheriting the static fallback; an empty picker is the honest render |
| Uncontrolled Select shows stale value after each add | verified-fixed — remounted via `key={chain.length}` |

**Every previous finding is resolved and the delta introduces nothing new.** No inline comments this round. I re-verified the delta files typecheck under `tsc -b` (clean; the remaining repo errors are the pre-existing embed-sdk workspace-link noise) and ran the tier-group unit suite at head (20 passed).

One cosmetic non-blocking note, not worth a thread: `key={chain.length}` resets the Select only when the chain actually changes, so a *rejected* add (400 — chain unchanged) leaves the failed alias displayed in the trigger; the error toast and the unchanged list make the state clear, and a controlled reset in `onSettled` would tidy it if the file is touched again.

PR #240, opened 2026-09-14
fix(agent-common, agent-runner): express every recursion budget in model calls

### LINKED ISSUE #239: Recursion budgets: unify on model calls and split the shared MAX_RECURSION_LIMIT (agent-common, agent-runner)
## Summary

#210 fixed the orchestrator's turn budget by expressing it in **model calls** (`ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`) and deriving the LangGraph super-step limit from the compiled graph. That fix was deliberately scoped to orchestrator-agent. The other consumers still express their bound directly in **super-steps**, via a single shared env name.

The result is that the *same sub-agent graph* carries two different, arbitrary bounds depending on how it is reached, and the knob the codebase tells operators to use doesn't reach one of the two paths.

## Evidence: identical graph, different bounds

Both call sites compile the same graph through `build_sub_agent_graph` ([graph_utils.py:2127](packages/agent-common/agent_common/core/graph_utils.py#L2127)) and then bind a recursion limit:

| path | call site | bound | env var read |
|---|---|---|---|
| delegated in-orchestrator | [dynamic_agent.py:1572](packages/agent-common/agent_common/agents/dynamic_agent.py#L1572) | **75** | `SUB_AGENT_RECURSION_LIMIT`, falling back to `MAX_RECURSION_LIMIT` ([:124](packages/agent-common/agent_common/agents/dynamic_agent.py#L124)) |
| scheduled job (agent-runner) | [core.py:1066](packages/agent-runner/agent/core.py#L1066) | **50** | `MAX_RECURSION_LIMIT` only ([:95](packages/agent-runner/agent/core.py#L95)) |

Same node shape, so the difference isn't tuning — it's two constants that drifted.

## The documented knob doesn't reach agent-runner

The legacy warning at [config.py:76-84](packages/orchestrator-agent/app/models/config.py#L76-L84) tells operators:

> To decouple them, set `SUB_AGENT_RECURSION_LIMIT` for the sub-agents and leave `MAX_RECURSION_LIMIT` to the sibling services.

`SUB_AGENT_RECURSION_LIMIT` is read in exactly one place — `dynamic_agent.py:124`. agent-runner never reads it. An operator following that advice raises the in-process sub-agents and silently leaves scheduled runs of the *same* agent at the tighter of the two bounds.

## Observed symptom

On a local stack with neither variable set (so both defaults apply), a scheduled job pointed at the `github-alloy-ch-org` sub-agent dies reproducibly:

```
langgraph.errors.GraphRecursionError: Recursion limit of 50 reached without hitting a stop condition.
  at packages/agent-runner/agent/core.py:1105 (_run_graph)
```

~33s, asking only for a repo description and default branch. At #210's measured ~8 nodes per lap, 50 super-steps is roughly **six model calls** — and the run spends some of those resolving 92 MCP tools. The same agent answers the same question fine when delegated from a conversation, where it gets 75.

Note this is *not* an approval-gating problem: agent-runner passes no `hitl_guarded_tools` to the builder, so a scheduled sub-agent has no HITL interrupt to park on. It simply did honest work and ran out of budget.

## Why the metric, not just the numbers

The argument for deriving is already written down in [step_budget.py](packages/orchestrator-agent/app/core/step_budget.py), and none of it is orchestrator-specific:

> the limit came to be 50 — a sensible number of model calls and a crippling number of steps

> it changes whenever a middleware carrying model hooks is added or removed, and nothing forces a hand-maintained constant to be updated with it

`dynamic_agent`'s 75 and agent-runner's 50 have exactly the same decay property. Any middleware added to `build_sub_agent_graph` silently tightens both, invisibly, with no test that would notice.

Expressing every bound in model calls would also make the numbers *comparable* — right now you cannot tell at a glance whether a sub-agent is more or less constrained than the orchestrator's 25.

## Proposed change

1. Move `step_budget.py` from orchestrator-agent into agent-common. The derivation is already generic — it classifies nodes by hook suffix off any compiled graph, with no orchestrator coupling. agent-runner already depends on `agent-common` as an editable path dep and imports from it.
2. Have `dynamic_agent` and agent-runner express their bound in model calls and derive the super-step limit, as the orchestrator does.
3. Give each consumer its own env name, so one variable no longer serves four.

## Caveats worth deciding up front

- **Conversion is not arithmetic.** The exchange rate is per-graph; a sub-agent's middleware stack isn't the orchestrator's, so the multiplier differs. Porting 50 and 75 by multiplication would bake in the same implied rate the change exists to remove. Sub-agent budgets should be chosen deliberately — which makes this a **behaviour change**, and it should not ride along inside an unrelated PR.
- **The shared name is a distinct defect from the shared unit.** One env var for four consumers stays wrong even if all four count model calls. Unifying the metric without splitting the names just makes the coupling less visible.
- **"All" is likely three of four.** `ringier-a2a-sdk` is an external published package and can't move on the same schedule, so the legacy name and its warning need to survive in some form.
- Worth confirming `build_sub_agent_graph`'s nodes follow the `<Middleware>.<hook>` naming the classifier keys on. The unknown-node fallback charges at the per-model-call rate, so the failure mode is a generous budget rather than a squeeze, but it should be checked rather than assumed.

## Context

Found while QA'ing #236 (ADR-0008, local sub-agents behind an in-process A2A server). Unrelated to that PR — the failing path is agent-runner, which never touches the orchestrator's A2A server.

aartaria:
closes #239

_Authored by Claude Code (agent), driven by @aartaria. Please review as agent-written code._

## What

#210 fixed the orchestrator's turn budget by expressing it in **model calls** and deriving the LangGraph super-step limit from the compiled graph, scoped deliberately to orchestrator-agent. This does the same for the two sub-agent paths and splits the shared env name.

The reported defect: `dynamic_agent` and `agent-runner` compile the **same** `build_sub_agent_graph` output and each bound its own hand-written super-step constant — 75 and 50. Not tuning; two numbers written down separately and left to drift. A scheduled job died reproducibly on `GraphRecursionError` where the identical sub-agent, delegated from a conversation, answered fine. And the knob the codebase advertised for decoupling them (`SUB_AGENT_RECURSION_LIMIT`) was read in exactly one of the two places, so an operator following that advice raised the in-process sub-agents and silently left scheduled runs at the tighter bound.

## Changes

1. **`step_budget.py` moves from orchestrator-agent into agent-common.** The derivation classifies nodes by hook suffix off any compiled graph and was never orchestrator-specific. Its unit tests move with it.
2. **`dynamic_agent` and `agent-runner` state their budget in model calls** and derive `recursion_limit` from their own compiled graph at the bind site — per sub-agent, since `extra_middlewares` (HITL, tool catalog, PTC) vary.
3. **Each consumer gets its own env name**: `SUB_AGENT_MAX_MODEL_CALLS_PER_TURN`, `AGENT_RUNNER_MAX_MODEL_CALLS_PER_TURN`, alongside the existing `ORCHESTRATOR_MAX_MODEL_CALLS_PER_TURN`. One name for four consumers was a defect distinct from the shared unit — unifying the metric without splitting the names would only have made the coupling less visible.
4. **The clamping, fallbacks and retired-name warnings are shared** via `resolve_max_model_calls()`, instead of living only in orchestrator config.

## On the caveats raised in the issue

- **Conversion is not arithmetic** — agreed, and not done. Both sub-agent paths default to **25 model calls**, matching the orchestrator, chosen rather than converted: the exchange rate is per-graph, so multiplying 75 or 50 would bake in the implied rate the change exists to remove. The argument for 25 is that a sub-agent is where the tool loop actually runs (discovery, retries, several rounds against an MCP server), so it has no business being more constrained than the planner delegating to it. **This is a behaviour change** and the reason this is its own PR.
- **Three of four, not all four** — confirmed. `ringier-a2a-sdk` is published externally and is untouched; it keeps reading `MAX_RECURSION_LIMIT` at 50.
- **Node naming** — checked rather than assumed. `build_sub_agent_graph` follows the `<Middleware>.<hook>` convention the classifier keys on, with an empty `unclassified` bucket on both middleware configurations. That is now a test, not a one-off observation.

## Operator-facing

`MAX_RECURSION_LIMIT` and `SUB_AGENT_RECURSION_LIMIT` are read by nothing in this repo any more. Setting either logs a warning naming what to set instead — and says explicitly that `MAX_RECURSION_LIMIT` is **not** safe to unset on the strength of that warning, because `ringier-a2a-sdk` still reads it. Neither is honoured here: inheriting a super-step value is the bug.

## Tests

- **agent-common** — the derivation's unit tests, plus marginal-cost and every-node-classified measurements against a real `build_sub_agent_graph` on both middleware configurations, plus a wiring test that drives the shipped `_build_graph` (not a re-implementation of it).
- **agent-runner** — pins that its default *is* the sub-agent default rather than a second number free to drift, and that `MAX_RECURSION_LIMIT=50` (the value that produced the observed failure) no longer moves it.
- **orchestrator-agent** — keeps what is specific to its own graph and env name.

Full suites green in all three packages: agent-common 1060 passed (9 pre-existing failures in `test_thinking_blocks_round_trip.py` and one `foundry_sdk` import error, both verified present on `main`), agent-runner 48 passed, orchestrator-agent 1027 passed. `ruff check` clean on every touched file.

## Checklist
- [x] Does this PR introduce a breaking change? 🚨 — **yes, deliberately.** Sub-agent turn budgets change (both paths now 25 model calls, replacing 75 and 50 super-steps), and `MAX_RECURSION_LIMIT` / `SUB_AGENT_RECURSION_LIMIT` stop having any effect on this repo's services. Deployments setting either will see a warning and should move to the per-service names.


packages/agent-common/agent_common/core/step_budget.py
```diff
@@ -136,3 +151,105 @@ def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
     buckets = classify_nodes(graph)
     _warn_unclassified(buckets)
     return _base_steps(buckets) + _steps_per_model_call(buckets) * max_model_calls
```
alloy-cicd: **`recursion_limit_for` has no floor, and a degenerate graph binds `recursion_limit=0`.** Verified: `recursion_limit_for(MagicMock(), 25)` returns `0` (a `MagicMock`'s `.nodes` iterates empty). This is already live in the existing suite — `tests/test_dynamic_agent.py:132/146/171` and `tests/test_dynamic_agent_skills.py:109` patch `build_sub_agent_graph` to return a `MagicMock`, so `_build_graph` in those tests now silently binds `recursion_limit=0` where it previously bound the constant 75. In production the same shape (a future non-Pregel graph type, or an already-wrapped runnable) would either raise `AttributeError` on a hot path or make every turn die on its first super-step.

Suggest a guard: `max(1, ...)` plus an explicit check that `core`/`per_model_call` is non-empty (or raising on an all-empty classification), so the degenerate case is a diagnosable error instead of a zero.

packages/agent-common/tests/test_step_budget.py
```diff
@@ -0,0 +1,397 @@
...
+    script = [_tool_turn(i) for i in range(model_calls - 1)] + [AIMessage(content="Done.")]
+    graph = _compiled_graph(
+        _ScriptedModel(responses=script),
+        exclude_deep_agents_middlewares=exclude_deep_agents_middlewares,
+    )
+
+    steps = 0
+    async for _ in graph.astream(
+        {"messages": [HumanMessage("go")]},
+        config={"configurable": {"thread_id": f"budget-{exclude_deep_agents_middlewares}-{model_calls}"}},
```
alloy-cicd: **`_measure_super_steps` runs with no `recursion_limit` bound, so LangGraph's default of 25 super-steps silently caps the measurement.** `_compiled_graph` returns the bare `build_sub_agent_graph` output (unlike the orchestrator suite, whose `_create_graph` applies `.with_config`). Today the deep-agents stack costs 3 + 4/call, so the n=4 run is ~19 steps — six steps under the default. One added middleware with model hooks pushes it over, and `test_marginal_cost_of_a_model_call_matches_the_derivation` dies with `GraphRecursionError` inside the `async for`, looking like a derivation break when the derivation is fine.

Pass `"recursion_limit": <something large>` in the config here so the measurement can never be truncated by the thing it measures.

packages/agent-common/tests/test_step_budget.py
```diff
@@ -0,0 +1,397 @@
...
+    runnable._cached_hitl_guarded = None
+    runnable._cached_context_gated_tools = None
+
+    bound = runnable._build_graph()
+
+    assert bound.config is not None
+    assert bound.config["recursion_limit"] == recursion_limit_for(graph, DEFAULT_SUB_AGENT_MAX_MODEL_CALLS)
+    # Not the retired constant: the point is that the number is no longer written
+    # down at the bind site, so a middleware change moves it.
+    assert bound.config["recursion_limit"] != 75
```
alloy-cicd: **`!= 75` holds by coincidence.** Measured: the lean stack (`exclude_deep_agents_middlewares=True`) derives per_model_call=3, base=0, so `recursion_limit_for(graph, 25) == 75` — exactly the retired constant this assertion says can no longer appear. The test happens to use the deep-agents graph (103), but if the deep stack ever loses one per-call hook plus its per-turn hooks, a *correct* derivation would fail this assertion — and conversely a genuine regression to a literal 75 on a lean stack would pass. The intent ("the number is derived, not written down") is already fully covered by the equality assertion on the previous line; suggest dropping this one.

packages/agent-common/tests/test_step_budget.py
```diff
@@ -0,0 +1,397 @@
...
+    runnable._cached_tools = []
+    runnable._cached_system_prompt = "p"
+    runnable._cached_response_format = None
+    runnable._cached_hitl_guarded = None
+    runnable._cached_context_gated_tools = None
+
+    bound = runnable._build_graph()
+
+    assert bound.config is not None
+    assert bound.config["recursion_limit"] == recursion_limit_for(graph, DEFAULT_SUB_AGENT_MAX_MODEL_CALLS)
```
alloy-cicd: **Ambient-env coupling:** the expectation is computed from `DEFAULT_SUB_AGENT_MAX_MODEL_CALLS`, but the bind site under test uses `_SUB_AGENT_MAX_MODEL_CALLS`, resolved from `SUB_AGENT_MAX_MODEL_CALLS_PER_TURN` at import time (`dynamic_agent.py:136`). With that env var set in the developer's shell — the exact tuning knob this PR introduces — this test fails, and because the constant is captured at import, no `monkeypatch.setenv` inside the test can reconcile the two. Either assert against `_SUB_AGENT_MAX_MODEL_CALLS` (what the bind site actually uses), or scrub the env var in a conftest.

packages/agent-common/agent_common/core/step_budget.py
```diff
@@ -136,3 +151,105 @@ def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
...
+
+    Ignoring it silently is the worse failure: an operator who pinned
+    ``MAX_RECURSION_LIMIT=50`` to tame one service would find it quietly doing
+    nothing here, with no line in the logs pointing at the variable they set.
+    """
+    for name in (LEGACY_RECURSION_LIMIT_ENV, LEGACY_SUB_AGENT_RECURSION_LIMIT_ENV):
+        raw = os.getenv(name)
+        if not raw or not raw.strip():
+            continue
+        logger.warning(
```
alloy-cicd, line 301: **The legacy warning emits the identical body for whichever variable is set, and fires once per consumer rather than once per process.** Two consequences:

1. An operator who set only `SUB_AGENT_RECURSION_LIMIT` is told, in one line, that `MAX_RECURSION_LIMIT` "is still read by ringier-a2a-sdk … so unsetting it may change an externally built agent's bound" and that `SUB_AGENT_RECURSION_LIMIT` "is read by nothing and can be removed" — advice about a variable they never set, formatted symmetrically with advice about the one they did. It also weakens the pinned-wording tests: both legacy names and "ringier-a2a-sdk" appear in *every* message, so `test_a_retired_super_step_env_var_is_ignored_but_warned_about` and `test_the_legacy_warning_does_not_orphan_the_sdk` cannot distinguish the two branches.
2. The orchestrator process calls `resolve_max_model_calls` at import time from both `app/models/config.py` and `agent_common/agents/dynamic_agent.py`, so with `MAX_RECURSION_LIMIT` set, startup logs two near-identical warnings whose only difference is contradictory "use `ORCHESTRATOR_…`" vs "use `SUB_AGENT_…`" advice, in import order.

Suggest: tailor the message to the variable actually set (the SDK caveat only applies to `MAX_RECURSION_LIMIT`), and consider a module-level `_warned: set` so each name warns once per process — being careful to keep the function re-entrant for the tests that call it repeatedly.

alloy-cicd (commented):

The extraction itself is clean: `app/core/step_budget.py` is gone with no stale imports, the removed orchestrator constants have no surviving importers, and no deployment manifest sets the retired env names. The derivation logic and its docs are solid. The findings below are ordered by severity; the first three are the ones worth fixing before merge.

**Findings on files outside this diff** (can't be inline-commented):

1. **`packages/ringier-a2a-sdk/ringier_a2a_sdk/agent/langgraph.py:54-59` — docstring falsified by this PR.** It still says `MAX_RECURSION_LIMIT` "is shared: agent-runner and this SDK default it to 50, agent-common's dynamic_agent to 75. Setting it moves all of them in a process that runs more than one." After this PR only the SDK reads that name. This is the one component the new legacy warning explicitly sends operators to, so the stale claim re-creates exactly the confusion `step_budget.py`'s warning was written to prevent. Updating those five lines to point at the three new `*_MAX_MODEL_CALLS_PER_TURN` names costs nothing.

2. **Derive-then-bind is repeated at every caller.** Both production callers of `build_sub_agent_graph` (`dynamic_agent.py:1542`, `agent-runner/agent/core.py:1079`) now hand-write the same compile → `recursion_limit_for` → log → `with_config` sequence. A `max_model_calls: int | None = None` parameter on `build_sub_agent_graph` that binds the derived limit before returning would make it impossible for a third call site to forget the bind — and forgetting is silent, because langgraph propagates the *parent's* recursion_limit into a child graph that binds none, which is precisely the bug shape this PR fixes.

**Question for the author (not blocking):** the removed hard ceilings (75 / 50 super-steps) are replaced by derived limits of ~103 (deep-agents stack at 25 calls) up to ~4x the old agent-runner bound, and no test bounds the absolute value — the new assertions (`derived >= measured`, slack < one call) both scale with the budget. Loosening the runaway guard 2–4x looks intentional given the module docstring, but worth a sentence in the PR description confirming it.

**Minor** (no inline comment): the test doubles are now duplicated across packages — `_ScriptedModel` exists ~6x (agent-common tests × 4, orchestrator tests × 2 incl. `tests/support/scripted_model.py`), and `_measure_super_steps`/`_tool_turn` are copied between the agent-common and orchestrator step-budget suites. A shared test-support module would keep the two measurements comparable. Also `test_an_implausibly_large_budget_is_honoured_but_flagged` hardcodes 2500 against an implicit `SANE_MAX_MODEL_CALLS`; `SANE_MAX_MODEL_CALLS + 1` would track the constant it tests.

packages/agent-common/agent_common/core/step_budget.py
```diff
@@ -136,3 +151,105 @@ def recursion_limit_for(graph: Any, max_model_calls: int) -> int:
...
+            "be removed.",
+            name,
+            raw,
+            service_env,
+            LEGACY_RECURSION_LIMIT_ENV,
+            LEGACY_SUB_AGENT_RECURSION_LIMIT_ENV,
+        )
+
+
+def _int_env(name: str, default: int) -> int:
```
alloy-cicd: **`_int_env` is now duplicated byte-for-byte** with `packages/orchestrator-agent/app/models/config.py:24`, which is still live (serves `AGENT_DISCOVERY_CACHE_TTL`, `MCP_DISCOVERY_CONCURRENCY`, `MCP_TOKEN_LEEWAY_SECONDS`). Same fallback semantics, same log line, same docstring intent — a future fix to env parsing applied to one copy leaves the two silently divergent, which is the drift class this PR exists to eliminate. `config.py` already imports from `agent_common.core.step_budget`; exporting this helper (drop the leading underscore) and deleting the orchestrator copy closes it.

packages/agent-common/agent_common/core/step_budget.py
```diff
@@ -98,8 +113,8 @@ def _warn_unclassified(buckets: dict[str, list[str]]) -> None:
     if buckets["unclassified"]:
         logger.warning(
```
alloy-cicd: **`_warn_unclassified` is an undeduplicated WARNING on a per-turn path.** `recursion_limit_for` is now called from `DynamicLocalAgentRunnable._build_graph`, which runs per invocation for sandbox-enabled agents and for any turn with attachments (dynamic_agent.py:1746/1763) — previously the limit was an import-time constant and nothing logged on this path. The day a LangGraph upgrade adds a node shape this module doesn't recognise, every sandboxed sub-agent turn emits this identical multi-line warning, several times per turn, burying the one signal ("teach the classifier") it was meant to raise. Suggest memoizing on the sorted unclassified-node tuple so each distinct shape warns once per process.

packages/agent-common/agent_common/agents/dynamic_agent.py
```diff
@@ -113,18 +114,27 @@
...
+# old number by multiplication would bake in exactly the implied rate this is
+# meant to remove. A sub-agent is where the tool loop actually runs — discovery,
+# retries, several rounds against an MCP server — so it has no business being more
+# constrained than the planner that delegates to it, which is what 25 says.
+#
+# Binding at all also matters: langgraph propagates `recursion_limit` into a
+# child graph that does not set its own, so without the `with_config` below every
+# sub-agent would inherit the orchestrator's.
+SUB_AGENT_MAX_MODEL_CALLS_ENV = "SUB_AGENT_MAX_MODEL_CALLS_PER_TURN"
+DEFAULT_SUB_AGENT_MAX_MODEL_CALLS = 25
```
alloy-cicd: **The default of 25 is now hand-written in three packages:** here, `agent-runner/agent/core.py` (`_DEFAULT_MAX_MODEL_CALLS = 25`), and orchestrator `app/models/config.py` (`DEFAULT_MAX_MODEL_CALLS_PER_TURN = 25`). The only thing tying any of them together is one assertion in agent-runner's test suite (which imports this heavy module cross-package just to read the int), and nothing at all ties the orchestrator's. That is the same drift mechanism as the 75-vs-50 defect this PR removes, one level up.

Since `step_budget.py` already owns the shared vocabulary (`MIN_MAX_MODEL_CALLS`, `SANE_MAX_MODEL_CALLS`, legacy names), a single `DEFAULT_MAX_MODEL_CALLS_PER_TURN = 25` there — or a defaulted `default` parameter on `resolve_max_model_calls` — writes the number once and makes `test_the_two_sub_agent_paths_no_longer_disagree` unnecessary rather than load-bearing.

packages/agent-runner/tests/test_step_budget.py
```diff
@@ -0,0 +1,56 @@
...
+    """`MAX_RECURSION_LIMIT=50` was the value that produced the observed failure.
+
+    It is read at import time, so the module is reloaded under the env var rather
+    than trusting the constant captured at collection.
+    """
+    monkeypatch.setenv(LEGACY_RECURSION_LIMIT_ENV, "50")
+
+    import agent.core as core
+
+    reloaded = importlib.reload(core)
```
alloy-cicd: **`importlib.reload(agent.core)` (twice) swaps every class/function object in the module for the rest of the pytest session.** `agent/__init__.py` binds `AgentRunner` at import, and sibling modules bind names at collection time (`tests/test_metadata_extraction.py:13`, `tests/test_agent_runner_security.py:24`), so after the reload those hold pre-reload objects while `sys.modules` holds post-reload ones — `isinstance`/`patch.object` across that boundary silently diverges depending on test order. If the `finally` reload itself raises under the modified env, the module is left half-populated for every later test.

The assertion needs none of this: `resolve_max_model_calls(_MAX_MODEL_CALLS_ENV, _DEFAULT_MAX_MODEL_CALLS)` under the monkeypatched env proves the same fact (the legacy name is not read) without reloading.

Relatedly, this module's assertions at lines 31/37/53 hold only while `AGENT_RUNNER_MAX_MODEL_CALLS_PER_TURN` is unset in the ambient environment — the exact knob this PR introduces — and there is no conftest scrubbing it; a `monkeypatch.delenv(..., raising=False)` guard would keep a tuned dev shell from failing these tests.

packages/agent-runner/agent/core.py
```diff
@@ -1062,7 +1075,14 @@ async def _run_graph(tools: list) -> None:
                 exclude_deep_agents_middlewares=False,
                 backend_factory=sandbox_backend_factory,
                 extra_middlewares=extra_middlewares,
-            ).with_config({"recursion_limit": _MAX_RECURSION_LIMIT})
+            )
+            recursion_limit = recursion_limit_for(graph, _MAX_MODEL_CALLS_PER_TURN)
+            logger.info(
```
alloy-cicd: **Log-level mismatch with the twin bind site:** this logs at INFO once per scheduled run, printing values that are constant for the process given the middleware stack, while `dynamic_agent.py:1543` logs the same fact at DEBUG. On a high-frequency scheduler this becomes the dominant INFO line in the runner's log, and level tuning that silences one site keeps the other. Suggest DEBUG here to match.

packages/agent-common/agent_common/core/step_budget.py
aartaria: Fixed in ad7c3763. `recursion_limit_for` now floors at one step per model call when the classification yields no per-model-call nodes, and logs a distinct warning naming the graph type — so the degenerate case is diagnosable instead of a silent 0. Not raising: this runs on a hot path (per invocation for sandbox-enabled sub-agents), and killing a turn over a test double would be worse than the floor. Covered by `test_a_graph_that_classifies_to_nothing_is_floored_and_warned` and `test_a_mock_graph_no_longer_binds_a_zero_budget`.

packages/agent-common/tests/test_step_budget.py
aartaria: Fixed in ad7c3763 — `_measure_super_steps` now passes `"recursion_limit": 10_000` explicitly, with a comment saying why. Good catch: the measurement being capped by the thing it measures would have surfaced as a derivation failure that was not one.

packages/agent-common/tests/test_step_budget.py
aartaria: Fixed in ad7c3763 — dropped. Confirmed the lean stack derives per_model_call=3, base=0, so `recursion_limit_for(graph, 25)` is exactly 75 there; the assertion held only because the test happens to use the deep-agents graph. The equality assertion above it already carries the intent.

packages/agent-common/tests/test_step_budget.py
aartaria: Fixed in ad7c3763 — the test now asserts against `_SUB_AGENT_MAX_MODEL_CALLS`, which is what the bind site actually uses, with a comment on why the module default is the wrong expectation here. Added the same ambient-env guard to agent-runner's suite as an autouse `monkeypatch.delenv` fixture.

packages/agent-common/agent_common/core/step_budget.py
aartaria: Partly fixed in ad7c3763.

Point 1 — agreed and fixed. Each retired name now carries its own advice (`_LEGACY_ENV_ADVICE`): the SDK caveat belongs only to `MAX_RECURSION_LIMIT`, and `SUB_AGENT_RECURSION_LIMIT` is simply told it can be removed. `test_the_advice_is_tailored_to_the_variable_actually_set` pins that the SDK caveat does *not* appear in the other branch, which also fixes the tests being unable to distinguish them.

Point 2 — not applying the per-process dedup. Two lines in a process hosting both the orchestrator and its in-process sub-agents is the correct reading, not a duplicate: both budgets really did lose their old name, and an operator who saw only one would leave the other on its default. With the messages now tailored the pair is no longer near-identical, and the "contradictory" advice is two different budgets naming their two different replacements. Documented as deliberate in the `warn_if_legacy_recursion_env_set` docstring. I did add dedup where it genuinely floods — the unclassified-node warning, which is on a per-turn path; this one fires only at import.

packages/agent-common/agent_common/core/step_budget.py
aartaria: Fixed in ad7c3763 — exported as `int_env` and the orchestrator copy deleted; `config.py` imports it under its old private name so the three unrelated call sites (`AGENT_DISCOVERY_CACHE_TTL`, `MCP_DISCOVERY_CONCURRENCY`, `MCP_TOKEN_LEEWAY_SECONDS`) are untouched.

packages/agent-common/agent_common/core/step_budget.py
aartaria: Fixed in ad7c3763 — memoized on the sorted unclassified-node tuple in a module-level `_warned_shapes`, so each distinct shape warns once per process. You are right that this path changed character: it was an import-time constant before this PR and is now per-invocation for sandboxed agents and attachment-carrying turns. The existing warning test discards its own shape first so it stays order-independent.

packages/agent-common/agent_common/agents/dynamic_agent.py
aartaria: Fixed in ad7c3763, and the best finding of the round — it is the PR's own defect one level up. `step_budget.DEFAULT_MAX_MODEL_CALLS_PER_TURN = 25` is now the single definition, `resolve_max_model_calls(env_name, default=...)` defaults to it, and all three packages call it with just their env name. `test_the_two_sub_agent_paths_no_longer_disagree` is gone along with the cross-package heavy import it needed — the fact it asserted is now structural.

packages/agent-runner/tests/test_step_budget.py
aartaria: Fixed in ad7c3763 — both reloads gone; the assertion now goes through `resolve_max_model_calls(_MAX_MODEL_CALLS_ENV)` under the monkeypatched env, and is parametrized over both retired names. Also took the second half: an autouse fixture now scrubs `AGENT_RUNNER_MAX_MODEL_CALLS_PER_TURN`, so a tuned dev shell cannot fail the module.

packages/agent-runner/agent/core.py
aartaria: Fixed in ad7c3763 — DEBUG here, matching `dynamic_agent`. Agreed on the reasoning: the values are constant for the process given the middleware stack, so once per scheduled run at INFO is noise on a busy scheduler.

aartaria:
Round 1 addressed in ad7c3763 — all ten inline findings replied to and resolved individually (eight fixed, one partly fixed, one branch declined with reasoning). Taking the four summary-body items here:

**The falsified `ringier-a2a-sdk` docstring** — fixed. `_get_default_recursion_limit` still claimed `MAX_RECURSION_LIMIT` is shared with agent-runner and dynamic_agent, which this PR made untrue. It now says both moved to model-call budgets under their own names, that the SDK keeps the name for compatibility as an externally published package, and points at `agent_common.core.step_budget` before anyone reinstates anything shared. Good catch on a file outside the diff.

**Folding derive-then-bind into `build_sub_agent_graph`** — not applying. The two lines look like duplication but the thing being duplicated is a *policy*, not a mechanism: each caller owns its budget, its env name and its log site, and only the caller knows which of the three it is. Pushing the bind into the builder would mean either passing the budget in (identical call-site line count, one more parameter) or having the builder read an env var, which reintroduces exactly the "one place decides for every caller" shape this PR removes. Happy to revisit if a third caller appears.

**The runaway guard being loosened, with no test bounding the absolute limit** — fair question, and deliberately not tested. The absolute super-step number is derived, so any assertion on it is an assertion about the current middleware stack that would need updating with every stack change — the hand-maintained constant this PR exists to delete, wearing a test's clothes. What *is* bounded is the unit that has meaning: `SANE_MAX_MODEL_CALLS = 200` warns that the guard is effectively disabled, and `MIN_MAX_MODEL_CALLS` clamps the other end. On the loosening itself: yes, 25 model calls is more than either path had, chosen because a sub-agent is where the tool loop actually runs and the observed bug was a correct run being killed mid-work. If 25 is too generous for scheduled jobs specifically, `AGENT_RUNNER_MAX_MODEL_CALLS_PER_TURN` now exists precisely so that can be tuned without touching anything else — I would rather set that from evidence than guess a lower number now.

**Test-double duplication** — acknowledged, not consolidated. `_ScriptedModel` exists in both suites because they measure *different graphs* (orchestrator vs. sub-agent) and the orchestrator's variant scripts `FinalResponseSchema` turns the sub-agent graph has no equivalent for. Sharing it would mean a cross-package test-helper import for ~30 lines, which trades one kind of coupling for a worse one.

packages/agent-common/agent_common/core/step_budget.py
```diff
@@ -0,0 +1,335 @@
...
+    on which every turn dies at its first super-step, silently and with nothing
+    pointing at the budget. One step per model call is a floor, not an estimate:
+    it cannot be right, but it is diagnosable, and it is warned about.
+    """
+    buckets = classify_nodes(graph)
+    _warn_unclassified(buckets)
+
+    per_call = _steps_per_model_call(buckets)
+    if per_call <= 0:
+        _warn_degenerate(graph, buckets)
```
alloy-cicd, line 205: **Minor, non-blocking:** `_warn_degenerate` sits on the same per-invocation path that motivated deduplicating `_warn_unclassified` in this very commit, but has no dedup itself. If a degenerate graph ever ships (the "future graph type whose nodes do not live on `.nodes`" case the docstring anticipates), this warns on every `_build_graph` — several times per turn for sandboxed sub-agents — and meanwhile the suites that patch `build_sub_agent_graph` with a `MagicMock` now emit it once per call as test-log noise. A one-line guard keyed on `type(graph).__name__` (or reusing `_warned_shapes` with a sentinel) would match the treatment its sibling just received. Fine to fold into a follow-up.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (b425a0a1bed9901d2bbba9b0dd1785b0301401f9..ad7c3763e8dbda0f94a9fcb3aa2955a593ceaf39)

**Ledger of round-1 findings** (all 10 threads resolved by the author; dispositions verified against the delta and by running the step-budget suites in all three packages — 27 + 4 + 13 passed):

1. `recursion_limit_for` floor — **verified-fixed** (floors at `max(max_model_calls, 1)` with a distinct warning; `recursion_limit_for(MagicMock(), 25)` no longer returns 0, covered by two new tests).
2. `_measure_super_steps` unbounded — **verified-fixed** (explicit `recursion_limit: 10_000` with rationale comment).
3. `!= 75` coincidence assertion — **verified-fixed** (dropped).
4. Ambient-env coupling in the bind-site test — **verified-fixed** (asserts against `_SUB_AGENT_MAX_MODEL_CALLS`; agent-runner got an autouse `delenv` fixture).
5. Legacy warning, point 1 (undifferentiated body) — **verified-fixed** (`_LEGACY_ENV_ADVICE` tailors per variable, pinned by a new branch-distinguishing test). Point 2 (once-per-consumer firing) — **rejected-by-author** with sound reasoning (two lines = two budgets that each lost their name, now documented in the docstring); accepted, not re-raised.
6. `_int_env` duplication — **verified-fixed** (exported as `int_env`, orchestrator copy deleted, call sites untouched; `os` remains used in config.py).
7. `_warn_unclassified` flood — **verified-fixed** (memoized per shape in `_warned_shapes`, with an order-independence guard in the test).
8. Default 25 in three packages — **verified-fixed** (single `DEFAULT_MAX_MODEL_CALLS_PER_TURN` in step_budget, defaulted parameter; the cross-package assertion test is gone because the fact is now structural; no stale references to the deleted constants remain).
9. `importlib.reload` in agent-runner tests — **verified-fixed** (both reloads gone, asserted via `resolve_max_model_calls`, parametrized over both retired names).
10. INFO/DEBUG mismatch — **verified-fixed** (DEBUG at the agent-runner bind site).

The round-1 summary items were also addressed: the `ringier-a2a-sdk` docstring is corrected (and now points at nannos#239 and `step_budget`). The derive-then-bind fold-in and the test-double dedup remain follow-up material, as flagged then — not blocking.

**Delta review:** one minor new observation (inline below); nothing blocking. The fixes introduce no correctness or security regressions.

packages/agent-common/agent_common/core/step_budget.py
aartaria: Fixed in 41daa10b — you are right, and it was an inconsistency introduced by the very commit that fixed the sibling. `_warn_degenerate` now dedups on `("__degenerate__", type(graph).__name__)` in the same `_warned_shapes` set, with a test asserting five calls produce one warning. Taking it now rather than as a follow-up since it is three lines and the reasoning was already established one function up.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (ad7c3763e8dbda0f94a9fcb3aa2955a593ceaf39..41daa10ba2d92dc7dac56955316498176e85ba14)

1. `_warn_degenerate` missing the dedup its sibling received — **verified-fixed** in 41daa10b: deduplicated on `("__degenerate__", type(graph).__name__)` in the shared `_warned_shapes` set (the sentinel keeps it disjoint from node-name shapes), with `test_the_degenerate_warning_is_not_repeated_per_invocation` asserting five calls produce one warning, and the existing degenerate-graph test updated to discard its own key for order-independence.

That was the only open finding. **The delta (one 32-line commit confined to `step_budget.py` and its test file) introduces nothing new** — verified by reading the full diff and running the agent-common step-budget suite at the new head (28 passed). No inline comments this round; from the automatic reviewer's side the PR is clean.

aartaria:
⚠️ **Correction to this PR's framing, and a migration note for existing deployments.**

I described the budget change as "a modest raise". That is true against the **defaults** the issue reported against — the repro was a local stack with neither variable set — and misleading for any deployment that has `MAX_RECURSION_LIMIT` set to a non-default value. Correcting it here rather than leaving it in the PR body.

Because the old bound is counted in super-steps, a deployment that raised it was buying a number whose meaning depends on the middleware stack's per-call node cost. Converting, at rates measured against real `build_sub_agent_graph` output:

| stack | super-steps per model call | base |
|---|---|---|
| sub-agent, no HITL guard (agent-runner's shape) | 4 | 3 |
| sub-agent with the HITL guard (dynamic_agent's shape) | 5 | 3 |

So `model_calls ≈ (old_recursion_limit - base) / per_call`. Any deployment whose converted figure is above this PR's default of 25 will be **tightened**, not loosened, when it rolls out — and since `MAX_RECURSION_LIMIT` stops being read here, that happens silently via the fallback.

**Migration, therefore: set the per-service names before or with the rollout** — `AGENT_RUNNER_MAX_MODEL_CALLS_PER_TURN` and `SUB_AGENT_MAX_MODEL_CALLS_PER_TURN` — to whatever budget was actually intended, rather than letting the fallback decide. Whether that is the converted figure or a deliberately lower one is a judgement call, but it should be a decision rather than a side effect of the merge.

It also strengthens the case for the change, unflatteringly: needing a large super-step number to make sub-agent runs complete is exactly the symptom this PR addresses, since such a number silently tightens every time a middleware carrying model hooks is added.

The code is unchanged by this comment — no new commit.

aartaria:
**Update (6ccdea02): the single default of 25 is replaced by three, one per kind of turn.**

25 was my invention with nothing behind it. These are set from operational experience of what each kind of turn actually needs, and defined together in `step_budget.py` so they can be compared at a glance:

| constant | default | the turn it sizes |
|---|---|---|
| `DEFAULT_ORCHESTRATOR_MAX_MODEL_CALLS` | 25 | planning — delegate and decide |
| `DEFAULT_SUB_AGENT_MAX_MODEL_CALLS` | 40 | delegated work, a user waiting |
| `DEFAULT_SCHEDULED_RUN_MAX_MODEL_CALLS` | 125 | unattended work |

They differ because the turns differ. A planning turn is a few calls around a decision, and a long one usually means a loop rather than hard work. A sub-agent is where the work happens, and the orchestrator can re-delegate if it stops short. A scheduled run has neither safeguard — nothing observes it stopping one call early, nothing re-delegates it, and a truncated result is silently useless rather than visibly incomplete — so erring large is plainly the cheaper mistake there.

**This largely retires the migration warning in my earlier comment.** A deployment that had raised `MAX_RECURSION_LIMIT` to make sub-agent runs complete now lands near where it already was when the variable stops being read, rather than silently dropping to a tighter budget. Converting your own value against the rates in that comment is still worth doing before rollout, but the fallback is no longer a cliff.

On the round-1 finding about the default being written in three places: co-locating preserves what that fix was protecting. The defect was never *a* number — it was the same number written separately per package and left to drift. These are deliberately different, defined side by side, and a test pins the ordering so a future edit cannot make a planner more generous than the sub-agents it delegates to. `resolve_max_model_calls` takes an explicit `default` again, since an implicit shared one would now hand a service the wrong kind of turn's budget.

`example-k8s-deployment` shows all three knobs at their defaults, so they are discoverable rather than something you find in the source.

Rebased onto the branch as it stands after the `main` advance; suites green (agent-common 1065, agent-runner 49, orchestrator-agent 1027, the 9 `test_thinking_blocks_round_trip.py` failures pre-existing on `main`).

PR #244, opened 2026-09-15
fix(agent-common, orchestrator-agent, clients): never send a retired model alias to the gateway

aartaria:
closes #

## What

A model alias retired on the Model Gateway could still be sent to it and rejected, ending the turn — while a live successor was configured and readable the whole time. Four independent defects had to line up, and each is worth fixing on its own.

### 1. Clients hardcoded a model alias

`client-slack`, `client-google-chat` and `client-email` each carried a `MODELS` constant and put a concrete alias into every request's metadata:

```ts
const MODELS = { ClaudeSonnet45: 'claude-sonnet-4.5', GPT4o: 'gpt4o' };
...
model: MODELS.ClaudeSonnet45,
```

That overrode the console's `chat` default slot for every user of those clients, including everyone who never picked a preferred model. When the alias was retired, every turn from those clients asked for a model that no longer existed. (The `gpt4o` entry was already a dead alias — no dashes — just never used.)

Sending no `model` at all lets the orchestrator fall back to `require_default_model()`, which resolves the `chat` slot an admin can repoint. `embed-sdk` already behaves this way: it sends `model` only when the user has actually chosen one.

### 2. `_resolve_alias` gave up after a single defaults read

Past the registry check the gateway has already told us the alias doesn't exist, so passing it through is a guaranteed rejection — a default is the only way out. But an empty defaults map with no recorded error is ambiguous: either nothing is configured, or that read landed badly. The read itself re-arms the cache as cold (see `_model_defaults`), so a second read refetches synchronously.

One unlucky first read was therefore enough to discard a default that was one call away. It now reads twice before giving up. A read that recorded an *error* still backs off rather than hammering an endpoint that is already failing.

### 3. Resolution ran only on a cache miss

`GraphFactory.get_graph` keyed both `_models` and `_graphs` on the **requested** alias, with resolution happening further down inside `create_model`. So a model instance built while an alias was still live — or during any window where resolution failed open — was served for the rest of the process lifetime; `_models` is never evicted and `_graphs` only on a document-store upgrade. A retired alias cached that way keeps being sent to the gateway long after the process knows its successor.

Resolving before the cache key costs a cached dict read (measured at ~0.19 µs/call on the hot path, no I/O, no lock) and lets a retirement heal on the next request instead of the next restart. Two aliases resolving to the same model now also share one compiled graph rather than building two.

### 4. The cause was invisible in logs

Cache refresh failures were logged at `debug`, so at the INFO level deployments typically run there was no record of *why* anything went wrong — only of the downstream rejections. Now `WARNING`. The give-up path in `_resolve_alias` is now `ERROR` and distinguishes an unreadable defaults endpoint (an outage) from a genuinely unset default (an admin task); the previous single message asserted the latter in both cases.

## Tests

New `packages/agent-common/tests/test_retired_alias_resolution.py` (7 cases): registered alias untouched, retired alias degrades, unreadable registry passes through, **empty-and-unerrored defaults are re-read**, errored defaults are not retried, a default still pointing at the alias being degraded away from is not treated as a successor, and multimodal embedding role preference.

The regression test fails against the previous `_resolve_alias` and passes with it.

Full suites green, measured against `origin/main` in a clean worktree: agent-common 1130 → **1141** passed (exactly the tests added here), orchestrator-agent 980 → **983**. Warning counts unchanged in both (114 / 106), so no warning gate moves. `ruff` clean on all touched files; the 10 findings it reports in agent-common are pre-existing and in files this PR does not touch.

## Note for reviewers

`_first_default` is called twice in `_resolve_alias`, and the second call only differs because the first one re-armed the defaults cache as cold. That coupling spans two functions and is carried by a comment. The explicit alternative is a `force=True` parameter on `_model_defaults` — happy to switch if reviewers prefer it.

Anything matching the log string `no gateway default set` will stop matching; both give-up messages were reworded.

## Checklist


packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -941,7 +942,17 @@ def get_graph(
...
+        # alias means resolution runs only on a cache miss, so a model instance built while an
+        # alias was still live — or during a window where resolution failed open — is served
+        # for the process lifetime: _models is never evicted, and _graphs only on a store
+        # upgrade. A retired alias cached that way keeps being sent to the gateway long after
+        # the process knows its successor. Resolving here costs a cached dict read (60s TTL,
+        # refreshed off-thread) and lets a retirement heal on the next request rather than the
+        # next restart. Two aliases resolving to the same model now also share one graph
+        # instead of building two.
+        requested: ModelType = model_type or require_default_model()
+        effective_model: ModelType = resolve_chat_model(requested)
```
alloy-cicd, line 960: **High — synchronous network I/O on the async hot path.** `get_graph` is reached from the async executor (`executor.py:1033`), and `resolve_chat_model` now runs on **every** request, before the graph-cache check. The "~0.19 µs cached dict read" claim only holds for the warm happy path. In the degraded state — requested alias not registered AND defaults map empty-and-unerrored — `_model_defaults` re-arms its cache cold on **every** read (`model_factory.py:585-586`), so each request performs up to two synchronous `urlopen` fetches (2 s timeout each) or a 3 s `cond.wait_for` (cold+inflight path in `_refresh_if_stale`) **inside the event loop**, with no backoff, stalling all concurrent turns.

Failure scenario: a user's `preferred_model` is a retired alias while console-backend legitimately returns `{}` (fresh deploy / defaults wiped) — every message from every user serializes behind repeated 2 s+2 s blocking fetches; streaming and unrelated in-flight turns stutter or time out, where previously only that one user's call failed fast at the gateway. Same class of problem via `_gateway_models`' own empty-re-arm (`model_factory.py:554-556`).

Suggestion: make the empty-but-fresh state non-blocking on the hot path — kick the refetch to the background thread (like the warm-stale path) and serve the cached answer, or rate-limit the cold re-arm (e.g. min 1 s between inline refetches).

packages/orchestrator-agent/app/core/graph_factory.py
```diff
@@ -941,7 +942,17 @@ def get_graph(
...
 
         cache_key = (effective_model, thinking_level)
```
alloy-cicd, line 962: **High — resolution now runs at two layers, and they can diverge.** The comment says "Resolve BEFORE the cache key, not inside create_model", but `create_model` still resolves internally (`model_factory.py:356`). On a graph-cache miss the same alias is resolved twice with a time gap; if the registry/defaults snapshot changes between the two reads (60 s TTL boundary), the graph is cached **permanently** under a key naming a different model than the one it actually calls — `_models`/`_graphs` are never evicted, so one hit persists until restart.

Failure scenario: `get_graph` resolves retired→`claude-sonnet-4-6`, keys the graph `(claude-sonnet-4-6, level)`; before `create_model` runs, that alias is itself retired and the inner resolve degrades to the new default. Forever after, every request resolving to `claude-sonnet-4-6` (including users who explicitly pick it once re-registered) silently runs the other model, with logs claiming otherwise — the exact stale-cache class this hunk's comment says it eliminates.

Suggestion: one resolution point — pass the already-resolved alias into a resolve-skipping path (e.g. `create_model(..., pre_resolved=True)` or an internal `_create_model_resolved`), keeping the public behaviour for other callers. That also removes the redundant second resolution on every miss.

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -626,18 +641,47 @@ def _resolve_alias(model_type: str, roles: tuple[str, ...], kind: str) -> str:
...
+    if found is not None:
+        role, default = found
+        logger.warning(
+            "%s model '%s' not registered on the gateway; falling back to default '%s' (role=%s)",
+            kind,
+            model_type,
+            default,
+            role,
+        )
+        return default
```
alloy-cicd, line 734: **Medium — the fallback default is returned without checking it is actually registered.** `models` is in scope (line 641) but never consulted for `default`. A stale default — admin retired the default's alias without repointing, or the defaults cache is up to 60 s behind the registry cache — is returned with a WARNING that reads as success ("falling back to default"), `get_graph` keys and caches a graph under that dead alias, and the gateway then 400s "Invalid model name" anyway: the exact failure this PR sets out to eliminate, now with a log trail claiming the degradation worked.

Suggestion: check `default in models` in `_first_default` (continue to the next role, and fall into the error branch if none qualifies).

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -626,18 +641,47 @@ def _resolve_alias(model_type: str, roles: tuple[str, ...], kind: str) -> str:
...
+    if _DEFAULTS_CACHE["last_error"] is not None:
+        logger.error(
+            "%s model '%s' is not registered and the model-defaults endpoint is unreadable (%s); "
+            "passing through — the gateway will reject it",
+            kind,
+            model_type,
+            _DEFAULTS_CACHE["last_error"],
+        )
+    else:
+        logger.error(
```
alloy-cicd, line 769: **Medium — this ERROR is false in the most likely real-world failure.** When the configured default **is set but points at the retired alias itself** (exactly the fleet state this PR fixes, and exactly what `test_default_pointing_at_the_retired_alias_is_not_a_successor` exercises), `_first_default` excludes it, the defaults map is non-empty so there is no retry, and this branch logs "no default is set for role(s) …" per request. The admin checks the console, sees a default configured, and is misdirected away from the actual fix (repoint the default) — the exact diagnosability the PR's own comments say this log split exists to provide.

Suggestion: a third case — "default for role(s) %s points at the retired alias itself; repoint it" — when some role's default equals `model_type`.

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -626,18 +641,47 @@ def _resolve_alias(model_type: str, roles: tuple[str, ...], kind: str) -> str:
...
+
+    # Past this point the registry was read SUCCESSFULLY and doesn't have the alias, so
+    # passing it through is a guaranteed rejection — a default is the only way out. Read the
+    # defaults TWICE before giving up: an empty map with no recorded error re-arms the cache
+    # as cold (see _model_defaults), so the second read re-fetches synchronously instead of
+    # re-serving the same empty snapshot. Without the retry a single first-use read that
+    # happens to land empty is indistinguishable from "no default is configured", and one
+    # unlucky fetch is enough to fail open on an alias the gateway has already disowned.
+    found = _first_default(roles, model_type)
+    if found is None and not _DEFAULTS_CACHE["defaults"] and _DEFAULTS_CACHE["last_error"] is None:
```
alloy-cicd: **Medium — the retry guard is a racy check-then-act, and a third hand-copy of the invariant.** The guard reads `_DEFAULTS_CACHE["defaults"]` and `["last_error"]` without `_DEFAULTS_LOCK` across two fields the refresh thread mutates non-atomically (`last_error` in `_run`'s try block, `ts`/`inflight` in the finally): thread A's first `_first_default()` returns None off an empty snapshot; a background `_run()` then stores a non-empty map containing the successor; A's guard now sees `not defaults == False`, **skips the retry, and passes the retired alias through to a guaranteed 400 even though the successor is sitting in the cache**. The same unlocked read at line 670 can pick the wrong error-log branch when `last_error` flips mid-call.

Separately, this predicate is the third hand-rolled copy of the empty-and-unerrored invariant (also at lines 585 and 621), and the retry only works via `_model_defaults`' cold re-arm two functions away — if that re-arm is ever conditioned (backoff, re-arm-N-times), the second read silently re-serves the same snapshot and this block becomes dead code.

The PR description already offers the fix: **`force=True` on `_model_defaults`, refetching synchronously under the cache's own locking** — please switch to it. It removes the duplicated predicate, the race, and the cross-function coupling in one move.

packages/client-slack/src/services/a2aClientService.ts
```diff
@@ -212,7 +207,10 @@ export class A2AClientService {
         slackChannelId: request.channelId,
         slackThreadTs: request.threadTs,
         slackMessageTs: request.messageTs,
-        model: MODELS.ClaudeSonnet45,
+        // No `model` is sent on purpose: the orchestrator then runs the console's
```
alloy-cicd, line 210: **Medium — deployment-order hazard: channel-client availability now hard-requires a populated `chat` default.** With no `model` in metadata and no `user.preferred_model`, `get_graph` calls `require_default_model()` (`model_factory.py:781`), which **raises `NoDefaultModelError`** — there is no alias fallback. Two states this converts from working to hard-failing:

1. *No default configured*: console-backend's `register_model` never auto-sets a default (no write to `ModelDefaultsService` — the model_factory comment claiming otherwise is wrong), so "models registered, default never clicked" is reachable. Previously such deployments worked as long as `claude-sonnet-4.5` was registered.
2. *Cold start during console-backend outage*: the defaults cache retains last-known values only within a process; an orchestrator restarted while console-backend is down has an empty cache with `last_error` set, so every Slack/email/GChat message fails with `NoDefaultModelError` until console-backend recovers — even though the LLM gateway is healthy. That coupled-outage class did not exist with the hardcoded alias.

The policy is intentional per the comments, but nothing sequences or checks it. Before rollout, confirm the `chat` default slot is populated in **every** environment (and consider a startup check or alert on `NoDefaultModelError`).

alloy-cicd (commented):

The four-defect analysis in the description is sound, and the fix direction is right. The review found **2 high-severity issues** on the new hot path (event-loop blocking in a degraded state, and a cache-key/model divergence from resolving at two layers), **3 medium** (fallback default not validated against the registry, a misleading give-up ERROR, a racy retry guard), and **1 deployment hazard** (channel clients now hard-require a populated `chat` default), plus minor cleanups. Details inline.

Two things that don't anchor to diff lines:

- **`packages/console-frontend/src/config/models.ts:29`** still hardcodes `claude-sonnet-4.5` in the static `MODEL_OPTIONS` fallback (used when the gateway list is empty/unreachable). Users can still persist a retired alias as `preferred_model`, and `modelSupportsThinking`/`getAvailableThinkingLevels` derive thinking capability from the retired alias rather than the model that will actually run after server-side degradation. Worth a follow-up.
- The claim (comment near `model_factory.py:581`) that registering the first model auto-sets it as the chat default is **not backed by code**: console-backend's `register_model` (`admin_model_gateway_router.py:388`) never writes to `ModelDefaultsService`; defaults are only set via the explicit `POST /models/{id}/default`. This matters for the deployment-hazard finding inline.

Verified clean: `_first_default` + retry is semantically equivalent to the removed per-role loop (same role order, same `default != model_type` exclusion); no `MODELS`/`GPT4o` references remain in any of the three TS packages; nothing in the repo (tests, alerts, monitors) matches the removed `no gateway default set` log string; voice-agent sends no model metadata; the debug→warning promotion cannot spam in production (ts is set even on failure, so one WARNING per 60s TTL per cache).

packages/client-email/src/services/a2aClientService.ts
```diff
@@ -114,10 +114,10 @@ function createAuthenticatedFetch(accessToken: string): typeof fetch {
   };
 }
 
-const MODELS = {
-  ClaudeSonnet45: 'claude-sonnet-4.5',
-  GPT4o: 'gpt4o',
-};
+// No `model` is sent in request metadata on purpose: the orchestrator then runs the
```
alloy-cicd: **Minor — orphaned comment placement, and third copy of the rationale.** This 4-line block sits at module scope where the `MODELS` constant used to be, attached to nothing — a developer re-adding `model:` at one of the four metadata sites (~line 230, 283, 336, 651) never sees it. client-slack and client-google-chat place the identical text inline inside the metadata object it protects; this copy has already drifted in wording and placement.

Suggestion: one canonical explanation next to `require_default_model` in agent-common (where the behaviour lives), with a one-line pointer at each client metadata site — or at least move this copy inline / into the class JSDoc so it guards the point of the mistake it exists to prevent.

packages/agent-common/tests/test_retired_alias_resolution.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def _seed_gateway(models: dict):
+    # Fresh ts so _refresh_if_stale serves this snapshot without firing a (network) refresh.
+    mf._GW_CACHE.clear()
+    mf._GW_CACHE.update({"ts": time.monotonic(), "models": models, "inflight": False, "last_error": None})
+
+
+def _seed_defaults(defaults: dict, last_error=None):
+    mf._DEFAULTS_CACHE.clear()
```
alloy-cicd: **Minor — third copy of the cache seed/restore fixtures.** `_seed_defaults` and the autouse restore fixture duplicate `tests/test_embedding_default_resolution.py:16-29` verbatim, and `_seed_gateway` is a third variant of `_cold_cache()` in `tests/test_gateway_cache_refresh.py:14`. The cache dict shape (`{ts, defaults/models, inflight, last_error}`) is an implicit contract now maintained in three private helpers; when a key is added, missing one file seeds a malformed cache that either fails obscurely or passes while exercising a state production can never be in. `agent-common/tests` has no `conftest.py` — this PR is a good moment to create one and hoist the seed/restore helpers.

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -506,7 +506,11 @@ def _run():
             cache["last_error"] = None
         except Exception as e:
             cache["last_error"] = e
-            logger.debug("Background refresh of '%s' failed: %s", key, e)
+            # WARNING, not debug: these caches decide which model every agent call runs on, so
+            # a failure here surfaces later as an unexplained wrong-model or rejected call. At
+            # the INFO level deployments typically run, debug would leave no record of the
+            # cause at all — only of the downstream damage.
+            logger.warning("Background refresh of '%s' failed: %s", key, e)
```
alloy-cicd: **Minor — new WARNING in gateway-less environments, and no de-duplication across an outage.** Production volume is bounded (ts is set even on failure, so ≤1/min per cache per process), but (a) environments deliberately running without `LLM_GATEWAY_URL`/`CONSOLE_BACKEND_URL` (unit tests, local dev) now emit this WARNING per cold cache, which can trip caplog no-warning assertions or CI warning gates; (b) a multi-hour outage across W workers still emits ~2W identical lines per minute with no recovery signal. Consider logging WARNING only on the failure **transition** (via the existing `last_error`), demoting repeats to debug, and an INFO on recovery.

packages/orchestrator-agent/app/core/graph_factory.py
aartaria: Fixed in 8f75f235. You are right that the ~0.19 µs figure only described the warm path, and the unbounded blocking was not even coming from the retry I added — it was the cold re-arm underneath, which fired on every read while the map stayed empty. The floor now lives in `_rearm_defaults_if_unconfirmed` (one per second, decided under the cache lock), so the degraded state costs at most one blocking re-population per second instead of one per request. `test_rearm_is_rate_limited` pins it. The `_gateway_models` empty-re-arm you note as the same class is untouched and pre-dates this PR — worth its own change rather than widening this one.

packages/orchestrator-agent/app/core/graph_factory.py
aartaria: Fixed in 8f75f235. `create_model` takes `pre_resolved` and `GraphFactory._create_model` passes it, so resolution happens once per creation; other callers keep the safety net unchanged. `test_create_model_is_told_the_alias_is_already_resolved` guards the wiring. The comment claiming a single resolution point is now true.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in 8f75f235. `_first_default` now takes the registry snapshot and requires `default in registered`, so an unregistered default is skipped rather than reported as a successful degrade. console-backend's `_retirement_fallback` already applied this rule ("only follow a successor that is actually live"); the two sides now agree. Covered by `test_unregistered_default_is_not_a_successor`.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in 8f75f235. Added the third case: "the default for role(s) … still points at it; repoint the default to a registered model". Your read of the consequence is the sharper one — the admin sees a default configured in the console and is steered away from the actual fix. `test_default_pointing_at_the_retired_alias_is_not_a_successor` now asserts on the message, not just the return value.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in 8f75f235, taking the `force=True` direction you endorsed, in the shape the second read made natural. `_rearm_defaults_if_unconfirmed` takes the empty-and-unerrored decision once, under `_DEFAULTS_LOCK`, so the interleaving you describe cannot skip the retry: either we observe the map the refresh wrote, or we observe empty and re-arm. `_resolve_alias`'s second read is now unconditional, so a refresh landing mid-call is picked up rather than lost — `test_concurrent_refresh_landing_mid_call_is_picked_up`. That also deletes the hand-copied predicate from `_resolve_alias` and from `_model_defaults`, which both now call the helper, so the re-arm and its rate limit have one owner.

packages/client-email/src/services/a2aClientService.ts
aartaria: Fixed in 8f75f235. Moved inline to all four metadata sites in this file, where the mistake would actually be made, and trimmed to three lines with a pointer to `require_default_model`. I did not consolidate the canonical explanation into agent-common in this PR — that is a reasonable follow-up, but it would mean touching a third package to relocate a comment.

packages/agent-common/tests/test_retired_alias_resolution.py
aartaria: Fixed in 8f75f235 — and this turned out not to be cosmetic, so thank you for it. Adding the re-arm floor introduced a second piece of process-global state, and running the full suite surfaced a real leak: one test's re-arm silently suppressed another's, failing `test_gateway_cache_refresh.py::test_empty_defaults_refetches_on_next_call` only in whole-suite runs. `tests/conftest.py` now carries an autouse restore of all three globals plus shared `seed_gateway`/`seed_defaults` fixtures, and the new file uses them. The two existing files keep their own seed helpers for now — migrating them is mechanical and safe under the autouse restore, but it is churn beyond this PR.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in 8f75f235. WARNING now fires on the failure transition only (`last_error is None`), repeats drop to debug, and a recovery logs INFO — so a multi-hour outage costs one line per cache per worker rather than ~2W per minute, and gateway-less environments emit one warning per cold cache instead of one per refresh. Verified the suite's warning count is unchanged against origin/main (114 in agent-common, 106 in orchestrator-agent), so no warning gate moves.

packages/client-slack/src/services/a2aClientService.ts
aartaria: Confirmed and partly fixed, but the substantive half is a rollout decision I am deliberately not taking alone — leaving this thread open as the visible to-do.

Verified your factual claim: nothing writes `model_defaults` except an explicit set, so "models registered, default never set" is reachable. The `model_factory` comment asserting otherwise is gone (it was in the block rewritten in 8f75f235), and the last copy of the claim — a test docstring in `test_gateway_cache_refresh.py` — is corrected in e8116bea.

On the behaviour itself: the hard dependency is intended. The alternative is what this PR removes — a hardcoded alias that appears to work right up until it is retired, which is exactly the failure that prompted the change. Trading a silent wrong-model for a loud misconfiguration error is the point. Your second state is the sharper one though: an orchestrator cold-starting while console-backend is down now fails every turn even with a healthy LLM gateway, and that coupled-outage class is genuinely new.

What I am not deciding here: whether that warrants a startup check, an alert on `NoDefaultModelError`, or a last-known-good persisted default. Those are operational trade-offs rather than correctness bugs, and they belong with whoever owns rollout. Flagged for the human on this PR.

packages/agent-common/tests/test_retired_alias_resolution.py
```diff
@@ -0,0 +1,136 @@
...
+    reads is used rather than lost to a check-then-act on the first result."""
+    seed_gateway(REGISTERED)
+    seed_defaults({})
+
+    def _land_the_refresh():
+        # Stand in for the refresh thread landing between the two reads.
+        mf._DEFAULTS_CACHE["defaults"] = {"chat": "claude-sonnet-4-6"}
+        return False
+
+    monkeypatch.setattr(mf, "_rearm_defaults_if_unconfirmed", _land_the_refresh)
```
alloy-cicd: **Minor — this test doesn't exercise what its name and docstring claim.** The monkeypatched `_rearm_defaults_if_unconfirmed` runs *inside* `_model_defaults`, which calls it **before** returning `_DEFAULTS_CACHE["defaults"]`. So the map is populated during the **first** read: `_first_default` → `_default_alias_for("chat")` → `_model_defaults()` → (hook populates) → returns the populated map — `found` is set on the first pass and the unconditional second read in `_resolve_alias` never runs. Delete the `if found is None: found = _first_default(...)` retry entirely and this test stays green.

To actually pin "a refresh landing between the two reads is picked up", populate on the second invocation only (a call counter in the stub), or stub `_first_default` to return `None` once and delegate afterwards.

packages/agent-common/agent_common/core/model_factory.py
```diff
@@ -626,18 +690,57 @@ def _resolve_alias(model_type: str, roles: tuple[str, ...], kind: str) -> str:
...
+    elif _default_points_at(roles, model_type):
+        logger.error(
+            "%s model '%s' is not registered and the default for role(s) %s still points at it; "
+            "repoint the default to a registered model — passing through, the gateway will reject it",
+            kind,
+            model_type,
+            ", ".join(roles),
+        )
+    else:
+        logger.error(
```
alloy-cicd, line 769: **Minor — the registered-check reintroduces one false variant of this message.** Now that `_first_default` skips an **unregistered** default (round-1 finding 3), a default that is *set but not registered* (e.g. `{"chat": "also-retired"}`, the exact state `test_unregistered_default_is_not_a_successor` seeds) falls through to this branch and logs "no default is set for role(s) …" — false: a default is set, it's just dead. Same misdirection class as the points-at-retired case fixed in this commit, and it arises from the same admin mistake (default left on a retired alias, only under a different name), so the "repoint the default" message is the right one for it too.

Suggestion: broaden `_default_points_at`'s condition (or add a fourth case) to cover "some role's default is set but not in `registered`" — e.g. "the default '%s' for role(s) %s is not registered; repoint it". `test_unregistered_default_is_not_a_successor` currently only asserts `"falling back" not in caplog.text`, so it would pass either way — worth asserting the message like its sibling test does.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (839e194b8902513aa4087d1482f6c28f88f7402e..e8116bea9b3d0b76a6f0ef908dfe0eb7039793ca)

| # | Finding | Disposition |
|---|---|---|
| 1 | Synchronous fetches on the async hot path | **verified-fixed** — `_rearm_defaults_if_unconfirmed` bounds the degraded state to ≤1 blocking re-population per second, decided under `_DEFAULTS_LOCK`; pinned by `test_rearm_is_rate_limited`. The `_gateway_models` empty-re-arm remains unbounded, deferred by the author as pre-existing — reasonable, noting that this PR's per-request resolution does make that path hotter than before, so the follow-up is worth scheduling. |
| 2 | Double resolution / cache-key–model divergence | **verified-fixed** — `create_model(..., pre_resolved=True)` wired from `GraphFactory._create_model`, guarded by `test_create_model_is_told_the_alias_is_already_resolved`. |
| 3 | Fallback default not validated against registry | **verified-fixed** — `_first_default` now requires `default in registered`; `test_unregistered_default_is_not_a_successor`. |
| 4 | Misleading "no default is set" ERROR | **verified-fixed** for the points-at-retired-alias case (third branch + message assertion). A residual variant exists — see inline. |
| 5 | Racy retry guard / triplicated predicate | **verified-fixed** — decision taken once under the lock, second read unconditional, predicate has one owner. The test claiming to pin the mid-call pickup doesn't actually exercise it — see inline. |
| 6 | Deployment hazard (chat default hard-required) | **still-open by author's deliberate choice** — facts confirmed, the false auto-set-default claim corrected in e8116bea, thread left open as the visible rollout decision for the human reviewer. Not re-raised. |
| 7 | Orphaned client-email comment | **verified-fixed** — inline at all four metadata sites. |
| 8 | Triplicated test fixtures | **verified-fixed** — `tests/conftest.py` with autouse restore (including the new process-global re-arm floor, which turned out to be load-bearing); migrating the two older files deferred, fine. |
| 9 | WARNING volume / gateway-less envs | **verified-fixed** — transition-only WARNING, debug repeats, INFO on recovery. |

**Delta review:** the fixes are correct and well-tested overall; two small new issues, both in or about the new code, posted inline. Nothing blocking.

packages/agent-common/tests/test_retired_alias_resolution.py
aartaria: Fixed in 76cdd080 — and you were right that it was vacuous, including about why: the stub ran inside the first read, so the map was populated before `_model_defaults` returned. The stub now answers empty once and populated thereafter, and asserts both reads happened. I checked it with exactly the falsification you proposed: deleting the `if found is None:` retry makes this test and `test_empty_unerrored_defaults_are_re_read_before_failing_open` both fail, and restoring it makes them pass.

packages/agent-common/agent_common/core/model_factory.py
aartaria: Fixed in 76cdd080. Added the fourth branch, naming the dead alias: "…the default(s) for role(s) chat are not registered either (also-retired); repoint to a registered model". `_default_points_at` is replaced by `_configured_defaults`, which separates "is a default set" from "is it usable" — the distinction the whole message split exists to make, and which my narrower predicate could not express. Good catch that my own round-1 fix re-opened the same misdirection one branch over. `test_unregistered_default_is_not_a_successor` now asserts the message (and the absence of "no default is set") rather than just the absence of the success line, so it would catch a regression either way.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (e8116bea9b3d0b76a6f0ef908dfe0eb7039793ca..76cdd08073095ac9dff6b8a6a71404ab8b4f7acf)

**Ledger of round-2 findings — both verified-fixed in 76cdd080:**

1. **Vacuous mid-call-refresh test** — verified-fixed. The stub now targets `_model_defaults` directly, answers empty on the first read and populated on the second, and asserts `len(reads) == 2`, so the unconditional second read in `_resolve_alias` is genuinely load-bearing for the test. The author verified it with exactly the proposed falsification (deleting the retry fails this test and `test_empty_unerrored_defaults_are_re_read_before_failing_open`; restoring it passes both).
2. **Set-but-unregistered default logged as "no default is set"** — verified-fixed. `_default_points_at` is replaced by `_configured_defaults` (separating "is a default set" from "is it usable"), a fourth branch names the dead alias with the same "repoint" instruction, and `test_unregistered_default_is_not_a_successor` now asserts the message and the absence of "no default is set", so it would catch a regression in either direction.

**Carried over:** the deployment-hazard thread on `client-slack/src/services/a2aClientService.ts` remains open by the author's deliberate choice as the visible rollout decision (startup check / alert / persisted last-known-good default) for the human reviewer — facts confirmed in round 2, not re-raised.

**Delta review: nothing new.** The only observation, below the threshold for a finding: in the rare mixed multi-role state (one role's default pointing at the retired alias, another role's default on a different unregistered alias) the "still points at it" branch lists all configured roles, not just the pointing one — the named fix is the same either way, so no action needed.

Every actionable finding from previous rounds is resolved and the delta introduces nothing new. No inline comments this round.

Issue #180, opened 2026-08-31
Extract the in-task-auth protocol reader: three hand-synced copies


aartaria:
The reader for `urn:nannos:a2a:in-task-auth:1.0` now exists three times, hand-synced:

- `packages/embed-sdk/src/transport/demux.ts` (`readAuthDataPart` / `readAuthRequired`)
- `packages/client-slack/src/utils/inTaskAuth.ts`
- `packages/client-google-chat/src/utils/inTaskAuth.ts`

The two chat clients' copies are byte-identical for ~130 lines (payload → metadata → prose precedence, the opaque-tool filter, the decision prose and DataPart shape); the SDK's differs only because it reads wire-shaped parts instead of `@a2a-js/sdk` types. A change to the precedence, the opaque-tool list, or the decision vocabulary has to be made in three places, and nothing fails if it is made in two.

## Why it wasn't done in the PR that created the second copy (#179)

`client-slack` and `client-google-chat` are not npm workspaces — the root `package.json` only lists `packages/embed-sdk` and `packages/console-frontend`. Each client installs independently and builds its own Docker image, so sharing code means a real packaging decision: a published/linked internal package, or workspace membership plus Dockerfile changes for both clients. That is bigger than the feature that surfaced it.

## Scope

- Decide the mechanism (workspace member vs. published internal package) — this is the actual question.
- Move the protocol half (part reading, precedence, opaque tools, decision vocabulary) into it; leave the presentation half (Block Kit / cardV2 builders) in each client.
- Keep the embed-sdk copy in step, or migrate it too if the wire/type mismatch can be bridged cheaply.

PR #246, opened 2026-09-16
feat(agent-runner, console-backend, clients): park a scheduled run on authorization instead of failing it

### LINKED ISSUE #245: Scheduler: a run blocked on a missing credential is failed or silently greened, and the owner is never asked
## Problem

A scheduled job runs unattended, but the credentials its tools need belong to a *person*.
When the MCP gateway answers a tool call with `need-credentials`, what is missing is the
job owner's consent — not a service, not a configuration, not anything the runtime can
supply. The run is blocked on a human, which is the one condition the scheduler has no
vocabulary for.

Today it has none at all. Three gaps compose:

1. **Nothing detects the error.** `AuthErrorDetectionMiddleware` is orchestrator-private,
   and agent-runner builds its sub-agent graph with `extra_middlewares=None`. The
   gateway's structured payload arrives as an ordinary `ToolMessage` and the model
   paraphrases it.
2. **Nothing can report it.** `SubAgentResponseSchema.task_state` is
   `completed | input_required | failed`; `_TERMINAL_TASK_STATE_NAMES` maps the same
   three; `JobRunStatus` has no auth state. The model's own verdict is the only signal.
3. **The verdict is destructive.** A model that says `failed` increments
   `consecutive_failures`, and the job is auto-paused at `max_failures`. A model that says
   `completed` instead leaves a silently green run that did nothing. Either way the owner
   is never asked for the credential, so the job cannot recover on its own — and the first
   outcome eventually disables it.

A missing credential is not evidence about the job. It is evidence about what the owner
has authorized, and it is fixable in one click — but only if somebody is asked.

## Proposed direction

A run blocked on authorization should **park**: keep its place in the schedule, ask its
owner on the delivery channel it already uses, and resume where it stopped when answered.
It is neither a success nor a failure, and should touch neither `consecutive_failures` nor
`next_run_at`.

Design is written up in `docs/adr/0009-authorization-parks-a-scheduled-run-it-does-not-fail-it.md`,
building on ADR-0008 (the in-task-auth vocabulary and durable parked tasks in
`agent_common.a2a`) and ADR-0007 (interrupted is not failed).

## Related

- #159 — auto-mode headless tool *approval* for scheduled runs is a different question and
  deliberately out of scope here: an approval has no per-owner key that a credential has,
  and a job that parks on every risky call is unusable against an unresponsive owner.
- #180, #181 — in-task-auth protocol readers and client-email coverage.

aartaria:
closes #245

> **Stacked PR.** Base is `fix/retired-model-alias-resolution` (#244), not `main` — the
> branch was cut from it. Review this diff alone; merge #244 first and this retargets to
> `main` on its own.

> 🤖 **Authored by Claude Code** (Opus 5) on @aartaria's machine, working from
> ADR-0009 — GitHub does not badge App attribution onto pull requests, so it is
> stated here instead (docs/adr/0003).

## What

A scheduled run that hits `need-credentials` is blocked on a *person*, and the scheduler
had no word for that. The model's paraphrase was the only signal: `failed` spent the job's
`max_failures` budget on a condition no retry can fix and eventually disabled the job,
`completed` left a green run that did nothing, and either way the owner was never asked for
the credential that would unblock it.

A blocked run now **parks**. It ends as `JobRunStatus.AUTH_REQUIRED` — neutral about the
job, touching neither `consecutive_failures` nor `next_run_at` — asks its owner on the
delivery channel the job already uses, and resumes where it stopped when answered.

Design: `docs/adr/0009-authorization-parks-a-scheduled-run-it-does-not-fail-it.md` (in this
PR), on top of ADR-0008 (in-task-auth vocabulary, durable parked tasks) and ADR-0007
(interrupted is not failed).

## The load-bearing choices

- **The park holds the job's schedule** rather than letting each occurrence park in turn.
  A blocked hourly job would otherwise burn 168 partial runs a week, each leaving a live
  checkpoint and its own answerable card — so pressing three of them does the job's work
  three times. Holding the schedule makes that impossible by construction instead of by a
  dedupe rule, a supersession sweep and a cancel path. Run-now refuses while a park is
  answerable, because it bypasses `claim_due_jobs`.
- **agent-runner's sub-agent construction converges with the orchestrator's** — the larger
  half of the diff. It stops assembling a graph inline and adopts
  `DynamicLocalAgentRunnable`, so one path builds every local sub-agent. Five behaviours
  that were agent-runner's alone survive the move (stall watchdog, tool-discovery failure
  policy, model-call budget — now a constructor argument, since an unattended run *should*
  get a larger one than a delegated turn — stateless catalogue listing, channel-aware
  formatting); three now benefit the orchestrator's delegated turns too.
  Worth a reviewer's attention: `LocalA2ARunnable.astream` catches everything and yields an
  `ErrorEvent` rather than raising, so the executor's `except` branch no longer runs for a
  local sub-agent. Run status therefore follows the sub-agent's **terminal task state**.
  Without that, this refactor would have recorded every crash as a success.
- **The durable task store is installed on both handlers.** A park leaves two task records
  — the outer one agent-runner's HTTP handler owns and the answer addresses by id, and the
  inner one `LocalA2AServer` owns. Making only the inner durable fails at exactly the moment
  this exists for: a restart between park and answer finds the sub-agent's task and not the
  outer one. It is agent-runner's *own* store, not the orchestrator's — both services
  default to the same schema in the same database, so a shared table would let either
  `tasks/get` and `tasks/cancel` the other's tasks.
- **The ask rides part zero of the scheduler payload** the clients already parse, instead of
  being shaped by the auth extension. A webhook POST is blind — there is no extension
  negotiation — and each client gives up silently on a part zero it cannot parse, so an
  extension-shaped status message would vanish in three clients at once. `agent_message`
  carries the authorize URL literally: upgraded clients render a card, un-upgraded ones post
  a working sentence.
- **The reply target is declared by the ask and validated by the client** against its own
  configuration. A pushed A2A task carries no statement of where its server lives, and a
  reply target blindly trusted off a webhook is a credential-forwarding primitive.
- **Resume goes through console-backend**, addressed to the parked outer task, as a new
  `RunTrigger.RESUMED` run. Letting the client answer agent-runner directly is the
  protocol-natural route and loses the run: nothing would create a run row, heartbeat it, or
  let the healer sweep an agent-runner that died mid-resume. Rejected alternatives are
  written up in ADR-0009 decision 6.

## Also in here

- A job that auto-pauses at `max_failures` now tells its owner, as a durable console
  notification (`NotificationType.SCHEDULED_JOB_PAUSED`). It cannot ride the delivery
  channel, which is optional and absent on exactly the jobs whose silence goes unnoticed.
- An empty tool whitelist means "everything" for the general-purpose agent, matching the
  orchestrator. agent-runner read it as "nothing", so the same agent had opposite capability
  depending on who started it. Strictly outside the decision, fixed here because a scheduled
  run cannot reach a credential-gated tool at all without it.
- `sqlalchemy[asyncio]` is now explicit in agent-runner: the async engine needs greenlet,
  which resolved locally by accident and not in the image.

## Migrations

`097` `auth_required` run status · `098` `parked_task_id` + `parked_payload` (JSONB) on
`scheduled_job_runs` · `099` the paused-job notification type.

## Known gaps (deliberate, named in the ADR)

- **A parked run has no expiry** — a job whose owner never answers stays stopped. Closing a
  park after some age is the natural follow-up and is not in this cut.
- **`delivered` on a run is inferred, not observed** — an ask that was never delivered is
  recorded as delivered, which matters more for an ask than a result. Predates this change
  (#191).
- **`agent-runner/agent/mcp_tools.py` is left unwired**, kept one release because its tests
  still document the console/gateway audience split. Port those, then delete it.
- **Unattended runs still execute sandbox code unguarded.** Unchanged here, but parking on
  an approval is now technically available, so the question is live — it needs its own
  decision (#159).

## Verification

All suites green locally: agent-common 1181 · agent-runner 63 · console-backend 1961 ·
orchestrator-agent 945 · client-slack 85 · client-google-chat 30. `tsc -b` clean on
console-frontend, `tsc --noEmit` clean on both clients.

Pre-existing ruff (console-backend, 22) and eslint (console-frontend, 171) errors are all in
files this PR does not touch — zero in touched files. Client eslint could not run locally:
`@alloy-ch/eslint-config-alloy` is not installed in this environment.

## Checklist

---

## Automatic review round 1 — 16 findings

Four were blocking bugs in the park/resume lifecycle, three of them contradicting ADR-0009's
own text. All fixed in `e9ec7a4e`, along with six smaller ones: **[1]** one park permanently
stopped the job's schedule (`claim_due_jobs` keyed on the status, which a parked run keeps
forever, instead of on `parked_task_id`) · **[2]** a park with no ask was recorded green ·
**[3]** an unparseable parked payload fell through to success, making the guard below it
unreachable · **[4]** the ask was consumed before the resumed run existed, and a `resumed`
run earned no retry when swept · **[5]** task-store durability was password-gated while the
checkpointer is host-gated — the split decision 3 forbids — and ignored `POSTGRES_SCHEMA`.
Then **[9]** 202 doc, **[10]** parked run-now badge, **[11]** duplicate `recordDecision`,
**[12]** stale comment, **[13]** an unreachable guard that would have escaped into a
background task.

**[7] is also fixed** (`a26cfe83`), after the finding was narrowed: the listing only ever
applied to general-purpose / `all_tools` scheduled sub-agents (`_is_full_catalogue_agent`),
and never to jobs with no sub-agent at all, so "jobs whose work needs no gateway tool"
overstated it. In that scope, `tools/list` now retries transient failures with backoff
(`is_retryable_mcp_error`, same policy as the orchestrator), a discovery failure is recorded
`INTERRUPTED` rather than `FAILED` — ADR-0007's "evidence about the runtime, not the job", so
it neither counts toward `max_failures` nor auto-pauses, and earns one fresh attempt — and the
two server listings run concurrently. A catalogue cache was deliberately *not* added: a
scheduled run has no warm parent turn to inherit from, so there is no analogue of the
orchestrator's `_pre_resolved_tools_for` and the only reuse available is across runs, which
schedules are too sparse to hit.

### Two duplications consolidated (`9f077c87`)

Both copies the review pointed at are now single implementations, done here rather than
deferred because neither needed a new dependency:

- **A2A task store → `ringier_a2a_sdk.agent.task_store`**, beside `postgres_checkpointer_mixin`
  (same concern, and the SDK already requires sqlalchemy + a2a-sdk). Two knobs carry the real
  difference: `table_name` — agent-runner names its own so neither service can `tasks/cancel`
  the other's — and `schema`, pinned for agent-runner's new table but *not* for the
  orchestrator's existing one, since pinning an existing table does not move it, it creates an
  empty one elsewhere and hides the tasks the service already had.
- **Discovery retry + concurrency cap → `agent_common.core.catalogue_ingest.fetch_with_retry`**,
  taking a coroutine factory so neither caller was reshaped. The process-wide cap moves with it,
  keeping both hard-won properties: process-wide because a per-call limit still allows
  `limit x N`, and the slot covers the fetch but not the backoff sleep.

What is deliberately *not* shared is what a failure means — the orchestrator tolerates a
partial catalogue, a scheduled run does not (ADR-0009) — which stays at each call site.

⚠️ **One behaviour change in the orchestrator**, worth a reviewer's eye since it is a second
service moving in a PR about scheduled runs: its task-store gate was `POSTGRES_HOST` AND
`POSTGRES_PASSWORD` and is now host alone, matching the checkpointer's documented convention.
It is the same latent split finding [5] identified in agent-runner; sharing the code made the
two copies disagree out loud. Its test asserted the old gate and now asserts the new one.

**[8] is fixed** (`941ac690`), on the assessor's side rather than by gating the harvesting —
teaching a shared transport function who is calling it is the worse fix, and the harvesting
helps every caller. The assessor now checks the run's terminal state before parsing what it
returned, so a non-completed run is recorded `assessment_unavailable` again instead of as a
real-looking "caution" verdict. The finding's parenthetical is fixed too: agent-runner refuses
to publish a park when there is no `scheduled_job_id`, because such a task is structurally
unanswerable (parking holds a *job's* schedule, the ask rides the *job's* channel, the answer
returns addressed to a *run*) and would now be a permanent orphan in the durable store.

**[14] is fixed** (`e9d0a5b8`): the threading was not intentional. Google Chat now sends the
ask's coordinates (a space and a thread, where Slack's are a channel and a ts) and threads the
resumed run's result under them — no backend change needed, since `reply_to` was already stored
opaquely. The two clients stay as two copies by decision (no shared TS package exists today,
and `a2aPushPayload.ts` is duplicated the same way), but they are back in step and the Google
Chat copy now has the tests the Slack one had.

**[16] is fixed** (`3e19176d`) after verifying the claim it rested on: the two derivations
were provably identical (single call chain, `context_id` forwarded unmodified, never
reassigned anywhere in the file, and `scheduled_run_task_id` is pure). Only `is_resume: bool`
now travels down, and the rule moved into `sub_agent_task_ids(context_id, *, is_resume)` —
one owner of "which id goes in which field", and directly testable, which the inline version
was not.

**[6] is fixed** (`62ceb7e8`). The race ran the way that is easy to read backwards: the
resumed run overwrote the owner. A schedule edit recomputes `next_run_at`, and the resumed
run then restated the value it had snapshotted at click time, so the edit was silently
reverted. The cause was that `next_run_at=NULL` meant both "keep the column" and "retire the
job", leaving no way to say "I have nothing to say about the schedule" —
`complete_job` now takes `leave_schedule`, and a resumed run sends no time at all, so there
is no stale write rather than a narrower window for one.

**Review round 2 (`62ceb7e8`) verified all 15 fixes against the new head and found one new
issue, [17]** — a regression from this PR's own concurrency change: a failed listing orphaned
its sibling, which then retried against a closing HTTP client and logged "Task exception was
never retrieved". Fixed in `181683f4` by awaiting every task before propagating, with
fail-don't-degrade unchanged. Round 2 also confirmed the resume path has no double
run-creation or double-finalize, the claim guard is closed on both sides of the commit, and
the once-job park→resume lifecycle retires correctly.

**All 17 findings are now closed.** Fifteen are fixed in this PR; **[15]** is filed as #250
— five independent parsers of the `auth_requirement` shape, with the console rendering tool
names the chat cards deliberately suppress. It spans four packages and is a disclosure
decision rather than a correctness fix, so it did not belong here. Every review thread is
replied to and resolved.

Gates after the fixes: console-backend 1967 · agent-runner 66 · agent-common 1181 ·
orchestrator-agent 945 · client-slack 85 · client-google-chat 30 · `tsc -b` clean.


alloy-cicd (commented):

Reviewed at head `53b3a5e8` (base PR #244 is still open and live, so the stacked diff is clean). 16 inline findings, each verified against the checked-out tree; numbered `[1]`–`[16]` for follow-up rounds.

**Blocking — the park/resume lifecycle has four correctness holes that compound:**
- **[1]** one park permanently stops the job's schedule, even after the ask is answered (`claim_due_jobs` matches the historical `auth_required` row forever);
- **[2]** a park with no ask is recorded as a green SUCCESS in agent-runner, contradicting its own warning (and a new test pins the wrong behavior);
- **[3]** console-backend's unparseable-payload fallback maps task state `auth_required` to SUCCESS, so the `parked_task_id` guard never protects the case it was written for;
- **[4]** the ask is committed away before the background resume exists — a crash in between strands the run unrecoverably.

**Should-fix:** [5] agent-runner's task store is in-memory whenever `POSTGRES_PASSWORD` is empty (durable checkpointer + volatile task store = ADR-0009's forbidden configuration) and ignores `POSTGRES_SCHEMA`; [6] the resume finalizes from a stale job snapshot and can rewind `next_run_at`; [7] full-catalogue scheduled jobs now fail (and march toward auto-pause) on any gateway outage; [8] the shared `dispatch_streaming` change silently alters the skill-security assessor's failure semantics.

**Smaller:** [9] resume endpoint returns the parked run's id while documenting the new one's; [10] parked run-now renders past-tense in the console; [11] duplicate `recordDecision` in the Slack handler; [12] stale contradictory comment + discarded message parts on the resume path; [13] dead `AlreadyAnswered` guard; [14] Google Chat resume service is an already-diverged copy of the Slack one (no threading, no tests); [15] fourth `auth_requirement` parser without opaque-tool filtering; [16] redundant `resume_task_id` plumbing.

**Notes without a diff anchor:**
- `packages/console-backend/AGENTS.md` §"Scheduled Run Vocabulary and Interruption" declares itself the contract but still enumerates three triggers (`scheduled`/`retry`/`manual`) and migrations 091/092 — this PR adds `RESUMED`, the schedule-holding `AUTH_REQUIRED` status, and migrations 097–099 without touching it.
- For the human reviewer's judgment: per ADR-0009 (decision 2 and Constraints), an unattended full-catalogue scheduled run now reaches every gateway tool with `risk_scorer=None` — no per-call approval. The ADR documents this as a deliberate posture ("Guarding unattended runs deserves its own decision"), so it is not reported as a bug, but it is a security-posture change worth an explicit sign-off given the interactive path does install the risk scorer.

Checked and NOT flagged (verified non-issues): `read_personal_file` is context-gated by design and was never usable in scheduled runs pre-PR; `_authorization_answer` is deliberately narrower than `answer_from_message` (using the shared helper would misroute text-only dispatches); the duplicate `recordDecision` cannot lose the answer; a fresh dispatch cannot carry an `authorization` DataPart from any current producer; `job.next_run_at` can never be NULL on the resume path.

packages/console-backend/console_backend/repositories/scheduled_job_repository.py
```diff
@@ -218,7 +220,8 @@ async def claim_due_jobs(self, db: AsyncSession, limit: int = 10) -> list[Claime
                   )
                   AND NOT EXISTS (
                         SELECT 1 FROM scheduled_job_runs r
-                        WHERE r.job_id = j.id AND r.status = 'running'
+                        WHERE r.job_id = j.id
+                          AND r.status IN ('running', 'auth_required')
```
alloy-cicd: **[1] Bug (blocking): one park permanently stops the job's schedule — even after it is answered.**

`clear_parked_task` only NULLs `parked_task_id` and deliberately keeps `status='auth_required'` forever, and no other writer ever rewrites that row's status (`complete_run`, `close_run_minimally`, `interrupt_stale_runs` are all gated on `status='running'`; the resumed run is a new row). So after the first park, this `NOT EXISTS … status IN ('running', 'auth_required')` matches the historical parked row forever and the job is never claimed again — still enabled, no pause notification, `next_run_at` in the past, completely silent. Only `Run now` still works (it checks `answerable_parked_run`, which correctly keys on `parked_task_id IS NOT NULL`).

The `auth_required` arm needs the same key: e.g. `(r.status = 'running' OR (r.status = 'auth_required' AND r.parked_task_id IS NOT NULL))`.

Secondary: neither partial index (`idx_scheduled_job_runs_running` from 093, `idx_scheduled_job_runs_parked` from 098 — whose comment claims it "mirrors" the running one) can serve a `status IN (…)` predicate, so the subquery scans the job's whole run history (via `idx_scheduled_job_runs_job`) per candidate job per tick. Splitting the predicate per status (the OR form above, or two NOT EXISTS) lets each arm be an index probe again.

packages/agent-runner/agent/core.py
```diff
@@ -770,8 +919,34 @@ def _meta_int(key: str) -> int | None:
...
+        # local agent any more. Left alone that turns "the agent blew up" into a success
+        # row whose result text happens to start with "Error:", which is exactly the
+        # silently-green run this ADR exists to abolish, arriving through the ADR's own
+        # refactor. The task state is the authority on how the run ended; the status
+        # follows it.
+        failed = sub_agent_task_state == "failed"
+
         result_meta = {
-            "scheduler_status": "success",
+            "scheduler_status": (
```
alloy-cicd, line 1039: **[2] Bug: a park with no ask is recorded as a SUCCESS, contradicting the warning a few lines up.**

`parked` (`sub_agent_task_state == "auth_required"`) and `failed` (`== "failed"`) are mutually exclusive, so with `parked=True, auth_payload=None` this ternary falls through to `"success"` and the final yield emits `TASK_STATE_COMPLETED` — while the log line above says "this is reported as a failure instead". console-backend then records SUCCESS and resets `consecutive_failures`: the silently-green run ADR-0009 exists to abolish, on exactly the path the warning anticipates (`_client_auth_payload` returning None — missing `auth_info` metadata, or `AuthPayload(**auth_info)` raising).

Note `test_a_park_with_no_ask_is_not_a_park` (`tests/test_agent_runner_security.py:672`) currently pins the wrong behavior: it asserts the state is not `auth_required` and is `TASK_STATE_COMPLETED`, never that the run failed.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -756,11 +995,31 @@ def _parse_result(self, data: dict[str, Any]) -> tuple[JobRunStatus, str | None,
...
-            status,
-            meta.get("agent_message"),
-            meta.get("error_message"),
-            conversation_id,
+        # A parked run carries the two things needed to reach it again. The task id is
+        # load-bearing: without it the run is stopped with nothing to address, so a
+        # payload claiming AUTH_REQUIRED without one is not a park at all and is
+        # recorded as a failure rather than silently stopping the job forever.
+        parked_task_id = meta.get("parked_task_id")
+        if status == JobRunStatus.AUTH_REQUIRED and not parked_task_id:
```
alloy-cicd, line 1023: **[3] Bug: this guard never fires for the case that actually produces the problem.**

It requires the status text to have parsed as JSON AND to name `scheduler_status: "auth_required"` while omitting `parked_task_id`. But the dangerous input is the payload that does NOT parse: `json.loads` fails → `meta = {"result_summary": text}` with no `scheduler_status` → the state fallback above maps A2A task state `auth_required` through its `else` branch to `"success"`. A parked run whose status text is prose or truncated (or comes from an older runner) is recorded as a plain SUCCESS, `consecutive_failures` resets, and the ask is dropped.

The fallback should map the task state `auth_required` explicitly — to failed, per this guard's own policy — instead of defaulting every non-`failed` state to success.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -1082,6 +1104,71 @@ async def resume_job(
...
+    # AUTH_REQUIRED status — it is a true record of how that occurrence ended — so the
+    # task id is what says a question is still outstanding, and clearing it is the claim.
+    # The check above is the cheap path; this conditional write is what actually decides,
+    # because two clicks can pass the check together.
+    if not await engine._repo.clear_parked_task(db, run_id):
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail="This run has already been answered.",
+        )
+    await db.commit()
```
alloy-cicd, line 1176: **[4] Bug (crash window): the ask is consumed before the resume exists.**

`clear_parked_task` + `commit()` land in-request; the resume runs in a `background_tasks.add_task` that may never execute (pod roll/OOM between commit and task, or `resume_parked_run` failing before `create_run`). The run is then `auth_required` with `parked_task_id = NULL`: the console card disappears (`answerable_parked_run` requires the id), a second click 409s as "already answered", no RESUMED run exists, and no sweep recovers it (`interrupt_stale_runs` touches only `running` rows). Combined with [1] the job is then dead silently, and the agent-runner task stays parked forever.

Consider clearing the parked id inside `resume_parked_run` once the new run row exists (keeping the conditional write as the race gate), or adding a startup sweep for answered-but-unresumed parks.

packages/agent-runner/agent/task_store.py
```diff
@@ -0,0 +1,62 @@
...
+
+def create_task_store() -> tuple[TaskStore, AsyncEngine | None]:
+    """Create the A2A task store: PostgreSQL-backed when configured, in-memory otherwise.
+
+    Returns the store and the engine backing it (None for the in-memory fallback);
+    the caller owns disposal. The in-memory fallback exists for local runs without a
+    database — a parked run cannot survive a restart there, which is a limitation of
+    running without Postgres and not a mode to deploy.
+    """
+    host = os.getenv("POSTGRES_HOST")
```
alloy-cicd: **[5] Bug: durability is password-gated while the checkpointer is host-gated, and `POSTGRES_SCHEMA` is ignored.**

(a) This falls back to `InMemoryTaskStore` unless both `POSTGRES_HOST` and `POSTGRES_PASSWORD` are non-empty, but the checkpointer in `core.py` (~129–173) requires only host (password defaults to `""`). An IAM/trust/.pgpass deployment therefore gets a durable checkpointer plus a volatile task store — the configuration ADR-0009 decision 3 says must not exist: a restart between park and answer loses the parked task, `message/send` fails with "Task not found", and the run is unresumable, with only a startup WARNING to distinguish it.

(b) Unlike the checkpointer (which passes `schema`), no schema/search_path is set here, so `agent_runner_tasks` is created in the role-default schema; replicas or environments with different role defaults would not see each other's parked tasks.

Also worth noting: this module is a near-verbatim copy of `orchestrator-agent/app/core/task_store.py` (same gate, pool sizing, driver, warning text). Lifting one `create_task_store(table_name=…)` into agent-common would keep the pool/driver/fallback policy from forking a third time.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -802,19 +1076,30 @@ async def _finalize(
...
+        # parameter, so a resumed run has to say the schedule it already has.
         try:
-            next_run_at = compute_next_run(
-                schedule_kind=job.schedule_kind,
-                cron_expr=job.cron_expr,
-                interval_seconds=job.interval_seconds,
-                run_at=job.run_at,
-                after=datetime.now(timezone.utc),
-                tz=job.timezone,
+            next_run_at = (
```
alloy-cicd, line 1118: **[6] Race (narrow): the resumed run finalizes with a job snapshot taken at click time.**

`job` is loaded in the router at request time and never re-read; `_finalize` writes `job.next_run_at` back through `complete_job`'s COALESCE. `claim_due_jobs` cannot advance it meanwhile (the RESUMED run's `running` row blocks claims), but a concurrent Run-now or a schedule edit can — and this write-back then rewinds `next_run_at` to the older timestamp, re-arming a past occurrence. Re-reading the job row inside `_finalize`, or passing an explicit "leave unchanged" marker, avoids it.

(The NULL→disable trap the comment above worries about is unreachable here: the column is `TIMESTAMPTZ NOT NULL`.)

packages/agent-runner/agent/core.py
```diff
@@ -1016,220 +1229,199 @@ async def _run_langgraph_agent(
...
+            resolver = McpToolResolver(
+                token_provider=token_provider,
+                gateway_url=_MCP_GATEWAY_URL,
+                gateway_client_id=_MCP_GATEWAY_CLIENT_ID,
+                console_mcp_url=f"{_CONSOLE_BACKEND_URL}/mcp",
+                console_client_id=_CONSOLE_BACKEND_CLIENT_ID,
+                timeout=timedelta(seconds=_MCP_TIMEOUT_SECONDS),
+                stateless_list=_MCP_CATALOGUE_STATELESS_LIST,
+            )
+            catalogue_tools = await resolver.resolve_all()
```
alloy-cicd: **[7] New failure mode: full-catalogue scheduled runs now hard-depend on the gateway.**

Pre-PR, an agent with an empty whitelist never touched the gateway; now `resolve_all()` (two full catalogue listings on the critical path, uncached across runs in the same tick) raises on any gateway or token-exchange outage, the outer handler records the run FAILED, and `consecutive_failures` marches toward auto-pause — including for jobs whose work needs no gateway tool at all. Fail-don't-degrade for discovery is ADR-0009's stated policy, so not a bug per se — but the coupling of auto-pause to gateway availability (and the per-run listing cost) deserves a deliberate decision: e.g. not counting discovery failures toward `max_failures`, and/or a short-TTL per-(server, user) catalogue cache.

packages/console-backend/console_backend/utils/a2a_dispatch.py
```diff
@@ -203,6 +241,10 @@ async def _dispatch_streaming(
                 result_context_id = ev.context_id or result_context_id
                 if ev.status.state in _TERMINAL_STATES:
                     final_state = _TERMINAL_STATES[ev.status.state]
+                if ev.status.state in _STATES_CARRYING_PAYLOAD_IN_STATUS and ev.status.HasField("message"):
```
alloy-cicd, line 244: **[8] Side effect on non-scheduler callers: the status harvesting is unconditional.**

`dispatch_streaming` is shared with `skill_security_service.py` and `debug_agent_service.py`. Shaping FAILED status text into `artifacts[-1]` means a failed assessor run no longer raises "Assessor agent returned no artifact": the scheduler-shaped JSON parses and `_parse_assessment_response` yields verdict "caution" with empty indicators and "No reasoning provided." — the same verdict as the old fallback, but the persisted record now looks like a real assessment and the `assessment_unavailable` indicator + warning log are lost. (An assessor that parks on AUTH_REQUIRED would also leave a permanently open task nobody answers, since `AuthErrorDetectionMiddleware` is installed regardless of `scheduled_job_id`.) Consider gating the harvesting on the scheduler caller, or having the assessor path check the final state.

packages/console-backend/console_backend/routers/scheduler_router.py
```diff
@@ -1082,6 +1104,71 @@ async def resume_job(
...
+@router.post(
+    "/jobs/{job_id}/runs/{run_id}/resume",
+    response_model=RunNowResponse,
+    status_code=202,
+    summary="Answer a run parked on the owner's authorization.",
+    description=(
+        "Continues a run that stopped because a tool needed the owner's credential. The "
+        "answer is delivered to the parked agent-runner task, the agent retries what was "
+        "blocked (or is told to stop, on a decline), and the result is delivered to the "
+        "job's channel like any other run. Returns 202 with the id of the new run that "
```
alloy-cicd: **[9] Doc bug: the 202 carries the PARKED run's id, not "the id of the new run that carries the continued work".**

The new run is only created later inside the background task, so the promised id cannot be known at response time. No current consumer reads the body's `run_id` (the frontend discards `resumeParkedRun`'s return; the Slack service checks only `response.status`), so the fix is to correct the description (or return the parked id labeled as such) before someone polls the parked row — which stays `auth_required` forever — for the resumed result.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -1152,7 +1280,7 @@ export function SchedulerJobDetailPage() {
                 runNowResult && (
                   <div className="flex flex-wrap items-center justify-between gap-2">
                     <div className="flex items-center gap-2">
-                      <RunStatusBadge status={runNowResult.status} />
+                      <RunStatusBadge run={runNowResult} />
```
alloy-cicd, line 1288: **[10] UI: a parked run-now renders in the past tense.**

`RunStatusBadge` keys "Waiting for authorization" on `run.parked_task_id`, but the socket payload built in `scheduler_engine.py` (~1208) carries no `parked_task_id` (this typechecks only because the Pick'd field is optional), so a manually-triggered run that parks shows "Stopped for authorization" while it is actually still waiting. The polled runs table corrects it moments later; include `parked_task_id` in the notification payload or special-case `auth_required` in this chip.

packages/client-slack/src/listeners/actions/inTaskAuthButton.ts
```diff
@@ -69,6 +74,55 @@ export function registerInTaskAuthActions(app: App, makeDeps: () => HandlerDepen
...
+        // address the parked task directly (ADR-0009).
+        if (replyTo) {
+          if (!resumeService) {
+            logger.error(
+              `Cannot answer scheduled run: CONSOLE_BACKEND_URL is not configured, so this ` +
+                `client has nowhere to send the authorization ${decision}`
+            );
+            return;
+          }
+          await recordDecision(
```
alloy-cicd: **[11] Duplicate call: `recordDecision` already ran unconditionally at ~line 67 with byte-identical arguments.**

This second invocation is dead weight — an extra Slack API round-trip per click before the resume. (Verified it cannot abort the resume: every failure path inside `recordDecision` is swallowed; and scheduled asks carry no `streamMessageTs`, so both calls take the plain `chat.update` branch.) Delete one of the two.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -297,6 +381,159 @@ async def run_job_now(self, job: ScheduledJob, run_id: int | None = None) -> Non
...
+                    # derived from the context, which the whole chain shares.
+                    _, metadata, push_config = await self._build_message_args(
+                        job, run_id, access_token, db
+                    )
+                    # Opaque to everything in between: the delivery channel put it on the
+                    # card and is the only thing that can read it back.
+                    if reply_to:
+                        metadata["reply_to_message"] = reply_to
+
+            # The run id in the metadata stays the PARKED run's. agent-runner derives the
```
alloy-cicd: **[12] Stale comment — it contradicts both the code and the comment 14 lines up.**

`_build_message_args(job, run_id, …)` stamps THIS run's id (as the comment at 453 correctly says), and `scheduled_run_task_id` keys on the context alone (its own docstring: "Keyed on the CONTEXT alone, not on the run"), not on `(context_id, scheduled_job_run_id)`. This block describes an abandoned design; delete or rewrite it.

Related, one line up: the `_` discard means `_build_message_args` builds full message parts only to throw them away — including `_resolve_voice_agent_id`'s extra SELECT on every resume of a voice job and, for a notification-shaped job, a potential LLM call. A split `_dispatch_metadata(job, run_id, db)` returning only `(metadata, push_config)` would keep that work off the resume path.

packages/console-backend/console_backend/services/scheduler_engine.py
```diff
@@ -297,6 +381,159 @@ async def run_job_now(self, job: ScheduledJob, run_id: int | None = None) -> Non
...
+        The caller has already CLAIMED the ask (``clear_parked_task``), so this is only
+        ever reached once per parked run — which is what stops a second click resuming a
+        task that has since gone terminal.
+
+        *decision* is passed through as the agent sees it. A decline resumes the task too:
+        the agent is told to stop and the run closes on its own terms, instead of staying
+        parked on a question that has been answered.
+        """
+        if parked_run.status != JobRunStatus.AUTH_REQUIRED or not parked_run.parked_task_id:
+            raise AlreadyAnswered(f"Run {parked_run.id} of job {job.id} is not waiting for an answer")
```
alloy-cicd: **[13] Dead guard: `AlreadyAnswered` can never fire on the production path.**

The router validates and clears `parked_task_id` in the DB but passes the pre-clear in-memory `run` object, so this attribute is always still set here. And if it ever did fire, it would escape unhandled into a Starlette background task — nothing catches it, and `run_id` is still None so no run row records the attempt. The real race gate is `clear_parked_task`'s conditional write plus the router's 409. Suggest dropping the exception class (only `test_scheduler_engine.py:1567` observes it) or downgrading this to a logged early-return.

packages/client-google-chat/src/services/scheduledRunResumeService.ts
```diff
@@ -0,0 +1,139 @@
...
+ * orchestrator would propose a fresh delegation task id on a sub-agent thread
+ * that is already parked, and the executor would reject it. See
+ * docs/adr/0009-authorization-parks-a-scheduled-run-it-does-not-fail-it.md.
+ *
+ * So the answer goes to console-backend, which owns the run: it re-resolves the
+ * owner's offline token, addresses the parked agent-runner task directly, and
+ * records the continued work as a run of its own. That last part is why the
+ * scheduler brokers it rather than this client calling agent-runner — a resumed
+ * run is minutes of tool calls that must be heartbeated and swept if the process
+ * executing them dies, and a floating promise in a Slack handler owns none of it.
```
alloy-cicd: **[14] Copy divergence: this is a line-for-line copy of the Slack service — this comment even still says "a floating promise in a Slack handler" — except the Slack copy sends `reply_to` (from `askMessage`) and this one doesn't.**

So Google Chat resume results arrive unthreaded (its `a2aNotificationHandler` has no `reply_to_message` handling), while ADR-0009 calls the chain "load-bearing". The console-backend contract (endpoint path, `{service, endpoint}` target validation, 409 semantics) is now asserted in two clients, with tests covering only the Slack copy. There is no shared TS client package today (`a2aPushPayload.ts` is already duplicated the same way), so at minimum: is the missing Google Chat threading intentional?

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
```diff
@@ -896,6 +917,104 @@ function EditForm({ job }: { job: ScheduledJob }) {
   );
 }
 
+// ---------------------------------------------------------------------------
+// The ask a parked run is waiting on
+// ---------------------------------------------------------------------------
+
+/** The authorize URL and the service it belongs to, out of the stored ask. */
+function readParkedAsk(run: ScheduledJobRun): { authUrl?: string; subject?: string } | null {
```
alloy-cicd, line 930: **[15] Fourth independent parser of the `auth_requirement` wire shape** (after slack/gchat `inTaskAuth.ts` and agent-common `event_translation.py`) — and this one skips the chat clients' `OPAQUE_TOOLS` filtering, so the same ask can render as "eval" in the console where the chat card deliberately shows generic text. A flat, producer-chosen `{authorize_url, service, resource}` (or one shared parse) would keep the four surfaces in step when the shape grows (scopes, device code, …).

packages/agent-runner/agent/core.py
```diff
@@ -722,11 +845,34 @@ def _meta_int(key: str) -> int | None:
...
         sub_agent_name: str | None = None
         prompt: str | None = None
+        auth_payload: dict[str, Any] | None = None
+
+        # An answer to a run this service parked earlier. The scheduler addressed THIS
+        # task by id — which the handler accepted only because the parked run left it
+        # non-terminal — so the sub-agent task to continue is the one that run derived,
+        # not a new one.
+        authorization = _authorization_answer(messages)
+        resume_task_id = (
```
alloy-cicd: **[16] Minor: `resume_task_id` is always `scheduled_run_task_id(task.context_id)`-or-None, and `_run_langgraph_agent` recomputes the identical value as `derived_task_id`;** `task_id = resume_task_id or None` (~1363) is a no-op on an already-optional str. Passing `is_resume: bool` and deriving the id once at the destination would leave a single site owning the "which id goes in which field" invariant instead of two files' worth of comments.

packages/console-backend/console_backend/repositories/scheduled_job_repository.py
aartaria: Fixed in e9ec7a4e: confirmed and it was the worst of the set — `auth_required` is kept for good by design, so testing the status alone meant the first park stopped the job permanently. The predicate now keys the park arm on `parked_task_id IS NOT NULL`, the same key `answerable_parked_run` uses, and is split into two `NOT EXISTS` clauses rather than one `status IN (…)` so each arm is an index probe again — thanks for the secondary catch. Migration 098 is unreleased in this PR, so its index gained the same `AND parked_task_id IS NOT NULL` in place rather than stacking a 100: `auth_required` accumulates forever, so indexing on it alone would grow without bound. Covered by two real-database tests in `test_scheduler_recovery.py` (an unanswered park holds the schedule, an answered one releases it and keeps its status).

packages/agent-runner/agent/core.py
aartaria: Fixed in e9ec7a4e: `failed` now includes the no-ask case, so the ternary can no longer fall through to `"success"`, and the payload carries an explicit `error_message` ("Stopped for authorization but produced no ask to answer") rather than forwarding whatever prose the model left behind. You were right that the test pinned the wrong behaviour — `test_a_park_with_no_ask_is_not_a_park` now asserts `TASK_STATE_FAILED` and that message, with the reasoning in its docstring: a park nobody can answer must not hold the schedule, and must not reset `consecutive_failures` either.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Fixed in e9ec7a4e. The fallback now maps task state `auth_required` explicitly instead of letting the `else` branch default it to success. It maps to `auth_required` rather than straight to failed, so the case resolves through the `parked_task_id` guard just below — which, with nothing parsed, finds no task id and records the failure with its own message. That keeps one policy in one place instead of two that could drift. Two tests added: the unparseable park is not recorded green, and a well-formed park that names its task still parks.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Fixed in e9ec7a4e, differently from both suggestions — the conditional write has to stay in-request (it is the race gate that makes a second click a 409), so instead the RESUMED run row is now created in the SAME transaction that consumes the ask. A death in that window then leaves an ordinary stale `running` row, which is a state the system already knows how to recover. That exposed a second half: `interrupt_stale_runs` granted a retry only to `trigger = scheduled`, so a `resumed` run would have been swept and earned nothing — contradicting ADR-0009 decision 7, which says a resumed run behaves like a scheduled one. Both are covered by a real-database test. Side effect: the 202 now genuinely carries the new run id, which resolves [9].

packages/agent-runner/agent/task_store.py
aartaria: Fixed in e9ec7a4e, both halves. (a) The gate is now `POSTGRES_HOST` alone, matching the checkpointer — you named the real cost precisely: the split fails only at the moment the store exists for, and announces itself as one startup WARNING nobody reads. (b) `POSTGRES_SCHEMA` now pins the connection `search_path`, via a small `schema_connect_args()` helper so the decision is unit-testable without standing up a second `DatabaseTaskStore` (two in one process collide on SQLAlchemy metadata). Three tests added. On the duplication with `orchestrator-agent/app/core/task_store.py`: agreed and NOT done here — lifting `create_task_store(table_name=…)` into agent-common touches a second service and belongs in its own PR rather than in a review round on this one. Raising it with the author.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Not applying in this PR — bringing it to the author instead. The analysis holds: `job` is a click-time snapshot and `_finalize` writes `next_run_at` back through `complete_job`, so a concurrent Run-now or schedule edit in that window is rewound. Two things make it a deliberate call rather than a quick fix: re-reading the row inside `_finalize` changes behaviour for every caller of a shared method, not just the resume, and the alternative ("leave unchanged" marker) adds a parameter to the same shared signature. Both are larger than the window they close, which needs a concurrent edit during a resume. Noted as open. Thanks for ruling out the NULL→disable trap explicitly — that saved a second look.

packages/agent-runner/agent/core.py
aartaria: Not applying — this is a product decision, and it is going to the author rather than being made in a review round. You framed it correctly: fail-don't-degrade for discovery is ADR-0009's stated policy, so the behaviour is intended; what is new and undesigned is that gateway availability now feeds `consecutive_failures` and therefore auto-pause, for jobs that may need no gateway tool at all. Both candidate fixes (not counting discovery failures toward `max_failures`; a short-TTL per-(server, user) catalogue cache) are their own decisions with their own blast radius. Recorded as an open question on the PR rather than silently resolved.

packages/console-backend/console_backend/utils/a2a_dispatch.py
aartaria: Not applying in this PR — raising with the author. Confirmed the coupling: `dispatch_streaming` is shared, the harvesting is unconditional, and a failed assessor run now parses as a real-looking "caution" verdict instead of raising, losing the `assessment_unavailable` indicator and its warning. That is a genuine behaviour change for a non-scheduler caller and it should be fixed — but which way is a judgement call (gate the harvesting on the scheduler caller vs. have the assessor check the final state), and it changes the security assessor's failure semantics, which is not something to decide unilaterally at the end of a review round. The parking observation is noted too: an assessor that hits `auth_required` would leave an unanswerable open task. Recorded as open.

packages/console-backend/console_backend/routers/scheduler_router.py
aartaria: Fixed in e9ec7a4e — by making the description true rather than by weakening it. [4]'s fix creates the RESUMED run row in-request, so the id in the 202 is now the new run's and is real and pollable at response time. The description also now says explicitly that the parked run keeps its own record and stays `auth_required` for good, which is the trap you were pointing at for whoever polls it next.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
aartaria: Fixed in e9ec7a4e: `parked_task_id` is now included in the socket notification payload, with a comment saying why it cannot be derived from the status. The two frontend interfaces (`SchedulerNotification`, `RunNowResult`) declare the field, so it actually reaches `RunStatusBadge` instead of being silently absent — which, as you noted, typechecked only because the Pick'd field is optional.

packages/client-slack/src/listeners/actions/inTaskAuthButton.ts
aartaria: Fixed in e9ec7a4e: the second call is deleted, keeping the unconditional one at the top. Thanks for verifying it could not abort the resume before suggesting the deletion — that is what made this safe to apply without a second look at the `streamMessageTs` branch.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Fixed in e9ec7a4e: the stale block is deleted. It described a task-id scheme that was abandoned — `scheduled_run_task_id` keys on the context alone, as its own docstring says — and it contradicted the correct comment 14 lines above it. On the related `_` discard: agreed that `_build_message_args` does real work (an extra SELECT for voice jobs, potentially an LLM call for notification-shaped ones) only to throw the message parts away. Not applying here — splitting out a `_dispatch_metadata(job, run_id, db)` is a refactor of a method with several callers and belongs in its own change. Raising it with the author.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Fixed in e9ec7a4e, following your reasoning to its conclusion: the class is gone. The second point decided it — even if the guard could fire, it ran inside a Starlette background task where the exception escapes into nothing and the run row would be left `running` forever. It is now a logged refusal that returns, and finalizes the run row as failed when it has one (which it now does, since [4]'s fix creates it in the router). The test at `test_scheduler_engine.py:1567` asserts the refusal instead of the exception, and that nothing is dispatched.

packages/client-google-chat/src/services/scheduledRunResumeService.ts
aartaria: Not applying — this is the "is it intentional?" question and it goes to the author. To answer the factual half: no, the missing `reply_to` is not a deliberate difference. ADR-0009 calls the chain load-bearing ("whatever the resumed run produces is delivered into the ask's thread — including a second ask"), so Google Chat losing the threading is a divergence from the design, not a channel limitation that was accepted. It is not a one-line fix though: `a2aNotificationHandler` on the Google Chat side has no `reply_to_message` handling at all, so it is client work plus the tests that copy currently lacks. Flagging the duplication as the real cause — two clients now assert the console-backend contract (endpoint path, target validation, 409 semantics) with tests on one of them, and `a2aPushPayload.ts` is already duplicated the same way. Both recorded as open.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
aartaria: Not applying — bringing it to the author. The count is right (this makes four parsers of the same wire shape) and the `OPAQUE_TOOLS` divergence is a real behavioural inconsistency rather than a tidiness point: the console can render "eval" where the chat card deliberately shows generic text, which is a deliberate disclosure decision being silently undone by a fourth reader. The fix worth making is the one you name — a flat producer-chosen shape, or one shared parse — and that is a cross-package change (console-frontend, two clients, agent-common) that would grow this PR well past what it was opened for. Recorded as open; it pairs naturally with #180, which already tracks the three hand-synced protocol readers.

packages/agent-runner/agent/core.py
aartaria: Not applying. Agreed the plumbing is redundant and that a single site owning the "which id goes in which field" invariant would be better. Holding off because the two derivation sites use different variables (`task.context_id` at the caller, `context_id` at `_run_langgraph_agent`), and I am not going to assert they are always identical on the strength of reading it — verifying that is the actual work here, and this is a readability change on the resume path at the end of a review round. It is a fair follow-up; noting it as open rather than pretending it is done.

packages/agent-runner/agent/core.py
aartaria: Fixed in a26cfe83 — and one correction to the finding first, because my earlier reply repeated it uncritically.

**"including for jobs whose work needs no gateway tool at all" is not accurate.** The listing is gated on `not config.mcp_tools and _is_full_catalogue_agent(sub_agent_cfg)`, and that predicate is `all_tools or name == "general-purpose"` (core.py). An ordinary sub-agent with an empty whitelist does not list — it gets no MCP tools, exactly as before. And the whole branch sits inside `if sub_agent_id:`, so a notification-only or watch-only job never constructs the resolver at all. The exposure was real but confined to general-purpose / explicitly `all_tools` scheduled sub-agents, for which the catalogue IS the capability and cannot be skipped.

What was fixed, in that narrower scope:

1. **Retry with backoff**, using your `is_retryable_mcp_error` — three attempts, 1s doubling, transient only; a 4xx or refused connection still fails fast. Worth recording that the retried unit is a whole server listing rather than the stateless POST: `fetch_catalogue` already answers a failed stateless attempt by falling back to the SDK session, so a raw 503 never reaches the retry loop — what does is the error raised once both paths are exhausted. My first attempt tested the wrong layer and hung against the SDK fallback, which is how this surfaced.
2. **INTERRUPTED rather than FAILED** for a discovery failure — your first suggestion, via the vocabulary that already exists rather than a new exemption flag. ADR-0007 makes INTERRUPTED "evidence about the runtime, not about the job": it leaves `consecutive_failures` alone in both directions and earns one fresh attempt. That is exactly right here, and better than merely not counting the failure, since by the next tick the gateway is usually back.
3. **The two listings now run concurrently**, which was not in the finding but is adjacent: independent endpoints, so the run was paying the slower plus the faster for nothing.

**Not doing the catalogue cache**, deliberately. It buys much less here than the framing suggests: unlike the orchestrator — which discovers once per user turn and hands sub-agents a pre-resolved subset (`_pre_resolved_tools_for`) — a scheduled run is the top of its own tree with no warm parent to inherit from, so the only reuse available is across runs, and schedules are too sparse for a short TTL to reach. It would also drag the entitlement-stamp fetch into agent-runner to do it. The per-run listing cost is now addressed by concurrency instead.

Eliminating the listing entirely would mean moving catalogue search server-side (today `ToolCatalogMiddleware` ranks client-side over the materialised catalogue, so deferring the listing would defeat `search_tools`). That is a real change to catalog mode affecting the orchestrator equally, and needs its own decision rather than riding along here.

aartaria: Follow-up on the concurrency, raised by @aartaria and correct: gathering the two listings without a bound was a regression, and 47b1d5d1 fixes it.

The dimension I had checked was the wrong one. Per run the fan-out is two servers, which is harmless — but the scheduler claims up to `claim_limit` jobs a tick and dispatches them together, so runs overlap, and agent-runner is the service that gets OOMKilled. Making one run list concurrently took it from 1 to 2 in-flight fetches, doubling the process peak on exactly the service that cannot afford it, to buy latency.

It now takes the same process-wide cap the orchestrator has, under the same `MCP_DISCOVERY_CONCURRENCY` env var and default (5), so the two are tuned together instead of drifting. Two details copied deliberately rather than re-derived: the cap is process-wide because a per-call limit still allows `limit x N` (the orchestrator learned that at ~31 concurrent fetches and an OOMKilled pod, 2026-08-15), and the slot covers the fetch only, released across the backoff sleep, so one flaky server cannot idle away a slot another run could use. The test drives five concurrent runs, ten fetches wanting to be in flight, and asserts the peak holds at the cap — it fails without the semaphore.

On sharing the code with the orchestrator: agreed in principle, and deliberately not done here. The retry policy, the cap and the predicate are now a second copy of one policy, which is the same drift this review already flagged for the task store in [5] — and agent-runner and the orchestrator do differ at the edges (many servers with `return_exceptions=True` and partial tolerance there, two servers and fail-fast here, per ADR-0009), so a shared helper has to parameterise that rather than assume it. Lifting it into agent-common touches a second service, which is the reason I gave for not doing the task-store lift in this PR either; doing one and not the other would be inconsistent. Worth one follow-up that consolidates both.

packages/agent-runner/agent/task_store.py
aartaria: Follow-up: the duplication you flagged here is now removed too, in 9f077c87 — @aartaria asked for both consolidations in this PR since they turned out to be clean.

`create_task_store` moved to `ringier_a2a_sdk.agent.task_store`, beside `postgres_checkpointer_mixin`, which is the same concern (a Postgres thing built from the service POSTGRES_* connection, gated the same way, same memory fallback). No new dependency anywhere — the SDK already requires sqlalchemy and a2a-sdk. Both services keep thin wrappers holding only what differs, and only two things do:

- **`table_name`** — agent-runner names its own, the orchestrator keeps the SDK default `tasks` it already owns. Your point about either service reaching the other's tasks is what this preserves.
- **`schema`** — pinned for agent-runner, deliberately not for the orchestrator. That asymmetry is not an oversight and the shared docstring says so: pinning an existing table does not move it, it creates an empty one in the configured schema and stops the service seeing the tasks it already had. A new store pins for free; an existing one keeps its resolution until someone migrates it on purpose.

**One behaviour change fell out of it, in the orchestrator.** Its gate was host AND password — the same latent split you identified here — so sharing the code made the two copies disagree out loud. It is now `POSTGRES_HOST` alone, which is what `postgres_checkpointer_mixin` already documents ("POSTGRES_HOST … Gates persistence"). Its test asserted the old gate and now asserts the new one, with the reasoning in the docstring. Worth a reviewer's eye since it is a second service changing behaviour in a PR about scheduled runs.

Gates after the move: orchestrator-agent 947, agent-runner 75, agent-common 1181, ringier-a2a-sdk 408.

packages/agent-runner/agent/core.py
aartaria: Sharing the discovery policy is done too, in 9f077c87 — @aartaria called it: if the refactor is clean, do it in this PR rather than deferring it.

The retry loop and the process-wide cap moved to `agent_common.core.catalogue_ingest.fetch_with_retry`. It takes a zero-argument coroutine factory, so neither caller had to be reshaped around a shared signature: the orchestrator passes its SDK-plus-stateless `_ingest_server`, agent-runner its `fetch_catalogue` call. `_get_catalogue_with_retry` and both `_DISCOVERY_SEMAPHORE` globals are gone.

Both properties that were learned the hard way moved with it, and are stated where they live now: the cap is process-wide because a per-call limit still allows `limit x N` (the ~31 concurrent fetches that OOMKilled the orchestrator on 2026-08-15), and the slot covers the fetch but not the backoff sleep, so a flaky server cannot idle away a slot a healthy fetch could be using.

What is deliberately **not** shared is what a failure means, which is the real difference between the two callers and the thing a shared helper could easily have flattened: the orchestrator gathers with `return_exceptions=True` and serves a partial catalogue, because a user turn is better served by most of their tools than by an error; a scheduled run fails the whole listing (fail-don't-degrade, ADR-0009). That stays at each call site, and both now say why.

Gates: orchestrator-agent 947, agent-runner 75, agent-common 1181, ringier-a2a-sdk 408, ruff clean on every touched file.

packages/console-backend/console_backend/utils/a2a_dispatch.py
aartaria: Fixed in 941ac690, on the assessor's side rather than by gating the harvesting — @aartaria's call, and I agree with it: teaching a shared transport function who is calling it is the worse of the two options you offered. The harvesting is a real improvement for every caller (before it, a failed run reached the scheduler with no error text at all), and the state the assessor needs is already in the result it is handed.

**The assessor now checks how the run ended before reading what it returned.** Your diagnosis was exact: "there is text" stopped implying "there is an assessment". Tested as `!= "completed"` rather than `== "failed"` so every other terminal state counts too, `auth_required` included — and `final_state` defaults to `"completed"` in `dispatch_streaming`, so a healthy run is never caught by it. That restores the existing path (raise → the `assessment_unavailable` fallback + its warning) with no new verdict plumbing. One bonus your finding made available: the raised error carries the harvested text, so the warning now says *why* the assessor was unavailable — detail that did not survive at all before this PR.

**Your parenthetical is fixed too, and it needed more than the state check.** An assessor that parked would leave a task nothing could ever close *or* answer: parking holds a JOB's schedule, the ask rides the JOB's delivery channel, and the answer comes back addressed to a RUN — none of which exist for a caller with no `scheduled_job_id`. Worse after this PR, since that task now sits in a durable store rather than being lost at the next restart, so it is a permanent orphan per occurrence. agent-runner therefore refuses to publish a park without a job and reports failed instead — the same shape as the existing `risk_scorer=None` constraint, enforced at the boundary rather than assumed from the caller's metadata.

That also surfaced a latent inconsistency worth naming: the park payload was gated on `parked and auth_payload`, which excludes only the no-ask case. A park with a good ask but no job would have announced `scheduler_status: "auth_required"` while being yielded as `TASK_STATE_FAILED`. There is now one `publishes_park` flag so the payload cannot disagree with its own task state.

Gates: console-backend 1970, agent-runner 76. Both guards were checked against their own absence — each test fails when its guard is disabled.

packages/client-google-chat/src/services/scheduledRunResumeService.ts
aartaria: Answering your question first: **no, the missing threading was not intentional** — confirmed by @aartaria. Fixed in e9d0a5b8, with the two copies deliberately kept as copies for now.

Your diff was exact: three lines had diverged, and each was a gap.

- The resume service now takes the ask's coordinates and sends them as `reply_to`. Google Chat's are a space and a thread where Slack's are a channel and a ts — and neither side of the trip interprets them, since console-backend stores `reply_to` as an opaque `dict[str, str]` and hands it back to whichever client delivers the result. **No backend change was needed**, which is worth noting: the contract was already channel-agnostic, only this client was not using it.
- The button handler passes this card's own space and thread, which is the only moment those coordinates exist — the ask went out before anyone had clicked anything.
- The notification handler threads the result under them, and only when the ask was posted in the space the notification is going to, since it is this client's own message it threads under.

And the stale comment you spotted is gone; it did still say "a floating promise in a Slack handler".

**On the duplication:** staying as two copies for now, by decision rather than by omission — there is no shared TS client package today, `a2aPushPayload.ts` is duplicated the same way, and inventing one is not this PR's job. What I have done is put them back in step: the two files now differ only in the channel-specific naming and in one paragraph explaining what each channel's coordinates are. Worth saying plainly that this is exactly the failure mode you identified — a copy drifting silently — so the pairing is a standing liability until a shared package exists, not something this fix retires.

**The test asymmetry is closed too**, which was the other half of your finding: the Google Chat copy now covers the result landing in the ask thread, a thread from another space being ignored, and an ordinary run threading under nothing. Each was checked against its own absence — I mutated the guard and confirmed the test fails.

Gates: client-google-chat 33, client-slack 85, `tsc --noEmit` clean on both.

packages/agent-runner/agent/core.py
aartaria: Retracting my "not applying" — @aartaria pushed back on it and was right. I declined on the grounds that the two sites derive from different variables and I would not assert they were always equal on the strength of reading it. Verifying that took four greps, and they are equal:

- `_execute_sub_agent` has exactly **one** caller, which passes `context_id=task.context_id`;
- it forwards `context_id=context_id` unmodified to its **single** `_run_langgraph_agent` call;
- nothing in the file ever **assigns** to `context_id` — it appears only as a parameter and as a keyword argument;
- `scheduled_run_task_id` is pure (`uuid5(namespace, context_id)`).

So `resume_task_id` was always exactly the `derived_task_id` recomputed downstream. Fixed in 3e19176d: only `is_resume: bool` travels down, and the id is derived where it is used.

I went one step further than the finding suggested, because of a problem your change surfaces: moving the decision downstream weakens the tests, which asserted the actual id at the `_execute_sub_agent` boundary and could now only assert a bool. So the rule itself moved into `sub_agent_task_ids(context_id, *, is_resume) -> (task_id, proposed_task_id)` — the single owner you asked for, and now directly testable, which the inline version was not. Four tests pin it: a resume addresses the parked task, fresh work proposes instead, both mean the same id, and no context means no id either way.

The resume test also pins the property the collapse rests on — that the `context_id` reaching `_execute_sub_agent` is this task's — so if that ever stops holding it fails loudly instead of silently addressing a task nobody opened.

Two files' worth of comments about which result belonged in which field are now one docstring. Gates: agent-runner 80, ruff clean.

packages/console-backend/console_backend/services/scheduler_engine.py
aartaria: Fixed in 62ceb7e8. @aartaria framed the requirement well — a schedule edit is a manual intervention and must be honoured — and that is exactly what the old code broke, in the direction that is easy to read backwards: it is the RESUMED RUN that overwrites the owner, not the other way round.

`update_job` recomputes `next_run_at` from the new definition, so the sequence was:

| Time | Event | `next_run_at` |
|---|---|---|
| 09:55 | owner clicks authorize; the router snapshots `job` | 11:00 |
| 10:00 | owner edits the schedule; recomputed from the new cron | **18:00** |
| 10:02 | resumed run finalises, echoing its 09:55 snapshot | **11:00** — the edit is gone |

The race is a symptom; the cause is that `next_run_at=NULL` carries two meanings at once. The COALESCE keeps the column and the CASE beside it retires the job, so a caller with nothing to say about the schedule had no way to say so — and the resumed run simulated "leave it alone" by restating a value it had read earlier. That read-then-write is the lost update.

So rather than either of your two suggestions, `complete_job` now takes `leave_schedule`, which suppresses only the retire branch. A resumed run passes it and sends no time at all, so the column resolves against the row own current value inside the same UPDATE. Re-reading the job inside `_finalize` would have narrowed the window; this removes the write, so there is no window. Whatever the edit computed simply stands.

Both callers that rely on NULL meaning "retire" are untouched — a once-job whose occurrence has passed, and the unresolvable-timezone pause — and a test pins that NULL still retires, since that was the trap in separating the two meanings.

Tested against a real database: an edit landing between the click and the finalise survives, and the flag is not read as "retire". The first test was checked against the old behaviour — restoring the echo makes it fail.

Thanks for ruling out the NULL→disable trap explicitly in the original finding; it is exactly the constraint the fix had to respect, and knowing it was unreachable *there* is what made it safe to disentangle *here*.

Gates: console-backend 1972, ruff clean.

packages/console-frontend/src/pages/SchedulerJobDetailPage.tsx
aartaria: Filed as #250 rather than fixed here — it spans four packages and the core of it is a product decision about disclosure, not a correctness fix that belonged in this PR.

Verifying it to write the issue turned up that the count is **five**, not four: `embed-sdk`'s `auth-required-card.tsx` is a fifth reader of the shape, and it carries its own verbatim copy of `OPAQUE_TOOLS`. So the set is duplicated three times (slack, google-chat, embed-sdk) and missing on the fourth rendering surface.

Your point about the console is also sharper than stated. `readParkedAsk` renders `requirement.service || requirement.resource`, and `resource` IS the tool name — so it is not merely unfiltered, it falls back to exactly the field the other three suppress. A parked ask that carries no service name renders the raw tool in the console, from the same payload a chat card would have shown generically.

(`agent-common`'s `event_translation.py` is the fifth reader but deliberately not a filtering site — it reconstructs the payload rather than rendering it, and its comment says carrying `resource` through is the point, since dropping it left consumers unable to say what was being authorized. So the suppression belongs at the producer or at each renderer, not there.)

The issue records both of your directions — a flat producer-chosen `{authorize_url, service, resource}`, or one shared parse per language — and names the blocker for the TypeScript half: there is no shared TS client package today, which is the same thing that keeps the resume service and `a2aPushPayload.ts` duplicated ([14]). It is linked to #180, which counts three hand-synced copies; this is the same problem with two more surfaces and one concrete divergence.

On the board with `Stream: Nannos`, Priority and Epic left for a human.

aartaria:
## 🤖 Heads-up for review round 2

Round 1's 16 findings are all closed — 15 fixed here, [15] filed as #250 — and every thread carries its disposition. This note is the part that does not live in any single thread: what was decided deliberately, what was compromised, and where a second pair of eyes is worth most.

**Read the diff against `fix/retired-model-alias-resolution`, not `main`.** This is a stacked PR (base is #244). Its four commits are not part of this change.

### Deliberate decisions — please don't re-raise these without new evidence

- **The orchestrator's task-store gate changed**, and it is the one behaviour change in a service this PR is not about. It was `POSTGRES_HOST` AND `POSTGRES_PASSWORD`, now host alone, matching the convention `postgres_checkpointer_mixin` documents. Same latent split finding [5] identified in agent-runner; sharing the code is what made the two copies disagree out loud. **This is the single highest-value thing to check**: it changes when a *different* service persists tasks.
- **The shared task store pins `schema` for agent-runner and deliberately not for the orchestrator.** Not an oversight. agent-runner's table is new, so pinning is free; the orchestrator's exists, and pinning would not move it — it would create an empty table in the configured schema and hide the tasks it already has. That is a migration, not a refactor.
- **The discovery retry policy is shared; the failure semantics are not.** The orchestrator gathers with `return_exceptions=True` and serves a partial catalogue; a scheduled run fails the listing (fail-don't-degrade, ADR-0009). A shared helper could easily have flattened that, so it is parameterised at the call sites on purpose.
- **A discovery failure is now `INTERRUPTED`, not `FAILED`.** It therefore earns a retry and does not count toward `max_failures`. Expect the run-status distribution to shift during a gateway outage — that is the intent, not a regression.
- **agent-runner refuses to park without a `scheduled_job_id`.** New refusal for non-scheduler callers (assessor, debug agent), because such a park is unanswerable and would orphan a task in the now-durable store.
- **Migration 098 was amended in place** rather than adding a 100, since it is unreleased in this PR. Its index gained `AND parked_task_id IS NOT NULL`.
- **`AlreadyAnswered` was deleted** rather than fixed: it could never fire, and would have escaped unhandled into a Starlette background task.

### Compromises — known, and deliberate

- **The two chat clients stay duplicated.** `scheduledRunResumeService.ts` and `a2aPushPayload.ts` are copies in client-slack and client-google-chat. [14] was exactly a silent drift between them; they are now back in step and both tested, but there is no shared TS client package and inventing one was out of scope. **The drift risk is not retired** — it is the same blocker as the TypeScript half of #250.
- **`agent-runner/agent/mcp_tools.py` is kept one release** though its listing moved to the shared path, because its tests still document the console/gateway audience split. ADR-0009 names this; port them, then delete it.
- **`delivered` on a run is still inferred, not observed.** An ask that was never delivered is recorded as delivered. Predates this change, matters more for an ask than a result.
- **A parked run has no expiry.** A job whose owner never answers stays stopped, by design in this cut.
- **`MCP_DISCOVERY_CONCURRENCY` is now read by two services** with very different fan-out shapes (the orchestrator lists many servers, agent-runner two), same name and default of 5 so they are tuned together. Deliberate, but worth a sanity check — and note agent-runner's deployment may not set it, so it takes the default.

### Considered and deliberately NOT done

- **A per-user catalogue cache** for agent-runner. Unlike the orchestrator, a scheduled run has no warm parent turn to inherit from (no analogue of `_pre_resolved_tools_for`), so the only reuse available is across runs — and schedules are too sparse for a short TTL to reach.
- **Server-side catalogue search**, which is the only thing that would remove the listing from the critical path entirely. It changes catalog mode for the orchestrator too and needs its own decision.

### Where a second pair of eyes is worth most

1. The **orchestrator task-store gate change** (above).
2. **`complete_job(leave_schedule=…)`** — it disentangles `next_run_at=NULL`, which meant both "keep the column" and "retire the job". Worth confirming the two callers that rely on the retire meaning (a `once` job, the unresolvable-timezone pause) really are the only ones.
3. **The resumed run row is now created in-request**, in the same transaction as the claim, so a crash between them leaves a sweepable `running` row rather than nothing. Worth checking that cannot produce a duplicate run on any path.
4. **`publishes_park`** — one flag now decides whether a park payload is emitted. Worth confirming no path emits park fields without it.

Every claim above was checked against the code rather than inferred, and each new guard was verified against its own absence (disable the guard, the test fails). Where I declined a round-1 finding on grounds I had not verified ([16]), @aartaria pushed back and the verification changed the answer — so do challenge reasoning that rests on an unstated assumption.

packages/agent-runner/agent/mcp_tools.py
```diff
@@ -138,15 +215,37 @@ def _server_for(name: str) -> str:
...
+            # and their own token, so the sequential loop this replaced paid the slower of
+            # the two plus the faster one on the run's critical path for nothing. Listing
+            # was already measured as the largest contributor to time-to-first-token in the
+            # orchestrator, which gathers for the same reason.
+            #
+            # Not return_exceptions: a partial catalogue is not a smaller catalogue, it is a
+            # silently less capable run. Fail-don't-degrade for discovery is ADR-0009's
+            # stated policy — the first failure propagates and the run reports it.
+            listed = await asyncio.gather(
+                *(self._list_server_with_retry(server, http_client) for server in connections)
```
alloy-cicd: **[17] Minor (non-blocking): a discovery failure orphans the sibling listing task.**

`asyncio.gather` without `return_exceptions` propagates the first exception but does not cancel the other listing task; the `async with httpx.AsyncClient` then closes the client under it. The orphaned task keeps retrying against a closed client for up to ~3s of backoff — its errors land in a never-awaited future, producing "Task exception was never retrieved" log noise on every discovery failure. The run outcome itself is correct (the real error still fails the run), so this is hygiene, not correctness. Fail-don't-degrade is preserved either way: `gather(..., return_exceptions=True)` + re-raise the first exception, or an `asyncio.TaskGroup` (which cancels siblings for free), keeps the policy without the orphan.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (53b3a5e86a41ec5f4ecf915d2eca2b715bb3214d..62ceb7e808c36e9fd0b55c02baa8d809124fbd6f)

**Ledger of round-1 findings — all 16 dispositions verified against the new head:**

- **Verified-fixed (15):** [1] claim guard split into two `NOT EXISTS` keyed on `parked_task_id IS NOT NULL`, with migration 098's index matched to it — pinned by real-database tests; [2] the no-ask park now fails with an explicit message and the test asserts the right behavior; [3] task state `auth_required` mapped explicitly and the unparseable park resolves through the `parked_task_id` guard to FAILED; [4] the RESUMED run row is created in the same transaction that consumes the ask, and `interrupt_stale_runs` now grants `resumed` runs the retry — a death in the window leaves an ordinary sweepable `running` row; [5] host-only gate + `POSTGRES_SCHEMA` search_path, then consolidated into `ringier_a2a_sdk.agent.task_store` with the two load-bearing knobs kept per service; [6] `complete_job(leave_schedule=…)` removes the stale write instead of narrowing its window — a concurrent schedule edit survives, NULL-retires semantics pinned for the untouched callers, and the once-job park→resume lifecycle is correct; [7] transient-only retry with backoff, discovery failure reported as INTERRUPTED (no `consecutive_failures` movement), concurrent listing bounded by the process-wide cap and the policy lifted into `agent_common.core.catalogue_ingest.fetch_with_retry` for both services; [8] fixed on the assessor's side — final-state check before parsing, and agent-runner refuses to publish a park with no `scheduled_job_id`, closing the structurally-unanswerable-task case; [9] the 202 now returns the real resumed run's id; [10] `parked_task_id` travels in the socket payload; [11] duplicate `recordDecision` deleted; [12] stale comment deleted; [13] `AlreadyAnswered` replaced by a logged refusal that finalizes the run; [14] Google Chat threads the resumed result under the ask, with the tests the Slack copy had; [16] `sub_agent_task_ids` is the single owner of the id derivation, and the equality claim it rests on was verified rather than assumed.
- **Deferred with a filed follow-up (1):** [15] → issue #250 (five parsers of the `auth_requirement` shape, `OPAQUE_TOOLS` disclosure divergence). Reasonable scoping; not re-raised.
- **Rejected-by-author: none.**

**Delta review:** one new minor finding, inline as [17] — the concurrent discovery `gather` orphans the sibling listing task on failure (log noise only; the run outcome stays correct). Everything else probed came back clean, including the areas most at risk from the fixes: no double run creation on the resume path, no double finalize, the pre/post-commit claim-guard race is closed from both sides, the `publishes_park` matrix is consistent across all combinations, the semaphore test genuinely exercises the cap, and the two resume-service copies are back in byte-step apart from channel naming.

**Follow-up notes (not blocking, carried from round 1):** `packages/console-backend/AGENTS.md` §"Scheduled Run Vocabulary and Interruption" still describes three triggers and migrations 091/092 — it now trails the shipped contract by `RESUMED`, the schedule-holding `AUTH_REQUIRED`, and migrations 097–099. And two deferred-by-thread items remain open by agreement: the `_dispatch_metadata` split ([12] thread) and the ADR-0009 unattended-posture sign-off, which stays with the human reviewer.

packages/agent-runner/agent/mcp_tools.py
aartaria: Fixed in 181683f4 — and correctly attributed: this was a regression introduced by this PR own concurrency change, not pre-existing.

`return_exceptions=True`, with the first exception re-raised immediately after, so every task is awaited before anything propagates and none outlives the HTTP client closing under it. Fail-don not-degrade is unchanged: a partial catalogue is still never served. The comment above that line previously argued *against* `return_exceptions`, conflating "await everything" with "tolerate a partial result" — rewritten, since it would otherwise read as a reversal of ADR-0009 policy.

I did not take the `TaskGroup` half of your suggested shape. It cancels the sibling rather than waiting, which is better on the failure path, but it wraps what propagates in an `ExceptionGroup` — changing the error a caller reports and the text that reaches `CatalogueDiscoveryError`. With two tasks, each bounded by the same timeout and awaited together on the success path anyway, waiting is the cheaper trade. Recorded in the code rather than left as a silent choice.

**The regression test is worth your attention next round, because my first version of it was vacuous** — it passed against the unfixed code. The other tests in that file stub `asyncio.sleep` via `catalogue_ingest.asyncio`, which resolves to the *global* module object, so the stub also swallowed the test own sleep: the sibling completed instantly whether or not it was awaited, and the assertion held either way. The committed tests therefore do not stub sleep at all (nothing retries — a 403 is not retryable), and the sibling is genuinely in flight when the failure propagates. Verified against its own absence: restoring the bare gather fails it.

That global-stub pattern is still present in the retry tests in the same file. It is harmless there — they are only skipping backoff and asserting call counts, and monkeypatch restores it — but it is a sharper tool than those tests look like they are reaching for, and it is what made my first attempt blind.

Gates: agent-runner 82, ruff clean.

PR #248, opened 2026-09-16
fix!: key delivery channels on the tenant, not the bot display name

### LINKED ISSUE #247: Delivery channels are keyed on the bot display name, so same-named installations collide
The Slack and Google Chat clients register one delivery channel per tenant, keyed by a client-supplied `installation_id`. Both derive that key from the bot's **display name** rather than from anything that actually identifies the tenant:

- `client-slack` — `installationId: bot.botName` in `packages/client-slack/src/services/installationRegistrar.ts`
- `client-google-chat` — `installationId: project.botName` in `packages/client-google-chat/src/services/installationRegistrar.ts`

Display names are not unique. Two Slack workspaces may each install a bot called `Nannos`, and the `bot_installations` schema explicitly allows it — `app_id` is the primary key and the table comment notes that "multiple bots may share the same team_id".

## Impact

Console-backend resolves channels by `(client_id, installation_id)`, and every Slack installation registers under the same `client_id`. So two installations sharing a bot name collapse onto **one** channel row:

1. **A tenant's delivery channel is silently overwritten.** Registration runs at startup over installations ordered by `created_at ASC`, so the newest same-named installation wins the upsert. The console then shows one channel whose label names the newcomer, and the other tenant has no channel at all.

2. **Notifications are delivered to the wrong tenant.** The per-installation notification secret is keyed on the same string, so both installations share one secret. The inbound callback authenticates a token by scanning installations for a matching secret and returning the first hit — which is the *oldest* same-named installation, not the one the channel belongs to. A scheduled job therefore reports success while its message is posted into a different workspace.

Both failures are silent: nothing logs a collision, and the console UI looks correct because the surviving row carries a plausible name.

`bot_name` is documented as cosmetic (`// Display name shown in messages`) while being load-bearing in three places — the channel key, the secret key, and the inbound routing key. Renaming a bot through the admin UI, which reads as a purely cosmetic edit, silently re-keys a tenant's secret and orphans its channel.

## Proposed fix

Key on the tenant's real identifier, which both clients already have:

- Slack: `app_id` — already the primary key of `bot_installations`, unique per workspace install, immutable, and already the routing key used by `inflight_tasks.app_id` / `pending_requests.app_id`.
- Google Chat: the GCP project name.

Leave `bot_name` as the human-readable channel label only.

Existing channel rows are referenced by `scheduled_jobs.delivery_channel_id`, so the change cannot simply create new rows under the new keys — that would leave every existing job delivering through a stale row whose secret the client no longer recognises. Registration should carry the previous key so console-backend can re-key the existing row in place, preserving its id.

## Unrelated bug found while reproducing

`PUT /api/v2/installations/{appId}` rejects every request the admin UI sends: the shared form includes `appId`, the update schema omits it (it is the path parameter), and zod emits `additionalProperties: false`. Editing an installation has never worked. Two aggravating factors in the same path:

- `OpenApiValidationError` is not an `Error` subclass, so Koa reports "non-error thrown" and serves an opaque 500 with no indication of which field was rejected.
- The error carries the raw request body, which for this endpoint contains the Slack bot token and signing secret — anything that formats the error writes both to the logs verbatim.
- The request logger reads `ctx.status` in a `.finally()` that runs before Koa assigns a status, so failed requests are logged as `404` while the caller receives a 500.

aartaria:
closes #247

Authored by Claude Code (Opus 5), driven by @aartaria.

## What was wrong

Both chat clients register one delivery channel per tenant, keyed by a client-supplied `installation_id` — and both derived that key from the bot's **display name**:

```ts
installationId: bot.botName,      // client-slack
installationId: project.botName,  // client-google-chat
```

Display names are not unique. `bot_installations` has `app_id` as its primary key and its own DDL comment notes that "multiple bots may share the same team_id", so two workspaces may each install a bot called `Nannos`. Since every Slack installation registers under the same `client_id`, and channels resolve by `(client_id, installation_id)`, those two installations collapsed onto **one** row.

Two failures followed, both silent:

1. **A tenant's channel is overwritten.** Registration runs at startup over installations ordered by `created_at ASC`, so the newest same-named installation wins the upsert. The console shows one channel labelled with the newcomer's workspace; the other tenant has none.

2. **Notifications go to the wrong tenant.** The notification secret is keyed on the same string, so both installations share one. The inbound callback authenticates a token by scanning installations for a match and returning the first hit — the *oldest* same-named installation. A scheduled job reports success while its message is posted into a different workspace.

Underneath it: `bot_name` is documented as cosmetic (`// Display name shown in messages`) while being load-bearing as the channel key, the secret key, and the inbound routing key. Renaming a bot in the admin UI — which reads as a purely cosmetic edit — silently re-keyed a tenant's secret and orphaned its channel.

## What this does

**Keys on the tenant.** Slack uses `app_id`: already the table's primary key, immutable, unique per workspace install, and already what `inflight_tasks.app_id` and `pending_requests.app_id` route on. Google Chat uses the GCP project name. `bot_name` goes back to being only the channel's human-readable label, and the type comments now say which field is which.

**Carries the new key across the A2A boundary.** The outbound `metadata.installation` both clients stamp on every message is threaded by the orchestrator into `list_channels_for_installation`. It sent `botName`, so re-keying the channels without it would have returned an empty channel list for every workspace and broken scheduled-job creation outright. Both clients now send the same key they register under.

**Accepts only the new key on the way back in.** Both inbound validators match a notification token against the secret keyed by the tenant's id and nothing else. There is no fallback to the old display-name secrets: that was the last place where a display name still decided which tenant a notification belonged to, and it resolved the colliding tenants exactly as wrongly as the keying this PR replaces.

## What this deliberately does not do

**Registration does not move existing channels onto the new keys.** An earlier revision had the clients send their old key so console-backend could re-key the row in place, preserving `scheduled_jobs.delivery_channel_id`. That is gone, for two reasons:

- It only worked where tenants had *not* collided. A shared row carries jobs from both tenants and only one can claim it, so for precisely the case this PR exists to fix, it preserved the bindings wrongly and kept the cross-tenant misdelivery alive.
- It put a tenant-moving operation in the request path, driven by an admin-editable display string. Renaming a bot to another workspace's App ID would have asked console-backend to hand over that workspace's channel and every job bound to it.

So a registration now creates or updates the channel for the key it presents and nothing else. `update_channel` likewise has no path to `installation_id`, with a test pinning that.

## Deploy-time step, not a follow-up

After the upgrade the old display-name rows remain as orphans, and any scheduled job provisioned before this change still points at one and still dispatches its old secret — which nothing now accepts. Such a job keeps running and keeps reporting success, because the push happens in `BasePushNotificationSender` after the run completes and that swallows the error, so **its notifications stop arriving and nothing says so**.

Those jobs therefore have to be re-pointed as part of shipping this, from the console's scheduler job editor. #249 carries the query, the re-pointing, the orphan cleanup and the stale-secret deletion — including the note that where two tenants collided, one channel's jobs belong to both and must be split, and that any job whose owner is not the tenant that channel really served has been misdelivering since before this fix.

Worth running #249's query before merging: if no job references a display-name channel, none of this applies.

## Also in here

`PUT /api/v2/installations/{appId}` rejected every request the admin UI sent, so **editing an installation has never worked**: the form is shared with the create flow and carries `appId`, the update schema omits it (it is the path parameter), and zod emits `additionalProperties: false`. It now sends only the declared fields. This is the endpoint you would reach for to work around the collision by renaming a bot, which is how it surfaced.

Three things made that failure hard to read, and one of them leaked credentials:

- **`OpenApiValidationError` carried the raw request body.** This endpoint's body holds the Slack bot token and signing secret, and Koa's `onerror` formats a thrown non-Error with `%j` — so a rejected edit wrote both to the logs verbatim. It now carries only rejected field paths and their messages, and a test pins that.
- **It was not an `Error` subclass**, so Koa reported "non-error thrown" and served an opaque 500. It now extends `Error` with `status = 400`, and a middleware answers with the field paths so the UI can name the bad field.
- **The request logger read `ctx.status` in a `.finally()`** that ran before Koa assigned a status, logging failed requests as Koa's default 404 while the caller received a 500. The error middleware is registered inside the logger, so the logged status is the one actually served.

## Testing

- `packages/client-slack` — 88 passing, 10 suites. New suites for the registrar (keys on `app_id`, same-named installations get distinct channels *and* distinct secrets, inactive skipped, per-installation failure isolation) and for request validation (rejects the unknown property, names the field including on a nested object, carries no submitted values, is a 400-bearing Error).
- `packages/client-google-chat` — 30 passing.
- `tsc` clean on `client-slack`, `client-google-chat`, `client-slack-frontend`; `eslint` clean on the changed frontend file.

## Not in here

The generated console SDK is untouched — the console never constructs a delivery channel, and regenerating pulled in unrelated drift from endpoints merged without a regen, which belongs in its own change.

## Breaking changes

One commit carries `!`, with a `BREAKING CHANGE:` footer: **`fix(clients)!: key delivery channels on the tenant, not the bot display name`**.

A channel's `installation_id`, the per-installation notification secret, and the `installation` value on outbound A2A message metadata all change meaning together — from the bot display name to the tenant's own id. A client and a console-backend on opposite sides of this resolve no channels for any tenant, so they deploy together.

**`console-backend` is not touched by this PR.** The re-key needs nothing from it: registration already resolves channels by `(client_id, installation_id)` and never re-keys, so only the clients change. (A test asserting `update_channel` cannot reach `installation_id` was written while exploring a re-keying approach that was dropped; it guards behaviour that already holds on `main`, so it will land separately rather than pull this package into a breaking change.)

## Checklist
- [x] Does this PR introduce a breaking change? 🚨


alloy-cicd (commented):

The core change is sound — keying delivery channels on the real tenant id (Slack App ID / GCP project name) with a locked legacy-adoption path is the right shape, and the new tests are good. But the re-key is only half-propagated: the outbound A2A `installation` metadata still carries `botName`, so after deploy the orchestrator's channel lookup matches nothing (comment 1 — this looks like a release blocker). Three more substantive issues: the adoption path is a standing cross-tenant re-key gated only on `client_id` and an admin-editable display name (2), the re-key UPDATE lacks the SAVEPOINT guard its sibling insert path has (3), and pre-existing scheduled jobs of a losing colliding tenant keep delivering into the winner's workspace — the log line saying jobs are "unaffected" is the problem, not the reassurance (4). Plus a rollout-window auth gap (5) and four smaller notes.

Checked and clear: `OpenApiValidationError` has no other catchers/readers of the removed `.input`/`.errors`; the `installation_id` kwarg on `update_channel` cannot leak through the public PATCH route; `previous_installation_id` is not written as a column; adoption is correctly scoped per `client_id`; conventions in root and console-backend `AGENTS.md` are followed (audited writes, typed signatures, `pg_session` tests).

Findings numbered in the inline comments, ordered by severity.

packages/client-slack/src/services/installationRegistrar.ts
```diff
@@ -67,7 +72,12 @@ export async function registerInstallations(deps: InstallationRegistrarDeps): Pr
   for (const bot of active) {
     try {
       await registerOne(deps, {
-        installationId: bot.botName,
+        // app_id, not bot_name: the Slack App ID is this table's primary key, is unique per
+        // workspace install, and is already what every other runtime path routes on. bot_name
+        // is a display string — two workspaces may legitimately both call their bot "Nannos",
+        // and keying on it silently collapsed them onto one channel and one secret.
+        installationId: bot.appId,
```
alloy-cicd, line 74: **1. Regression: the outbound A2A `installation` metadata was not migrated with the key.**

Registration now keys the channel on `bot.appId`, but every chat turn still stamps `metadata.installation = request.botName` (`a2aClientService.ts:209`, and the type comment at `:37` still says it "must match the botName used to register delivery channels"). The orchestrator threads that value into `console_list_delivery_channels` → `DeliveryChannelRepository.list_channels_for_installation`, which filters `WHERE installation_id = :installation_id`.

After this deploys and the row is re-keyed to `A0123…`, a lookup scoped to `installation="Nannos"` returns `[]` — so "send me this report every morning" finds no delivery channel for **every** workspace, not just the colliding ones. Same on the Google Chat side: `messageHandler.ts:726` still resolves `botName` while the registrar sends `project.projectName`.

Both `a2aClientService.ts` files (and the gchat `messageHandler`) need to forward the new tenant key. Verify: grep for `installation:` in the two `a2aClientService.ts` files.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -245,11 +281,26 @@ async def upsert_channel_by_installation(
...
+        secret the client no longer recognises.
         """
         if data.installation_id is None:
             raise ValueError("installation_id required for upsert")
 
         existing = await self.get_by_installation(db, client_id, data.installation_id)
+
+        rekey_from: str | None = None
+        if existing is None and data.previous_installation_id:
+            legacy = await self._lock_by_installation(db, client_id, data.previous_installation_id)
```
alloy-cicd: **2. The adoption path is a standing, unauthenticated re-key — gated only on `client_id` and a free-text display name.**

`previous_installation_id` is sent by the clients on **every** boot forever (no version gate, no retirement), and its value is `bot.botName` — admin-editable free text (`BotInstallationsController.update` accepts `body.botName`). All Slack workspaces share one Keycloak client, so:

- An operator who renames any bot to a value that happens to match another channel's current `installation_id` (e.g. another workspace's App ID, or a not-yet-migrated `"Nannos"`) causes the next boot to adopt that row — overwriting its `secret`, `webhook_url` and key, and dragging every `scheduled_jobs.delivery_channel_id` bound to it onto the wrong tenant. Silently.
- Conversely, renaming a bot *before* its first post-upgrade boot makes its legacy row unmatchable: orphaned, with jobs still bound to it and a secret no client accepts.

At minimum, guard the adoption (only accept a `previous_installation_id` that doesn't collide with a live key / only while a rollout flag is set), give the field a retirement owner, and consider whether a one-shot data migration (`bot_name → app_id` per client) wouldn't fit this bounded, known set of rows better than permanent request-path machinery.

Related cleanup: the re-key is threaded through `update_channel` as an out-of-band `installation_id` kwarg whose only guard is a docstring. Doing the re-key as one explicit `self.update(..., fields={"installation_id": ...})` inside this adoption branch keeps the tenant-moving capability out of the shared mutator that a future PATCH wiring could reach.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -276,7 +327,22 @@ async def upsert_channel_by_installation(
             secret=data.secret,
             message_formatting=data.message_formatting,
         )
-        updated = await self.update_channel(db=db, actor=actor, channel_id=existing.id, data=update)
+        updated = await self.update_channel(
+            db=db,
+            actor=actor,
+            channel_id=existing.id,
+            data=update,
+            installation_id=data.installation_id if rekey_from else None,
```
alloy-cicd: **3. The re-key UPDATE can hit `delivery_channels_client_installation_uidx` with no SAVEPOINT — unlike the insert path right above it.**

The insert branch wraps its race in `begin_nested()` + `except IntegrityError` and converges. The adoption branch does a bare UPDATE of `installation_id`: if a concurrent replica inserted the same `(client_id, installation_id)` between this transaction's `get_by_installation` miss and the UPDATE, the unique-index violation escapes, aborts the whole request transaction, and the client gets a 500. Worse, the failed adopter's legacy row survives still keyed to the old value — and since the *new* key now exists, every future boot takes the `existing is not None` path and never looks at the legacy key again, so that row (with its scheduled jobs) is permanently orphaned.

Same `begin_nested()` treatment as the insert path, with a sensible fallback (drop the adoption and update the row the winner committed), closes it. Verify: compare with the guarded insert at the `except IntegrityError` block ~30 lines up.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -276,7 +327,22 @@ async def upsert_channel_by_installation(
...
+        )
+        if rekey_from:
+            logger.info(
+                "Re-keyed delivery channel %s from installation_id=%s to %s (client_id=%s); "
+                "scheduled jobs bound to it are unaffected",
```
alloy-cicd: **4. For the colliding tenants — the actual bug being fixed — pre-existing scheduled jobs keep misdelivering, and this log line claims otherwise.**

Where two workspaces shared `"Nannos"`, one row served both, and *both* tenants' `scheduled_jobs.delivery_channel_id` point at it. After deploy, whichever installation boots first adopts the row; the loser gets a fresh channel — but the loser's existing jobs stay bound to the adopted row and now deliver into the winner's workspace with the winner's secret. The cross-tenant leak this PR exists to fix survives for all pre-existing jobs, boot order decides which tenant is hit, and `test_only_the_first_claimant_adopts_a_shared_legacy_key` blesses the row behaviour without covering the job consequence.

Re-pointing those jobs is data work the registration call can't do — it needs a migration/ops script (which is also an argument for doing the whole re-key as a migration, per comment 2). At minimum, drop "scheduled jobs bound to it are unaffected" from this log message — it's only true when the legacy key wasn't shared — and consider logging a warning when more than one client presents the same `previous_installation_id`.

packages/client-slack/src/slackApp.ts
```diff
@@ -92,10 +92,13 @@ export async function startSlackApp(config: Config) {
       for (const bot of installations) {
         if (!bot.isActive) continue;
         try {
-          const secret = await installationSecretService.get(bot.botName);
+          // Keyed by app_id to match InstallationRegistrar. Keying on bot_name gave two
+          // installations that share a display name one secret, so this loop returned
+          // whichever was created first and delivered the notification to the wrong workspace.
+          const secret = await installationSecretService.get(bot.appId);
```
alloy-cicd, line 104: **5. Rollout window: inbound notifications are rejected until registration succeeds, and the old `botName` secret is orphaned in SSM.**

The verifier now checks only the `appId`-keyed secret, but that secret is first provisioned inside `registerOne` — which early-returns when `config.consoleBackend` is unset, runs fire-and-forget after `app.start()`, and swallows per-installation failures. Until it completes, the channel row in console-backend still holds the old `botName`-keyed secret, so every scheduler push mismatches and 403s; the 60s negative cache in `installationSecretService.get()` extends each miss. If registration keeps failing (console-backend briefly unreachable during the deploy), the outage persists silently.

A transitional fallback — also try the `botName` key when the `appId` lookup misses, mirroring `previous_installation_id` on the server side — would close the window. And nothing deletes or migrates the old `${prefix}/${botName}` SSM parameters, which remain live-looking credentials nobody rotates. Same applies to the Google Chat middleware.

packages/client-google-chat/src/middleware/a2aNotificationAuth.ts
```diff
@@ -26,7 +26,7 @@ export function createA2ANotificationAuthMiddleware(
 
     for (const project of googleChatConfigs) {
       try {
-        const secret = await installationSecretService.get(project.botName);
+        const secret = await installationSecretService.get(project.projectName);
```
alloy-cicd, line 32: **6. Nit: the `catch` below still logs `bot=${project.botName}` while the lookup key is now `project.projectName`** — when a lookup fails, the log names a value that no longer identifies the secret being resolved, which is exactly the kind of misdirection this PR removes elsewhere. Log the `projectName` (the module docstring's "keyed by botName" wording is stale too, if it's still there).

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
```diff
@@ -231,6 +240,33 @@ async def get_by_installation(
             return None
         return _row_to_response(row)
 
+    async def _lock_by_installation(
```
alloy-cicd: **7. Simplification: `_lock_by_installation` is `get_by_installation` byte-for-byte plus ` FOR UPDATE`.**

Two raw SQL strings over the same table/predicate now have to be edited together; a predicate change landing in one but not the other would make the adoption path select rows the normal path considers invisible. The repo already has the exact pattern for this: `RateCardRepository._find_rate_card_id(..., *, for_update: bool = False)` builds the query as `"SELECT ..." + (" FOR UPDATE" if for_update else "")` — its docstring even names the read-then-write re-key use case. Suggest `get_by_installation(..., for_update=True)` with the same keyword-only flag.

packages/client-slack/src/routes/openApiRouter.ts
```diff
@@ -3,18 +3,54 @@ import Router from '@koa/router';
...
+
+/**
+ * Reduce AJV errors to field paths and messages. `params` is included only for
+ * `additionalProperties`, whose param is the offending *key* name (not its value) and is the
+ * one detail that makes an otherwise unactionable error obvious.
+ */
+const toDetails = (errors: ValidateFunction['errors']): ValidationDetail[] =>
+  (errors ?? []).map((e) => {
+    const additional = (e.params as { additionalProperty?: string } | undefined)?.additionalProperty;
+    const path = e.instancePath || (additional ? `/${additional}` : '');
```
alloy-cicd: **8. Minor: for a *nested* `additionalProperties` failure the offending key name is dropped.**

`e.instancePath` is truthy for nested objects (e.g. `/config`), so the `additional` branch never runs and the detail reads `{path: '/config', message: 'must NOT have additional properties'}` — the caller can't tell which field was rejected, which is the stated purpose of `toDetails`. Today's schemas are flat so nothing hits this, but the helper is generic:

```ts
const path = additional ? `${e.instancePath}/${additional}` : e.instancePath;
```

packages/client-slack-frontend/src/pages/EditInstallationPage.tsx
```diff
@@ -33,8 +33,15 @@ export function EditInstallationPage() {
 
   const handleSubmit = async (data: InstallationFormData) => {
     if (!appId) return
+    // Spelled out rather than spread from `data`: the form is shared with the create flow, so
+    // it carries `appId`, which here belongs in the path and is rejected as an unknown property
+    // by the update schema. Listing the fields keeps that contract visible at the call site.
     const body = {
-      ...data,
+      teamId: data.teamId,
```
alloy-cicd, line 40: **9. Minor: the hand-listed fields trade one silent failure for another.**

Spelling the body out fixes the `appId` rejection, but now a field added to `InstallationFormData` and the update schema compiles cleanly here while never being sent — the form shows the value, the PUT drops it, and only the create flow persists it. `const { appId: _appId, ...body } = data` keeps the single source of truth in the type (the new router test uses exactly this idiom), or express the split once in the schema: `installationSchema.omit({ appId: true }).partial()`. Fine to keep as-is if the explicitness is preferred — flagging the trade-off so it's a decision, not an accident.

packages/client-slack/src/services/installationRegistrar.ts
aartaria: Fixed in 5680fae9: confirmed, and it was a release blocker — `metadata.installation` feeds `forwarded_installation`, so `list_channels_for_installation` would have returned empty for every workspace after deploy. Both clients now send the key they register under: `A2ASlackBasedRequest.installationId = appId`, Google Chat sends `projectName`.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Fixed in 5680fae9: the adoption UPDATE now gets the same `begin_nested`/`IntegrityError` treatment as the insert path. Agreed on reachability — the row lock only serialises claimants of the *same* legacy key, so a client whose previous key changed between boots can insert under the new key concurrently. On conflict it now converges on the winner and leaves the legacy row untouched for whoever still needs it.

packages/client-slack/src/slackApp.ts
aartaria: Fixed in 5680fae9: both inbound validators now fall back to the pre-re-key `bot_name` secret after every new-key secret has missed, and log when they do. Running it only as a second pass keeps the display-name ambiguity out of the normal path. Marked transitional — to remove once every environment has registered under the new key.

packages/client-google-chat/src/middleware/a2aNotificationAuth.ts
aartaria: Fixed in 5680fae9: now logs `project=${project.projectName}`.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Fixed in 5680fae9: folded into `get_by_installation` behind a `for_update` flag, matching the `rate_card_repository` pattern you pointed at. Helper removed, lock rationale moved into that docstring.

packages/client-slack/src/routes/openApiRouter.ts
aartaria: Fixed in 5680fae9: the offending key is now appended to `instancePath` rather than substituted, so a nested rejection reports `/config/extra`. Test added.

packages/client-slack-frontend/src/pages/EditInstallationPage.tsx
aartaria: Not applying, but it is a fair trade to name. The spread was my first version; eslint rejects the `const { appId: _appId, ...fields }` form under `@typescript-eslint/no-unused-vars`, so the options were a disable comment or an explicit list. I took the list because a field silently *dropped* on update is a visible, debuggable bug, whereas a field silently *sent* is rejected wholesale by `additionalProperties: false` — which is exactly the failure this PR exists to fix, and it takes down the entire edit. Happy to revisit if the form grows; a shared `InstallationUpdateBody` type derived from the route schema would settle it properly, and that belongs with the SDK regen rather than here.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Confirmed, and not fixable at registration time — leaving this thread open as the visible to-do. Where two tenants genuinely shared a key, the row carries jobs created from both and only one can adopt it; the other tenant's jobs keep delivering into the adopter's. Splitting them needs a data migration that can attribute each job to a tenant, which registration has no basis to do. I have corrected the misleading log line (it now says jobs already bound to the channel continue to use it, not that they are "unaffected") and stated the limitation in the upsert docstring in 5680fae9. Raising the migration with the author for a follow-up issue rather than growing this PR.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Leaving open for the author to call — this is a security trade-off, not a correctness bug. Two things bound it: adoption only fires when the caller has *no* channel under its own key, so it cannot touch an already-migrated tenant; and the actor able to rename a bot is a Slack-admin-UI admin who can already edit any installation's token directly, so the authority is not new. What is real is the accidental case: a cosmetic rename that happens to match a not-yet-migrated legacy name silently adopts that row. Options are (a) accept for one release and drop `previous_installation_id` from the clients afterwards, (b) gate adoption behind an env flag defaulted off after rollout, or (c) drop adoption and accept orphaned rows plus a manual re-point of existing jobs. I did not pick one unilaterally since it changes the rollout story.

aartaria: Fixed in 5290a284 — and you were right that my first reply understated it. I had argued adoption was bounded because it only fires when the caller has no channel under its own key; that does nothing about the case you named, where the renamed value matches another workspace's *live* App ID. That row is adopted just as readily as a stale one.

All four points:

1. **Live-key collision guard.** The clients now withhold a legacy key that matches any live tenant key (`liveKeys` set over active installations in both registrars). Only the client side holds the full mapping, so console-backend cannot make this call — it sees rows, not the set of keys that are current. The affected installation still registers its own channel; only the adoption offer is dropped, and it is logged with the remedy.
2. **Rollout flag.** `DELIVERY_CHANNEL_LEGACY_REKEY` (`DeliveryChannelConfig.legacy_rekey`, default on) gates the adoption branch; when off it logs the ignored key and falls through to a normal insert. Test added.
3. **Retirement owner.** #249, referenced from both the flag and the field docstring. It lists the removal steps (field, flag, adoption branch, the transitional `bot_name` secret fallbacks, orphaned secrets) and records the limitation from finding 4.
4. **Shared-mutator cleanup.** `update_channel` can no longer change `installation_id` at all — the kwarg is gone. Re-keying is `_rekey_installation`, called only from the adoption branch. Test asserts `update_channel` leaves `installation_id` untouched.

On the one-shot migration: I looked at it and it does not fit, because the `bot_name` → `app_id` mapping does not exist in console-backend's database. It lives in client-slack's own database and, for Google Chat, in that client's environment config — so a SQL migration has nothing to join against. The mapping is only knowable client-side at boot, which is why it arrives on the registration request. What the flag buys is the property you actually wanted from a one-shot: the machinery stops being permanent, and can be switched off from the server the moment rollout completes rather than when the last client redeploys.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Follow-up now filed as #249, which records this limitation alongside the retirement steps for the transitional machinery — including the point that it is easier to attribute jobs while the legacy rows are still identifiable, so the decision is better made before `DELIVERY_CHANNEL_LEGACY_REKEY` is switched off. Leaving this thread open until @aartaria decides whether that migration lands before or after this PR.

packages/client-slack/src/services/installationRegistrar.ts
```diff
@@ -64,10 +69,30 @@ export async function registerInstallations(deps: InstallationRegistrarDeps): Pr
     return;
   }
 
+  // Every app_id that currently owns a channel. A bot_name equal to one of these must never be
+  // offered as a legacy key: bot_name is admin-editable free text, so renaming a bot to another
+  // workspace's App ID would otherwise ask console-backend to hand that workspace's live channel
+  // — and every scheduled job bound to it — to the renamed bot. Only this side knows the full set
+  // of tenant keys, so the check belongs here rather than in console-backend.
+  const liveKeys = new Set(active.map((b) => b.appId));
```
alloy-cicd: **New (round 2): the collision guard only covers *active* installations, but inactive ones can still own live channels.**

`liveKeys` is built from `active`, yet a channel registered while an installation was active stays in console-backend keyed under that installation's App ID after the installation is deactivated (nothing deletes the row on deactivation, and deactivation is exactly the state where nobody is watching its deliveries). Renaming another bot to that App ID passes this guard — `previous_installation_id` is offered, and console-backend adopts the deactivated installation's channel along with every scheduled job bound to it: the precise hijack this guard was added to prevent, through the one class of key it doesn't cover.

One-word fix: build the set from `installations` (the full `listAll()` result) instead of `active` — an inactive installation's App ID is still a live tenant key on the console-backend side. The Google Chat twin doesn't have this gap (its config list is the complete key set), so only this file needs the change.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 2 (e0183aadf6b6583fd5b9c85191ed6fea622d375f..5290a2849f225d232c1c3a1bbe29d3659367fe36)

Ledger of round-1 findings:

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | A2A `installation` metadata still `botName` | **verified-fixed** (5680fae9) — both clients now send the key they register under; the only `A2ASlackBasedRequest` constructor sets `installationId: appId`, gchat sends `projectName` |
| 2 | Standing unauthenticated re-key via free-text bot name | **verified-fixed** (5290a284) — client-side live-key collision guard, `DELIVERY_CHANNEL_LEGACY_REKEY` flag, retirement tracked in #249, and the `update_channel` kwarg removed in favour of `_rekey_installation` reachable only from the adoption branch. One residual gap in the new guard — see the single inline comment |
| 3 | Re-key UPDATE unguarded against the unique index | **verified-fixed** (5680fae9) — same `begin_nested()`/`IntegrityError` treatment as the insert path, converging on the winner and leaving the legacy row for whoever still needs it |
| 4 | Colliding tenants' pre-existing jobs keep misdelivering | **still-open by design** — correctly deferred to a data migration, recorded in #249 and the upsert docstring; the misleading log line is fixed. The open thread remains the visible to-do for the human reviewer |
| 5 | Rollout window: inbound notifications rejected until registration lands | **verified-fixed** (5680fae9) — transitional `bot_name` secret fallback in both validators, second-pass only, logged, marked for removal in #249 |
| 6 | Stale `bot=` log label in gchat middleware | **verified-fixed** |
| 7 | `_lock_by_installation` duplicating `get_by_installation` | **verified-fixed** — folded behind `for_update`, matching `rate_card_repository` |
| 8 | Nested `additionalProperties` dropping the key name | **verified-fixed** — appends to `instancePath`, test added |
| 9 | Hand-listed update-body fields | **rejected-by-author** with sound reasoning (eslint `no-unused-vars` vs. a visible-vs-wholesale failure trade-off); accepted, not re-raised |

The delta itself is clean apart from one narrow gap in the new collision guard (inline). The transitional secret fallback deliberately reintroduces display-name ambiguity as a second pass only — acknowledged in the code and bounded by #249, fine as is.

packages/client-slack/src/services/installationRegistrar.ts
aartaria: Fixed in 607f20fd: `liveKeys` is now built from all installations rather than the active ones, with a comment explaining why deactivation does not retire a key — it stops re-registration but leaves the channel row in place under that App ID. Test added covering an inactive App ID as the collision target. Agreed the Google Chat twin has no equivalent gap, since its key set is the configured project list and that is complete by construction.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Resolved differently, in 660b0493: the adoption machinery is gone entirely, so the log line this thread objected to no longer exists.

Your diagnosis is what killed it. Adoption only preserved job bindings correctly for tenants that had *not* collided — and a shared row carries jobs from both tenants while only one can claim it, so for exactly the case this PR exists to fix it preserved them wrongly and kept the misdelivery alive. Paying for that with a tenant-moving operation in the request path, keyed on an admin-editable display string, was the wrong trade. Registration now creates or updates the channel for the key it presents and nothing else; 398 lines removed.

The consequence is explicit rather than papered over: old display-name rows stay as orphans with existing jobs bound to them, those jobs keep delivering via the transitional `bot_name` secret fallback, and they are re-pointed from the console scheduler job editor (`delivery_channel_id` is on the job update model and the detail page has a channel selector) — no migration script required. #249 now covers identifying them, re-pointing, deleting the orphans, and only then removing the fallback and the stale secrets. It also records your point that any job whose owner is not the tenant the orphaned row really served has been misdelivering since before this fix.

packages/console-backend/console_backend/repositories/delivery_channel_repository.py
aartaria: Superseded by 660b0493: rather than bounding this surface, it is gone. `previous_installation_id`, the `DELIVERY_CHANNEL_LEGACY_REKEY` flag, `_rekey_installation`, the `FOR UPDATE` lookup and the clients' live-key guards are all removed — there is no tenant-moving operation in the request path left to guard. Your instinct in the last paragraph of this comment was the right one: the machinery was not worth its risk, and the bounded set of rows is better handled as deliberate data work (#249). Recording it here since my previous reply described a guard and a flag that no longer exist.

alloy-cicd (commented):
🤖 **Automatic review** — Claude Code `/code-review`, incremental round 3 (5290a2849f225d232c1c3a1bbe29d3659367fe36..266d891d35f81897f3390bcbc947309ff3d93681)

Note: the branch was rebased since round 2 (new SHAs, BREAKING CHANGE footers added); the ledger below is against the content delta, which I verified commit by commit.

Ledger:

| Finding | Disposition |
|---------|-------------|
| Round 2: collision guard built from active installations only | **verified-fixed** (e8379ea9 built `liveKeys` from all installations), then **superseded** — e64a3119 removes the guard along with the entire adoption machinery it protected |
| Round 1 #2 (adoption as a standing re-key surface) | **verified-fixed by removal** (e64a3119) — `previous_installation_id`, the flag, `_rekey_installation`, the `FOR UPDATE` lookup, and the client guards are all gone; `test_update_channel_cannot_rekey` stays. I confirmed no stray references remain in any of the three packages (one exception, inline comment below) |
| Round 1 #4 (colliding tenants' jobs, left open) | **resolved by removal** — with adoption gone, re-pointing legacy-bound jobs is now an explicit, documented deploy step (BREAKING CHANGE footers + #249) rather than something adoption half-did. The author's reasoning that adoption preserved bindings *wrongly* for exactly the colliding case is correct |

One deliberate reversal the human reviewer should weigh, flagged once here and not as a finding: 266d891d also removes the transitional `bot_name` secret fallback that round 1's finding 5 introduced. The commit message owns the consequence explicitly — every pre-existing job bound to a display-name channel is now silently rejected at the callback (the push runs after the run reports SUCCESS, so the job shows no failure) until re-pointed, and that re-pointing is therefore a **blocking deploy step**, not cleanup. The rationale (the fallback's first-match-wins pass was the last code path where a display name decided tenant routing, i.e. the bug itself) is sound; the operational bet — that #249's re-pointing actually happens at deploy time in every environment — is the author's and their reviewer's to accept. The BREAKING CHANGE footer says the right thing.

The delta introduces nothing else: the removals are complete and symmetrical, the rewritten comments describe the final behaviour accurately (one stale exception below), imports that the removed code used are all still needed by remaining code, and the surviving tests pin the right invariants.

packages/client-google-chat/src/config/config.ts
```diff
@@ -95,9 +95,9 @@ export async function getConfigFromEnv(): Promise<Config> {
     if (!process.env[envVarName]) {
       throw new Error(`Please provide ${envVarName}`);
     }
-    // bot_name is load-bearing: it is the installation_id used for delivery-channel
-    // registration and the SSM key for the inbound notification secret. A missing
-    // value would silently break both, so fail fast at startup instead.
+    // bot_name is only a display string now (projectName is the tenant key), but it still
+    // labels the channel in the console and is sent as previous_installation_id so an
```
alloy-cicd: **Nit (round 3): this comment still describes the removed adoption mechanism.** `previous_installation_id` no longer exists anywhere — e64a3119 removed it from both registrars and the console-backend model — so "is sent as previous_installation_id so an already-registered channel is re-keyed rather than orphaned" now points a reader at machinery that's gone. Only the label half is still true; something like: *"bot_name is only a display string now (projectName is the tenant key), but it still labels the channel in the console. Fail fast if it is missing."* The Slack twin of this rationale was already cleaned up in the same commits, so this is the last stale reference.

aartaria: Fixed in b7250453: the comment no longer mentions `previous_installation_id`. It now says bot_name is a display string that still labels the channel in the console, which is why an absent one fails fast rather than registering an unidentifiable channel.

Note the branch was rebuilt into two commits since round 3, so that SHA is the new history rather than a commit on top of what you reviewed. The tree is identical to the round-3 head apart from this comment and the removal of the console-backend changes (they were docstring-only plus one test, and their presence would have pulled that package into a breaking-change version bump for no behaviour change).

